PR #114, opened 2015-01-29
Update example in doc.go

jnjackins:
The function argument to ebiten.Run takes an *ebiten.Image,
and must return an error.


hajimehoshi:
Nice catch! Thank you!

PR #134, opened 2016-01-18
Check was missing from context_js

DrJosh9000:
Run calls Check whether or not gopherjs is used, and it was not implemented here.


hajimehoshi:
Aha, I forgot to add Check for JS at 633ee4096e753c6567cc123e6c088a4de9324198 because .travis.yml was broken. Thank you!

PR #274, opened 2016-10-05
examples/font: simplify code

martinlindhe:


martinlindhe:
on second look, not sure this is correct.
it no longer can use ebitenutil\file_js.go OpenFile(), so i guess it breaks in browsers?

hajimehoshi:
I guess `ioutil.ReadFile` works only with node, but not sure. `ebiten.OpenFile` uses XHR so that it works on browsers. Have you checked that?

martinlindhe:
i have not checked with browsers, can do that, but i'' having doubts if this was a valid fix in the first place :-) might just drop it

hajimehoshi:
Sure, let me close this once. Please feel free to reopen this if you confirm this is a problem.

PR #276, opened 2016-10-13
examples/sprites: Reuse op

tSU-RooT:
Hi, I have a question.

ebiten/examples/sprites/main.go
```go
	op := &ebiten.DrawImageOptions{
		ImageParts: sprites,
	}
	op.ColorM.Scale(1.0, 1.0, 1.0, 0.5)
	if err := screen.DrawImage(ebitenImage, op); err != nil {
		return err
	}
```
this sample codes generates new instance of  option on every update, 
Is this needed?

### DIFF
--- a/examples/sprites/main.go
+++ b/examples/sprites/main.go
@@ -102,7 +102,10 @@ const (
 	MaxSprites = 50000
 )
 
-var sprites = &Sprites{make([]*Sprite, MaxSprites), 500}
+var (
+	sprites *Sprites
+	op      *ebiten.DrawImageOptions
+)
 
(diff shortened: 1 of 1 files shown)


hajimehoshi:
This is not needed actually, but your fix adds a new global variable. This is a kind of trade-off between code-readability and performance, and I adopted the latter. IMHO performance is not a problem here.

hajimehoshi:
Let me close this. Feel free to reopen this if you find there is actually a performance problem.

tSU-RooT:
> IMHO performance is not a problem here.  

I got it.  

Thank you for reply.

PR #318, opened 2017-02-06
Hide console window on Windows

gonutz:
When compiling on Windows without the -ldflags "-Hwindowsgui" flag, a
console window pops up when running a compiled program.
This change hides the console after the program starts. The console is
visible for a moment before the actual game window shows.

### DIFF
--- a/internal/ui/hide_console_windows.go
+++ b/internal/ui/hide_console_windows.go
@@ -1,0 +1,16 @@
+package ui
+
+import "syscall"
+
+var (
+	kernel32         = syscall.NewLazyDLL("kernel32.dll")
+	user32           = syscall.NewLazyDLL("user32.dll")
+	getConsoleWindow = kernel32.NewProc("GetConsoleWindow")
+	showWindowAsync  = user32.NewProc("ShowWindowAsync")
+)
+
+func hideConsoleWindowOnWindows() {
+	windowHandle, _, _ := getConsoleWindow.Call()
+	const SW_HIDE = 0
+	showWindowAsync.Call(windowHandle, SW_HIDE)
+}
--- a/internal/ui/hide_console_windows.go
+++ b/internal/ui/hide_console_windows.go
@@ -14,17 +14,95 @@
 
 package ui
 
-import "syscall"
-
-var (
-	kernel32         = syscall.NewLazyDLL("kernel32.dll")
-	user32           = syscall.NewLazyDLL("user32.dll")
-	getConsoleWindow = kernel32.NewProc("GetConsoleWindow")
-	showWindowAsync  = user32.NewProc("ShowWindowAsync")
+import (
+	"syscall"
+	"unsafe"
 )
 
+// hideConsoleWindowOnWindows will hide the console window that is showing when
+// compiling on Windows without specifying the '-ldflags "-Hwindowsgui"' flag.
+//
+// The hiding is done if the console window exists for less than 1 second
+// because during development you will want to run from the command line and it
+// is not supposed to be hidden whenever you 'go run' or 'go test' the program.
+// That is why the work-around is to assume that the developer console is always
+// open for >= 1 sec begore running the program. The end user will double-click
+// the executable, it will pop up a console window and then call this function
+// right away so the time between creating the console and running this function
+// should never surpass 1 second. In that case the console is then hidden.
 func hideConsoleWindowOnWindows() {
-	windowHandle, _, _ := getConsoleWindow.Call()
-	const SW_HIDE = 0
-	showWindowAsync.Call(windowHandle, SW_HIDE)
+	consoleWindow, _, _ := getConsoleWindow.Call()
+	if consoleWindow == 0 {
+		return // no console attached
+	}
+
+	var consoleProcessID uint32
+	getWindowThreadProcessId.Call(
+		consoleWindow,
+		uintptr(unsafe.Pointer(&consoleProcessID)),
+	)
+	if consoleProcessID == 0 {
+		return // error retrieving the console process ID
+	}
+
+	const (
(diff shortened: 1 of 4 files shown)


gonutz:
I tested this code on Windows and Linux, on Windows it hides the console and on Linux it does nothing, just like it should.

hajimehoshi:
Thank you!

1. This doesn't change anything in case when launching an Ebiten app from console like `go run`, is this correct?
2. Please add the license comment (see https://github.com/hajimehoshi/ebiten/commit/6d2692b0db1b5a06c3ba20034f41ac82ebcdf758).
3. I prefer `hide_console_dummy.go` to `hide_console_notwindows.go`

hajimehoshi:
Also please check the CI result: this fails to compile with GopherJS.

gonutz:
OK, you gave me something to think about there :-)
To answer your questions:

1. It did always hide the console at first which is why I now check if the attached console has been open for more than 1 second and in that case assume this is the developer running the program from the command line, so do not close the console. If on the other hand the console was open for less than 1 second, this is assumed to be the user double-clicking on the .exe in which case the console is hidden.

2. License comments were added.

3. The non-Windows file was renamed to `hide_console_notwindows.go`.

4. There was a compile error which I fixed.

internal/ui/hide_console_windows.go
```diff
@@ -14,17 +14,95 @@
...
+		PROCESS_QUERY_INFORMATION = 0x0400
+		FALSE                     = 0
+	)
+	consoleProcess, _, _ := openProcess.Call(
+		PROCESS_QUERY_INFORMATION,
+		FALSE,
+		uintptr(consoleProcessID),
+	)
+	if consoleProcess == 0 {
+		return // error retrieving the console process handle
```
hajimehoshi: If this is unexpected, return error value (e.g. `errors.New("ui: OpenProcess failed")` is fine)

internal/ui/hide_console_windows.go
```diff
@@ -14,17 +14,95 @@
...
+	var creationTime, ignore filetime
+	ok, _, _ := getProcessTimes.Call(
+		uintptr(consoleProcess),
+		uintptr(unsafe.Pointer(&creationTime)),
+		uintptr(unsafe.Pointer(&ignore)),
+		uintptr(unsafe.Pointer(&ignore)),
+		uintptr(unsafe.Pointer(&ignore)),
+	)
+	if ok == 0 {
+		return // error retrieving the process creation time
```
hajimehoshi: ditto.

gonutz:
I think the function should not return an error for two reasons.

1. Hiding the console is a convenience and if it fails for some esoteric reason and the console stays open, this does not really hurt anybody. It would not be worth the effort to even check the error that would be returned as it is not clear how to handle it anyway. You would of course not want the program to exit if the console does not close. The only thing to do would be logging an error which is not necessary anyway. If you as the developer see that the console does not close but you know it ought to, you will see that right away. So no need for logging it.

2. If the window does not have a console attached because you actually did compile with the -windowsgui flag, then the process ID will be 0 or the process handle will be nil so the early return is actually correct for this case.

gonutz:
I think that would be very involved, I mean this solution is already involved but I think it works fine in reality.
You know how it is, you try your best to write the perfect program and then reality hits you.
The 1 second solution works for me for double-clicking, running and testing from cmd and running from my IDE.
The only case where the console would not be hidden would be if you open the command line, type 'go run ...' and the program builds and starts all in less than 1 second. This should not happen.

hajimehoshi:
> Hiding the console is a convenience and if it fails for some esoteric reason and the console stays open, this does not really hurt anybody. It would not be worth the effort to even check the error that would be returned as it is not clear how to handle it anyway. You would of course not want the program to exit if the console does not close. The only thing to do would be logging an error which is not necessary anyway. If you as the developer see that the console does not close but you know it ought to, you will see that right away. So no need for logging it.
> 
> If the window does not have a console attached because you actually did compile with the -windowsgui flag, then the process ID will be 0 or the process handle will be nil so the early return is actually correct for this case.

That makes sense. Thank you for explanation.

hajimehoshi:
> I think that would be very involved, I mean this solution is already involved but I think it works fine in reality.
You know how it is, you try your best to write the perfect program and then reality hits you.
> The 1 second solution works for me for double-clicking, running and testing from cmd and running from my IDE.
> The only case where the console would not be hidden would be if you open the command line, type 'go run ...' and the program builds and starts all in less than 1 second. This should not happen.

Yeah I know your PR works in reality, but I thought checking parent processes is a more robust way.

OK, I'll accept your PR. I might fix checking-parent way later if I have time.

I appreciate your work! lgtm

gonutz:
Cool, thanks!

hajimehoshi:
Hmm, I have tested this but on my machine (Vaio Z Canvas), the delta time was more than 1000[ms] and fails to hide the console. I'll adopt checking-parent way anyway.

gonutz:
I just noticed a bug in the code that I wrote, the time is not converted correctly in:
`const ms = 1000 // to convert dt to milliseconds`
it actually would have to be:
`const ms = 10000 // to convert dt to milliseconds`
so in the old case the timeout was 100 ms and not 1 second.

I realized this because I added the same code to my `prototype` library.

Anyway, I saw that you now use the parent process ID, maybe I will adopt that, I am going to have to test it now on my machine and when running from Windows+R.

gonutz:
Your solution works great, that was a very good idea, I think I might steal it from you... ;-)

hajimehoshi:
> Your solution works great, that was a very good idea, I think I might steal it from you... ;-)

Thank you, but I updated further better way. Feel free to copy this :-)

gonutz:
Very nice, I just updated my code with this solution, thank you very much for pointing it out!

PR #401, opened 2017-08-05
Initial keyboard effort

j7b:
Haven't tested js yet.

### DIFF
--- a/examples/runes/main.go
+++ b/examples/runes/main.go
@@ -1,0 +1,26 @@
+// +build example
+
+package main
+
+import (
+	"github.com/hajimehoshi/ebiten"
+	"github.com/hajimehoshi/ebiten/ebitenutil"
+)
+
+var runes = []rune("Type some stuff on your keyboard:\n")
+
+func update(screen *ebiten.Image) error {
+	runes = append(runes, ebiten.Keyboard()...)
+	if ebiten.IsKeyPressed(ebiten.KeyEnter) {
+		if l := len(runes); l > 0 {
+			if runes[l-1] != '\n' {
+				runes = append(runes, '\n')
+			}
+		}
+	}
+	return ebitenutil.DebugPrint(screen, string(runes))
+}
+
+func main() {
+	ebiten.Run(update, 320, 240, 2.0, "Hi")
+}
--- a/examples/runes/main.go
+++ b/examples/runes/main.go
@@ -12,10 +12,8 @@ var runes = []rune("Type some stuff on your keyboard:\n")
 func update(screen *ebiten.Image) error {
 	runes = append(runes, ebiten.Keyboard()...)
 	if ebiten.IsKeyPressed(ebiten.KeyEnter) {
-		if l := len(runes); l > 0 {
-			if runes[l-1] != '\n' {
-				runes = append(runes, '\n')
-			}
+		if len(runes) > 0 && runes[len(runes)-1] != '\n' {
+			runes = append(runes, '\n')
 		}
 	}
 	return ebitenutil.DebugPrint(screen, string(runes))
--- a/internal/ui/keyboard.go
+++ b/internal/ui/keyboard.go
@@ -1,0 +1,7 @@
+package ui
+
+import "sync"
+
+var runebuffer []rune
+
+var rblock = new(sync.Mutex)
--- a/internal/ui/keyboard.go
+++ b/internal/ui/keyboard.go
@@ -4,4 +4,4 @@ import "sync"
 
 var runebuffer []rune
 
-var rblock = new(sync.Mutex)
+var rblock sync.Mutex
--- a/internal/ui/keyboard_js.go
+++ b/internal/ui/keyboard_js.go
@@ -2,28 +2,17 @@
 
 package ui
 
-import "github.com/gopherjs/gopherjs/js"
+import (
+	"unicode"
 
-var doc = js.Global.Get("document")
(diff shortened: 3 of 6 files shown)


internal/ui/keyboard.go
```diff
@@ -0,0 +1,7 @@
+package ui
+
+import "sync"
+
+var runebuffer []rune
+
+var rblock = new(sync.Mutex)
```
hajimehoshi: `var rblock sync.Mutex` sounds enough.

internal/ui/keyboard_js.go
```diff
@@ -0,0 +1,39 @@
...
+
+package ui
+
+import "github.com/gopherjs/gopherjs/js"
+
+var doc = js.Global.Get("document")
+
+func push(char rune) {
+	go func() {
+		rblock.Lock()
```
hajimehoshi: I don't think lock is needed on JS: `runebuffer` never cause racing

examples/runes/main.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/hajimehoshi/ebiten/ebitenutil"
+)
+
+var runes = []rune("Type some stuff on your keyboard:\n")
+
+func update(screen *ebiten.Image) error {
+	runes = append(runes, ebiten.Keyboard()...)
+	if ebiten.IsKeyPressed(ebiten.KeyEnter) {
+		if l := len(runes); l > 0 {
+			if runes[l-1] != '\n' {
```
hajimehoshi: How about `if len(runes) > 0 && runes[len(runes)-1] != '\n'`?

internal/ui/keyboard_js.go
```diff
@@ -0,0 +1,39 @@
+// +build js
+
+package ui
+
+import "github.com/gopherjs/gopherjs/js"
+
+var doc = js.Global.Get("document")
+
+func push(char rune) {
+	go func() {
```
hajimehoshi: Is goroutine needed here? (There are no caller of `push`, so probably this is a temporary hack?)

internal/ui/keyboard_js.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func init() {
+	if doc == nil {
+		return
+	}
+	kp := doc.Get("onkeypress")
+	switch kp.Length() {
+	case 0:
+		doc.Set("onkeypress", push)
```
hajimehoshi: `addEventListener` is preferable.

internal/ui/keyboard_js.go
```diff
@@ -0,0 +1,39 @@
...
+		runebuffer = append(runebuffer, char)
+		rblock.Unlock()
+	}()
+}
+
+func init() {
+	if doc == nil {
+		return
+	}
+	kp := doc.Get("onkeypress")
```
hajimehoshi: Which environment is keypress event unavailable?

internal/ui/keyboard.go
j7b: Initing lock types as pointers prevents inadvertent copies, but np.

internal/ui/keyboard_js.go
j7b: ya redid

internal/ui/keyboard_js.go
j7b: yep not

internal/ui/keyboard_js.go
j7b: All have keypress, the weird thing is charCode is deprecated in favor of the unimplemented char

internal/ui/keyboard_js.go
j7b: Yes, actually had to refactor because preventdefault in keydown prevents keypress from firing, keypress might need preventdefault but I think keyup's preventdefault is doing the right thing.

examples/runes/main.go
j7b: did

internal/ui/keyboard.go
```diff
@@ -0,0 +1,7 @@
+package ui
+
+import "sync"
+
+var runebuffer []rune
+
+var rblock sync.Mutex
```
hajimehoshi: You can move this to keyboard_glfw.go

internal/ui/keyboard.go
```diff
@@ -0,0 +1,7 @@
+package ui
+
+import "sync"
+
+var runebuffer []rune
```
hajimehoshi: I'd dup this to keyboard_*.go and remove this file.

internal/ui/keyboard_js.go
```diff
@@ -0,0 +1,26 @@
...
+
+func keypress(e *js.Object) {
+	if runebuffer != nil {
+		if r := rune(e.Get("charCode").Int()); unicode.IsPrint(r) {
+			runebuffer = append(runebuffer, r)
+		}
+	}
+}
+
+func Keyboard() []rune {
```
hajimehoshi: So Keyboard has side effect to clear the buffer? This is a different behavior from Touches or other API. Is it possible to clear the buffer when a new frame comes instead of when Keyboard is called?

hajimehoshi: BTW, I feel like there should be a better name than Keyboard, but not sure. Keyboard is confusing with existing APIs like IsKeyPressed. How about InputChars?

j7b:
Clearing the buffer requires wrapping updatefunc or adding it to ui_* as part of loop

PR #424, opened 2017-08-28
js: Detecting pulse input on browser

1l0:
This is not intended to merge. Just showing how I tried to detect pulse input on browsers.
Tap works but there is a glitch for drag arised from this.
Sorry for commit messages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,9 @@ script:
   - go test -v github.com/hajimehoshi/ebiten
   - go test -v github.com/hajimehoshi/ebiten/internal/math
   - go test -v github.com/hajimehoshi/ebiten/internal/restorable
-  - gopherjs build --tags example -v github.com/hajimehoshi/ebiten/examples/blocks
+
+# Suspend building with GopherJS temporarily (#422)
+# - gopherjs build --tags example -v github.com/hajimehoshi/ebiten/examples/blocks
 
 # Looks like testing GL on node is hard.
 # - gopherjs test -v github.com/hajimehoshi/ebiten github.com/hajimehoshi/ebiten/internal/graphics
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 go:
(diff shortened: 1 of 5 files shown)


hajimehoshi:
> Sorry for commit messages.

No problem: all commits will be squashed so please make commits as you like :-)

hajimehoshi:
> Just showing how I tried to detect pulse input on browsers.
> Tap works but there is a glitch for drag arised from this.

I don't understand the intention of this PR yet.  Why can this detect pulse input?

1l0:
Check the difference of behaviour of examples/paint on a browser with tap on macOS trackpad.

hajimehoshi:
Tried but I couldn't notice the difference...

hajimehoshi:
@1l0 Could you tell me how different they are?

1l0:
On [official example](https://hajimehoshi.github.io/ebiten/examples/paint.html), single tap doesn't work sometimes for me.

hajimehoshi:
* Do you use a Magic Trackpad? I'm using a trackpad on a MacBook Pro, and I don't think there is a difference between them.
* Perhaps, did you enable 'tap to click'? I didn't, and I'll enable it and try again later. http://osxdaily.com/2014/01/28/enable-tap-to-click-mac-trackpad/

1l0:
Tap to click is enabled using MacBook trackpad.

hajimehoshi:
Confirmed that your PR changed the situation.

* In both cases, taps were ignored in most cases.
* It looks like more taps could be recognized with your PR than the original state.
* Strokes didn't end until the next tap with your PR

hajimehoshi:
I understand a tap is treated as a click, but should this be treated as mousedown/up or drag events? I'm a little confused.

1l0:
That's strange, taps are not ignored on this PR for me. Anyway yes this PR is not a sort of solution to fix something but my demand was just simply to detect a tap on my environment so perhaps you don't need to bother with it at all.

PR #467, opened 2017-12-31
Allow vertex colors

seebs:
This is more applicable for things like line-drawing, but specifying
colors per-vertex allows OpenGL to handle gradients in hardware,
which makes them a lot prettier than if you had to do them in
your images, also quite a lot faster.

Signed-off-by: Seebs <seebs@seebs.net>

(This is related to my interest in using scaled "images" consisting of a single white pixel as a way to rapidly draw lines with actual thickness, etc., unlike what DrawLine and DrawRect can do.)

### DIFF
--- a/image.go
+++ b/image.go
@@ -61,6 +61,12 @@ func (i *Image) Fill(clr color.Color) error {
 	return nil
 }
 
+func (i *Image) SetColor(index int, clr color.Color) error {
+	rgba := color.RGBAModel.Convert(clr).(color.RGBA)
+	i.restorable.SetColor(index, rgba)
+	return nil
+}
+
 // DrawImage draws the given image on the image i.
 //
 // DrawImage accepts the options. For details, see the document of DrawImageOptions.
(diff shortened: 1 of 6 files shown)


seebs:
whoops, missed the test case. added the missing color values to images_test.go:vertices().

hajimehoshi:
Thank you for the PR.

Hmm, I'm against adding `SetColor` to the image since changing color is already done by `ColorM`. Is it possible to check if the given `ColorM` is just scaling without translating?

As this adds the amount of vertices data, I'm worried that this might affect a performance impact for apps that don't use colors.

seebs:
I think ColorM is solving a different problem -- it's an image-wide effect, rather than being different ones per point. I'm looking for the ability to have gradients. I guess they could be ColorM gradients rather than RGBA gradients, but I'd be concerned that the computational load would be a lot higher, because at that point you'd have to try to interpolate between points, and I'm not sure how easy that is. Straight multiplicative colors, GL will handle without any additional effort.

Cost of the additional vertex data appears to be very small.

hajimehoshi:
I understand you want to draw shapes with gradient color, but could you wait for my designing how to implement vector graphics including what you want?

seebs:
seems reasonable, sure. if you want second opinions on design, etc, ping me in gitter, or have a look at faiface/pixel which i think has a pretty good API for a lot of what it does.

hajimehoshi:
Thank you for your advice :-)

PR #477, opened 2018-01-07
Seebs/prevertex

seebs:
Updated/cleaned up a bit. This produces actual correct results, unlike a previous iteration. It's still noticably faster under load; doing the very simple subset of matrix multiplication needed to apply affine transformations on the CPU is not very expensive, and is very close to being cheaper than even copying the bits over to the GPU at all. Also frees up six positions in the vertex format in case of future enhancements like colors.


hajimehoshi:
Could you merge this with master branch? Thank you!

vertices.go
```diff
@@ -70,85 +97,25 @@ func vertices(sx0, sy0, sx1, sy1 int, width, height int, geo *affine.GeoM) []flo
 		g.Concat(geo)
 		geo = &g
 	}
+	x0, y0 := float64(0), float64(0)
```
hajimehoshi: `0.0` is preferable to `float64(0)`

vertices.go
```diff
@@ -43,6 +43,33 @@ func (v *verticesBackend) get() []float32 {
 	return s
 }
 
+// compute lowest power of two equal to or greater than n.
```
hajimehoshi: Start with `npo2 is` for Go convention

BTW, I'd use `internal/math.NextPowerOf2Int`. I believe performance is not so different. Have you compared this?

vertices.go
```diff
@@ -70,85 +97,25 @@ func vertices(sx0, sy0, sx1, sy1 int, width, height int, geo *affine.GeoM) []flo
...
-	vs[46] = g4
-	vs[47] = g5
+	x, y := geo.Apply32(x0, y0)
+	oneVertex(vs[0:6], x, y, u0, v0, u1, v1)
+	x, y = geo.Apply32(x1, y0)
+	oneVertex(vs[6:12], x, y, u1, v0, u0, v1)
+	x, y = geo.Apply32(x0, y1)
+	oneVertex(vs[12:18], x, y, u0, v1, u1, v0)
+	x, y = geo.Apply32(x1, y1)
+	oneVertex(vs[18:24], x, y, u1, v1, u0, v0)
```
hajimehoshi: These create new slices: Is the performance on browsers fine?

seebs: I haven't tried, I'm not sure how. It may be worse, in which case probably don't do this?

vertices.go
seebs: I would expect it to be noticably worse, in most cases. Branches are horrendously expensive, and loops that will usually loop several times are going to be *way* more expensive than a small number of operations. Not that this is a very expensive part of the rendering as a whole, but for the actual cost of this operation, I'd expect probably around an order of magnitude difference. The bit operation thing ends up being about 12 instructions, and can be inlined (at least in non-desktop go) because there's no loops. The loop version can't be inlined, and while it might be a couple fewer instructions, for even a small image (say, 64px), it's going to be looping at least six times.

vertices.go
hajimehoshi: As we discussed on Slack, this would affect the performance on browsers since a slice is represented by ArrayBuffer. Please keep these assignments flat.

vertices.go
hajimehoshi: As there is a benchmark `BenchmarkDrawImage` in image_test.go, let's compare the performance. I admit theoretically it is possible that your code is faster, but this seems a too-early optimization to me.

hajimehoshi: On second thought, as the current code doesn't use `NextPowerOf2Int` anyway, I'm fine with your npo2. Thank you for discussing at Slack.

Please fix the comment anyway.

hajimehoshi:
ping?

vertices.go
hajimehoshi: I started to think about the performance of `npo2` on browsers: on browsers, a lot of range checks are inserted in interger operations. I don't think this is not too much in this case, but am not sure.

vertices.go
hajimehoshi: I'll fix this later.

Issue #537, opened 2018-03-07
graphics.MaxImageSize for framebuffer too small for 5K displays


nf:
I've been playing a little with `ebiten` and I've found that running the `spriteshd` example I get this panic:

```
imac:examples nf$ go run spriteshd/main.go 
panic: ebiten: width (5120) must be less than or equal to 4096

goroutine 6 [running]:
github.com/hajimehoshi/ebiten.checkSize(0x1400, 0xb40)
	/Users/nf/go/src/github.com/hajimehoshi/ebiten/image.go:576 +0x236
github.com/hajimehoshi/ebiten.newImageWithScreenFramebuffer(0x1400, 0xb40, 0x0)
	/Users/nf/go/src/github.com/hajimehoshi/ebiten/image.go:555 +0x39
github.com/hajimehoshi/ebiten.(*graphicsContext).SetSize(0xc42097c000, 0x780, 0x438, 0x4005555555555555)
	/Users/nf/go/src/github.com/hajimehoshi/ebiten/graphicscontext.go:62 +0x103
github.com/hajimehoshi/ebiten/internal/ui.(*userInterface).updateGraphicsContext(0x42c10c0, 0x4213240, 0xc42097c000)
	/Users/nf/go/src/github.com/hajimehoshi/ebiten/internal/ui/ui_glfw.go:451 +0x103
github.com/hajimehoshi/ebiten/internal/ui.(*userInterface).update(0x42c10c0, 0x4213240, 0xc42097c000, 0xc4200ae100, 0xc4204f4080)
	/Users/nf/go/src/github.com/hajimehoshi/ebiten/internal/ui/ui_glfw.go:475 +0x10a
github.com/hajimehoshi/ebiten/internal/ui.(*userInterface).loop(0x42c10c0, 0x4213240, 0xc42097c000, 0x0, 0x0)
	/Users/nf/go/src/github.com/hajimehoshi/ebiten/internal/ui/ui_glfw.go:507 +0x85
github.com/hajimehoshi/ebiten/internal/ui.Run(0x780, 0x438, 0x3ff0000000000000, 0x41f61f6, 0x18, 0x4213240, 0xc42097c000, 0x42c8ce0, 0x41cd660)
	/Users/nf/go/src/github.com/hajimehoshi/ebiten/internal/ui/ui_glfw.go:397 +0x165
[...]

PR #538, opened 2018-03-07
internal/graphics: increase defaultViewportSize to 5120

nf:
This makes it possible to run ebiten applications in full screen on 5k
displays, such as Apple's 27" iMac.

Fix issue #537


hajimehoshi:
Thank you!

* The bug is a real concern, so we should fix this. Thank you for reporting!
* The const affects a texture size, 
  * The number should be power of 2 for portability and ease to calculation. Could you fix this to 8192?
  * I'm not sure how widely 8192-size texture is supported. Checking the maximum texture size with GL_MAX_TEXTURE_SIZE would be required. I'll add this check, so you don't have to care about this in this PR.

nf:
@hajimehoshi I changed it to 8192, and also expanded the comment.

hajimehoshi:
lgtm, thank you for contribution!

PR #555, opened 2018-03-15
genkeys: Add Pause and Print Screen buttons

silbinarywolf:
Add support for Pause and Print Screen buttons.

Tested on Windows builds only so far.


hajimehoshi:
Awesome! That's great since JS version is also implemented.

Would it be possible to add other keys like NumLock? https://godoc.org/github.com/go-gl/glfw/v3.2/glfw#Key

silbinarywolf:
Sure. Do you want me to just go through and fill in the gaps for everything? (Numlock, Numpads keys, etc)

hajimehoshi:
Yes, as many (GLFW) keys as possible (if the key doesn't exist on browsers, please skip that).

silbinarywolf:
@hajimehoshi Is there a simple input testing app? ie. Something where I can press any key and it just tells me in the key in the top left corner or something?

I'd just like to be able to run an Ebiten project in the browser to ensure those keys are supported before adding them in.

If there isn't already such a tool in the project, should there be?

silbinarywolf:
Is it all good I update examples/keyboard to print the current key being held in the top left corner? Would help with testing now and into the future.

If not, I'll just put together my own program for doing this :)

hajimehoshi:
Yeah, I'm fine with fixing example/keyboard to show texts in the top left corner :-) This would be good since anyone else can test the new keys easily.

hajimehoshi:
Now I've fixed `examples/keyboard`, what we have to do is to add keys to `genkeys.go` and `go generate`?

silbinarywolf:
The only key left seemed to be the ScrollLock key.

Also with testing I found that PrintScreen didn't print when pressed, so Im assuming that key just doesn't work in browsers (at least in Chrome + Edge), however it worked on native Windows builds so I just commented out the keyCodeToNameEdge item.

genkeys.go
```diff
@@ -70,6 +72,7 @@ func init() {
 		"RightBracket": {"BracketRight"},
 		"GraveAccent":  {"Backquote"},
 		"NumLock":      {"NumLock"},
+		"ScrollLock":   {"ScrollLock"},
```
hajimehoshi: You don't have to care the order: please add all the new keys here

genkeys.go
```diff
@@ -130,12 +134,16 @@ func init() {
...
 		0x08: "Backspace",
 		0x90: "NumLock",
+		0x91: "ScrollLock",
 		0x6e: "KPDecimal",
 		0x6f: "KPDivide",
 		0x6a: "KPMultiply",
 		0x6d: "KPSubtract",
 		0x6b: "KPAdd",
 
+		// On Chrome and Edge, this does not work seem to be captured. At least on Windows.
```
hajimehoshi: You don't have to care the order: please add all the new keys here

Also fix the comment: Chrome doesn't use this map.

genkeys.go
```diff
@@ -130,12 +134,16 @@ func init() {
...
+		//0x2C: "PrintScreen",
```
hajimehoshi: You don't have to comment this out. This should not be harmful. I'll check this later.

silbinarywolf:
I'll be looking into:
- Adding the ContextMenu key if it's on my keyboard
- Will uncomment the PrintScreen edge keycode. My assumption was we explicitly don't want to expose keys that don't work.

I also want to note that on my Windows 10 machine, if you press and hold PrintScreen, it will flicker quickly that it's being pressed in the top-left corner than go away. We might want this to act more consistently with other operating systems if they allow you to press and hold it?

silbinarywolf:
Done.

So the "Menu" key doesn't seem to work in Chrome at all, however it will work in IE11 and Edge with the following weirdness. (ie. highlighting the "-" key on the keyboard display, but displaying in text that correct keyname. If you press the "-" key, it won't be on the keyboard display but will say "Minus" in the top-left corner)

menu_highlights_minus_in_ie11_and_edge

genkeys.go
hajimehoshi: You missed this?

genkeys.go
```diff
@@ -130,11 +135,14 @@ func init() {
...
 		0x08: "Backspace",
 		0x90: "NumLock",
+		0x91: "ScrollLock",
+		0x2C: "PrintScreen",
 		0x6e: "KPDecimal",
 		0x6f: "KPDivide",
 		0x6a: "KPMultiply",
 		0x6d: "KPSubtract",
 		0x6b: "KPAdd",
+		0x5d: "Menu",
```
hajimehoshi: Please add all the new keys here.

genkeys.go
```diff
@@ -70,6 +72,8 @@ func init() {
 		"RightBracket": {"BracketRight"},
 		"GraveAccent":  {"Backquote"},
 		"NumLock":      {"NumLock"},
+		"ScrollLock":   {"ScrollLock"},
+		"Menu":         {"Menu"},
```
hajimehoshi: `"Menu": {"ContextMenu"}` (GLFW key is the key, KeyboardEvent.code is the value)

hajimehoshi:
> Will uncomment the PrintScreen edge keycode. My assumption was we explicitly don't want to expose keys that don't work.

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

Oh, if Edge behaves same way as IE, PrintScreen works only on keyup event, and PrintScreen doesn't work on Edge well. OK, let's comment it out with comments. Sorry for confusion.

silbinarywolf:
Anything else?
Not sure what you mean by "You missed this?" for ScrollLock.

hajimehoshi:
> You don't have to care the order: please add all the new keys here

genkeys.go
```diff
@@ -70,6 +73,7 @@ func init() {
 		"RightBracket": {"BracketRight"},
 		"GraveAccent":  {"Backquote"},
 		"NumLock":      {"NumLock"},
+		"Menu":         {"ContextMenu"},
```
hajimehoshi: Ah sorry, could you move all the new keys here? I think this is my last request.

hajimehoshi:
Oh, you need to rebase this PR...

genkeys.go
```diff
@@ -136,6 +143,9 @@ func init() {
 		0x6d: "KPSubtract",
 		0x6b: "KPAdd",
 
+		// NOTE: On Edge, this key does not work. hajimehoshi stated: "PrintScreen works only on keyup event, and PrintScreen doesn't work on Edge well."
```
hajimehoshi: `hajimehoshi stated` is not needed... I'll fix this laster.

genkeys.go
```diff
@@ -136,6 +143,9 @@ func init() {
 		0x6d: "KPSubtract",
 		0x6b: "KPAdd",
```
hajimehoshi: The new comment should have been here.

silbinarywolf:
Done

genkeys.go
hajimehoshi: nvm, you did the right work :-)

hajimehoshi:
Thank you!

PR #559, opened 2018-03-18
Remove use of defer in simple use cases

explodes:
`defer` is being used in simple functions, I have removed them for a small performance win.

Removing `defer` from draw functions will have a much larger cumulative performance win.


hajimehoshi:
As there is a test for panic in `image_test.go`, defer usage is the must in `image.go` unfortunately.

image.go
```diff
@@ -150,6 +150,9 @@ func (i *Image) isDisposed() bool {
 //
 // DrawImage always returns nil as of 1.5.0-alpha.
 func (i *Image) DrawImage(img *Image, options *DrawImageOptions) error {
+	if i == img {
+		panic("ebiten: cannot draw image onto itself")
```
hajimehoshi: Why is the new panic needed?

explodes: `image_test TestImageSelf` looks for a panic when you try to draw an image onto itself. This simply panics early, instead of when the underlying system panics when you try to do so. This allows  the panicking behavior to exist while also being able to remove the `defer` from image.go

hajimehoshi: Ah ok. Probably https://github.com/hajimehoshi/ebiten/blob/master/internal/shareable/shareable.go#L112 is no longer needed (if `i != img`, the images MUST have different backend images after `ensureNotShared` is called). Could you fix the panic in shareable.go from `shareable: Image.DrawImage: img must be different from the receiver` to just `not reached`?

hajimehoshi:
You might want to rebase this. Many locks have gone.

PR #587, opened 2018-04-22
Do not allocate a touch slice if no touches have been made

explodes:
Touches() will always allocate a new slice to copy touches into even if no touches are reported, for example on desktop. Adding a simple check makes sure that this unnecessary allocation does not happen.

### DIFF
--- a/input.go
+++ b/input.go
@@ -122,16 +122,13 @@ type Touch interface {
 
 // Touches returns the current touch states.
 //
-// Touches will return nil when there are no touches.
+// Touches returns nil when there are no touches.
 // Touches always returns nil on desktops.
 func Touches() []Touch {
-	t := ui.AdjustedTouches()
-	if len(t) == 0 {
-		return nil
+	touches := ui.AdjustedTouches()
+	var copies []Touch
+	for _, touch := range touches {
+		copies = append(copies, touch)
 	}
-	tt := make([]Touch, len(t))
-	for i := 0; i < len(tt); i++ {
-		tt[i] = t[i]
-	}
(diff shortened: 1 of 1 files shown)


hajimehoshi:
I understand what you mean but does the cost really matter?

explodes:
Of course! This method is likely to be called every frame, resulting in potentially 60 allocations per second. 
It isn't a major perf problem, but it is one nonetheless. 

The question is, is a check her needlessly complicated code versus the actual perf gain.

hajimehoshi:
Hmm, considering compiler optimization, I don't think we should care.

However, I don't strongly disagree to merge this. So it's ok. Only one concern is this might be a breaking change, but I think this is acceptable.

Could you fix the comment of Touches func that this might return nil?

input.go
```diff
@@ -122,9 +122,13 @@ type Touch interface {
 
 // Touches returns the current touch states.
 //
+// Touches will return nil when there are no touches.
```
hajimehoshi: `returns`? (A function is singular)

input.go
```diff
@@ -122,9 +122,13 @@ type Touch interface {
...
 // Touches always returns nil on desktops.
 func Touches() []Touch {
 	t := ui.AdjustedTouches()
+	if len(t) == 0 {
+		return nil
+	}
```
hajimehoshi: Another way to avoid `if` clause would be:

```go
ts := ui.AdjustedTouches()
var ts2 []Touch
for _, t := range ts {
    ts2 = append(ts2, t)
}
return ts2
```

PR #628, opened 2018-06-16
Enable mouse wheel input

PSteinhaus:
This change simply adds the funcionality to read mouse wheel input, which (at least for me) is a very necessary feature. My solution is clearly not perfect, as I do rely on somewhat of a workaround to reset the scrolling value to 0 while no scroll events happen. That should probably be changed, but I hope  you forgive me for that, as I am absolutely new to go and haven't been able to find an alternative to this yet, though it's probably pretty easy.

### DIFF
--- a/input.go
+++ b/input.go
@@ -52,7 +52,7 @@ func CursorPosition() (x, y int) {
 	return ui.AdjustedCursorPosition()
 }
 
-// MouseScroll returns the x and y offset of the scroll wheel.
+// MouseWheel returns the x and y offset of the scroll wheel.
 // It returns 0 if the wheel isn't being rolled.
 func MouseWheel() (xoff, yoff float64) {
 	return input.Get().MouseWheel()
--- a/input.go
+++ b/input.go
@@ -52,8 +52,8 @@ func CursorPosition() (x, y int) {
 	return ui.AdjustedCursorPosition()
 }
 
-// MouseScroll returns the x and y offset of the scroll wheel or scrolling device.
-// For a scroll wheel it returns 0 if the wheel isn't being rolled.
+// MouseScroll returns the x and y offset of the scroll wheel.
+// It returns 0 if the wheel isn't being rolled.
 func MouseScroll() (xoff, yoff float64) {
 	return input.Get().MouseScroll()
 }
--- a/input.go
+++ b/input.go
@@ -55,7 +55,7 @@ func CursorPosition() (x, y int) {
 // MouseScroll returns the x and y offset of the scroll wheel.
 // It returns 0 if the wheel isn't being rolled.
 func MouseWheel() (xoff, yoff float64) {
-	return input.Get().MouseScroll()
+	return input.Get().MouseWheel()
 }
 
 // IsMouseButtonPressed returns a boolean indicating whether mouseButton is pressed.
--- a/input.go
+++ b/input.go
@@ -52,8 +52,9 @@ func CursorPosition() (x, y int) {
 	return ui.AdjustedCursorPosition()
 }
 
-// MouseScroll returns the y offset of the mouse wheel
-func MouseScroll() float64 {
+// MouseScroll returns the x and y offset of the scroll wheel or scrolling device.
+// For a scroll wheel it returns 0 if the wheel isn't being rolled.
+func MouseScroll() (xoff, yoff float64) {
 	return input.Get().MouseScroll()
 }
 
--- a/input.go
+++ b/input.go
@@ -52,6 +52,11 @@ func CursorPosition() (x, y int) {
 	return ui.AdjustedCursorPosition()
 }
 
+// MouseScroll returns the y offset of the mouse wheel
+func MouseScroll() float64 {
+	return input.Get().MouseScroll()
+}
+
 // IsMouseButtonPressed returns a boolean indicating whether mouseButton is pressed.
 //
 // This function is concurrent-safe.
--- a/input.go
+++ b/input.go
@@ -54,7 +54,7 @@ func CursorPosition() (x, y int) {
 
 // MouseScroll returns the x and y offset of the scroll wheel.
 // It returns 0 if the wheel isn't being rolled.
-func MouseScroll() (xoff, yoff float64) {
+func MouseWheel() (xoff, yoff float64) {
 	return input.Get().MouseScroll()
 }
 
--- a/internal/input/input_glfw.go
+++ b/internal/input/input_glfw.go
@@ -30,9 +30,9 @@ import (
 type Input struct {
 	keyPressed         map[glfw.Key]bool
 	mouseButtonPressed map[glfw.MouseButton]bool
-	scrollX		   float64
-	scrollY		   float64
-	scrollTime	   time.Time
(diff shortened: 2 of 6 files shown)


internal/input/input_glfw.go
```diff
@@ -22,13 +22,16 @@ package input
...
+	"time"
 
 	glfw "github.com/go-gl/glfw/v3.2/glfw"
 )
 
 type Input struct {
 	keyPressed         map[glfw.Key]bool
 	mouseButtonPressed map[glfw.MouseButton]bool
+	scrollY			   float64
+	scrollTime		   time.Time
```
hajimehoshi: Please run `gofmt -s -w .` or the CI test doesn't pass.

input.go
```diff
@@ -52,6 +52,11 @@ func CursorPosition() (x, y int) {
 	return ui.AdjustedCursorPosition()
 }
 
+// MouseScroll returns the y offset of the mouse wheel
```
hajimehoshi: From the comment, I couldn't understand when this function return 0 or non-0 value.

input.go
```diff
@@ -52,6 +52,11 @@ func CursorPosition() (x, y int) {
...
+func MouseScroll() float64 {
+	return input.Get().MouseScroll()
```
hajimehoshi: You should implement this function not only at `input_glfw.go` but also `input_js.go` and `input_mobile.go`. I think it should be possible to implement this for browsers. For mobiles, just return 0 is fine.

input.go
hajimehoshi: BTW, is Y axis enough? Doesn't wheeling happen both on X an Y axes?

PSteinhaus: I see your point. I oriented that comment on the variable names in the callback (xoff, yoff). In my mind it works that way that the offset of a normal mouse wheel is always set back to 0 when it's not used. I'll add that.

For me, Y axis was enough, but i guess I should add it while I'm at it.

input.go
PSteinhaus: Ok, will do!

PSteinhaus: I implemented the function at `input_mobile.go`  but failed to do so at `input_js.go`. Sry, I'm too inexperienced to see through that.

internal/input/input_glfw.go
PSteinhaus: Thx, that solved that problem.

PSteinhaus:
I am sorry, I have no idea why the CI test fails. Do you have any?

PSteinhaus:
I added a `MouseScroll()` function to `input_js.go` , but it's disfunctional and I don't have the knowledge necessary to implement a working version, sorry.
Tell me if you're ok with that. I will of course close this pull request if not.

hajimehoshi:
No problem: could you reave a TODO comment at `input_js.go`? I'll take a look later.

input.go
```diff
@@ -52,6 +52,12 @@ func CursorPosition() (x, y int) {
 	return ui.AdjustedCursorPosition()
 }
 
+// MouseScroll returns the x and y offset of the scroll wheel or scrolling device.
+// For a scroll wheel it returns 0 if the wheel isn't being rolled.
+func MouseScroll() (xoff, yoff float64) {
```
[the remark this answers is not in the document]
hajimehoshi: As far as I know, GLFW 3.2 doesn't accept touches. So just renaming should be fine :-)

internal/input/input_glfw.go
```diff
@@ -101,6 +111,17 @@ func (i *Input) Update(window *glfw.Window, scale float64) {
 				i.m.Unlock()
 			}
 		})
+		// I added this in here so that it would only be run once too
+		window.SetScrollCallback(func(w *glfw.Window, xoff float64, yoff float64) {
+			i.m.Lock()
+			i.scrollX = xoff
+			i.scrollY = yoff
+			i.scrollTime = time.Now() // this is a rather inelegant way of setting the scroll value to 0 if no scrolling is happening
```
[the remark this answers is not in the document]
PSteinhaus: Thanks for the tip, it's changed now.

input.go
```diff
@@ -52,6 +52,12 @@ func CursorPosition() (x, y int) {
 	return ui.AdjustedCursorPosition()
 }
 
+// MouseScroll returns the x and y offset of the scroll wheel.
+// It returns 0 if the wheel isn't being rolled.
+func MouseScroll() (xoff, yoff float64) {
```
hajimehoshi: Rename this to `MouseWheel`

PSteinhaus: Agreed, that's more fitting. Done.

internal/input/input_js.go
```diff
@@ -91,6 +96,11 @@ func (i *Input) IsMouseButtonPressed(button MouseButton) bool {
 	return false
 }
 
+// TODO: Mouse scroll functionality is not yet implemented in js
+func (i *Input) MouseWheel() (xoff, yoff float64) {
```
hajimehoshi: Define the counterpart in `input_mobile.go`

internal/input/input_js.go
```diff
@@ -49,6 +49,11 @@ func (i *Input) ClearRuneBuffer() {
 	i.runeBuffer = nil
 }
 
+func (i *Input) ResetScrollValues() {
```
hajimehoshi: I think `ResetScrollValues` is not needed in `input_js.go`

internal/input/input_js.go
PSteinhaus: Done.

internal/input/input_js.go
```diff
@@ -91,6 +91,11 @@ func (i *Input) IsMouseButtonPressed(button MouseButton) bool {
 	return false
 }
 
+// TODO: Mouse scroll functionality is not yet implemented in js
```
hajimehoshi: Move this TODO comment in the function body

internal/input/input_mobile.go
```diff
@@ -36,6 +38,11 @@ func (i *Input) IsKeyPressed(key Key) bool {
 	return false
 }
 
+// TODO: Scrolling is not yet implemented on mobile.
```
hajimehoshi: Remove this comment: I don't think I can implement this for mobiles.

internal/input/input_js.go
PSteinhaus: True, I'll add a TODO comment in `ui_js.go`

internal/input/input_glfw.go
```diff
@@ -101,6 +115,13 @@ func (i *Input) Update(window *glfw.Window, scale float64) {
 				i.m.Unlock()
 			}
 		})
+		// I added this in here so that it would only be run once too
+		window.SetScrollCallback(func(w *glfw.Window, xoff float64, yoff float64) {
```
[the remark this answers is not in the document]
PSteinhaus: But using the method with `i.callbacksInitialized` makes it easier to understand and maintain the code, so it's probably better, I agree.

hajimehoshi:
Thank you for using and contributing to my lib! Detecting mouse wheel state sounds an important feature that I didn't realize before you created this PR :-)

PR #672, opened 2018-08-26
ebitenutil/file: Fix OpenFile() so it works with wasm

silbinarywolf:


silbinarywolf:
Closing for now until I ensure this is working correctly.

PR #674, opened 2018-08-26
ebitenutil/file: Fix OpenFile() so it works with wasm

silbinarywolf:
* Change to use `js.NewCallback`
* Use `uint8contentWrapper := js.Global().Get("Uint8Array").New(content)`, without this all the bytes in `data` stays zeroed. (TypedArray cannot use `set` on an ArrayBuffer)

### DIFF
--- a/ebitenutil/file_js.go
+++ b/ebitenutil/file_js.go
@@ -39,7 +39,7 @@ func OpenFile(path string) (ReadSeekCloser, error) {
 	req := js.Global().Get("XMLHttpRequest").New()
 	req.Call("open", "GET", path, true)
 	req.Set("responseType", "arraybuffer")
-	req.Call("addEventListener", "load", func() {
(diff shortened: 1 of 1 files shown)


hajimehoshi:
This change leaks callback objects. The callback objects should be one, or be deleted correctly.

silbinarywolf:
@hajimehoshi I didnt understand what you meant initially, but I just CTRL+F'd through the Golang source code for `js.NewCallback` to see how they used it.

Is this fine?

hajimehoshi:
I'm not sure what code you are pointing out, but in this PR's case, `NewCallback` is called whenever `OpenFile` is called, right?

hajimehoshi:
Ah sorry I missed them

hajimehoshi:
Updated .travis.yml to use the latest Go (and GopherJS). Could you rebase this PR and let TravisCI run again? Thanks!

PR #723, opened 2018-10-23
Speed up DrawTriangles

seebs:
DrawTriangles is expensive and slow because of massive memory
allocation and garbage collection costs. This patch moves from ~47TPS
on my laptop (with ~24k triangles) to 60TPS. The first part
is just allocating the right size of vertex buffer up front; that
got to about 55TPS. The second part replaces the frequent
allocations of []float32 in Vertex() calls with writing the
desired values into a provided destination slice.

Time spent in drawing triangles for 1,000 frames:
	13.07s	baseline
	11.09s	preallocate whole buffer to avoid resizing
	6.13s	use new PutVertex function

This might need some cleanup, but I think it's good evidence that
the design change is viable.

### DIFF
--- a/image.go
+++ b/image.go
@@ -519,10 +519,10 @@ func (i *Image) DrawTriangles(vertices []Vertex, indices []uint16, img *Image, o
 		filter = graphics.Filter(img.filter)
 	}
 
-	vs := []float32{}
+	vs := make([]float32, len(vertices)*10)
 	src := img.mipmap.original()
-	for _, v := range vertices {
-		vs = append(vs, src.Vertex(float32(v.DstX), float32(v.DstY), v.SrcX, v.SrcY, v.ColorR, v.ColorG, v.ColorB, v.ColorA)...)
+	for idx, v := range vertices {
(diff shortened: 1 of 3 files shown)


internal/graphicsutil/vertices.go
```diff
@@ -136,6 +136,33 @@ func QuadIndices() []uint16 {
...
+	vs[3] = sy / hf
+	vs[4] = -1
+	vs[5] = -1
+	vs[6] = cr
+	vs[7] = cg
+	vs[8] = cb
+	vs[9] = ca
+}
+
 func Vertex(width, height int, dx, dy, sx, sy float32, cr, cg, cb, ca float32) []float32 {
```
hajimehoshi: I think `Vertex` is no longer used? :-)

hajimehoshi: My suggestion is to remove `Vertex` and rename `PutVertex` to `Vertex`.

seebs:
I've updated this and rebased to master. I decided to keep the PutVertex name because the behavior is otherwise-surprising; most of the time, functions like this return things, rather than pushing them into a provided slice.

hajimehoshi:
I might change the name or signature, but `PutVertex` is ok :-)

Issue #729, opened 2018-11-06
devicescale: Heavy on Linux


hajimehoshi:
They call a command every frame. Reported by Ben.

Issue #713, opened 2018-10-11
Use 'main thread' pattern


hajimehoshi:
This reduces most of global functions:

```go
func main() {
    ebiten.Main(func(env *ebiten.Env) {
        img := env.NewImage(100, 100)
        scale := env.DeviceScaleFactor()
        // ...

        env.SetScreenSize(100, 100)
        env.Run(func(screen *ebiten.Image) error {
            // ...
        })
    })
}
```

This solves several problems:

1. Some functions like `DeviceScaleFactor` must be called on the main OS thread before `Run`. This pattern conforms users to do that.
2. `Image.At` must not be called in `init` functions since JVM is not initialized yet at that time. With this pattern, creating images would be impossible in `init`.

PR #731, opened 2018-11-06
Added monitor cache and scale cache

lologarithm:
resolves https://github.com/hajimehoshi/ebiten/issues/729

In Linux (and at least when using Gnome) requesting the monitors and scaling factor is very time consuming.

We can cache the list of all monitors and simply subscribe to a window change event to recalculate the window cache. This will get rid of needing to re-calculate monitors on every call to 'currentMonitorImpl' (at least in unix). The cache is implemented for all desktop clients because it could be helpful to other OSes (not just unix)

Additionally the scale can be cached for at least a single frame. It is possible that different monitors have different scale factor so the cache will probably need to be upgraded to be per-monitor once that is supported in unix.

I am not sure how to tell when a frame begins and ends (might be easier with https://github.com/hajimehoshi/ebiten/issues/713) so this is implemented with a time check. 

Here are pictures of profiles comparing old to new code, see the change so that GetAt is taking much less time per frame.
Before:
old_profile
After:
new_profile

### DIFF
--- a/internal/devicescale/impl_unix.go
+++ b/internal/devicescale/impl_unix.go
@@ -19,11 +19,13 @@
 package devicescale
 
 import (
+	"math"
 	"os"
 	"os/exec"
 	"regexp"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"time"
 )
 
@@ -39,10 +41,27 @@ const (
 )
 
 var (
-	cachedScale  float64
-	cachedAt     int64
-	scaleExpires = int64(time.Second / 60)
+	cachedScale     uint64 // use atomic to read/write as multiple goroutines touch it.
+	cacheUpdateWait = time.Millisecond * 100
 )
+
+func init() {
+	go scaleUpdater()
+}
+
+func impl(x, y int) float64 {
+	return math.Float64frombits(atomic.LoadUint64(&cachedScale))
+}
+
+// run as goroutine. Will keep the desktop scale up to date.
+// This can be removed once the scale change event is implemented in GLFW 3.3
+func scaleUpdater() {
+	for {
+		s := getscale(0, 0)
+		atomic.StoreUint64(&cachedScale, math.Float64bits(s))
+		time.Sleep(cacheUpdateWait)
+	}
+}
 
 func currentDesktop() desktop {
 	tokens := strings.Split(os.Getenv("XDG_CURRENT_DESKTOP"), ":")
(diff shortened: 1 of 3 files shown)


internal/ui/ui_glfw.go
```diff
@@ -117,6 +121,44 @@ func initialize() error {
...
 
+type cachedMonitor struct {
+	m  *glfw.Monitor
+	vm *glfw.VidMode
+	// Pos of monitor in virtual coords
+	x int
+	y int
+}
+
+// monitors is the cache for unix monitor list.
```
hajimehoshi: Not only for unix but also all desktops using GLFW?

internal/devicescale/impl_unix.go
```diff
@@ -37,6 +38,12 @@ const (
 	desktopXfce
 )
 
+var (
+	cachedScale    float64
+	cachedAt       int64
+	scaleExpiresMS = int64(16)
```
hajimehoshi: How about using nano time? The code would be:

```
scaleExpires = time.Second / 60

// ...

now := time.Now().UnixNano()
if now-cachedAt < scaleExpiresMS {

// ...

cachedAt = now // now can be reused here: I don't think now should be updated.
```

Wdyt?

lologarithm: Implementing this now.

lologarithm:
@hajimehoshi Another idea I just had this evening is having a background thread calling the gsettings application every 100ms or so. They would use sync.atomic package to read/write the cached value. 

This would result in at most a 100ms latency before resizing the application after changing desktop scaling (which seems reasonable). Additionally you never have the main thread waiting for the gsettings application. You do incur the atomic.LoadIntXX (however we choose to represent the float value, probably something like X decimal places as an integer).

internal/devicescale/impl_unix.go
lologarithm: Oh also, I recalculated 'now' because of how long gsettings was taken.

hajimehoshi: OK so the time how long gsettings is taken cannot be ignored. Thanks!

hajimehoshi:
Having another goroutine sounds a better idea!

>  (however we choose to represent the float value, probably something like X decimal places as an integer).

That's true. Probably `LoadInt64` and `math.Float64bits` will be used? Another idea is to use mutex (`sync.Mutex`). I think mutex is also fine here.

PR #757, opened 2018-12-20
Clarify the type of passing parameters

wasedaigo:
I was getting the following error on High Sierra
```
program_source:141:7: error: call to 'clamp' is ambiguous
```

This change fixes the compile error

### DIFF
--- a/internal/graphicsdriver/metal/driver.go
+++ b/internal/graphicsdriver/metal/driver.go
@@ -170,7 +170,7 @@ fragment float4 FragmentShader(VertexOut v [[stage_in]],
   }
   c = (color_matrix_body * c) + color_matrix_translation;
   c *= v.color;
-  c = clamp(c, 0, 1);
+  c = clamp(c, 0.0f, 1.0f);
   c.rgb *= c.a;
   return c;
 }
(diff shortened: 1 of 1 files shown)


wasedaigo:
I have realized I do not need  to add ".f" for floating point to make it compile, what is the standard?

internal/graphicsdriver/metal/driver.go
```diff
@@ -170,7 +170,7 @@ fragment float4 FragmentShader(VertexOut v [[stage_in]],
   }
   c = (color_matrix_body * c) + color_matrix_translation;
   c *= v.color;
-  c = clamp(c, 0, 1);
+  c = clamp(c, 0.0f, 1.0f);
```
hajimehoshi: Can we add f to the other line?

```
     float2 rate = clamp(((fract(p0 * source_size) - rate_center) * scale) + rate_center, 0.0, 1.0);
```

hajimehoshi:
> @hajimehoshi I have realized I do not need to add ".f" for floating point to compile this, what is the coding standard?

I've not decided coding style in Metal program.

hajimehoshi:
As we discussed offline, let's remove `f` for consistency.

wasedaigo:
removed

PR #779, opened 2019-01-11
use Opengl instead on early MacOSX platform which does not support Metal

mebusy:
Those early MacOS devices such like `iMac (21.5-inch, Mid 2011)`  do not support Metal.  
"Metal is not supported on this system"

I added a patch to detect whether this Mac support Metal , and use Opengl instead if not.

### DIFF
--- a/internal/graphicscommand/driver_mac.go
+++ b/internal/graphicscommand/driver_mac.go
@@ -20,8 +20,26 @@ package graphicscommand
 import (
 	"github.com/hajimehoshi/ebiten/internal/graphicsdriver"
 	"github.com/hajimehoshi/ebiten/internal/graphicsdriver/metal"
+	"github.com/hajimehoshi/ebiten/internal/graphicsdriver/metal/mtl"
+	"github.com/hajimehoshi/ebiten/internal/graphicsdriver/opengl"
 )
 
+// patch: some earlier Apple Mac product dose
+// NOT support Metal
+var is_MacOS_GPU_support_Metal = true
+
+func init() {
+	// on 1st initialization , detect Metal feature
+	_, err := mtl.CreateSystemDefaultDevice()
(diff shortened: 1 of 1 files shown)


internal/graphicscommand/driver_mac.go
```diff
@@ -20,8 +20,26 @@ package graphicscommand
 import (
 	"github.com/hajimehoshi/ebiten/internal/graphicsdriver"
 	"github.com/hajimehoshi/ebiten/internal/graphicsdriver/metal"
+	"github.com/hajimehoshi/ebiten/internal/graphicsdriver/metal/mtl"
+	"github.com/hajimehoshi/ebiten/internal/graphicsdriver/opengl"
 )
 
+// patch: some earlier Apple Mac product dose
+// NOT support Metal
+var is_MacOS_GPU_support_Metal = true
```
hajimehoshi:
```go
// isMetalSupported represents whether Metal is supported.
var isMetalSupported = true
```

internal/graphicscommand/driver_mac.go
```diff
@@ -20,8 +20,26 @@ package graphicscommand
...
+
+func init() {
+	// on 1st initialization , detect Metal feature
```
hajimehoshi: I think we don't need this comment.

internal/graphicscommand/driver_mac.go
```diff
@@ -20,8 +20,26 @@ package graphicscommand
...
+	_, err := mtl.CreateSystemDefaultDevice()
+	if err != nil {
```
hajimehoshi:
```go
if _ err := mtl.CreateSystemDefaultDevice(); err != nil {
```

hajimehoshi:
Thank you!

PR #840, opened 2019-04-11
Change docs about go-bindata into rakyll/statik

eihigh:
go-bindata is not appropriate because no longer maintained.
Now I suggest rakyll/statik insteadly which is actively maintained.

NOTE: I confirmed that go-bindata is documented only here.


eihigh:
Thanks!

Issue #926, opened 2019-08-22
Add event API to allow low-level events (e.g. input)


beoran:
As per our discussion on Slack I'd like to propose that Ebiten exposes the low level events it receives from the underlying platform (key presses, joystick, etc) in a separate package "event". 

Arguments in favor: 

1) It's more efficient than converting the state based API.
2) It's easier to use for certain applications, such as GUI's.
3) In some cases it's more correct.

The API could be something like it is in Allegro, https://liballeg.org/a5docs/trunk/events.html,  but without the C-ism like unions and an event queue, which could be replaced simply with a Go channel.

hajimehoshi:
Thanks! I agree to introduce a new package for lower level input. The existing input APIs would rely on the new 'event' package.

Note that gomobile has an event system by channels: https://godoc.org/golang.org/x/mobile/app

I'll reply more details about your idea later.

hajimehoshi:
My assumption is that all event, not only inputs but also e.g., updating frames, should be in the same event loop, so the event package would treat all such events. (and the existing APIs would be the wrapper).  Do you think it'd make sense?

beoran:
Yes, that makes sense. All events should be handled in the same loop. In Allegro we also have things like timer events, or events when the screen becomes not drawable, etc. The existing APi is the wrapper, and if you like, you can also request that the events are not sent only to the wrapper but also directly to your own channel.

Yes, the design of the event system could probably look like it is in x/mobile/app , although the events would be a bit more high level (wrapping the platform specific events).  In that library there are events for lifecycle, mouse, paint, size, touch. We will also need events for joystick, keyboard, window. Perhaps timers are not needed as Go already provides these.

hajimehoshi:
Thanks. We're now on the same page :) Then, what we should do will be:

1) Rewrite driver.UI, driver.UIContext, driver.Input, (e.g., https://github.com/hajimehoshi/ebiten/blob/master/internal/driver/ui.go) event-based
2) Rewite all the driver implementation
3) Create the new package github.com/hajimehoshi/ebiten/event to use the new driver API
4) Rewrite the existing package github.com/hajimehoshi/ebiten to use the event package

Could you write a doc about your suggestion of the new API in 'event' package?

beoran:
It's easier for me to write some code than a design document, so I wrote this documented code in the hope it will be clear enough to show what I mean.  Feel free to make suggestions, the file is here:

[...]

PR #930, opened 2019-08-27
PR for Event package as discused in #926

beoran:
Let's first implement this event package, so it can then be used gradually where needed.


event/event.go
```diff
@@ -0,0 +1,181 @@
+// event is a package that models events that occur during the execution of a program
```
hajimehoshi: `Package event is ...`

hajimehoshi: Add a license comment: https://github.com/hajimehoshi/ebiten/blob/master/CONTRIBUTING.md

event/event.go
```diff
@@ -0,0 +1,181 @@
...
+// It is empty for now because there are no general methods
+// required of events yet.
+type Event interface {
+}
+
+// KeyCharacter is an event that occurs when a character is actually typed on
+// the keyboard. This may be provided by an input method.
+type KeyCharacter struct {
+	// Code of the key pressed or released
+	Code int
```
hajimehoshi: What kind of code value is inside? I guess we'd need an enum value.

event/event.go
```diff
@@ -0,0 +1,181 @@
...
+	// Key board modifiers
+	Modifiers int
```
hajimehoshi: ditto

event/event.go
```diff
@@ -0,0 +1,181 @@
...
+type GamepadAxis struct {
+	// Which gamepad caused the event
+	ID int
+	// Axis of the game pad that changed, if any
+	Axis int
+	// Position of the axis after the change
+	Position float32
+}
+
+//GamepadButtonDown is a game pad button press event.
```
hajimehoshi: Add space after `//`

event/event.go
```diff
@@ -0,0 +1,181 @@
...
+type GamepadButtonDown struct {
+	// Which gamepad caused the event
+	ID int
+	// Button that was pressed, if any.
+	Button int
+	// Position of the button after the change.
+	Position float32
```
hajimehoshi: What is position?

event/event.go
```diff
@@ -0,0 +1,181 @@
...
+	Code int
+	// Key board modifiers
+	Modifiers int
+}
+
+// KeyUp is an event that occurs when a key is released on the keyboard.
+// The data is the same as for a KeyDown event
+type KeyUp KeyDown
+
+// GamepadAxis is for event where an axis on a game pad changes.
```
hajimehoshi: `game pad` -> `gamepad`

event/event.go
```diff
@@ -0,0 +1,181 @@
...
+	Position float32
+}
+
+//GamepadButtonDown is a game pad button release event.
+//The data is identical to a GamePadButtonDown event.
+type GamepadButtonUp GamepadButtonDown
+
+//GamepadAttach happens when a new game pad is attached.
+type GamepadAttach struct {
+	// Which gamepad caused the event
```
hajimehoshi: `ID represents ...`

event/event.go
```diff
@@ -0,0 +1,181 @@
...
+	ID int
+	// Amount of axes the game pad has.
```
hajimehoshi: ditto

event/event.go
```diff
@@ -0,0 +1,181 @@
...
+	Axes int
+	// Amount of buttons it has.
```
hajimehoshi: ditto

event/event.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+//GamepadButtonDown is a game pad button release event.
+//The data is identical to a GamePadButtonDown event.
+type GamepadButtonUp GamepadButtonDown
+
+//GamepadAttach happens when a new game pad is attached.
+type GamepadAttach struct {
+	// Which gamepad caused the event
+	ID int
```
hajimehoshi: Could a gamepad identifier an integer?

event/event.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+//MouseAxes is a mouse axis event.
+type MouseAxis struct {
+	// X position of the event
+	X float32
+	// Y position of the event
+	Y float32
+	// Wheel is the position of the mouse wheel
+	Wheel float32
```
hajimehoshi: Wheel should have multiple directions.

event/event.go
```diff
@@ -0,0 +1,181 @@
...
+	// DeltaX is the change in X since last event
+	DeltaX float32
+	// DeltaY is the change in Y since last event
+	DeltaY float32
+	// DeltaWheelis the change in the wheel position since last event
+	DeltaWheel float32
+	// Button that was pressed.
+	Button int
```
hajimehoshi: What kind of value is used here?

event/event.go
```diff
@@ -0,0 +1,181 @@
...
+	// Y position of the event
+	Y float32
+	// Change in X since last touch event
+	DeltaX float32
+	// Change in Y since last touch event
+	DeltaY float32
+	// Pressure of applied touch
+	Pressure float32
+	// Is the touch event primary or not
+	Primary bool
```
hajimehoshi: Primary represents whether

examples/event/main.go
```diff
@@ -0,0 +1,73 @@
+package main
```
hajimehoshi: This file is great to know how event package is used, but we don't need this file in Ebiten.

event/event.go
beoran: OK, I will do this.

beoran: OK, I will do this.

event/event.go
beoran: OK, I wixll make a type KeyCode, and enumerate all values. this will be a large file, so probably it is best to do this in a separate file keycode.go. If you agree with this idea then I will do this.

event/event.go
beoran: OK, I will do this (see previous comment).

event/event.go
beoran: Ok, I will do this everywhere.

event/event.go
beoran: On some game pads, buttons are analog, and can report how "hard" they are pressed. Maybe the word Pressure is better? Pressure normally varies between 0.0 (not pressed) and 1.0 (fully pressed).

event/event.go
beoran: Ok, I will fix this everywhere.

event/event.go
beoran: OK, I will fix this everywhere

event/event.go
beoran: OK, will fix this.

event/event.go
beoran: OK will fix this too. :)

event/event.go
beoran: Sorry I don't quite undserstand your question? In Allegro we used an integer to identify a gamepad. For example, if you get a GamepadAttach event with ID 7, it means that from now on you can expect to revceive other gamepad events with ID also set to 7. Would perhaps you like to use a string ID in stead?

event/event.go
beoran: In that case, I think it's better to split up wheel again in  W and Z, and DeltaWheel in DetaW and DeltaZ? Do you agree this is a good idea?

event/event.go
beoran: It's the number of the mouse button, much like the number of the gamepad button. Nowadays, some gaming mice have up to 15 mouse buttons, so not just 3 buttons.  Would you like this too to become an enum of MouseButton, with MouseButtonLeft=1, MouseButtonRight=2, MouseButtonMiddle=3,   MouseButtonExtra1 = 4 , etc  or such?

event/event.go
beoran: OK, will fix this.

examples/event/main.go
beoran: OK, I will remove this.

event/event.go
beoran: Or, perhaps I should copy ebiten/key.go?

AUTHORS
```diff
@@ -2,6 +2,7 @@
 Acid147 <norman.rauschen@web.de>
 Andrew Gerrand <nf@wh3rd.net>
 Ben Echols <lologarithm@gmail.com>
+Beoran <beoran@gmail.com>
```
hajimehoshi: It's OK if they are `.gitconfig`'s name and email :D

event/event.go
```diff
@@ -0,0 +1,211 @@
...
+// required of events yet.
+type Event interface {
+}
+
+// KeyCharacter is an event that occurs when a character is actually typed on
+// the keyboard. This may be provided by an input method.
+type KeyCharacter struct {
+	// Code of the key pressed or released.
+	Code int
+	// Key board modifiers.
```
hajimehoshi: `Modifiers is`...

event/event.go
hajimehoshi: NVM, I confirmed identifiers are represented as integers in GLFW and Web gamepad API.

event/event.go
hajimehoshi: https://github.com/hajimehoshi/ebiten/blob/master/internal/driver/keys.go I thought you could use `driver.Key` but unfortunately this is an internal package.

OK let's keep as int in this PR and leave TODO comment that we will define key enum and use it. I'll add key definitions to this event package later.

event/event.go
```diff
@@ -0,0 +1,211 @@
...
+// MouseAxes is a mouse axis event.
+type MouseAxis struct {
+	// X position of the mouse pointer.
+	X float32
+	// Y position of the mouse pointer.
+	Y float32
+	// W is the vertical position of the mouse wheel.
+	W float32
+	// Z is the horizontal position of the mouse wheel.
+	Z float32
```
hajimehoshi: The order should be XYZW? Do we use WZ btw?

event/event.go
```diff
@@ -0,0 +1,211 @@
...
+
+// GamepadDetach happens when a gamepad is detached.
+type GamepadDetach struct {
+	// ID represents which gamepad caused the event.
+	ID int
+}
+
+// MouseAxes is a mouse axis event.
+type MouseAxis struct {
+	// X position of the mouse pointer.
```
hajimehoshi: The comment needs to be fixed.

event/event.go
```diff
@@ -0,0 +1,211 @@
...
+// TouchBegin occurs when a touch begins.
+type TouchBegin struct {
+	// Touch ID that caused the touch event.
+	ID int
+	// X position of the event.
+	X float32
+	// Y position of the event.
+	Y float32
+	// Change in X since last touch event.
+	DeltaX float32
```
hajimehoshi: TouchBegin doesn't have to include DeltaX and DeltaY. Let's leave only necessary members.

event/event.go
hajimehoshi: Pressure sounds better :)

event/event.go
hajimehoshi: What about separating Wheel event and Button-Pressing event?

event/event.go
hajimehoshi: We should have enums. Please leave a TODO comment same as Keys :)

event/event.go
hajimehoshi: I replied to the comment about wheel.

event/event.go
beoran: OK!

event/event.go
beoran: OK!

event/event.go
beoran: Great idea! I will make a mouse wheel event with Horizontal, Vertical, DeltaHorizontal and DeltaVertical. Unless you thing X, Y, DeltaX and DeltaY is better?

event/event.go
beoran: OK, will do!

AUTHORS
beoran: You mean I don't need to do this? It's automati? If so, I should leave out this change then?

event/event.go
beoran: Indeed, I will do so. This does mean that TouchEnd will have the same data as TouchMoved, not as TouchBegin

AUTHORS
hajimehoshi: Basically I'lld do this change and you don't have to add :-)

event/event.go
```diff
@@ -0,0 +1,214 @@
...
+	Buttons int
+}
+
+// GamepadDetach happens when a gamepad is detached.
+type GamepadDetach struct {
+	// ID represents which gamepad caused the event.
+	ID int
+}
+
+// MouseAxes is a mouse axis event.
```
hajimehoshi: MouseAxis?

event/event.go
```diff
@@ -0,0 +1,214 @@
...
+	// DeltaY is the change in Y since the last mouse event.
+	DeltaY float32
+}
+
+// MouseWheel is a mouse wheel event.
+type MouseWheel struct {
+	// Vertical is the vertical position of the mouse wheel.
+	Vertical float32
+	// Horizontal is the horizontal position of the mouse wheel.
+	Horizontal float32
```
hajimehoshi: I prefer x and y instead of vertical and horizontal (we might add other axes in the future?).

event/event.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// TouchEnd occurs when a touch ends.
+// The data is the same as for a TouchMoved event.
+type TouchEnd TouchMoved
+
+// TouchCancel occurs when a touch is canceled.
+type TouchCancel struct {
+	// ID of the touch that caused the touch event.
+	ID int
```
hajimehoshi: What's the difference between TouchEnd and TouchCancel BTW?

event/event.go
```diff
@@ -0,0 +1,214 @@
...
+// Event is an interface that custom events should implement.
+// It is empty for now because there are no general methods
+// required of events yet.
+type Event interface {
+}
+
+// KeyCharacter is an event that occurs when a character is actually typed on
+// the keyboard. This may be provided by an input method.
+type KeyCharacter struct {
+	// Code of the key typed.
```
hajimehoshi: Let's make this sentence (Code is/represents ...)

event/event.go
hajimehoshi: Or MouseMove?

event/event.go
```diff
@@ -0,0 +1,214 @@
...
+// MouseButtonDown is a mouse button press event.
+type MouseButtonDown struct {
+	// X position of the mouse pointer.
+	X float32
+	// Y position of the mouse pointer.
+	Y float32
+	// DeltaX is the change in X since the last mouse event.
+	DeltaX float32
+	// DeltaY is the change in Y since the last mouse event.
+	DeltaY float32
```
hajimehoshi: Do we need DeltaX and DeltaY for MouseButtonDown/Up?

event/event.go
```diff
@@ -0,0 +1,214 @@
...
+type MouseButtonUp MouseButtonDown
+
+// MouseEnter occurs when the mouse enters the view window.
+type MouseEnter struct {
+	// X position of the mouse pointer.
+	X float32
+	// Y position of the mouse pointer.
+	Y float32
+	// DeltaX is the change in X since last mouse event.
+	DeltaX float32
```
hajimehoshi: Do we need DeltaX and DeltaY for MouseEnter and MouseLeave?

event/event.go
hajimehoshi: X, Y is better (I left a new comment).

event/event.go
```diff
@@ -0,0 +1,214 @@
...
+	// Position of the axis after the change.
+	// It varies between -1.0 and 1.0.
+	Position float32
+}
+
+// GamepadButtonDown is a gamepad button press event.
+type GamepadButtonDown struct {
+	// ID represents which gamepad caused the event.
+	ID int
+	// Button that was pressed, if any.
```
hajimehoshi: Let's make this sentence.

event/event.go
```diff
@@ -0,0 +1,214 @@
...
+	Button int
+	// Pressure applied to the gamepad button.
```
hajimehoshi: ditto

event/event.go
```diff
@@ -0,0 +1,214 @@
...
+type MouseButtonDown struct {
+	// X position of the mouse pointer.
+	X float32
+	// Y position of the mouse pointer.
+	Y float32
+	// DeltaX is the change in X since the last mouse event.
+	DeltaX float32
+	// DeltaY is the change in Y since the last mouse event.
+	DeltaY float32
+	// Button that was pressed.
```
hajimehoshi: Let's make this sentence.

event/event.go
```diff
@@ -0,0 +1,214 @@
...
+	// TODO: this should change later from an int to an enumeration type
+	Button int
+	// Pressure applied on the mouse click.
```
hajimehoshi: ditto.

event/event.go
beoran: Ok, but see later for MouseAxis -> MouseMove.

event/event.go
beoran: Ok

event/event.go
beoran: MouseMove sounds better, yes.

event/event.go
beoran: Ok, I will make it so.

event/event.go
beoran: It's convenient to have it, but if we also send a MouseMove on click, it is not strictly necessary. So I will drop these fields.

event/event.go
beoran: Again, this is not strictly needed if we also send a MouseMove event. So I will remove them.

event/event.go
beoran: On Android a touch can be cancelled. See http://stackoverflow.com/questions/11960861/what-causes-a-motionevent-action-cancel-in-android for some more details.  Similarly for iOS: https://developer.apple.com/documentation/uikit/uiresponder/1621116-touchescancelled . So let's keep this event.

event/event.go
beoran: Ok, the comments should be full sentences. I will make it so everywhere.

event/event.go
beoran: Ok

event/event.go
beoran: Ok

beoran:
Ok

AUTHORS
beoran: Ok, So I won't then. :)

event/event.go
beoran: I also renamed TouchMoved to TouchMove to make it consistent.

event/event.go
```diff
@@ -0,0 +1,208 @@
...
+// ViewSize occurs when the size of the application's view port changes.
+type ViewSize struct {
+	// Width is the actual, real width of the view.
+	Width float32
+	// Height is the actual, real height of the view.
+	Height float32
+	// X is the X position of the view on the physical screen.
+	X float32
+	// Y is the Y position of the view on the physical screen.
+	Y float32
```
hajimehoshi: * Use `int` instead of `float32`.
* Add explanation that the unit is device-independent pixel.
* X, Y are not needed so far (Ebiten does not care these values)

event/event.go
hajimehoshi: Ack.

event/event.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// TouchBegin occurs when a touch begins.
+type TouchBegin struct {
+	// ID identifies the touch that caused the touch event.
+	ID int
+	// X is the X position of the touch.
+	X float32
+	// Y is the Y position of the touch.
+	Y float32
```
hajimehoshi: * Use int?
* Let's add explanation that the unit is device-independent pixels.

event/event.go
```diff
@@ -0,0 +1,208 @@
...
+// MouseMove is a mouse movement event.
+type MouseMove struct {
+	// X is the X position of the mouse pointer.
+	X float32
+	// Y is the Y position of the mouse pointer.
+	Y float32
+	// DeltaX is the change in X since the last MouseMove event.
+	DeltaX float32
+	// DeltaY is the change in Y since the last MouseMove event.
+	DeltaY float32
```
hajimehoshi: * Use int? Or some devices usually use float32 precision?
* Let's add explanation that the unit is device-independent pixels.

event/event.go
beoran: Ok, for the screen, int is fine since this is in device independent pixels which are always a whole number for the screen. I will remove the X and Y. since they are not used yet by ebiten.

event/event.go
beoran: For X and Y measurements, like touch or mouse, the resolution of the underlying system is often much higher than those of the device independent pixels. For example, If I have a device independent resolution of 320x240 on a real full screen of 4 time the size, we can detect mouse movements of 0.25 device independent pixels. Rounding this to int gives "jerky" results and makes the input less precise. Due to rounding, the result could be 1 pixel off. The DeltaX and DeltaY will loose even more precision, up to 2pixels, due to the sampling theorem (https://en.wikipedia.org/wiki/Nyquist%E2%80%93Shannon_sampling_theorem).

Furthermore, on some platforms, such as Windows, it's possible to get high resolution measurements from the mouse. This is especially useful for pen tablets (e.g Wacom), which have such a high resolution. 

Therefore I would recommend that for all measurements this stays a float32, but as you say, expressed in device independent pixels.

event/event.go
beoran: Let's use float32 expressed as device independent pixels, for the reasons mentioned above.

event/event.go
```diff
@@ -0,0 +1,238 @@
...
+    DeltaX float32
+    // Deltay is the change in Y since last touch event.
+    // This value is expressed in device independent pixels.
+    DeltaY float32
+    // Pressure of applied touch.
+    // It varies between 0.0 for not pressed, and 1.0 for completely pressed.
+    Pressure float32
+    // Primary represents whether the touch event is the primary touch or not.
+    // If it is true, then it is a primary touch.
+    // If it is false then it is a secondary touch.
```
hajimehoshi: `Otherwise, it is false` is more precise (there can be third touch)

event/event.go
hajimehoshi: Thanks, acknowledged.

event/event.go
```diff
@@ -0,0 +1,238 @@
...
+    Primary bool
+}
+
+// TouchEnd occurs when a touch ends.
+// The data is the same as for a TouchMoved event.
```
hajimehoshi: `TouchMove` ?

event/event.go
```diff
@@ -0,0 +1,238 @@
...
+    // X is the X position of the mouse pointer.
+    // This value is expressed in device independent pixels.
+    X float32
+    // Y is the Y position of the mouse pointer.
+    // This value is expressed in device independent pixels.
+    Y float32
+}
+
+// MouseLeave occurs when the mouse leaves the view window.
+// The data is identical to MouseEnter
```
hajimehoshi: Add `.`

event/event.go
beoran: ok

event/event.go
beoran: indeed!

event/event.go
beoran: ok

beoran:
> Thank you! Please use gofmt for indentation.

I did, but then I replaced the tabs by spaces because I didn't realize before that gofmt indents with tabs. I learned something today. I will run it through gofmt again.

beoran:
Finally! Thanks for your corrections, you have a fine eye for detail.

hajimehoshi:
Thank you for your dedicated work!

I'll add key enums later. Please wait for a while. As I'm a little busy with other works, it might take 2 or 3 days, sorry.

Please let me know if you find what you want to do :-)

beoran:
Well, I think I will port some low level Gamepad drivers I wrote for Allegro to Go. Probably in a separate repository, though.

hajimehoshi:
We should use the Ebiten's current implementation (GLFW for desktops for example) to minimize the required tasks. My plan is, as I mentioned, to fix the current `driver.UI` definition and its implementations.

What do you think?

beoran:
Hmm, that is probably a good approach in the short term, although its has some problems. 

GLFW has only state-driven joystick input for now, the call back based one is in development since 2016. Allegro5 and SDL offer more features, but they might be too heavy weight. All are based on C and require a cgo wrapper (I wrote one a cgo for Allegro by the way). It's not an ideal situation. I know C very well, but I'd like to use more Go and less C, and less cgo as well.

But anyway, for the purpose of getting event driven input in Ebiten without having to doo too much work, fixing driver.UI is a good next step.

hajimehoshi:
That's a good point I did not realize. Indeed, GLFW's gamepad input is not event driven.

I agree that fixing the current UI is the first step and introducing a new gamepad infrastructure would be the second step. The next step would be a separate issue.

> All are based on C and require a cgo wrapper (I wrote one a cgo for Allegro by the way).

Note that on Windows Ebiten never uses Cgo for convenience :-)

PR #933, opened 2019-09-02
event: Add Key

hajimehoshi:
This also renames event names from `Key*` to `KeyboardKey*`.

Updates #926

@beoran , please take a look.


event/event.go
```diff
@@ -25,32 +25,30 @@ package event
...
 	Character rune
 }
 
-// KeyDown is an event that occurs when a key is pressed on the keyboard.
-type KeyDown struct {
-	// Code is the key code of the key pressed or released.
-	// TODO: this should change later from an int to an enumeration type.
-	Code int
+// KeyboardKeyDown is an event that occurs when a key is pressed on the keyboard.
+type KeyboardKeyDown struct {
```
beoran: The name KetboardKeyDown is longer, but consistent with names like MouseButtonDown, so I would say this is an improvement after all.

event/keys.go
```diff
@@ -0,0 +1,129 @@
+// Copyright 2013 The Ebiten Authors
```
beoran: Yes, this file also looks fine.

genkeys.go
```diff
@@ -306,6 +306,24 @@ const (
 )
 `
 
+const eventKeysTmpl = `{{.License}}
```
beoran: This generator code also looks OK.

hajimehoshi:
Thank you for your reviewing!

PR #935, opened 2019-09-10
event: Auto-generate event definitions

hajimehoshi:
As event package needs to depend on driver package, and drvier
package needs to emit events, we need event definitions in both
side. This change add a generator for event definitions.

Updates #926


hajimehoshi:
@beoran Could you take a look? Thanks!

event/event.go
```diff
@@ -40,46 +32,64 @@ type KeyboardKeyCharacter struct {
...
 type GamepadAxis struct {
 	// ID represents which gamepad caused the event.
 	ID int
+
 	// Axis is the axis of the game pad that changed position.
 	Axis int
-	// Position is the psoition of the axis after the change.
-	// It varies between -1.0 and 1.0.
+
+	// Position is the psoition of the axis after the change. It varies between -1.0 and 1.0.
```
beoran: psosition -> position

genevents.go
```diff
@@ -0,0 +1,367 @@
...
+	return strings.Split(e.Comment, " ")[0]
+}
+
+type Member struct {
+	Comment string
+	Type    string
+}
+
+func (m *Member) Name() string {
+	return strings.Split(m.Comment, " ")[0]
```
beoran: `return strings.SplitN(m.Comment, " ", 2)[0]`
will be more efficient.

genevents.go
```diff
@@ -0,0 +1,367 @@
...
+					Comment: "Height is the height of the view. This value is expressed in device independent pixels.",
+					Type:    "int",
+				},
+			},
+		},
+	}
+)
+
+const (
+	license = `// Copyright 2013 The Ebiten Authors
```
beoran: I think legally this should be 2013-2019 to cover the whole development range. But I am not a lawyer.

event/event.go
hajimehoshi: Done.

genevents.go
hajimehoshi: Good point. Done.

genevents.go
hajimehoshi: As far as I know, only one year when the file is created should be fine (e.g., Chromium does so). The problem is that auto-generated file cannot determine the year, then I adopted 2013 which Ebiten started.

hajimehoshi:
Thank you!

PR #952, opened 2019-10-06
Add new example: squirals

dbriemann:
Squirals is a funny little graphics experiment inspired by the xscreensaver `squirals`.


examples/squiral/main.go
```diff
@@ -0,0 +1,371 @@
...
+}
+
+func update(screen *ebiten.Image) error {
+	reset := false
+
+	if ebiten.IsKeyPressed(ebiten.KeyEscape) {
+		return quit
+	}
+
+	if ebiten.IsKeyPressed(ebiten.KeyB) {
```
hajimehoshi: Use `inpututil.IsKeyJustPressed`

examples/squiral/main.go
```diff
@@ -0,0 +1,371 @@
...
+	canvas.Fill(colornames.Black)
+
+	auto = &automaton{}
+	auto.init()
+}
+
+func update(screen *ebiten.Image) error {
+	reset := false
+
+	if ebiten.IsKeyPressed(ebiten.KeyEscape) {
```
hajimehoshi: As the window can be closed by `Alt+F4` on Windows and `Command + Q` on Mac, I don't think we need to add this.

examples/squiral/main.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+
+	if ebiten.IsKeyPressed(ebiten.KeyB) {
+		if background == colornames.White {
+			background = colornames.Black
+		} else {
+			background = colornames.White
+		}
+		reset = true
+	} else if ebiten.IsKeyPressed(ebiten.KeyT) {
```
hajimehoshi: ditto

examples/squiral/main.go
```diff
@@ -0,0 +1,371 @@
...
+		selectedPalette = (selectedPalette + 1) % len(palettes)
+		reset = true
+	}
+
+	if ebiten.IsMouseButtonPressed(ebiten.MouseButtonLeft) {
```
hajimehoshi: Use `inpututil.IsMouseButtonJustPressed`

examples/squiral/main.go
```diff
@@ -0,0 +1,371 @@
...
+	colors []color.Color
+}
+
+var (
+	quit = fmt.Errorf("quit")
+
+	background = colornames.Black
+
+	palettes = []palette{
+		palette{
```
hajimehoshi: Use `gofmt -s -w` and `palette{` will be `{`

examples/squiral/main.go
```diff
@@ -0,0 +1,371 @@
...
+		selectedPalette = (selectedPalette + 1) % len(palettes)
+		reset = true
+	}
+
+	if ebiten.IsMouseButtonPressed(ebiten.MouseButtonLeft) {
+		reset = true
+	}
+
+	if reset {
+		screen.Fill(background)
```
hajimehoshi: I don't think we need to fill `screen` since the screen will be rendered with `canvas` anytime.

examples/squiral/main.go
```diff
@@ -0,0 +1,371 @@
...
+import (
+	"fmt"
+	"image/color"
+	"math/rand"
+	"sync"
+	"time"
+
+	"github.com/hajimehoshi/ebiten"
+	"github.com/hajimehoshi/ebiten/ebitenutil"
+	"golang.org/x/image/colornames"
```
hajimehoshi: Use `color.Black` and `color.White` and remove the dependency on `golang.org/x/image/colornames`.

examples/squiral/main.go
```diff
@@ -0,0 +1,371 @@
...
+	amount    = width / 32
+)
+
+const (
+	east direction = iota
+	south
+	west
+	north
+
+	horizontal direction = iota
```
hajimehoshi: That sounds a little odd to share the same type here. We don't use `horizontal` and `vertical` anyway, right?

.gitignore
```diff
@@ -2,3 +2,4 @@
 *~
 *.aar
 *.apk
+.vscode
```
hajimehoshi: Is this intentional? If this is common, I'm fine.

examples/squiral/main.go
```diff
@@ -0,0 +1,371 @@
...
+	ebitenutil.DebugPrintAt(
+		screen,
+		"esc: quit",
+		1, 64,
+	)
+}
+
+func main() {
+	ebiten.SetMaxTPS(250)
+	if err := ebiten.Run(update, wwidth, wheight, 2, "Squirals"); err != nil {
```
hajimehoshi: Isn't this `Squirrels`? Or, is `Squirals` a common work for this kind of graphics?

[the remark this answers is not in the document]
dbriemann: I just copied this and didn't even realise.. I will adapt this to my scaling constant.

go.sum
```diff
@@ -1,4 +1,5 @@
 github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802/go.mod h1:IVnqGOEym/WlBOVXweHU+Q+/VP0lqqI8lqeDx9IjBqo=
+github.com/go-gl/glfw v0.0.0-20190409004039-e6da0acd62b1 h1:QbL/5oDUmRBzO9/Z7Seo6zf912W/a6Sr4Eu0G/3Jho0=
```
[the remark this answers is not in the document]
dbriemann: I think this happened when I tried the intellij ide .. will remove that.

examples/squiral/main.go
dbriemann: Usually I really prefer to quit demos with `esc` over `alt+f4` but it's not really important so I will remove it.

examples/squiral/main.go
dbriemann: I don't like this simplification but it's your repo so removed :)

Usually when reading a large slice, I don't see the type in-between and then need to scroll up.. I don't see the benefit of this.

examples/squiral/main.go
dbriemann: Ah yes I chose colors before randomly and that's why I used the package.. now I have predefined themes. Looks much better.

.gitignore
dbriemann: This should stay (imo). If you use vscode and you have configurations for the workspace (e.g. for modules) you need this folder.

examples/squiral/main.go
hajimehoshi: I understand your idea, but for example, there is no 'termination' on browsers or mobiles. Then, not having a way to terminate the app is more portable.

dbriemann: I see. That's a good argument! :)

examples/squiral/main.go
```diff
@@ -0,0 +1,352 @@
...
+
+	// east, south, west, north
+	dirs   = [4]tup{tup{1, 0}, tup{0, 1}, tup{-1, 0}, tup{0, -1}}
+	outers = [2][2]tup{
+		[2]tup{tup{0, 1}, tup{0, -1}}, // east, west
+		[2]tup{tup{1, 0}, tup{-1, 0}}, // south, north
+	}
+)
+
+type tup struct {
```
hajimehoshi: What about renaming this to `vec2`?

examples/squiral/main.go
```diff
@@ -0,0 +1,352 @@
...
+	blocker         = color.RGBA{0, 0, 0, 254}
+	free            = color.RGBA{0, 0, 0, 0}
+
+	rotors = [2][3]direction{
+		[3]direction{1, 0, 3},
+		[3]direction{3, 0, 1},
+	}
+
+	// east, south, west, north
+	dirs   = [4]tup{tup{1, 0}, tup{0, 1}, tup{-1, 0}, tup{0, -1}}
```
hajimehoshi: Let's define a method on `direction`

```
func (d direction) vec2() tup {
  switch d {
  case east:
    return tup{1,0}
  case south:
    return tup{0, 1}
  case west:
    return tup{-1,0}
  case north:
    return tup{0, -1}
  default:
    panic("not reached")
  }
}
```

examples/squiral/main.go
```diff
@@ -0,0 +1,352 @@
...
+				color.RGBA{0xDD, 0xE8, 0x0C, 0xFF}, //#DDE80C
+				color.RGBA{0x74, 0x9A, 0x0D, 0xFF}, //#749A0D
+			},
+		},
+	}
+	selectedPalette = 0
+	colorCycle      = 0
+	canvas          *ebiten.Image
+	auto            *automaton
+	once            sync.Once
```
hajimehoshi: Where is this `once` used?

examples/squiral/main.go
```diff
@@ -0,0 +1,352 @@
...
+	"github.com/hajimehoshi/ebiten/ebitenutil"
+	"github.com/hajimehoshi/ebiten/inpututil"
+)
+
+type direction int
+
+const (
+	width  = 800
+	height = 600
+	scale  = 2
```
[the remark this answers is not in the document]
dbriemann: Arr.. I actually changed it back by accident.. will fix.

examples/squiral/main.go
```diff
@@ -0,0 +1,352 @@
...
+	amount = width / 32
+)
+
+const (
+	east direction = iota
+	south
+	west
+	north
```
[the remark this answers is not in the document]
dbriemann: Oh oops.. thought only horizontal and vertical were not used.. 
In an old version I used them somehow but not needed anymore.

Need to check my linter settings.. wonder why it doesn't show me unused consts.

examples/squiral/main.go
```diff
@@ -0,0 +1,352 @@
...
+}
+
+var (
+	background = color.Black
+
+	palettes = []palette{
+		{
+			name: "sand dunes",
+			colors: []color.Color{
+				color.RGBA{0xF2, 0x74, 0x05, 0xFF}, //#F27405
```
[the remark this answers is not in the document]
dbriemann: Actually I would love to keep them because my editor highlights these with actual colors. See screenshot:

colors

examples/squiral/main.go
```diff
@@ -0,0 +1,352 @@
...
+	selectedPalette = 0
+	colorCycle      = 0
+	canvas          *ebiten.Image
+	auto            *automaton
+	once            sync.Once
+	blocker         = color.RGBA{0, 0, 0, 254}
+	free            = color.RGBA{0, 0, 0, 0}
+
+	rotors = [2][3]direction{
+		[3]direction{1, 0, 3},
```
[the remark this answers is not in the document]
dbriemann: These values are actually offsets to be used with the dirs slice... Let's just keep the numbers here and remove the consts.

examples/squiral/main.go
dbriemann: well ok :)

examples/squiral/main.go
dbriemann: I would prefer the array. It's faster and less code. I also removed the `direction` type because it does not have much use anymore.

examples/squiral/main.go
dbriemann: Something I forgot to remove. Good catch.

examples/squiral/main.go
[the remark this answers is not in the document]
hajimehoshi: Hmm, I tend not to want to leave comments only for specific editors...

examples/squiral/main.go
```diff
@@ -0,0 +1,341 @@
...
+		1, 32,
+	)
+	ebitenutil.DebugPrintAt(
+		screen,
+		fmt.Sprintf("[t]: cycle theme (current: %s)", palettes[selectedPalette].name),
+		1, 48,
+	)
+	ebitenutil.DebugPrintAt(
+		screen,
+		"[esc]: quit",
```
hajimehoshi: Remove this.

examples/squiral/main.go
```diff
@@ -0,0 +1,341 @@
...
+
+	return nil
+}
+
+func draw(screen *ebiten.Image) {
+	screen.DrawImage(canvas, nil)
+	ebitenutil.DebugPrintAt(
+		screen,
+		fmt.Sprintf("TPS: %0.2f, FPS: %0.2f", ebiten.CurrentTPS(), ebiten.CurrentFPS()),
+		1, 1,
```
hajimehoshi: 1, 0? or, other lines should be shifted by 1 in Y.

examples/squiral/main.go
```diff
@@ -0,0 +1,341 @@
...
+}
+
+func setpix(xy vec2, col color.Color) {
+	canvas.Set(xy.x, xy.y, col)
+	auto.colorMap[xy.x][xy.y] = col
+}
+
+func init() {
+	rand.Seed(time.Now().UnixNano())
+	c, err := ebiten.NewImage(width, height, ebiten.FilterNearest)
```
hajimehoshi: You can ignore the error form `NewImage`. This is always nil. This error type is just for backward compatibility :-)

Also please use `ebiten.FilterDefault`.

Then

```
canvas, _ = ebiten.NewImage(width, height, ebiten.FilterDefault)
```

is fine.

examples/squiral/main.go
```diff
@@ -0,0 +1,341 @@
...
+				s.dead = true
+				return
+			}
+		}
+	}
+
+	s.speed = rand.Intn(5) + 1
+	s.pos.x = rx
+	s.pos.y = ry
+	s.dir = rand.Intn(5)
```
hajimehoshi: Why is this 5? Not 4?

examples/squiral/main.go
```diff
@@ -0,0 +1,341 @@
...
+		},
+	}
+	selectedPalette = 0
+	colorCycle      = 0
+	canvas          *ebiten.Image
+	auto            *automaton
+	blocker         = color.RGBA{0, 0, 0, 254}
+	free            = color.RGBA{0, 0, 0, 0}
+
+	rotors = [2][3]int{
```
hajimehoshi: Does this mean something about clockwise and counterclockwise? Could you leave a comment?

examples/squiral/main.go
```diff
@@ -0,0 +1,341 @@
...
+		[3]int{1, 0, 3},
+		[3]int{3, 0, 1},
+	}
+
+	// east, south, west, north
```
hajimehoshi: Please follow a standard Go guideline. The comment should start with `dirs is` in this case.

examples/squiral/main.go
```diff
@@ -0,0 +1,341 @@
...
+	dirs   = [4]vec2{vec2{1, 0}, vec2{0, 1}, vec2{-1, 0}, vec2{0, -1}}
+	outers = [2][2]vec2{
```
hajimehoshi: Please leave a comment to explain this.

examples/squiral/main.go
```diff
@@ -0,0 +1,341 @@
...
+				color.RGBA{0xDD, 0xE8, 0x0C, 0xFF}, //#DDE80C
+				color.RGBA{0x74, 0x9A, 0x0D, 0xFF}, //#749A0D
+			},
+		},
+	}
+	selectedPalette = 0
+	colorCycle      = 0
+	canvas          *ebiten.Image
+	auto            *automaton
+	blocker         = color.RGBA{0, 0, 0, 254}
```
hajimehoshi: Isn't this 255, or is this intentional?

go.sum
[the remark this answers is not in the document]
hajimehoshi: ping

examples/squiral/main.go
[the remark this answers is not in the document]
dbriemann: There are plugins for nearly every existing editor to highlight colors in hex notation. What is so bad about these comments? It also enables copy&paste of the colors. Is there a go std function that takes a hex color to create a `color.Color`? I didn't find one when I implemented this.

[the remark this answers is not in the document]
hajimehoshi: > There are plugins for nearly every existing editor to highlight colors in hex notation.

VSCode, or other editors? I'm not sure how common this is.

> What is so bad about these comments?

The comments don't add any additional information.

>  Is there a go std function that takes a hex color to create a color.Color?

Unfortunately there is no such function. You can make this, but in this case, the function would be like `hexToColor(0xf27405)`, and this would not be what you wanted.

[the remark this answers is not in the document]
dbriemann: > > There are plugins for nearly every existing editor to highlight colors in hex notation.
> 
> VSCode, or other editors? I'm not sure how common this is.

VSCode and other editors. It is very common for web developers. 


> > What is so bad about these comments?
> 
> The comments don't add any additional information.

Well depends.. only if the editor supports it. I understand you would not want these comments in a library e.g. but in a demo I don't see the harm.

> > Is there a go std function that takes a hex color to create a color.Color?
> 
> Unfortunately there is no such function. You can make this, but in this case, the function would be like `hexToColor(0xf27405)`, and this would not be what you wanted.

The option would be a function that consumes strings of the format `#rrggbb` but I am reluctant to write it just to get rid of a few comments :)

examples/squiral/main.go
hajimehoshi: Please leave a comment explaining the intention.

PR #1016, opened 2019-12-12
Add cursor capture functionality

### LINKED ISSUE #1014: ui: Capturing cursors
https://www.glfw.org/docs/3.3/input_guide.html#cursor_mode

neuronpool:
Closes #1014 

The API is separate from the cursor visibility function for backwards compatibility. Combining the APIs into one function (SetCursorMode, or something similar) may be more elegant.

### DIFF
--- a/cursormodes.go
+++ b/cursormodes.go
@@ -16,13 +16,13 @@ package ebiten
 
 import "github.com/hajimehoshi/ebiten/internal/driver"
 
-// An InputCursorMode represents
+// A CursorModeType represents
 // a render and coordinate mode of a mouse cursor.
-type InputCursorMode int
+type CursorModeType int
 
 // Cursor Modes
 const (
-	CursorModeVisible  = InputCursorMode(driver.CursorModeVisible)
-	CursorModeHidden   = InputCursorMode(driver.CursorModeHidden)
-	CursorModeCaptured = InputCursorMode(driver.CursorModeCaptured)
+	CursorModeVisible  = CursorModeType(driver.CursorModeVisible)
+	CursorModeHidden   = CursorModeType(driver.CursorModeHidden)
+	CursorModeCaptured = CursorModeType(driver.CursorModeCaptured)
 )
--- a/cursormodes.go
+++ b/cursormodes.go
@@ -1,4 +1,4 @@
-// Copyright 209 The Ebiten Authors
+// Copyright 2019 The Ebiten Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/cursormodes.go
+++ b/cursormodes.go
@@ -1,0 +1,28 @@
+// Copyright 209 The Ebiten Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ebiten
+
+import "github.com/hajimehoshi/ebiten/internal/driver"
+
+// An InputCursorMode represents
+// a render and coordinate mode of a mouse cursor.
+type InputCursorMode int
+
+// Cursor Modes
+const (
+	CursorModeVisible  = InputCursorMode(driver.CursorModeVisible)
+	CursorModeHidden   = InputCursorMode(driver.CursorModeHidden)
+	CursorModeCaptured = InputCursorMode(driver.CursorModeCaptured)
+)
--- a/internal/uidriver/glfw/ui.go
+++ b/internal/uidriver/glfw/ui.go
@@ -491,13 +491,17 @@ func (u *UserInterface) CursorMode() driver.CursorMode {
 	if !u.isRunning() {
 		return u.getInitCursorMode()
 	}
-	v := driver.CursorModeVisible
+	var v driver.CursorMode
 	_ = u.t.Call(func() error {
 		switch u.window.GetInputMode(glfw.CursorMode) {
+		case glfw.CursorNormal:
+			v = driver.CursorModeVisible
 		case glfw.CursorHidden:
 			v = driver.CursorModeHidden
 		case glfw.CursorDisabled:
 			v = driver.CursorModeCaptured
+		default:
+			panic("Invalid cursor mode")
 		}
 		return nil
 	})
@@ -519,7 +523,7 @@ func (u *UserInterface) SetCursorMode(mode driver.CursorMode) {
 		case driver.CursorModeCaptured:
 			c = glfw.CursorDisabled
 		default:
-			return nil
+			panic("Invalid cursor mode")
 		}
 		u.window.SetInputMode(glfw.CursorMode, c)
 		return nil
--- a/internal/uidriver/glfw/ui.go
+++ b/internal/uidriver/glfw/ui.go
@@ -493,7 +493,8 @@ func (u *UserInterface) CursorMode() driver.CursorMode {
 	}
 	var v driver.CursorMode
 	_ = u.t.Call(func() error {
-		switch u.window.GetInputMode(glfw.CursorMode) {
+		mode := u.window.GetInputMode(glfw.CursorMode)
+		switch mode {
 		case glfw.CursorNormal:
 			v = driver.CursorModeVisible
 		case glfw.CursorHidden:
@@ -501,7 +502,7 @@ func (u *UserInterface) CursorMode() driver.CursorMode {
 		case glfw.CursorDisabled:
 			v = driver.CursorModeCaptured
 		default:
-			panic("Invalid cursor mode")
+			panic(fmt.Sprintf("invalid cursor mode: %d", mode))
 		}
 		return nil
 	})
@@ -514,7 +515,7 @@ func (u *UserInterface) SetCursorMode(mode driver.CursorMode) {
 		return
 	}
 	_ = u.t.Call(func() error {
-		c := glfw.NoAPI
+		var c int
 		switch mode {
 		case driver.CursorModeVisible:
 			c = glfw.CursorNormal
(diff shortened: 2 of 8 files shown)


run.go
```diff
@@ -285,6 +285,31 @@ func SetCursorVisibility(visible bool) {
 	SetCursorVisible(visible)
 }
 
+// IsCursorCaptured returns a boolean valued indicating whether
```
hajimehoshi: `IsCursorCaptured reports whether...`

run.go
```diff
@@ -285,6 +285,31 @@ func SetCursorVisibility(visible bool) {
...
+// IsCursorCaptured is concurrent-safe.
+func IsCursorCaptured() bool {
+	return uiDriver().IsCursorCaptured()
+}
+
+// SetCursorCaptured changes the capture state of the cursor.
+// The system cursor is invisible and locked to the game window when captured.
+// Setting captured to false sets the cursor to visible.
+//
+// SetCursorCaptured does nothing on mobile or JS.
```
hajimehoshi: s/JS/browsers/

since we have Wasm version :-)

internal/uidriver/glfw/ui.go
```diff
@@ -514,6 +540,21 @@ func (u *UserInterface) SetCursorVisible(visible bool) {
 	})
 }
 
+func (u *UserInterface) SetCursorCaptured(captured bool) {
+	if !u.isRunning() {
+		u.setInitCursorCaptured(captured)
+		return
+	}
+	_ = u.t.Call(func() error {
+		c := glfw.CursorNormal
```
hajimehoshi: Hmm, I'm not sure setting `normal` is fine when the cursor is hidden. If cursor visibility and capturing state is orthogonal, we would need to keep the states in boolean values and switch the GLFW state correctly. Wdyt?

run.go
```diff
@@ -285,6 +285,30 @@ func SetCursorVisibility(visible bool) {
 	SetCursorVisible(visible)
 }
 
+// IsCursorCaptured reports whether the cursor is captured or not.
+//
+// IsCursorCaptured always returns false on mobiles/JS/browsers.
```
hajimehoshi: `mobiles or browsers` is fine.

run.go
```diff
@@ -285,6 +285,30 @@ func SetCursorVisibility(visible bool) {
...
+// IsCursorCaptured is concurrent-safe.
+func IsCursorCaptured() bool {
+	return uiDriver().IsCursorCaptured()
+}
+
+// SetCursorCaptured changes the capture state of the cursor.
+// The system cursor is invisible and locked to the game window when captured.
+// Setting captured to false sets the cursor to visible.
+//
+// SetCursorCaptured does nothing on mobiles/JS/browsers.
```
hajimehoshi: ditto

internal/uidriver/glfw/ui.go
neuronpool: I was thinking that as well. A combined API for all cursor states would be better, but I didn't want to cause any breaking changes. 

I think I could just add a cursor state, so toggling cursor capture would revert to the previous state.

Let me think of the best way to do this and I'll post an update.

hajimehoshi: I agree we should not break the compatibility. Another idea is to deprecate `Is/SetCursorVisibile` and add a new function to take enum values. I'm fine with both (adding the new functions as you did with some tricks, or adding a brand new function).

Thank you for considering!

neuronpool: The cursor visibility API is now deprecated

neuronpool:
The cursor visibility API is not deprecated in favor of a cursor mode API.

We'll probably going to want to update the examples if some are using the cursor visibility functions.

cursormodes.go
```diff
@@ -0,0 +1,28 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ebiten
+
+import "github.com/hajimehoshi/ebiten/internal/driver"
+
+// An InputCursorMode represents
+// a render and coordinate mode of a mouse cursor.
+type InputCursorMode int
```
hajimehoshi: `CusorMode`?

internal/uidriver/glfw/ui.go
```diff
@@ -487,27 +487,39 @@ func (u *UserInterface) adjustPosition(x, y int) (int, int) {
...
-	v := false
+	v := driver.CursorModeVisible
 	_ = u.t.Call(func() error {
-		v = u.window.GetInputMode(glfw.CursorMode) == glfw.CursorNormal
+		switch u.window.GetInputMode(glfw.CursorMode) {
+		case glfw.CursorHidden:
+			v = driver.CursorModeHidden
+		case glfw.CursorDisabled:
+			v = driver.CursorModeCaptured
+		}
```
hajimehoshi:
```
default:
  panic("not reached")
```

internal/uidriver/glfw/ui.go
```diff
@@ -487,27 +487,39 @@ func (u *UserInterface) adjustPosition(x, y int) (int, int) {
...
+func (u *UserInterface) SetCursorMode(mode driver.CursorMode) {
 	if !u.isRunning() {
-		u.setInitCursorVisible(visible)
+		u.setInitCursorMode(mode)
 		return
 	}
 	_ = u.t.Call(func() error {
-		c := glfw.CursorNormal
-		if !visible {
+		c := glfw.NoAPI
```
hajimehoshi: `NoAPI` as a default value does not make sense. Let's panic when `mode` is an invalid value.

internal/uidriver/glfw/ui.go
```diff
@@ -487,27 +487,39 @@ func (u *UserInterface) adjustPosition(x, y int) (int, int) {
...
+		switch mode {
+		case driver.CursorModeVisible:
+			c = glfw.CursorNormal
+		case driver.CursorModeHidden:
 			c = glfw.CursorHidden
+		case driver.CursorModeCaptured:
+			c = glfw.CursorDisabled
+		default:
+			return nil
```
hajimehoshi: `panic("not reached")`

internal/uidriver/mobile/ui.go
```diff
@@ -383,11 +383,11 @@ func (u *UserInterface) adjustPosition(x, y int) (int, int) {
 	return int(float64(x)/s - ox/as), int(float64(y)/s - oy/as)
 }
 
-func (u *UserInterface) IsCursorVisible() bool {
-	return false
+func (u *UserInterface) CursorMode() driver.CursorMode {
+	return 0
```
hajimehoshi: I prefer returning `CursorModeHidden`

internal/uidriver/js/ui.go
```diff
@@ -123,12 +123,25 @@ func (u *UserInterface) adjustPosition(x, y int) (int, int) {
...
-func (u *UserInterface) SetCursorVisible(visible bool) {
+func (u *UserInterface) SetCursorMode(mode driver.CursorMode) {
+	var visible bool
+	switch mode {
+	case driver.CursorModeVisible:
+		visible = true
+	case driver.CursorModeHidden:
+		visible = false
+	default:
+		visible = canvas.Get("style").Get("cursor").String() != "none"
```
hajimehoshi: Just early `return` would work.

run.go
```diff
@@ -261,26 +261,46 @@ func ScreenScale() float64 {
...
-// SetCursorVisible does nothing on mobiles.
+// SetCursorMode does nothing on mobiles.
 //
-// SetCursorVisible is concurrent-safe.
+// SetCursorMode is concurrent-safe.
+func SetCursorMode(mode InputCursorMode) {
+	uiDriver().SetCursorMode(driver.CursorMode(mode))
+}
+
+// IsCursorVisible is deprecated as of 1.11.0-alpha.2. Use CursorMode instead.
```
hajimehoshi: `1.10.0-alpha` is fine

run.go
```diff
@@ -261,26 +261,46 @@ func ScreenScale() float64 {
...
+func IsCursorVisible() bool {
+	return CursorMode() == CursorModeVisible
+}
+
+// SetCursorVisible is deprecated as of 1.11.0-alpha.2. Use SetCursorMode instead.
```
hajimehoshi: ditto

cursormodes.go
neuronpool: I wasn't so sure, as the current function to get cursor mode is ebiten.CursorMode(). I didn't want to use ebiten.GetCursorMode() for consistency. Thoughts?

run.go
neuronpool: 1.10.0-alpha or 1.11.0-alpha?

cursormodes.go
hajimehoshi: Ha, good point. What about `CursorModeType` instead of `InputCusorMode`?

run.go
hajimehoshi: Ah sorry, I meant `1.11.0-alpha`.

(I might remove all `-alpha` suffix from 'as of' later, but let's keep the consistency so far.)

cursormodes.go
neuronpool: That sounds good to me!

internal/uidriver/glfw/ui.go
neuronpool: I have added a panic here. However, I have left the initial assignment of c := glfw.NoAPI. This is so I can guarantee the resolution of const to the correct type. If you have other suggestions, I can make changes.

internal/uidriver/glfw/ui.go
```diff
@@ -487,27 +487,43 @@ func (u *UserInterface) adjustPosition(x, y int) (int, int) {
...
+		c := glfw.NoAPI
+		switch mode {
+		case driver.CursorModeVisible:
+			c = glfw.CursorNormal
+		case driver.CursorModeHidden:
 			c = glfw.CursorHidden
+		case driver.CursorModeCaptured:
+			c = glfw.CursorDisabled
+		default:
+			panic("Invalid cursor mode")
```
hajimehoshi: Please start with a small case (for consistency).

`fmt.Sprintf("invalid cursor mode: %d", mode)` is better.

internal/uidriver/glfw/ui.go
hajimehoshi: `var c int` would make sense as `SetInputMode` takes an integer value.

neuronpool:
I made another revision and pushed, but it's not showing up here. Weird

hajimehoshi:
> I made another revision and pushed, but it's not showing up here. Weird

You mean https://github.com/neuronpool/ebiten/commit/786adfc98b0ef805a15673204c588e57ffbbb4be ? Weird. Let's wait and see...

neuronpool:
> 
> 
> > I made another revision and pushed, but it's not showing up here. Weird
> 
> You mean [neuronpool@786adfc](https://github.com/neuronpool/ebiten/commit/786adfc98b0ef805a15673204c588e57ffbbb4be) ? Weird. Let's wait and see...

It still isn't appearing. Maybe you need to apply a review first before the new commit shows?

hajimehoshi:
What about adding an empty commit? I'll squash commits before merging anyway

neuronpool:
> 
> 
> What about adding an empty commit? I'll squash commits before merging anyway

That seems to have worked ._.

Issue #1034, opened 2019-12-27
gamepad: Add GamepadGUID and GamepadName functions


silbinarywolf:
I want to add GamepadName so I can detect whether someone is using an Xbox Controller or a PS4 Controller and display more accurate controls for those devices.

PR #1035, opened 2019-12-27
gamepad: add GamepadGUID and GamepadName functions

silbinarywolf:
Here's my first attempt at #1034.

I'm not sure if I can make the string returning functions more robust or not, might be an issue constantly allocating these?

### DIFF
--- a/input.go
+++ b/input.go
@@ -92,6 +92,29 @@ func IsMouseButtonPressed(mouseButton MouseButton) bool {
 	return uiDriver().Input().IsMouseButtonPressed(driver.MouseButton(mouseButton))
 }
 
+// GamepadGUID returns a string with the uuid.
+//
+// GamepadGUID is concurrent-safe.
+//
+// GamepadGUID always returns an empty string on browsers and mobiles.
+func GamepadGUID(id int) string {
+	return uiDriver().Input().GamepadGUID(id)
+}
+
+// GamepadName returns a string with the name.
+// This function may vary in how it returns descriptions for the same device across platforms
+// for example the following drivers/platforms see a Xbox One controller as the following:
+// - Windows: "Xbox Controller"
+// - Chrome: "Xbox 360 Controller (XInput STANDARD GAMEPAD)"
+// - Firefox: "xinput"
+//
+// GamepadName is concurrent-safe.
+//
+// GamepadName always returns an empty string on mobiles.
+func GamepadName(id int) string {
+	return uiDriver().Input().GamepadName(id)
+}
+
 // GamepadIDs returns a slice indicating available gamepad IDs.
 //
 // GamepadIDs is concurrent-safe.
--- a/internal/driver/input.go
+++ b/internal/driver/input.go
@@ -16,6 +16,8 @@ package driver
 
 type Input interface {
 	CursorPosition() (x, y int)
+	GamepadGUID(id int) string
+	GamepadName(id int) string
 	GamepadAxis(id int, axis int) float64
 	GamepadAxisNum(id int) int
 	GamepadButtonNum(id int) int
--- a/internal/glfw/glfw_notwindows.go
+++ b/internal/glfw/glfw_notwindows.go
@@ -235,6 +235,14 @@ func CreateWindow(width, height int, title string, monitor *Monitor, share *Wind
 	return theWindows.add(w), nil
 }
 
+func (j Joystick) GetGUID() string {
+	return glfw.Joystick(j).GetGUID()
+}
+
+func (j Joystick) GetName() string {
+	return glfw.Joystick(j).GetName()
+}
+
 func (j Joystick) GetAxes() []float32 {
 	return glfw.Joystick(j).GetAxes()
 }
--- a/internal/glfw/glfw_windows.go
+++ b/internal/glfw/glfw_windows.go
@@ -311,6 +311,40 @@ func CreateWindow(width, height int, title string, monitor *Monitor, share *Wind
 	return theGLFWWindows.add(w), nil
 }
 
+func (j Joystick) GetGUID() string {
+	ptr := glfwDLL.call("glfwGetJoystickGUID", uintptr(j))
+	panicError()
+	var backed [256]byte
+	as := backed[:0]
+	for i := int32(0); ; i++ {
+		b := *(*byte)(unsafe.Pointer(ptr))
+		ptr += unsafe.Sizeof(byte(0))
+		if b == 0 {
+			break
+		}
+		as = append(as, b)
+	}
+	r := string(as)
+	return r
+}
+
+func (j Joystick) GetName() string {
+	ptr := glfwDLL.call("glfwGetJoystickName", uintptr(j))
+	panicError()
+	var backed [256]byte
+	as := backed[:0]
+	for i := int32(0); ; i++ {
+		b := *(*byte)(unsafe.Pointer(ptr))
+		ptr += unsafe.Sizeof(byte(0))
+		if b == 0 {
+			break
+		}
+		as = append(as, b)
+	}
+	r := string(as)
+	return r
+}
+
 func (j Joystick) GetAxes() []float32 {
 	var l int32
 	ptr := glfwDLL.call("glfwGetJoystickAxes", uintptr(j), uintptr(unsafe.Pointer(&l)))
--- a/internal/uidriver/glfw/input.go
+++ b/internal/uidriver/glfw/input.go
@@ -29,6 +29,8 @@ import (
 
 type gamePad struct {
 	valid         bool
+	guid          string
+	name          string
 	axisNum       int
 	axes          [16]float64
 	buttonNum     int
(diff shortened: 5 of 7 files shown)


input.go
```diff
@@ -92,6 +92,29 @@ func IsMouseButtonPressed(mouseButton MouseButton) bool {
 	return uiDriver().Input().IsMouseButtonPressed(driver.MouseButton(mouseButton))
 }
 
+// GamepadGUID returns a string with the uuid.
+//
+// GamepadGUID is concurrent-safe.
+//
+// GamepadGUID always returns an empty string on browsers and mobiles.
+func GamepadGUID(id int) string {
+	return uiDriver().Input().GamepadGUID(id)
```
hajimehoshi: https://developer.mozilla.org/en-US/docs/Web/API/Gamepad/id should be available on browsers.

input.go
```diff
@@ -92,6 +92,29 @@ func IsMouseButtonPressed(mouseButton MouseButton) bool {
...
+// This function may vary in how it returns descriptions for the same device across platforms
+// for example the following drivers/platforms see a Xbox One controller as the following:
+// - Windows: "Xbox Controller"
+// - Chrome: "Xbox 360 Controller (XInput STANDARD GAMEPAD)"
+// - Firefox: "xinput"
+//
+// GamepadName is concurrent-safe.
+//
+// GamepadName always returns an empty string on mobiles.
+func GamepadName(id int) string {
```
hajimehoshi: Comment that `GamepadName is concurrent-safe`.

input.go
```diff
@@ -92,6 +92,29 @@ func IsMouseButtonPressed(mouseButton MouseButton) bool {
 	return uiDriver().Input().IsMouseButtonPressed(driver.MouseButton(mouseButton))
 }
 
+// GamepadGUID returns a string with the uuid.
+//
+// GamepadGUID is concurrent-safe.
+//
+// GamepadGUID always returns an empty string on browsers and mobiles.
```
hajimehoshi: Comment that `GamepadGUID is concurrent-safe.`

hajimehoshi:
OK now I've updated .travis.yml so that this PR is tested correctly. Could you rebase or merge your PR?

input.go
silbinarywolf: ID is more analogous to glfw's GamepadName than GamepadGUID, so if you look at this PR, I'm actually already using ID.

`ebiten\internal\uidriver\js\input.go`
```
i.gamepads[id].name = gamepad.Get("id").String()
```

input.go
silbinarywolf: It does have that comment?

input.go
silbinarywolf: It does have that comment?

silbinarywolf:
Have done so. We'll see how this goes :)

input.go
hajimehoshi: Add a comment `GamepadGUID returns an empty string before the main loop starts`

input.go
```diff
@@ -92,6 +92,29 @@ func IsMouseButtonPressed(mouseButton MouseButton) bool {
...
+// GamepadName returns a string with the name.
+// This function may vary in how it returns descriptions for the same device across platforms
+// for example the following drivers/platforms see a Xbox One controller as the following:
+// - Windows: "Xbox Controller"
+// - Chrome: "Xbox 360 Controller (XInput STANDARD GAMEPAD)"
+// - Firefox: "xinput"
+//
+// GamepadName is concurrent-safe.
+//
+// GamepadName always returns an empty string on mobiles.
```
hajimehoshi: Add a comment `GamepadName returns an empty string before the main loop starts`

input.go
hajimehoshi: Ah ok so there is no way to get GUID on browsers...

input.go
```diff
@@ -92,6 +92,29 @@ func IsMouseButtonPressed(mouseButton MouseButton) bool {
...
+func GamepadGUID(id int) string {
+	return uiDriver().Input().GamepadGUID(id)
+}
+
+// GamepadName returns a string with the name.
+// This function may vary in how it returns descriptions for the same device across platforms
+// for example the following drivers/platforms see a Xbox One controller as the following:
+// - Windows: "Xbox Controller"
+// - Chrome: "Xbox 360 Controller (XInput STANDARD GAMEPAD)"
+// - Firefox: "xinput"
```
hajimehoshi: In the first place, what is the purpose to get the name of the game pad? It looks like the name is not portable, then this looks not useful to detect the gamepad type.

silbinarywolf: I want the in-game instructions with game controls to display A,B,X,Y if we detect that its a Xbox or Xbox 360 Controller. Likewise if we detect that a PS4 controller is in, then we want the controls to reflect that too.

If we don't fuzzily detect either of those controller types, then we fallback to displaying buttons as the less intuitive "Button 0", "Button 1", etc.

hajimehoshi: As there are various possibilities of the names, I was wondering how you could detect the gamepad is Xbox's or not.

input.go
```diff
@@ -92,6 +92,29 @@ func IsMouseButtonPressed(mouseButton MouseButton) bool {
 	return uiDriver().Input().IsMouseButtonPressed(driver.MouseButton(mouseButton))
 }
 
+// GamepadGUID returns a string with the uuid.
```
hajimehoshi: Question: How do we use GUID of the gamepads? This does not work on browsers anyway.

silbinarywolf: I mostly implemented this because I wasn't sure what glfw methods would give me the information I needed. I don't actually utilize this anywhere.

However, its not crazy to think I might want to leverage this in the future for complex controller mapping.
ie. Users can setup certain profiles for certain controllers, the GUID could be used to differentiate different controller types.

input.go
silbinarywolf: I've put a code snippet of what I'm going to do for now. I'm not claiming this would be the most robust / error-free way to do it but I have no better ideas. I'm yet to test with a PS3/PS4 controller so I haven't added that logic in yet.

Worth noting that I also check for what input was last pressed (keyboard/gamepad) and I swap the buttons shown on-screen accordingly.

Finally, here's a resource I'll be referring to in the future to make this check cover more cases:
https://github.com/sriharshachilakapati/GLFW-Controller-Database

```
const deviceId = 0

type GamepadShape int32

const (
	GamepadShapeStandard    GamepadShape = 0
	GamepadShapeXbox        GamepadShape = 1
	GamepadShapePlaystation GamepadShape = 2
)

func updateLoop() {
	// Detect controller
	gamepadShape = GamepadShapeStandard
	description := gml.GamepadGetDescription(deviceId)
	if description != "" {
		if strings.HasPrefix(description, "Xbox Controller") ||
			strings.HasPrefix(description, "Xbox 360 Controller") ||
			strings.HasPrefix(description, "xinput") {
			gamepadShape = GamepadShapeXbox
		}
	}
}
```

input.go
hajimehoshi: > However, its not crazy to think I might want to leverage this in the future for complex controller mapping.

Then, how about adding this when we think this is necessary?

silbinarywolf: Sure, I'm happy to take this out of the PR. Want it omitted completely or just commented out? My thinking being the latter having less friction to bring back in later if someone wants or needs to.

silbinarywolf: Hm actually, upon further investigation, it looks like this GUID value could be useful in helping me utilize an extensive / existing open-source gamepad mapping file:
https://github.com/gabomdq/SDL_GameControllerDB/blob/master/gamecontrollerdb.txt

hajimehoshi: Well I understood GUID works, but you'd need to use `GamepadName` for potability in any cases. I prefer GUID in general but using names is inevitable so far.

silbinarywolf: In my opinion, since GamepadGUID is more robust, ideally you want to use that *if* you have it and then fallback to using GamepadName if its unavailable / empty string.

hajimehoshi: OK that makes sense. I hope Gamepad API will have a function for GUID in the future...

input.go
hajimehoshi: I was wondering if there is a good list including browsers.

hajimehoshi: If there is, I'd want to add a link in the comment. What do you think?

hajimehoshi: ping

silbinarywolf: I'm not aware of anything comprehensive!
I just inlined what I saw with some of my manual testing with the controllers I have.

Sorry for the delayed reply :)

hajimehoshi: Acknowledged.

hajimehoshi:
Please rebase this PR onto the current master.

silbinarywolf:
Done :)

hajimehoshi:
I read Chromium code and found that the gampad string is determined from USB devices' properties:

https://source.chromium.org/chromium/chromium/src/+/master:device/gamepad/gamepad_device_linux.cc;drc=0761512df18f31aaffee9b7c10ec19e7a974afe8;bpv=0;bpt=1;l=406-440
https://source.chromium.org/chromium/chromium/src/+/master:device/gamepad/gamepad_data_fetcher.cc;drc=0761512df18f31aaffee9b7c10ec19e7a974afe8;bpv=0;bpt=1;l=91-96

WebUSB API might solve the issue we can't get GUIDs, but I am not sure.

hajimehoshi:
I read the implementation of GLFW, and it looks like GUID is generated based on USB properties or DirectInput results:

https://github.com/glfw/glfw/blob/51bb76c7c3de934b2dd3affcdc297a32e4817835/src/win32_joystick.c#L452-L467
https://github.com/glfw/glfw/blob/781fbbadb0bccc749058177b1385c82da9ace880/src/cocoa_joystick.m#L169-L183

The GUID way just followed SDL's way. I'm now convinced that browsers would not have APIs for GUID. I was wondering if we could get USB properties directly so that we can determine vendors or device IDs.

hajimehoshi:
Let's continue to discuss at https://github.com/hajimehoshi/ebiten/issues/1048

PR #1045, opened 2020-01-05
Fix initial world randomization

dericmiller:
Because golang executes the init function after establishing
the variables in the var block, the random number seed wasn't
initializing until after the initial world state had gotten
established (leading to an identical game of Life on every run).
To fix this, we establish an empty world in the var block,
and then populate it in the init function after the random
number generator has been seeded.

### DIFF
--- a/examples/life/main.go
+++ b/examples/life/main.go
@@ -157,7 +157,7 @@ const (
 )
 
 var (
-	world  = NewWorld(screenWidth, screenHeight, 0)
+	world  *World
 	pixels = make([]byte, screenWidth*screenHeight*4)
 )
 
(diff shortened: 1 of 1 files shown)


examples/life/main.go
```diff
@@ -156,7 +157,7 @@ const (
 )
 
 var (
-	world  = NewWorld(screenWidth, screenHeight, int((screenWidth*screenHeight)/10))
+	world  = NewWorld(screenWidth, screenHeight, 0)
```
hajimehoshi: I prefer `var world *World` without initializing.

dericmiller: Concur.

Issue #1048, opened 2020-01-09
Is there a general API to get gamepad identifers?: Gamepad GUID is not general


hajimehoshi:
Discussion from https://github.com/hajimehoshi/ebiten/pull/1035

I read the implementation of GLFW, and it looks like GUID is generated based on USB properties or DirectInput results:

https://github.com/glfw/glfw/blob/51bb76c7c3de934b2dd3affcdc297a32e4817835/src/win32_joystick.c#L452-L467
https://github.com/glfw/glfw/blob/781fbbadb0bccc749058177b1385c82da9ace880/src/cocoa_joystick.m#L169-L183

The GUID way just followed SDL's way. I'm now convinced that browsers would not have APIs for GUID. I was wondering if we could get USB properties directly so that we can determine vendors or device IDs. If so, we should replace the current GUID API with this.

hajimehoshi:
CC @silbinarywolf

hajimehoshi:
@beoran might be interested in this

hajimehoshi:
Suggestion: I'd like to rename `GamepadGUID` to `GamepadSDLID` to make the situation clear. @silbinarywolf What do you think?

beoran:
I have looked into the matter but basically,   there are so many USB game controllers, of which some have the same if but still a different physical form. And the support for getting a USB or blootooth I'd from a device is highly platform dependent.

So, all in all, I consider that it is a rather hopeless matter to try to automatically map a physical input device to a virtual controller. In stead, your application should provide a game pad configuration dialog that allows the user to configure this manually, and to calibrate the joysticks on the device.

hajimehoshi:
Thanks. I understood there is not a good way to get identifiers of gamepads. Detecting Xbox or PS or Nintendo Switch and offering a suitable UI for configuration would be helpful, but I agree we should also offer configuration UI working for any gamepads.

Emulating the SDL's way would be the most feasible way apparently. This does not guarantee detecting gamepads 100%, but there is a good community database to determine gamepads from IDs.

hajimehoshi:
I'll mark this wontfix. Let's revisit this when we find a nice solution.

PR #1049, opened 2020-01-10
input: Rename GamepadGUID -> GamepadSDLID

hajimehoshi:
Gamepad GUID is a SDL specific notion and, strictly speaking,
they are not GUID (UUID) since they don't follow UUID's
specifications.

Renaming the function makes the situation clearer.

Updates #1048


hajimehoshi:
@silbinarywolf Please take a look.

hajimehoshi:
I plan to merge this tonight if there is no reply :-)

silbinarywolf:
All good! Go ahead!

PR #1050, opened 2020-01-10
Fix delayed DrawImage()

hiroebe:
`resolvePendingPixels` must be called in `delayedCommands`.
Otherwise, following `init` function does not work:

```go
func init() {
    srcImg, _ := ebiten.NewImage(w, h, ebiten.FilterDefault)
    dstImg, _ := ebiten.NewImage(w, h, ebiten.FilterDefault)

    // Set some pixels
    srcImg.Set(16, 16, clr)

    // Previous Set() is ignored, and nothing is drawn on dstImg.
    dstImg.DrawImage(srcImg, nil)
}
```

### DIFF
--- a/internal/buffered/image.go
+++ b/internal/buffered/image.go
@@ -222,6 +222,8 @@ func (i *Image) DrawImage(src *Image, bounds image.Rectangle, a, b, c, d, tx, ty
 	delayedCommandsM.Lock()
 	if needsToDelayCommands {
 		delayedCommands = append(delayedCommands, func() {
+			src.resolvePendingPixels(true)
+			i.resolvePendingPixels(false)
 			i.img.DrawImage(src.img, bounds, g, colorm, mode, filter)
 		})
 		delayedCommandsM.Unlock()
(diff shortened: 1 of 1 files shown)


hajimehoshi:
Thanks! Would it be possible to add tests at `image_test.go`?

hiroebe:
OK, I will.

hajimehoshi:
I realized that the test should be done before the main loop, then this might be tricky. I'll do if you are not sure.

hiroebe:
I think I understand the trickiness, but adding tests to `image_test.go` caused panic in glfw and I'm struggling with it. So could you please add test instead of me?

internal/buffered/image.go
```diff
@@ -222,6 +222,8 @@ func (i *Image) DrawImage(src *Image, bounds image.Rectangle, a, b, c, d, tx, ty
 	delayedCommandsM.Lock()
 	if needsToDelayCommands {
 		delayedCommands = append(delayedCommands, func() {
+			src.resolvePendingPixels(true)
+			i.resolvePendingPixels(false)
```
hajimehoshi: Could you add the same thing in `DrawTriangles`?

hiroebe: Done.

PR #1051, opened 2020-01-10
Remove duplicate check for io.Seeker

hiroebe:


hajimehoshi:
What is duplicated with this?

hiroebe:
Type assertion at L528-530 is duplicate, because the same assertion is done at L538-541.

https://github.com/hajimehoshi/ebiten/blob/dcad3d21c9f4ab8f158e7a8a6996ff192694c1e7/audio/audio.go#L527-L541

PR #1058, opened 2020-01-14
Add IsForeground API to glfw driver

### LINKED ISSUE #1037: ui: Add IsForeground
I would like to be able to detect when the game window loses focus.

Possible ways:
 - Allow the registration of hooks for focus loss and focus gain.
 - Allow the polling of focus state.

Use cases:
 - Bringing up a "pause" menu, so the user has time to prepare before resuming play.
 - Pausing concurrent operations which must run on a separate timing than `Update()`.

zacharyburkett:
Partially closes #1037  
Proper implementations for web and mobile still required

### DIFF
--- a/internal/uidriver/glfw/ui.go
+++ b/internal/uidriver/glfw/ui.go
@@ -51,6 +51,7 @@ type UserInterface struct {
 	origPosY             int
 	runnableInBackground bool
 	vsync                bool
+	foreground           bool
 
 	lastDeviceScaleFactor float64
 
@@ -252,6 +253,19 @@ func (u *UserInterface) setInitWindowDecorated(decorated bool) {
 	u.m.Unlock()
 }
 
+func (u *UserInterface) isForeground() bool {
+	u.m.RLock()
+	v := u.foreground
+	u.m.RUnlock()
+	return v
+}
+
+func (u *UserInterface) setIsForeground(foreground bool) {
+	u.m.Lock()
+	u.foreground = foreground
+	u.m.Unlock()
+}
+
 func (u *UserInterface) isRunnableInBackground() bool {
 	u.m.RLock()
 	v := u.runnableInBackground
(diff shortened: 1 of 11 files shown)


internal/uidriver/js/ui.go
```diff
@@ -76,6 +76,10 @@ func (u *UserInterface) IsFullscreen() bool {
 	return false
 }
 
+func (u *UserInterface) IsForeground() bool {
+	return true
```
hajimehoshi: Please leave a TODO comment.

internal/uidriver/mobile/ui.go
```diff
@@ -369,6 +369,10 @@ func (u *UserInterface) SetFullscreen(fullscreen bool) {
 	// Do nothing
 }
 
+func (u *UserInterface) IsForeground() {
```
hajimehoshi: ditto

internal/uidriver/glfw/ui.go
```diff
@@ -580,6 +598,10 @@ func (u *UserInterface) createWindow() error {
 		u.reqHeight = height
 	})
 
+	u.window.SetFocusCallback(func(_ *glfw.Window, focused bool) {
+		u.setIsForeground(focused)
+	})
```
hajimehoshi: Do we need to use `SetFocusCallback`? 

I think checking `u.window.GetAttrib(glfw.Focused)` should work. See `update` function.

internal/uidriver/js/ui.go
hajimehoshi: `TODO: Implement this` is clearer :-)

internal/uidriver/glfw/ui.go
```diff
@@ -402,6 +402,18 @@ func (u *UserInterface) SetFullscreen(fullscreen bool) {
 	u.setWindowSize(w, h, fullscreen, u.vsync)
 }
 
+func (u *UserInterface) IsForeground() (foreground bool) {
```
hajimehoshi: I prefer not using a named returning variable here. When the game does not start, let's return false.

```go
func (u *UserInterface) IsForeground() bool {
    if !u.isRunning() {
        return false
    }
    var foreground bool
    _ = u.t.Call(func() error {
        foreground = u.window.GetAttrib(glfw.Focused) == glfw.True
        return nil
    })
    return foreground
}
```

PR #1085, opened 2020-02-20
Reduce heap allocations in thread.go and input.go

corfe83:
I've been doing some profiling of a very simple ebiten project, and noticed that thread.go was doing a bunch of unnecessary allocations to accomplish its work. This change seeks to reduce GC work.

Input.go was also doing some unnecessary allocations.

The thread.go change reduces the total number of allocations per frame from 1342 to 852 (~36% reduction). The input.go change reduces it further to 752 (~44% total reduction). Perf tests were done on windows.


corfe83:
> Nice catch, LGTM!
> 
> nits: Please gofmt.

Oops, thanks, go fmt change pushed.

input.go
```diff
@@ -52,13 +52,13 @@ func IsKeyPressed(key Key) bool {
...
+		keys = []driver.Key{driver.KeyLeftAlt, driver.KeyRightAlt}
 	case KeyControl:
-		keys = append(keys, driver.KeyLeftControl, driver.KeyRightControl)
+		keys = []driver.Key{driver.KeyLeftControl, driver.KeyRightControl}
 	case KeyShift:
-		keys = append(keys, driver.KeyLeftShift, driver.KeyRightShift)
+		keys = []driver.Key{driver.KeyLeftShift, driver.KeyRightShift}
 	default:
-		keys = append(keys, driver.Key(key))
+		keys = []driver.Key{driver.Key(key)}
```
hajimehoshi: Just out of curiosity, I was wondering why this reduced heap allocation. What was the difference? @corfe83

hajimehoshi: NVM: From https://go.godbolt.org/z/3gbhBz and https://go.godbolt.org/z/9w2ptQ, slice values are on the stack.

PR #1088, opened 2020-02-21
Fix missing js build tag

tomlister:
Added js to build tags as suggested by @hajimehoshi


tomlister:
Awesome, looking forward to contributing in the future.

PR #1089, opened 2020-02-21
Fix more unnecessary allocations per frame

corfe83:
This simple change brings my simple test project from 752 allocations per frame to 474 allocations per frame. It seems a shame that go's escape analysis is not smart enough to leave this variable on the stack.

GeoM is a 24-byte struct and there is a slight perf difference that we are storing it in stack, but also copying it around with this change (instead of an 8-byte pointer). This could make things faster (due to stack / CPU cache) or slower (due to copying more memory) - when I try a stress test (drawing 100K images per frame), I can't see any actual performance difference (but I do see 100K fewer allocations, and GC is no longer running almost all the time).

### DIFF
--- a/internal/buffered/image.go
+++ b/internal/buffered/image.go
@@ -214,7 +214,7 @@ func (i *Image) DrawImage(src *Image, bounds image.Rectangle, a, b, c, d, tx, ty
 		panic("buffered: Image.DrawImage: src must be different from the receiver")
 	}
 
-	g := &mipmap.GeoM{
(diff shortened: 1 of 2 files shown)


hajimehoshi:
Thank you for your work!

corfe83:
> AFAIK, copying structs is heavy on GopherJS and Ebiten used to avoid such copying. However, as we already have WebAssembly and this works well, I'm fine to accept this change.

Thanks, that is helpful to know!

PR #1092, opened 2020-02-23
Xbox360 gamepad bindings

tracer8086:
I've added convenient bindings to access all the axes and buttons of XBOX 360 Gamepad (the most common PC gamepad) separately.


hajimehoshi:
Sorry but I'd want to avoid adding enums for a specific gamepad in Ebiten.

Does your example work on other OSes than Windows or browsers correctly? AFAIK, the button assignments vary and it is almost impossible to expect which key is assigned to which the gamepad's key.

PR #1094, opened 2020-02-25
Reduce allocs further by passing things directly to UI thread instead of through heap

corfe83:
Some of our commonly called function calls to UI thread still require transferring parameters and return values via heap. This fixes several frequently called functions to pass them directly instead of through heap. The code in thread.go is intended to be extendable.

This change brings allocs per frame in my simple project from 474 to 250.

I'm probably going to stop focusing on reducing allocs after this change, because the others feel like golang compiler bugs where escape analysis should be smarter.

### DIFF
--- a/internal/thread/thread.go
+++ b/internal/thread/thread.go
@@ -28,18 +28,19 @@ const (
 )
 
 const MaxPublicParams = 2
+
 type call struct {
-	funcType functionType
+	funcType          functionType
 	func0ReturnsError func() error
-	func1ReturnsBool func(uintptr) bool
-	func2ReturnsBool func(uintptr, uintptr) bool
-	funcReturns2 func() (uintptr, uintptr)
-	params [MaxPublicParams]uintptr
+	func1ReturnsBool  func(uintptr) bool
+	func2ReturnsBool  func(uintptr, uintptr) bool
+	funcReturns2      func() (uintptr, uintptr)
+	params            [MaxPublicParams]uintptr
 }
 
 type result struct {
-	err error
-	flag bool
+	err     error
+	flag    bool
 	result1 uintptr
 	result2 uintptr
 }
@@ -98,7 +99,7 @@ loop:
 // Call panics when Loop already ends.
 func (t *Thread) Call(f func() error) error {
 	thisCall := call{
-		funcType: type0ParamsReturnError,
+		funcType:          type0ParamsReturnError,
 		func0ReturnsError: f,
 	}
 	select {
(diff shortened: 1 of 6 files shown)


hajimehoshi:
Thank you!

Please address on the bot failures. It looks like `go vet` failed.

hajimehoshi:
Apparently this PR includes another person's changes. Is this intended?

internal/thread/thread.go
```diff
@@ -18,10 +18,37 @@ import (
...
+type functionType int
+
+const (
+	type0ParamsReturnError functionType = iota
+	type1ParamReturnBool
+	type2ParamsReturnBool
+	typeReturn2
+)
+
+const MaxPublicParams = 2
```
hajimehoshi: This doesn't have to be exposed.

corfe83:
> Apparently this PR includes another person's changes. Is this intended?

My apologies, these are personal changes, I did not intend to make any of the changes from my work account >_<. I'm learning go and decided I might as well contribute to help learn.

That person is the same as me, but I've force pushed the branch to overwrite that identity for privacy reasons.

I'm looking into the go vet issues now.

internal/uidriver/glfw/ui.go
```diff
@@ -670,11 +670,10 @@ func (u *UserInterface) run(context driver.UIContext) error {
 }
 
 func (u *UserInterface) updateSize(context driver.UIContext) {
-	var w, h int
-	_ = u.t.Call(func() error {
-		w, h = u.windowWidth, u.windowHeight
-		return nil
+	r1, r2 := u.t.CallReturn2(func() (uintptr, uintptr) {
```
hajimehoshi: If this call matters in terms of heap allocation, what about using sync.Mutex or atomic.Value?

internal/uidriver/glfw/ui.go
```diff
@@ -756,11 +755,11 @@ func (u *UserInterface) update(context driver.UIContext) error {
 	}
 
 	// Update the screen size when the window is resizable.
-	var w, h int
-	_ = u.t.Call(func() error {
-		w, h = u.reqWidth, u.reqHeight
-		return nil
+	r1, r2 := u.t.CallReturn2(func() (uintptr, uintptr) {
```
hajimehoshi: ditto

internal/uidriver/glfw/input.go
```diff
@@ -243,19 +244,18 @@ func (i *Input) IsKeyPressed(key driver.Key) bool {
 	if !i.ui.isRunning() {
 		return false
 	}
-	var r bool
-	_ = i.ui.t.Call(func() error {
+	return i.ui.t.BoolCall2(uintptr(unsafe.Pointer(i)), uintptr(key), func(param1, param2 uintptr) bool {
```
hajimehoshi: I feel like introducing a sync.Mutex would be better than adding variations of Call.

corfe83:
Yeah, perhaps this is overzealous. I'm going to close the PR.

PR #1107, opened 2020-03-21
ui: Add an optional function Draw function to Game interface

### LINKED ISSUE #1104: Add an optional function `Draw` to `Game` interface
```go
type Game interface {
    Layout(int, int) (int, int)
    Update(screen *Image)
    Draw(screen *Image)
}
```

Maybe? As Game is added at v1.11.0-alpha, this is the last chance to add this function.

CC @zacharyburkett

hajimehoshi:
This change adds an optional function Draw to the Game interface.
With Draw function, the game logic and rendering are separate.
There are some benefits:

  * The API is clearer and easier to understand.
  * When TPS < FPS, smoother rendering can be performed without
    changing the game logic depending on TPS.
  * Porting to XNA, which has separate functions Update and Draw,
    would be a little easier.

Draw is optional due to backward compatibility. Game interface was
already used before v1.11.x in mobile packages, and adding a
function would break existing code unfortunately. Then, we adopted
switching the behavior based on whether Draw is implemented or not
by type assertions.

IsDrawingSkipped will always return true when Draw is implemented.

Fixes #1104

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -20,20 +20,16 @@
 //     type Game struct{}
 //
 //     // Update proceeds the game state.
-//     // Update is called every frame (1/60 [s]).
+//     // Update is called every tick (1/60 [s] by default).
 //     func (g *Game) Update(screen *ebiten.Image) error {
-//
 //         // Write your game's logical update.
-//
-//         if ebiten.IsDrawingSkipped() {
-//             // When the game is running slowly, the rendering result
-//             // will not be adopted.
-//             return nil
-//         }
-//
-//         // Write your game's rendering.
-//
 //         return nil
+//     }
+//
+//     // Draw draws the game screen.
+//     // Draw is called every frame (typically 1/60[s] for 60Hz display).
+//     func (g *Game) Update(screen *ebiten.Image) error {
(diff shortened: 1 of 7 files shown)


hajimehoshi:
This is a relatively important design change. @zacharyburkett, could you take a look when you are available? Thanks!

hajimehoshi:
I found a bug: dumping a screenshot doesn't work with Draw. I'll fix this soon.

hajimehoshi:
Fixed.

zacharyburkett:
This seems to work well. Frame based interpolation is now possible.

Maybe it's worth adding delta time to the Draw function? If not, it's still very easy to check delta time yourself in the game.

hajimehoshi:
Thank you for taking a look!

> Maybe it's worth adding delta time to the Draw function? If not, it's still very easy to check delta time yourself in the game.

That's a good idea! However, my concern is that Update would have to have a time-delta argument if Draw had, and it is impossible to change the Update argument so far. I'll be able to add such change Ebiten 2.0 or 3.0 :-)

Issue #1151, opened 2020-05-10
Need help with drawn string bounding box calculation


SamusAranX:
I just spent a few hours trying to implement a method that measures drawn text size, but I can't get it to work reliably in conjunction with [the multiface package](https://github.com/AndreKR/multiface) that I use to get fallback font support. I need this for a small UI framework I'm working on for a personal project, and I need to be able to measure strings *before* drawing them to enable text alignment in UI controls. Currently, it works well enough for my default sans serif font, but utterly fails with the monospace one: 
![image](https://user-images.githubusercontent.com/676069/81512546-91f24900-9321-11ea-955e-b2be52880226.png)

I was hoping to open a PR later on, but I got stuck. Here's the relevant commit in my fork of ebiten: https://github.com/SamusAranX/ebiten/commit/d8f1dbdc21ec5b06fafbb8978d0ae924cd5b4e96

Here's my helper file that handles font loading and multiface construction: https://gist.github.com/SamusAranX/2672324335cca4fd80e26c1b4fb40c1a

And here's the code responsible for drawing what's in the screenshot: 
```go
func DrawTextInBox(dst *ebiten.Image, str string, font font.Face, x, y int) (int, int) {
	w, h := text.MeasureString(str, res.UIFont)
	wf, hf := float64(w), float64(h)

	m := font.Metrics()

	ebitenutil.DrawRect(dst, float64(x), float64(y - (m.Height.Round() - m.Descent.Round())), wf, hf - float64(m.Descent.Round()), color.Black)

	text.Draw(dst, str, font, x, y, color.White)

	return w, h + m.Descent.Round()
}

func (s *DebugScreen) Render(image *ebiten.Image) {
	fps := int(math.Round(ebiten.CurrentFPS()))
	w, h := image.Size()

	c := colorful.Hsv(s.hue, 1, 0.5)

	_ = image.Fill(c)

	strings := []string{
		s.GetTitle(),
		"Line 1",
		"    Line 2 with four spaces",
		"Line 3 abcdefg \nダミーテキスト",
	}

	strX := 10
	strY := 32

	for _, s := range strings {
		_, boxH := DrawTextInBox(image, s, res.UIFont, strX, strY)
		strY += boxH
	}

	strY += 4

	for _, s := range strings {
		_, boxH := DrawTextInBox(image, s, res.MonoFont, strX, strY)
		strY += boxH
	}

[...]

PR #1152, opened 2020-05-11
Adds a MeasureString method

SamusAranX:
(sorta-kinda continuation from #1151)

This PR adds a `MeasureString(string, font.Face)` method next to the `Draw()` method in ebiten's `text` package. It should work with any supported font.Face, even "non-standard" faces like those produced by packages like https://github.com/AndreKR/multiface. MeasureString's code was adapted from the Draw method, so it's just as reliable.

I added this method because I needed a reliable way of measuring text to align it and Go's built-in [BoundString](https://pkg.go.dev/golang.org/x/image/font?tab=doc#Drawer.BoundString) method isn't that.

MeasureString only returns width and height as integers. The code below shows how to draw boxes that encompass a drawn string, along with the output:

```go
func DrawTextInBox(dst *ebiten.Image, str string, face font.Face, x, y int) (int, int) {
	w, h := text.MeasureString(str, face)
	wf, hf := float64(w), float64(h)

	m := face.Metrics()
	mHeight := float64(m.Height.Round())

	ebitenutil.DrawRect(dst, float64(x), float64(y) - mHeight, wf, hf, color.Black)
	text.Draw(dst, str, face, x, y, color.White)

	return w, h
}

func (s *DebugScreen) Render(image *ebiten.Image) {
	fps := int(math.Round(ebiten.CurrentFPS()))
	w, h := image.Size()

	_ = image.Fill(color.RGBA{R: 127, G: 127, B: 127, A: 255})

	strings := []string{
		s.GetTitle(),
		"    Line 1: Four spaces and Äccéñtş\nalso a newline with Äccéñtş",
		"Line 2 gcdefg\nダミーテキスト",
	}

	strX := 10
	strY := 32

	for _, s := range strings {
		_, boxH := DrawTextInBox(image, s, res.UIFont, strX, strY)
		strY += boxH + 2 // +2 for a bit of margin between boxes
	}

	strY += 2

	for _, s := range strings {
		_, boxH := DrawTextInBox(image, s, res.MonoFont, strX, strY)
		strY += boxH + 2 // +2 for a bit of margin between boxes
	}

	_ = ebitenutil.DebugPrint(image, fmt.Sprintf("%02d FPS | %d×%d", fps, w, h))
}
```

the generated image


.gitignore
```diff
@@ -3,3 +3,4 @@
 *.aar
 *.apk
 .vscode
+.idea
```
hajimehoshi, line 6: I don't think this is necessary for this change.

text/text.go
```diff
@@ -300,3 +302,51 @@ func Draw(dst *ebiten.Image, text string, face font.Face, x, y int, clr color.Co
...
+		w = int(math.Max(float64(w), float64(fx.Round())))
+		h = int(math.Max(float64(h), float64(fy.Round() + faceHeight.Round())))
+
+		prevR = r
+	}
+
+	textM.Unlock()
+
+	return w, h + faceDescent.Round()
+}
```
hajimehoshi, line 352: Add a new line. (gofmt should do this :-)

text/text.go
```diff
@@ -300,3 +302,51 @@ func Draw(dst *ebiten.Image, text string, face font.Face, x, y int, clr color.Co
...
+//
+// Be careful that the passed font face is held by this package and is never released.
+// This is a known issue (#498).
+//
+// MeasureString is concurrent-safe.
+func MeasureString(text string, face font.Face) (int, int) {
+	textM.Lock()
+
+	x, y := 0, 0
+	w, h := 0, 0
```
hajimehoshi, line 319: I'd keep `w`, `h` as `fixed.Int26_6`, and convert them to floats at last. What do you think?

text/text.go
```diff
@@ -300,3 +302,51 @@ func Draw(dst *ebiten.Image, text string, face font.Face, x, y int, clr color.Co
 
 	textM.Unlock()
 }
+
+// MeasureString measures the size of a given string using a given font.
```
hajimehoshi, line 306: Rather than returning sizes, what about returning bounds?

text/text.go
```diff
@@ -300,3 +302,51 @@ func Draw(dst *ebiten.Image, text string, face font.Face, x, y int, clr color.Co
...
+//
+// text is the string that's being measured.
+// face is the font for text rendering.
+//
+// Be careful that the passed font face is held by this package and is never released.
+// This is a known issue (#498).
```
hajimehoshi, line 312: Just to confirm: this is because `glyphAdvance`, right?

text/text.go
```diff
@@ -300,3 +302,51 @@ func Draw(dst *ebiten.Image, text string, face font.Face, x, y int, clr color.Co
...
+
+		fx += glyphAdvance(face, r)
+
+		w = int(math.Max(float64(w), float64(fx.Round())))
+		h = int(math.Max(float64(h), float64(fy.Round() + faceHeight.Round())))
+
+		prevR = r
+	}
+
+	textM.Unlock()
```
hajimehoshi, line 349: Use `defer` and move this just after `textM.Lock()`

.gitignore
SamusAranX: Oh yeah, this could probably be taken out. I added this to .gitignore because I was getting tired of seeing GoLand's garbage files in my diffs.

text/text.go
SamusAranX: With "bounds", do you mean Go's image.Rectangle? I intentionally didn't use that because it comes with X and Y coordinates that would be left unused since they'd always be (0, 0).

text/text.go
SamusAranX: I couldn't tell you. I copied this comment from `Draw()` because the code is very similar, so it would make sense that the same bug would happen there too.

text/text.go
SamusAranX: Makes sense, I forgot about defer.

text/text.go
SamusAranX: I swear I ran gofmt on this. 😅 
Will do.

text/text.go
SamusAranX: I intentionally return ints because this method is supposed to return text size in whole pixels. I suppose it might make sense to have these start out as float64s and only cast them to ints when returning, but that *might* lead to rounding errors. I'll investigate.

text/text.go
hajimehoshi: >  I intentionally didn't use that because it comes with X and Y coordinates that would be left unused since they'd always be (0, 0).

(X, Y) can be a relative position when assuming (0, 0) as a dot position, like what BoundString does.

text/text.go
hajimehoshi: I don't think you have to treat float64 in the calculations. Keeping `fixed.Int26_6` and converting them to ints at the end should be fine.

.gitignore
SamusAranX: Took it out in 481985a.

text/text.go
SamusAranX: A newline was added in 481985a.

text/text.go
SamusAranX: Added defer in 481985a. I added the same defer to the Draw method as well since it has the same structure.

text/text.go
SamusAranX: `math.Max` only returns float64s, I'd have to cast that back to `fixed.Int26_6` if I declared `w` and `h` as `fixed.Int26_6`. Functionally, there's no real difference whether the method uses one or the other.

text/text.go
SamusAranX: In 78a34f9, I made MeasureString return two `image.Point`s, one for the origin and one for the bounds. I tried using an `image.Rectangle` instead, but since it'd almost always have a negative Y coordinate, retrieving the height would be really unintuitive.
This way, one can simply do `origin, bounds := text.MeasureString(str, face)` and get everything.

SamusAranX: As of 78a34f9, a DrawTextInBox method could look like this:
```go
func DrawTextInBox(dst *ebiten.Image, str string, face font.Face, x, y int) image.Point {
	origin, bounds := text.MeasureString(str, face)

	xf, yf := float64(x + origin.X), float64(y + origin.Y)
	wf, hf := float64(bounds.X), float64(bounds.Y)

	ebitenutil.DrawRect(dst, xf, yf, wf, hf, color.Black)
	text.Draw(dst, str, face, x, y, color.White)

	return bounds
}
```

hajimehoshi: > since it'd almost always have a negative Y coordinate, retrieving the height would be really unintuitive.

I am not sure how unintuitive it would be: `(b.Max.Y - b.Min.Y)` should work.

1. To be exact, my suggestion is that the returned value would be [Rectangle26_6](https://pkg.go.dev/golang.org/x/image/math/fixed?tab=doc#Rectangle26_6)

2. You can calculate a float64 value from Int26_6 like this

```
func toFloat(x fixed.Int26_6) float64 {
    return float64(x >> 6) + float64(x & ((1 << 6) - 1)) / float64(1 << 6)
}
```

2. With the rectangle, the code would be like this:

```
func DrawTextInBox(dst *ebiten.Image, str string, face font.Face, x, y int) image.Point {
	bounds := text.MeasureString(str, face)

	xf, yf := float64(x) + toFloat(bounds.Min.X), float64(y) + toFloat(bounds.Min.Y)
	wf, hf := toFloat(bounds.Max.X - bounds.Min.X), toFloat(bounds.Max.X - bounds.Min.X) 

	// ...

	return bounds
}
```

text/text.go
hajimehoshi: Unfortunately, `Round()` would lose some information.

I'd create min/max function for Int26_6.

.gitignore
hajimehoshi: This still remains. Could you remove this?

hajimehoshi: Sorry this was removed.

.gitignore
```diff
@@ -2,4 +2,4 @@
 *~
 *.aar
 *.apk
-.vscode
+.vscode
```
hajimehoshi, line 5: Could you add a new line?

hajimehoshi:
So the implementation should be similar to the existing `font.BoundString`, and the difference is how to consider new lines. Is that correct?

SamusAranX:
Pretty much. MeasureString takes some additional things (like font ascenders) into account though. BoundString doesn't do that.

text/text.go
SamusAranX: IMO it would unnecessarily bloat code for a neglegible increase in precision.
Int26_6 is unintuitive in that it's impossible to just cast it to another type, requiring bit fiddling or rounding to be used in other contexts.
Seeing as this would be an API for game devs who want something they can work with instead of font creators who might want the return value to have the same type as the freetype package, I'd rather stick with normal integers.

.gitignore
SamusAranX: Added in 83cd779.

text/text.go
SamusAranX: In b03ea20, I changed MeasureString so that it uses fixed.Int26_6 internally and only rounds at the end. This should avoid accumulation of rounding errors for very long strings.

I also took the liberty of changing the fixed26_6ToFloat64 function to mirror your toFloat method, since there's little point in having two functions to almost exactly the same thing.

hajimehoshi: > Seeing as this would be an API for game devs who want something they can work with instead of font creators who might want the return value to have the same type that's used throughout the freetype package, I'd rather stick with normal integers.

Aha, I understand your points. OK let's avoid returning fixed26_6.

1. Sorry for going back and forth, but could we have this function to return only the size? If we want the dot position, we can calculate it by ourselves. We could have another function, but let's consider this later.
2. Instead of `Round`, let's use `Ceil` so that the region would be safer to cover the text. (Or, retuning float64 might be another good option. As long as text.Draw takes integers, I'm fine with ints though.)

> I also took the liberty of changing the fixed26_6ToFloat64 function to mirror your toFloat method,

Oh I didn't realize that. Thank you for taking care of this!

SamusAranX: > Sorry for going back and forth

No problem! Better to be safe than sorry 😄

> ...could we have this function to return only the size? If we want the dot position, we can calculate it by ourselves.

To be able to locate a piece of drawn text in top-left-origin space, you need the font's height so you can subtract it from the text's starting dot position. This necessitates a call to `face.Metrics()`, which does [a bunch of stuff](https://github.com/golang/freetype/blob/master/truetype/face.go#L254) and afaik isn't cached anywhere. 
Since MeasureString already calls `face.Metrics()` as part of its normal operation, I'd say it's actually helpful to devs to do that work for them and return a ready-to-use origin point along with a size. It's probably a form of premature optimization, yes, but it saves devs a function call and keeps their code clean.

> Instead of Round, let's use Ceil

Good call. I pushed this change in 5ce4ba9.

text/text.go
SamusAranX: I'll resolve this because all calls to math.Max were removed in b03ea20.

text/text.go
hajimehoshi: > Since MeasureString already calls face.Metrics() as part of its normal operation, I'd say it's actually helpful to devs to do that work for them and return a ready-to-use origin point along with a size. It's probably a form of premature optimization, yes, but it saves devs a function call and keeps their code clean.

I understand the performance concern, but calculating `Metrics` should not so heavy task. Then this sounds like, as you indicated, premature optimization. In the users' perspective, clean API is preferable to clean implementation. So I'd want to add this as another function if necessary.

SamusAranX: I removed the origin point in 4aca1b0 and edited the function's documentation comment a bit.

text/text.go
```diff
@@ -297,6 +300,59 @@ func Draw(dst *ebiten.Image, text string, face font.Face, x, y int, clr color.Co
 
 		prevR = r
 	}
+}
+
+// MeasureString returns the measured size of a given string using a given font.
```
hajimehoshi, line 305: Please add a comment about how to treat new lines (Metrics().Height of the given font is used).

SamusAranX: I'm not sure I understand. MeasureString returns a Point with X *and* Y values, implying that both width and height are correctly measured.

hajimehoshi: Considering multiple lines is an important feature of this function. It is not trivial to users what line-height is adopted.

SamusAranX: I can't come up with a fitting comment that doesn't also sound really stilted. 

This method does all the right things automatically. The user shouldn't have to worry about how newlines work because they are already familiar with how they work from the Draw method.
If I explain in the comment how the method handles newlines, I'd also have to explain how it handles special characters, multiple font faces, ascenders, descenders, and so on.

hajimehoshi: I agree you suggestion makes sense, thanks!

hajimehoshi:
Please rebase or merge with the master branch.

I'll go to bed soon. I'll re-review your PR tomorrow, and merge this if there are no additional things.

text/text.go
hajimehoshi: Could you add the comment you suggested?

PR #1169, opened 2020-05-24
use slices for maps where keys are small contiguous integers

seebs:
Durations for key presses and gamepad button presses were being stored
in map[int]int, where the key values were always small integers, and
furthermore, were always contiguous and contained the whole set. It's
much faster and cheaper to do those with slices.

We could probably do this with mouse buttons, but I was wary of that
code because I don't see any handling for more-than-three-button
mice. Similar logic could perhaps apply to touch events, but there
it might make more sense to use a dynamic allocation of some kind.

It might be more efficient to just swap the two slices back and
forth, computing the new values from the previous values and
then swapping the slice arrays themselves, but that seemed like
a more intrusive change.

### DIFF
--- a/inpututil/inpututil.go
+++ b/inpututil/inpututil.go
@@ -24,8 +24,8 @@ import (
 )
 
 type inputState struct {
-	keyDurations     map[ebiten.Key]int
-	prevKeyDurations map[ebiten.Key]int
+	keyDurations     []int
+	prevKeyDurations []int
 
 	mouseButtonDurations     map[ebiten.MouseButton]int
 	prevMouseButtonDurations map[ebiten.MouseButton]int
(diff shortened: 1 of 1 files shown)


seebs:
So far as I can tell, the CI failure is in the shader code and unrelated.

hajimehoshi:
> So far as I can tell, the CI failure is in the shader code and unrelated.

Sorry I've already fixed. Could you rebase your PR?

hajimehoshi:
> We could probably do this with mouse buttons, but I was wary of that
code because I don't see any handling for more-than-three-button
mice.

You mean we should have MouseButtonMax?

seebs:
I would be in favor of a mousebuttonmax, I'm not sure what it would be. I think my current mouse generates events for Button 11.

PR #1227, opened 2020-06-30
Stereo audio panning example

### LINKED ISSUE #1224: Stereo Sound Panning
I implemented a "pan-able" audio stream. It changes the stereo channels signal amplitude as the bytes are being read.
I can open a PR if there's a need for it to be in the library.

```go
package audio

import (
	"bytes"
	"io"

	"github.com/hajimehoshi/ebiten/audio"
)

// StereoPanStream is an audio buffer that changes the stereo channel's signal
// based on the Panning.
type StereoPanStream struct {
	*bytes.Reader
	pan float64 // -1: left; 0: center; 1: right
}

func (s *StereoPanStream) Read(p []byte) (n int, err error) {
	n, err = s.Reader.Read(p)
	if err != nil {
		return
	}
	ls := min(s.pan*-1+1, 1)
	rs := min(s.pan+1, 1)
	for i := 0; i < len(p); i += 4 {
		lc := int16(float64(int16(p[i])|int16(p[i+1])<<8) * ls)
		rc := int16(float64(int16(p[i+2])|int16(p[i+3])<<8) * rs)

		p[i] = byte(lc)
		p[i+1] = byte(lc >> 8)
		p[i+2] = byte(rc)
		p[i+3] = byte(rc >> 8)
	}
	return
}

func (s *StereoPanStream) Close() error {
	return nil
}

func (s *StereoPanStream) SetPan(pan float64) {
	s.pan = min(max(-1, pan), 1)
}

func (s *StereoPanStream) Pan() float64 {
	return s.pan
}

// NewStereoPanStream returns a new StereoPanStream with a shared buffer src.
// The src's format must be linear PCM (16bits little endian, 2 channel stereo)
// without a header (e.g. RIFF header). The sample rate must be same as that
// of the audio context.
//
// The src can be shared by multiple buffers.
func NewStereoPanStream(src []byte) *StereoPanStream {
	return &StereoPanStream{
		Reader: bytes.NewReader(src),
	}
}

// NewStereoPanStreamFromReader returns a new StereoPanStream with a copied buffer src.
// The buffer needs to be copied because the pan algorithm changes the wave
// amplitude as the bytes are read.
//
// The src's format must be linear PCM (16bits little endian, 2 channel stereo)
// without a header (e.g. RIFF header). The sample rate must be same as that
// of the audio context.
func NewStereoPanStreamFromReader(src audio.ReadSeekCloser) *StereoPanStream {
	src.Seek(0, io.SeekStart)
	buf := new(bytes.Buffer)
	io.Copy(buf, src)
	return &StereoPanStream{
		Reader: bytes.NewReader(buf.Bytes()),
	}
}

// test that it fulfills Ebiten's ReadSeekCloser
var _ audio.ReadSeekCloser = &StereoPanStream{}

func min(a, b float64) float64 {
	if a < b {
		return a
	}
	return b
}

func max(a, b float64) float64 {
	if a > b {
		return a
	}
	return b
}
```

gabstv:
This example simulates an audio source (sprite) and an audio listener (camera). The panning is relative to the sprite's position.

Closes #1224 

Edit: The example should be run with stereo speakers or headsets 😅

Example:
https://cquiz-us.s3.amazonaws.com/gtest/index.html

### DIFF
--- a/examples/audiopanning/main.go
+++ b/examples/audiopanning/main.go
@@ -166,20 +166,14 @@ type StereoPanStream struct {
 	pan float64 // -1: left; 0: center; 1: right
 }
 
-const sqrt2div2 = 1 / math.Sqrt2 // math.sqrt(2)/2.0
-const deg45 = math.Pi / 4        // 45º
-
 func (s *StereoPanStream) Read(p []byte) (n int, err error) {
 	n, err = s.ReadSeekCloser.Read(p)
 	if err != nil {
 		return
 	}
 
-	angle := deg45 * s.pan
-	acos := math.Cos(angle)
-	asin := math.Sin(angle)
-	ls := sqrt2div2 * (acos - asin)
-	rs := sqrt2div2 * (acos + asin)
+	ls := math.Min(s.pan*-1+1, 1)
+	rs := math.Min(s.pan+1, 1)
 	for i := 0; i < len(p); i += 4 {
 		lc := int16(float64(int16(p[i])|int16(p[i+1])<<8) * ls)
 		rc := int16(float64(int16(p[i+2])|int16(p[i+3])<<8) * rs)
--- a/examples/audiopanning/main.go
+++ b/examples/audiopanning/main.go
@@ -172,6 +172,11 @@ func (s *StereoPanStream) Read(p []byte) (n int, err error) {
 		return
 	}
 
+	// This implementation uses a linear scale, for stereo or mono sounds.
+	// If pan = 0.0, the balance for the sound in each speaker is at 100% left and 100% right.
+	// When pan is -1.0, only the left channel of the stereo sound is audible, when pan is 1.0,
+	// only the right channel of the stereo sound is audible.
+	// https://docs.unity3d.com/ScriptReference/AudioSource-panStereo.html
 	ls := math.Min(s.pan*-1+1, 1)
 	rs := math.Min(s.pan+1, 1)
 	for i := 0; i < len(p); i += 4 {
--- a/examples/audiopanning/main.go
+++ b/examples/audiopanning/main.go
@@ -166,6 +166,12 @@ func (s *StereoPanStream) Read(p []byte) (n int, err error) {
 	if err != nil {
 		return
 	}
+
+	// This panning implementation uses a simpler algorithm that doesn't
+	// mix the left and right channels. For a more accurate implementation,
+	// there's a method called "constant power panning":
+	// https://dsp.stackexchange.com/questions/21691/algorithm-to-pan-audio/21736
+
 	ls := math.Min(s.pan*-1+1, 1)
 	rs := math.Min(s.pan+1, 1)
 	for i := 0; i < len(p); i += 4 {
--- a/examples/audiopanning/main.go
+++ b/examples/audiopanning/main.go
@@ -161,19 +161,20 @@ type StereoPanStream struct {
 	pan float64 // -1: left; 0: center; 1: right
 }
 
+const sqrt2div2 = 0.7071067811865476 // math.sqrt(2)/2.0
+const rad45 = math.Pi / 4            // 45º
+
 func (s *StereoPanStream) Read(p []byte) (n int, err error) {
 	n, err = s.Reader.Read(p)
 	if err != nil {
 		return
 	}
 
-	// This panning implementation uses a simpler algorithm that doesn't
-	// mix the left and right channels. For a more accurate implementation,
-	// there's a method called "constant power panning":
-	// https://dsp.stackexchange.com/questions/21691/algorithm-to-pan-audio/21736
-
-	ls := math.Min(s.pan*-1+1, 1)
-	rs := math.Min(s.pan+1, 1)
+	angle := rad45 * s.pan
+	acos := math.Cos(angle)
+	asin := math.Sin(angle)
+	ls := sqrt2div2 * (acos - asin)
+	rs := sqrt2div2 * (acos + asin)
 	for i := 0; i < len(p); i += 4 {
 		lc := int16(float64(int16(p[i])|int16(p[i+1])<<8) * ls)
 		rc := int16(float64(int16(p[i+2])|int16(p[i+3])<<8) * rs)
--- a/examples/audiopanning/main.go
+++ b/examples/audiopanning/main.go
@@ -64,8 +64,8 @@ type Game struct {
 	// 1: 0% left channel, 100% right channel
 	panning float64
 
-	r    float64
-	xpos float64
+	count int
+	xpos  float64
 }
 
 func (g *Game) initAudio() {
@@ -96,10 +96,16 @@ func (g *Game) initAudio() {
 	g.player.Play()
 }
 
+// time is whithin the 0 ... 1 range
+func lerp(a, b, t float64) float64 {
+	return a*(1-t) + b*t
+}
+
 func (g *Game) Update(screen *ebiten.Image) error {
 	g.initAudio()
-	g.r += ((1.0 / 60.0) * 2 * math.Pi) * 0.1 // full cycle every 10 seconds
-	g.xpos = (float64(screenWidth) / 2) + math.Cos(g.r)*(float64(screenWidth)/2)
+	g.count++
+	r := float64(g.count) * ((1.0 / 60.0) * 2 * math.Pi) * 0.1 // full cycle every 10 seconds
+	g.xpos = (float64(screenWidth) / 2) + math.Cos(r)*(float64(screenWidth)/2)
 	g.panning = lerp(-1, 1, g.xpos/float64(screenWidth))
 	g.panstream.SetPan(g.panning)
 	return nil
@@ -160,8 +166,8 @@ type StereoPanStream struct {
 	pan float64 // -1: left; 0: center; 1: right
 }
 
-const sqrt2div2 = 0.7071067811865476 // math.sqrt(2)/2.0
-const rad45 = math.Pi / 4            // 45º
+const sqrt2div2 = 1 / math.Sqrt2 // math.sqrt(2)/2.0
+const deg45 = math.Pi / 4        // 45º
 
 func (s *StereoPanStream) Read(p []byte) (n int, err error) {
 	n, err = s.ReadSeekCloser.Read(p)
@@ -169,7 +175,7 @@ func (s *StereoPanStream) Read(p []byte) (n int, err error) {
 		return
 	}
 
-	angle := rad45 * s.pan
+	angle := deg45 * s.pan
 	acos := math.Cos(angle)
 	asin := math.Sin(angle)
 	ls := sqrt2div2 * (acos - asin)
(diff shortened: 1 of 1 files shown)


hajimehoshi:
Thank you! Let's continue and reach agreement at #1224 first.

examples/audiopanning/main.go
```diff
@@ -0,0 +1,232 @@
...
+	player    *audio.Player
+	panstream *StereoPanStream
+
+	// panning goes from -1 to 1
+	// -1: 100% left channel, 0% right channel
+	// 0: 100% both channels
+	// 1: 0% left channel, 100% right channel
+	panning float64
+
+	r    float64
```
hajimehoshi, line 68: `r` should an integer representing a count in order to avoid numeric errors. Maybe renaming it to `count` sounds fine. So the logic will be like this:

```go
g.count++
r := float64(g.count) * ((1.0 / 60.0) * 2 * math.Pi) * 0.1
g.xpos = (float64(screenWidth) / 2) + math.Cos(r)*(float64(screenWidth)/2)
```

examples/audiopanning/main.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	angle := rad45 * s.pan
+	acos := math.Cos(angle)
+	asin := math.Sin(angle)
+	ls := sqrt2div2 * (acos - asin)
+	rs := sqrt2div2 * (acos + asin)
+	for i := 0; i < len(p); i += 4 {
+		lc := int16(float64(int16(p[i])|int16(p[i+1])<<8) * ls)
+		rc := int16(float64(int16(p[i+2])|int16(p[i+3])<<8) * rs)
```
hajimehoshi, line 180: Would it be possible to have mixing the left and right channel like Chromium does? (Though ragtime.ogg is a mono-channel audio...)

examples/audiopanning/main.go
```diff
@@ -0,0 +1,232 @@
...
+}
+
+// StereoPanStream is an audio buffer that changes the stereo channel's signal
+// based on the Panning.
+type StereoPanStream struct {
+	*bytes.Reader
+	pan float64 // -1: left; 0: center; 1: right
+}
+
+const sqrt2div2 = 0.7071067811865476 // math.sqrt(2)/2.0
```
hajimehoshi, line 164: `1 / math.Sqrt2` should be the same constant.

examples/audiopanning/main.go
```diff
@@ -0,0 +1,232 @@
...
+const rad45 = math.Pi / 4            // 45º
```
hajimehoshi, line 165: `deg45`?

examples/audiopanning/main.go
```diff
@@ -0,0 +1,232 @@
...
+func NewStereoPanStreamFromReader(src audio.ReadSeekCloser) *StereoPanStream {
+	src.Seek(0, io.SeekStart)
+	buf := new(bytes.Buffer)
+	io.Copy(buf, src)
+	return &StereoPanStream{
+		Reader: bytes.NewReader(buf.Bytes()),
+	}
+}
+
+func lerp(a, b, t float64) float64 {
```
hajimehoshi, line 230: This function should be moved above `(*Game).Update`.

examples/audiopanning/main.go
```diff
@@ -0,0 +1,232 @@
...
+	return s.pan
+}
+
+// NewStereoPanStream returns a new StereoPanStream with a shared buffer src.
+// The src's format must be linear PCM (16bits little endian, 2 channel stereo)
+// without a header (e.g. RIFF header). The sample rate must be same as that
+// of the audio context.
+//
+// The src can be shared by multiple buffers.
+func NewStereoPanStream(src []byte) *StereoPanStream {
```
hajimehoshi, line 208: This function is not used anywhere?

examples/audiopanning/main.go
hajimehoshi: So the implementation would be like this:

```go
panrad := math.Abs(s.pan) * math.Pi / 2
lgain := math.Cos(panrad)
rgain := math.Sin(panrad)
for i := 0; i < len(p); i += 4 {
    l := float64(int16(p[i])|int16(p[i+1])<<8)
    r := float64(int16(p[i+2])|int16(p[i+3])<<8)
    var lc, rc int16
    if (s.pan <= 0) {
        lc = int16(l + r * lgain)
        rc = int16(r * rgain)
    } else {
        lc = int16(l * lgain)
        rc = int16(r + l * rgain)
    }
    ...
}
```

examples/audiopanning/main.go
gabstv: I copied from my implementation, but it isn't used in the example. Removed

examples/audiopanning/main.go
gabstv: I tested this, and when it gets closes to -1 or 1, the audio becomes clipped (some frequencies get clipped to max), so I adjusted to 
```go
if s.pan <= 0 {
	lc = int16((l + r*lgain) / 2)
	rc = int16(r * rgain)
} else {
	lc = int16(l * lgain)
	rc = int16((r + l*rgain) / 2)
}
```

gabstv: It solves clipping, but still doesn't do a proper mixing of `lc` or `rc` (since `lgain` and `rgain` max out at cos/sin of 45º). Another problem is that it reduces the difference between different pans (if pan is <= 0, lc is getting half of the original value regardless of the pan value).

hajimehoshi: Thank you for testing! What about capping the value after the calculation?

```go
    l := float64(int16(p[i])|int16(p[i+1])<<8)
    r := float64(int16(p[i+2])|int16(p[i+3])<<8)
    var lc, rc int16
    if (s.pan <= 0) {
        lc = int16(l + r * lgain)
        rc = int16(r * rgain)
    } else {
        lc = int16(l * lgain)
        rc = int16(r + l * rgain)
    }
    lc = min(max(lc, 1.0), -1.0)
    rc = min(max(rc, 1.0), -1.0)
```

I'd like to keep the general algorithm as possible...

gabstv: The best way to ensure, even if the user imports a heavily mixed true stereo sound for panning, that the sound is mixed would be:
```go
	angle := deg45 * s.pan
	acos := math.Cos(angle)
	asin := math.Sin(angle)
	ls := sqrt2div2 * (acos - asin)
	rs := sqrt2div2 * (acos + asin)
	for i := 0; i < len(p); i += 4 {
		l := float64(int16(p[i]) | int16(p[i+1])<<8)
		r := float64(int16(p[i+2]) | int16(p[i+3])<<8)
		mixed := (l + r) / 2 // we mix left and right to enforce that the source is mono
		lc := int16(mixed * ls)
		rc := int16(mixed * rs)

		p[i] = byte(lc)
		p[i+1] = byte(lc >> 8)
		p[i+2] = byte(rc)
		p[i+3] = byte(rc >> 8)
	}
```
What do you think?

hajimehoshi: Hmm, does this mean all the audio sources are treated as mono-channel? I'd want to keep the source stereo.

BTW, in the Chromium implementation, deg45 does not exist, right? (I think you are referring the implementation in the StackOverflow question, but I'd want to avoid this since I don't understand the algorithm.)

gabstv: By the way, I tested with capping the values to min/max int16 but clipping still occurs:
```go
	pan := s.pan
	if pan <= 0 {
		pan += 1
	}
	panrad := pan * math.Pi / 2
	lgain := math.Cos(panrad)
	rgain := math.Sin(panrad)
	for i := 0; i < len(p); i += 4 {
		l := float64(int16(p[i]) | int16(p[i+1])<<8)
		r := float64(int16(p[i+2]) | int16(p[i+3])<<8)
		var lc, rc int
		if s.pan <= 0 {
			lc = int(l + r*lgain)
			rc = int(r * rgain)
		} else {
			lc = int(l * lgain)
			rc = int(r + l*rgain)
		}
		lc16 := int16(maxint(minint(lc, math.MaxInt16), math.MinInt16))
		rc16 := int16(maxint(minint(rc, math.MaxInt16), math.MinInt16))

		p[i] = byte(lc16)
		p[i+1] = byte(lc16 >> 8)
		p[i+2] = byte(rc16)
		p[i+3] = byte(rc16 >> 8)
	}
```

Just to show that implementations may differ, Unity does this:

> AudioSource.panStereo. 
> Description
> 
> Pans a playing sound in a stereo way (left or right). This only applies to sounds that are Mono or Stereo.
> 
> This pan is applied before 3D panning calculations are considered. In other words, stereo panning affects the left right balance of the sound before it is spatialised in 3D.
> 
> Mono sounds are panned from left to right using constant power panning (non linear fade). This means when pan = 0.0, the balance for the sound in each speaker is 71% left and 71% right, not 50% left and 50% right. This gives (audibly) smoother pans.
> 
> Stereo sounds heave each left/right value faded up and down according to the specified pan position. This means when pan = 0.0, the balance for the sound in each speaker is 100% left and 100% right. When pan = -1.0, only the left channel of the stereo sound is audible, when pan = 1.0, only the right channel of the stereo sound is audible.
> 
> Values range from -1.0 to 1.0.
> 
> -1.0 = Full left 0.0 = center 1.0 = full right
> 
> Default = 0.0.

TLDR: If mono, use 45 degrees formula; If stereo: use my original implementation (use original left and right channels, linear gain instead of 15degrees).

In my opinion, since Ebiten doesn't keep track if the original sound was mono or stereo, I would force into mono and do the 45 degrees formula.

hajimehoshi: Ah ok, referring an existing API sounds great! I'm fine as long as the algorithm is well documented and widely used. Could you leave a comment that the implementation is in the same way of [AudioSource.panStereo](https://docs.unity3d.com/ScriptReference/AudioSource-panStereo.html)? 

> In my opinion, since Ebiten doesn't keep track if the original sound was mono or stereo, I would force into mono and do the 45 degrees formula.

I slightly prefer to always treat the source as stereo, but treating them as mono would make sense as an example. Please also comment about this.

gabstv: If you want, I can revert the algorithm back to the one that uses the original channels (like Unity does when it's stereo). For mono, it would't make much difference (it would only be a little less smooth).

The 45 degrees formula ensures that the `sqrt(lgain*lgain + rgain*rgain)` stays at 1:
img

hajimehoshi: > If you want, I can revert the algorithm back to the one that uses the original channels (like Unity does when it's stereo). For mono, it would't make much difference (it would only be a little less smooth).

Yes, the stereo way is nicer to me. Sorry for going back and forth.

The unclear thing to me is that how the volumes are interpolated: if the panning is -0.9, would the right channel's volume be 0.1? Or, would sin/cos be used here?

> The 45 degrees formula ensures that the sqrt(lgain*lgain + rgain*rgain) stays at 1:

I understand the total power of the waves is always 1. If the panning is -1, would the degree -90°?

gabstv: The range is always between -45º and 45º, so -90º would never happen (since panning is clamped between -1 and 1).
If the panning is -0.9, the angle would be translated to `-40.5º`:
```
// -40.5 to radians: -0.7068583470577035
// cos(-40.5º) : 0.7604059656000309
// sin(-40.5º): -0.6494480483301837
lgain = sqrt2div2 * (cos - sin) // = 0.996917333733128
rgain = sqrt2div2 * (cos + sin) // = 0.07845909572784494
```
I may have miscalculated the true left and right gains of my explanation image 😅

I'll revert the algorithm back to stereo + linear then!

gabstv: I reverted to stereo+linear algorithm

hajimehoshi: OK thank you for explaining, and reverting the implementation! I'll take a look.

hajimehoshi: Please leave a comment about the algorithm and the link to Unity.

gabstv: added!

hajimehoshi:
Let me remove the intro-loop part. I think removing this should not affect the core value of the example.

gabstv:
oh sure, I used the intro/loop example as base

hajimehoshi:
Done: dcba3808270d2a5063768bff57ddf15c697514a4

I appreciate your contribution!

I might change the panning algorithm in the future if I find a nice algorithm. Through the discussion with you, I now understand about the power of waves better :-) Thank you!

gabstv:
Thank you! It's your comment that made me go after the true mono panning algorithm (45deg) and actually understand it! Feel free to change the algorithm when the time comes.

PR #1233, opened 2020-07-02
Invert ColorM

### LINKED ISSUE #548: affine: ColorM.Invert function
The counterpart of `ColorM` for #547. Inverting a 5-dimension affine matrix is pretty much harder.

gabstv:
Closes #548


internal/affine/colorm.go
```diff
@@ -167,6 +167,229 @@ func (c *ColorM) UnsafeElements() ([]float32, []float32) {
...
+	det := c.det()
+	if det == 0 {
+		panic("ebiten: c is not invertible")
+	}
+
+	if !c.isInited() {
+		c.body = make([]float32, 16)
+		c.translate = make([]float32, 4)
+		copy(c.body, colorMIdentityBody)
+		copy(c.translate, colorMIdentityTranslate)
```
hajimehoshi, line 260: Can we early-return here without setting the body and the translate?

internal/affine/colorm.go
```diff
@@ -167,6 +167,229 @@ func (c *ColorM) UnsafeElements() ([]float32, []float32) {
 	return eb, et
 }
 
+func (c *ColorM) det() float32 {
+	if c == nil {
+		return 0
```
hajimehoshi, line 172: When c is nil, c is an identity. Then this should return 1. (This should be included in `isInited`).

internal/affine/colorm.go
```diff
@@ -167,6 +167,229 @@ func (c *ColorM) UnsafeElements() ([]float32, []float32) {
...
+
+// IsInvertible returns a boolean value indicating
+// whether the matrix c is invertible or not.
+func (c *ColorM) IsInvertible() bool {
+	return c.det() != 0
+}
+
+// Invert inverts the matrix.
+// If c is not invertible, Invert panics.
+func (c *ColorM) Invert() {
```
hajimehoshi, line 250: As ColorM is an affine matrix, you don't have to calculate inversion of a 5x5 matrix. Calculating an inversion of 4x4 matrix is enough! Assuming that M is an affine 5x5 matrix that consists of M_body (5x5 only with the body) and M_translate (5x5 only with the translate), M^(-1) can be calculated like this:

```
M^(-1) = (M_body * M_translate)^(-1)
       = M_translate^(-1) * M_body^(-1)
```

`M_translate^(-1)` can be calculated easily: just multiply -1 to M_translate (imagine that moving something by an offset X. The opposite operation is to move it by -X).

`M_body^(-1)` can be calculated in the same way of inverting 4x4 matrix, even though M_body is a 5x5 matrix :-)

BTW, I realized that GeoM's det doesn't actually calculate 3x3 det, so I renamed this :-) 5bc09acb41fa10a1b436780daeaaf034b15483f6

hajimehoshi, line 250: ColorM in this package is immutable. Please return a new `ColorM` object (or nil) by this function.

internal/affine/colorm.go
gabstv: should I just return a new identity matrix if `!isInited` (or if is nil)? Based on the immutability

internal/affine/colorm.go
gabstv: I assumed that ColorM was always a 5x5 by reading the issue lol. Initially I generated a 4x4 inversion, but by inspecting the `SetElement(i, j int, element float32)`, the `c.translate` seemed to be the last row of the 5x5 matrix to invert, then I also assumed that the last column had to exist at the net and Invert funcs (`0 0 0 0 1`).

So, the best way to do this is to generate a 4x4 inversion for `c.body` and just multiply the `c.translate` by -1?

internal/affine/colorm.go
hajimehoshi: In this case you should return nil. nil indicates an identity color matrix.

internal/affine/colorm.go
hajimehoshi: > ... I also assumed that the last column had to exist at the det and Invert funcs (0 0 0 0 1).

Yes, this assumption is correct. Your calculation seems totally fine, but I'm suggesting a more efficient way.

> So, the best way to do this is to generate a 4x4 inversion for c.body and just multiply the c.translate by -1?

Very close. As I indicated, the multiplying order is `M_translate^(-1) * M_body^(-1)`.

gabstv: I removed all the unnecessary multiplications (by the last column 0 0 0 0 1) and now the `det` is 4x4. The inversion still uses the `c.translate`, but since the last column was removed, it's more efficient.

gabstv: Benchmarks:
```
goos: darwin
goarch: amd64
pkg: github.com/hajimehoshi/ebiten/internal/affine
BenchmarkColorMInvert-4   	 5123756	       215 ns/op	     128 B/op	       3 allocs/op
PASS
ok  	github.com/hajimehoshi/ebiten/internal/affine	1.527s
```

internal/affine/colorm.go
```diff
@@ -167,6 +167,164 @@ func (c *ColorM) UnsafeElements() ([]float32, []float32) {
...
+		m01*(m10*b234234-m12*b034234+
+			m13*b024234) +
+		m02*(m10*b134234-m11*b034234+
+			m13*b014234) -
+		m03*(m10*b124234-m11*b024234+
+			m12*b014234)
+	if det == 0 {
+		return 0
+	}
+	return 1 / det
```
hajimehoshi, line 210: Why is det() returns 1 / det?

internal/affine/colorm_test.go
```diff
@@ -79,3 +80,75 @@ func TestColorMScaleOnly(t *testing.T) {
 		}
 	}
 }
+
+func TestColorMIsInvert(t *testing.T) {
```
hajimehoshi, line 84: `TestColorMIsInvertible`

hajimehoshi, line 84: Can we add tests for Invert?

internal/affine/colorm.go
```diff
@@ -167,6 +167,164 @@ func (c *ColorM) UnsafeElements() ([]float32, []float32) {
...
+// Invert inverts the matrix.
+// If c is not invertible, Invert panics.
+func (c *ColorM) Invert() *ColorM {
+	if !c.isInited() {
+		return nil
+	}
+
+	det := c.det()
+	if det == 0 {
+		panic("ebiten: c is not invertible")
```
hajimehoshi, line 228: `affine: c is not invertible`

(There might be some panic messages starting with `ebiten:` in this package, but sorry, they were outdated)

internal/affine/colorm.go
gabstv: It's just that Invert actually uses 1/det. I will change this to return det and invert it before using on the Invert() func then

internal/affine/colorm.go
gabstv: I copied from GeoM. My bad!

internal/affine/colorm.go
hajimehoshi: Yeah, that makes more sense.

internal/affine/colorm.go
```diff
@@ -167,6 +167,158 @@ func (c *ColorM) UnsafeElements() ([]float32, []float32) {
 	return eb, et
 }
 
+func (c *ColorM) det() float32 {
```
hajimehoshi, line 170: `det4x4` would be more precise as this doesn't mean det for the 5x5 ColorM.

internal/affine/colorm.go
```diff
@@ -167,6 +167,158 @@ func (c *ColorM) UnsafeElements() ([]float32, []float32) {
...
+	m20 := c.body[8]
+	m21 := c.body[9]
+	m22 := c.body[10]
+	m23 := c.body[11]
+	m30 := c.body[12]
+	m31 := c.body[13]
+	m32 := c.body[14]
+	m33 := c.body[15]
+
+	b234234 := m22*m33 - m23*m32
```
hajimehoshi, line 192: I'm not sure the naming rule here. What does '4' mean?

internal/affine/colorm.go
```diff
@@ -167,6 +167,158 @@ func (c *ColorM) UnsafeElements() ([]float32, []float32) {
...
+	m31 := c.body[13]
+	m32 := c.body[14]
+	m33 := c.body[15]
+
+	m40 := c.translate[0]
+	m41 := c.translate[1]
+	m42 := c.translate[2]
+	m43 := c.translate[3]
+
+	a2334 := m32*m43 - m33*m42
```
hajimehoshi, line 248: ditto

internal/affine/colorm_test.go
```diff
@@ -79,3 +80,94 @@ func TestColorMScaleOnly(t *testing.T) {
...
+	a = a.SetElement(0, 2, 0).SetElement(1, 2, 0).SetElement(2, 2, 1).SetElement(3, 2, 2)
+	a = a.SetElement(0, 3, 0).SetElement(1, 3, 0).SetElement(2, 3, 0).SetElement(3, 3, 1)
+	a = a.SetElement(0, 4, 0).SetElement(1, 4, 0).SetElement(2, 4, 0).SetElement(3, 4, 0)
+
+	ia := (&ColorM{}).SetElement(0, 0, 1).SetElement(1, 0, -8).SetElement(2, 0, 9).SetElement(3, 0, 7)
+	ia = ia.SetElement(0, 1, 0).SetElement(1, 1, 1).SetElement(2, 1, 0).SetElement(3, 1, -4)
+	ia = ia.SetElement(0, 2, 0).SetElement(1, 2, 0).SetElement(2, 2, 1).SetElement(3, 2, -2)
+	ia = ia.SetElement(0, 3, 0).SetElement(1, 3, 0).SetElement(2, 3, 0).SetElement(3, 3, 1)
+	ia = ia.SetElement(0, 4, 0).SetElement(1, 4, 0).SetElement(2, 4, 0).SetElement(3, 4, 0)
+
```
hajimehoshi, line 141: Defining elements by a slice and initializing ColorMs with them would be more readable.

internal/affine/colorm.go
gabstv: It's from a matrix inversion generator. I'm not sure of the naming pattern, but I don't have an idea for a replacement either. It is used to avoid calculating this more than once (since it's used more than once).

internal/affine/colorm.go
gabstv: As I replied above. I could replace it with m32m43m33m42 (it doesn't look good but at least it makes more sense). Let me know what you think

internal/affine/colorm.go
hajimehoshi: > It's from a matrix inversion generator.

What generate did you use?

internal/affine/colorm_test.go
gabstv: But the test package is `package affine_test` so it doesn't have access to the private `body` and `translate`. I would still have to call SetElement multiple times, right?

hajimehoshi: Yes, so you'd need to call SetElement by for-loop.

```
es := [4][5]float32{ { ... }, { ... }, ... }
for j := 0; j < 5; j ++ {
  for i := 0; i < 4; i++ {
    ia = ia.SetElement(i, j, es[j][i])
  }
}
```

internal/affine/colorm.go
gabstv: https://github.com/willnode/N-Matrix-Programmer

hajimehoshi: https://github.com/willnode/N-Matrix-Programmer/blob/master/Info/Matrix_4x4.txt '4' doesn't come from this generator.

gabstv: but it's the 5x5:
https://github.com/willnode/N-Matrix-Programmer/blob/master/Info/Matrix_5x5.txt

hajimehoshi: I think we adopted the way to invert the body part (4x4) of the matrix.

hajimehoshi: Or, if the way I suggested and the way of inverting a 5x5 matrix are not so different, that's fine. I'm not sure. What do you think?

hajimehoshi: At least, this function calculates a det of a 4x4 matrix, right?

gabstv: It's the 5x5 without the last column. I removed the multiplications by 0 and 1 (m44).

gabstv: The det is 4x4

hajimehoshi: Let me confirm: I think we have adopted the way of `M_translate^(-1) * M_body^(-1)`, right?

gabstv: You suggested it, but I removed the 5th row and doing a 4x4 inversion, inverting a 4x1 and then multiplying them now doesn't seem to be necessary. It's about the same number of ops combined.

hajimehoshi: Ah OK, acknowledged. As long as the number of ops doesn't change, both are fine. Thank you for explanation!

Then, could you revert det4x4 to det, which is more straight forward? Sorry for the confusion.

gabstv: Are you sure? The det is indeed technically 4x4 because it uses the body (that is 4x4). Let me know

hajimehoshi: Actually this calculates the det of the 5x5 matrix, but with optimization assuming the last column is (0,0,0,0,1), right? The local variable names actually indicates this process as you said. We should rename this function's name back, or, rename the local variables to represent inversion of a 4x4 matrix. What do you think?

gabstv: I renamed the function to avoid further confusion

hajimehoshi:
FYI: I'm adding some refactoring: 2e8e35750dad6f25f70e3e259c7139eac93cf76f

PR #1251, opened 2020-07-18
Add CompositeModeMultiply

### LINKED ISSUE #410: Support blend modes like 'multiply'
https://drafts.fxtf.org/compositing-2/#blending

Ebiten supports composite mode (e.g. xor) but not blend mode (e.g. multiply). Is it possible to do the same way as CSS with OpenGL?

nanoslayer:
This pull request adds a new composite mode called `CompositeModeMultiply`, which multiplies the source color with the destination color.

I have tested my changes on Linux and Windows only.

Closes #410

### DIFF
--- a/graphics.go
+++ b/graphics.go
@@ -85,4 +85,8 @@ const (
 	// Sum of source and destination (a.k.a. 'plus' or 'additive')
 	// c_out = c_src + c_dst
 	CompositeModeLighter CompositeMode = CompositeMode(driver.CompositeModeLighter)
+
+	// The product of source and destination (a.k.a 'multiply blend mode')
+	// c_out = c_src * c_dst
+	CompositeModeMultiply CompositeMode = CompositeMode(driver.CompositeModeMultiply)
 )
--- a/internal/driver/compositemode.go
+++ b/internal/driver/compositemode.go
@@ -35,8 +35,9 @@ const (
 	CompositeModeDestinationAtop
 	CompositeModeXor
 	CompositeModeLighter
+	CompositeModeMultiply
 
-	CompositeModeMax = CompositeModeLighter
+	CompositeModeMax = CompositeModeMultiply
 )
 
 type Operation int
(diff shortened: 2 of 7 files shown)


nanoslayer:
I have tested my changes on Linux. I cannot test on Windows or macOS. Mobile and JS support is also currently unimplemented.

I also haven't tested the blend modes with the shader, as I haven't need to use a shader yet.

hajimehoshi:
Let's reach an agreement at #410 first.

internal/driver/compositemode.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 The Ebiten Authors
+// Copyright 2020 The Ebiten Authors
```
hajimehoshi, line 1: You don't have to update the existing file's year.

graphics.go
```diff
@@ -85,4 +85,7 @@ const (
 	// Sum of source and destination (a.k.a. 'plus' or 'additive')
 	// c_out = c_src + c_dst
 	CompositeModeLighter CompositeMode = CompositeMode(driver.CompositeModeLighter)
+
+	// The product of source and destination (a.k.a 'multiply blend mode')
```
hajimehoshi, line 89: Add this line

`c_out = c_src * c_dst`

internal/driver/compositemode.go
```diff
@@ -35,8 +35,7 @@ const (
 	CompositeModeDestinationAtop
 	CompositeModeXor
 	CompositeModeLighter
-
-	CompositeModeMax = CompositeModeLighter
+	CompositeModeMultiply
```
hajimehoshi, line 38: Keep `CompositeModeMax`.

nanoslayer:
> Would it be possible to implement this for Metal? If not, I'll do that.

I'm not sure how to do that exactly.

hajimehoshi:
> > Would it be possible to implement this for Metal? If not, I'll do that.
> I'm not sure how to do that exactly.

OK I'll do this after merging this.

PR #1257, opened 2020-07-21
Add various build manifests

nanoslayer:
This pull request adds various build manifests that will be used by sourcehut builds. It tests the installation of Ebiten on the following distros:

- Alpine Linux
- Arch Linux
- Debian
- Fedora
- Ubuntu

If you configure sourcehut dispatch with this repository, they will be run after every commit.

### DIFF
--- a/.builds/alpine.yml
+++ b/.builds/alpine.yml
@@ -1,0 +1,17 @@
+image: alpine/latest
+packages:
+  - alsa-lib-dev
+  - libx11-dev
+  - libxrandr-dev
+  - libxcursor-dev
+  - libxinerama-dev
+  - libxi-dev
+  - mesa-dev
+  - pkgconf
+  - go
+sources:
+  - https://github.com/hajimehoshi/ebiten
+tasks:
+  - build: |
+      cd ebiten
+      go build . ./audio
--- a/.builds/arch.yml
+++ b/.builds/arch.yml
@@ -1,0 +1,16 @@
+image: archlinux
+packages:
+  - alsa-lib
+  - mesa
+  - libxrandr
+  - libxcursor
+  - libxinerama
+  - libxi
+  - pkg-config
+  - go
+sources:
+  - https://github.com/hajimehoshi/ebiten
+tasks:
+  - build: |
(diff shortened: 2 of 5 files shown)


hajimehoshi:
Thank you!

I don't know anything about sourcehut builds unfortunately. The files seem concise, but I don't want to add dependent services if possible. I thought Dockerfile would be more portable, but what do you think?

BTW, Dockerfile can work on GitHub Actions (https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action).

nanoslayer:
I'm not sure, but Docker does seem more complicated.

hajimehoshi:
OK I understand converting them into Dockerfiles would be an expensive work.

Let's use sourcehut builds as a good start, and revisit this work later when we really want to use Docker.

@nanoslayer I'd be happy if you could give me links to documentations how to use sourcehut builds. Thanks!

nanoslayer:
See the documentation for sourcehut builds: https://man.sr.ht/builds.sr.ht/#integrations
You'll have to use https://dispatch.sr.ht to setup the integration with Github.

The documentation for dispatch.sr.ht is available here: https://man.sr.ht/dispatch.sr.ht/
There is a section specifically for Github: https://man.sr.ht/dispatch.sr.ht/github.md

-- Email domain proudly hosted at https://migadu.com

.builds/alpine.yml
```diff
@@ -0,0 +1,17 @@
...
+  - libxi-dev
+  - mesa-dev
+  - pkgconf
+  - go
+sources:
+  - https://github.com/hajimehoshi/ebiten
+tasks:
+  - build: |
+      cd ebiten
+      go build
```
hajimehoshi, line 17: We also need `go build ./audio` to test the audio. `go build ./...` would be better?

nanoslayer: I'm not sure what `./...` does. `go build . ./audio/` seems to do the job.

hajimehoshi: ... means all the defendant directories.

PR #1291, opened 2020-08-10
example: dissolve shader

kyeett:
Noise image used:
noise

Result
dissolve


examples/shader/dissolve.go
```diff
@@ -0,0 +1,35 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build ignore
+
+package main
+
+var Time float
+var Cursor vec2
+var ImageSize vec2
```
hajimehoshi, line 21: This was reverted to `ScreenImage`. Sorry!

examples/shader/dissolve.go
```diff
@@ -0,0 +1,35 @@
...
+
+func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
+	// Triangle wave to go 0-->1-->0...
+	Limit := abs(2*fract(Time/3) - 1)
```
hajimehoshi, line 25: Use small caps for local variables.

examples/shader/dissolve.go
```diff
@@ -0,0 +1,35 @@
...
+	Level := image3TextureAt(texCoord).x
+
+	// Add a white border
+	if Level < Limit && Level > Limit-0.1 {
+		Alpha := image0TextureAt(texCoord).w
+		return vec4(1) * Alpha
```
hajimehoshi, line 31: `vec4(alpha)`?

examples/shader/dissolve.go
kyeett: since this is not used, this is only for consistency?

examples/shader/dissolve.go
kyeett: yeah. I tried to convey the meaning "draw white, if alpha is zero", but it is hard to understand from the code above anyway :-) 

I will use vec4(alpha)

examples/shader/dissolve.go
kyeett: done

examples/shader/dissolve.go
hajimehoshi: This is only for consistency.

hajimehoshi:
BTW, would it be possible to use OpenSimplex noise instead of adding a new image?

https://en.wikipedia.org/wiki/OpenSimplex_noise
https://github.com/ojrac/opensimplex-go/blob/master/opensimplex_base.go#L17-L128

kyeett:
It is possible, but perhaps not very convenient. In a real world case, I would use this approach of sending in a pre-rendered image, as it is faster (and straight forward)

If we don't want to send in an image we must
1. Send it in as a uniform (not sure what format) 
OR
2. Make the simplex function inside the shader. 

You suggest 2nd option?

hajimehoshi:
I suggested 2. OK I understand that the shader would be more complex. Let's go with binding-an-image way.

BTW, what is the license of the noise.png? Could you modify examples/resources/images/license.md?

kyeett:
I don't remember where I got it from! I'll look, or generate a new one with a new source.

kyeett:
Created a new noise image using OpenSimplex. It is not exactly the same, but I like this look as well.

noise

dissolve2

PR #1292, opened 2020-08-11
net/http OpenFile - OpenNetFile

tomlister:
#1154 
Testing performed with:
```go
stream, err := ebitenutil.OpenNetFile("http://mysafeinfo.com/api/data?list=englishmonarchs&format=json")
if err != nil {
	log.Fatal(err)
}
body, err := ioutil.ReadAll(stream)
if err != nil {
	log.Fatal(err)
}
fmt.Println(string(body))
```

### DIFF
--- a/ebitenutil/file_notjs.go
+++ b/ebitenutil/file_notjs.go
@@ -20,6 +20,8 @@
 package ebitenutil
 
 import (
+	"io"
+	"net/http"
 	"os"
 	"path/filepath"
 )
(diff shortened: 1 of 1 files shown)


hajimehoshi:
My intention was to fix OpenFile of the _js.go version.

tomlister:
> My intention was to fix the _js.go version of OpenFile.

Oh right my bad. What exactly's wrong with the _js.go version?

hajimehoshi:
https://github.com/hajimehoshi/ebiten/blob/master/ebitenutil/file_js.go

This uses XMLHttpRequest and is unnecessarily complicated.

Noofbiz:
Has net/http been ported to wasm yet? The reason it's like that is likely because gopherjs didn't support net/http :P

hajimehoshi:
Wasm supports it, but GopherJS may not. Good point!

hajimehoshi:
Looks like GopherJS supports this?

https://github.com/gopherjs/gopherjs/blob/master/compiler/natives/src/net/http/http.go

tomlister:
Gotcha. I'll implement it tomorrow morning.

Issue #1132, opened 2020-04-09
Add a mode to skip clearing the screen at the beginning of a frame


hajimehoshi:
EDIT: The title was:  skip sending vertices when they are out of the region

Inspired by https://github.com/hajimehoshi/ebiten/issues/1130. In theory, we can skip sending the vertices to GPU if they are out of the region. I'm not sure how efficiently we can do it.

hajimehoshi:
Idea: Calculate a circumscribed circle of the region rectangle and a circumscribed circle of a triangle, and see whether those circles are excluded or not.

hajimehoshi:
Ah, in the Pixel version, rendering onto the screen happens only when necessary, maybe?

https://github.com/jrcichra/gollercoaster/blob/62570d8957c033fe121c111bc883e7df919eb24a/game/game.go#L111-L127

jrcichra:
Yeah and camera movement in pixel, it was easier for me to move around without redrawing.

jrcichra:
Because of this line not taking any time: https://github.com/jrcichra/gollercoaster/blob/62570d8957c033fe121c111bc883e7df919eb24a/game/game.go#L129

hajimehoshi:
Thanks!

> Yeah and camera movement in pixel, it was easier for me to move around without redrawing.

Ebiten doesn't have a mode not to clear the screen at the beginning of a frame. I'll consider this.

For the camera movement, Ebiten can do a similar thing by `(*GeoM).Concat`. I think this is a more suitable solution for Ebiten than introducing a global matrix.

beoran:
Actually I didn't realize Ebiten does this, but this is normally not needed. If you use a tile map, then this will cover the background completely. Clearing the screen at the beginning of the frame should only happen if the back ground is set to a solid color or similar.

tomlister:
I have a commit that adds isClearingSkipped and SetClearingSkipped. Would you like me to PR?

hajimehoshi:
Yes, I appreciate it!

I prefer the name `IsClearingScreenSkipped` and `SetClearingScreenSkipped`.

tomlister:
At the moment this is how it's implemented.
```go
if game, ok := c.game.(interface{ Draw(*Image) }); ok {
	if !IsClearingSkipped() {
		c.offscreen.Clear()
	}
	game.Draw(c.offscreen)
}
```
Is there any other clear instruction that should be skipped? I noticed that there were a couple of them in uicontext.

[...]

PR #1302, opened 2020-08-13
Implement switchable screen clearing.

tomlister:
#1132


run.go
```diff
@@ -101,6 +102,22 @@ func setDrawingSkipped(skipped bool) {
 	atomic.StoreInt32(&isDrawingSkipped, v)
 }
 
+// SetClearingScreenSkipped enables or disables the clearing of the screen at the beginning of each frame.
+func SetClearingScreenSkipped(skipped bool) {
```
hajimehoshi, line 106:
```
// The default value is false and the screen is cleared each frame by default.
//
// SetClearingScreenSkipped is concurrent-safe.
```

Issue #1367, opened 2020-09-26
Add a single-thread mode by introducing a new build tag


jakecoffman:
This project is super cool so I finally started porting the cp-examples over to ebiten. With vsync off I noticed the [ebiten tumble](https://github.com/jakecoffman/cpebiten/blob/gh-pages/tumble/tumble.go) ran much slower than the [cp-example tumble](https://github.com/jakecoffman/cp-examples/blob/master/tumble/tumble.go) that builds the triangles every single frame. 

The cp-example was getting ~1600 FPS and ebiten was at ~600 FPS. 

So I did some profiling and noticed some channel related locks. From what I understand most game engines are single threaded, so I [removed the threading](https://github.com/jakecoffman/ebiten/commit/1d10970a8c48fbcb01860d6ea9ee90d97f2087cf) and the ebiten version is now getting ~2100 FPS. 

Is there a specific design decision that lead to threading in the GLFW code? If there's a good reason maybe add an option to disable threading to unlock peak performance?

hajimehoshi:
Hi,

Thank you for the information!

OpenGL functions (and probably Metal functions) must be called on the specific OS thread. Without this syncing, the functions can be called on a wrong OS thread since goroutines can use arbitrary threads.

hajimehoshi:
So probably we can disable the threading only when `GOMAXPROCS=1`. Let me think.

jakecoffman:
I am aware of that restriction. I think if ebiten allows the user to disable threading and document doing this will mean that any goroutines they spawn will need to communicate back to the main thread then it's on the user to obey or get a crash.

hajimehoshi:
My concern is that disabling threading for performance is counterintuitive and doesn't sound a way to go for optimization.

Just an idea: In the implementation, the exported functions in the `internal/shareable` package already processes with a mutex. Then, we can create a big main loop inside `internal/shareable` so that we can reduce bunch of `internal/thread.Thread.Call`.

jakecoffman:
I tried this on a Mac and found that using threads the tumble demo used near 50% CPU, and without threads it used 15% CPU.

## With threads

<img width="956" alt="Screen Shot 2020-09-26 at 2 58 44 PM" src="https://user-images.githubusercontent.com/886768/94349351-36db9d00-0009-11eb-8365-b089c86c7a91.png">

## Without threads

[...]

PR #1396, opened 2020-10-18
ebiten: single thread via build tag

jakecoffman:
updates #1367


audio/audio.go
```diff
@@ -35,9 +35,9 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"github.com/hajimehoshi/ebiten/v2/internal/sync"
```
hajimehoshi, line 38: Could you move the import path `github.com/hajimehoshi/ebiten...` to a different below group for consistency?

internal/thread/thread.go
```diff
@@ -31,6 +34,7 @@ func New() *Thread {
 	return &Thread{
 		funcs:   make(chan func() error),
 		results: make(chan error),
+		ch:      make(chan struct{}),
```
hajimehoshi, line 37: When is this channel used?

hajimehoshi (commented):
Thanks!

audio/audio.go
```diff
@@ -35,9 +35,9 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"github.com/hajimehoshi/ebiten/v2/sync"
```
hajimehoshi, line 38: I meant 

```
import (
	"bytes"
	"errors"
	"fmt"
	"io"
	"runtime"
	"sync"
	"time"

	"github.com/hajimehoshi/ebiten/v2/internal/sync"
)
```

Please keep `sync` an internal package.

jakecoffman: Ah, will do!

internal/shareable/image.go
```diff
@@ -18,14 +18,14 @@ import (
...
 	"runtime"
-	"sync"
 
 	"github.com/hajimehoshi/ebiten/v2/internal/affine"
 	"github.com/hajimehoshi/ebiten/v2/internal/driver"
 	"github.com/hajimehoshi/ebiten/v2/internal/graphics"
 	"github.com/hajimehoshi/ebiten/v2/internal/hooks"
 	"github.com/hajimehoshi/ebiten/v2/internal/packing"
 	"github.com/hajimehoshi/ebiten/v2/internal/restorable"
+	"github.com/hajimehoshi/ebiten/v2/internal/sync"
```
hajimehoshi, line 28: This package is tricky due to `runtime.SetFinalizer`. Finalizers run in different goroutines. Is there a guarantee that the main goroutines and the finalizer goroutines never race?

internal/thread/thread_noop.go
```diff
@@ -0,0 +1,50 @@
...
+	"errors"
+)
+
+// Thread represents an OS thread.
+type Thread struct{}
+
+// New creates a new thread.
+//
+// It is assumed that the OS thread is fixed by runtime.LockOSThread when New is called.
+func New() *Thread {
```
hajimehoshi, line 29: Remove the comments since they don't explain this Thread type.

internal/thread/thread.go
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// +build !ebitensinglethread
```
hajimehoshi, line 15: Instead of specifying a build tag here, I'd create `NoopThread`, rename `Thread` to `OSThread`, `New` to `NewOSThread`, and define `Thread` interface that includes `Loop` and `Call` function.

internal/uidriver/glfw/run.go
```diff
@@ -0,0 +1,61 @@
+// Copyright 2020 The Ebiten Authors
```
hajimehoshi, line 1: Rename `run.go` to `run_notsinglethread.go`.

internal/sync/sync_noop.go
```diff
@@ -0,0 +1,33 @@
...
+
+func (m *Mutex) Lock()   {}
+func (m *Mutex) Unlock() {}
+
+type RWMutex struct{}
+
+func (m *RWMutex) Lock()    {}
+func (m *RWMutex) Unlock()  {}
+func (m *RWMutex) RLock()   {}
+func (m *RWMutex) RUnlock() {}
```
hajimehoshi, line 31: BTW, how much would the performance be improved without mutex?

internal/shareable/image.go
jakecoffman: There is no guarantee, I will switch it back to the stdlib sync package.

internal/sync/sync_noop.go
jakecoffman: I will do some performance testing before proceeding.

hajimehoshi: OK. I recommend to focus only on the 'run-loop' part, and then consider the mutex part in another PR later. What do you think?

internal/uidriver/glfw/run_notsinglethread.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/hajimehoshi/ebiten/v2/internal/driver"
+	"github.com/hajimehoshi/ebiten/v2/internal/graphicscommand"
+	"github.com/hajimehoshi/ebiten/v2/internal/thread"
+)
+
+func (u *UserInterface) Run(uicontext driver.UIContext) error {
+	u.context = uicontext
+
+	// Initialize the main thread first so the thread is available at u.run (#809).
+	u.t = thread.NewOSThread()
```
hajimehoshi, line 32: `thread.New` is also called in internal/uidriver/mobile so please fix that. Note that mobiles cannot enable the single-thread mode since a rendering thread is given externally. Then, just renaming it is fine. Don't forget to rename `*thread.Thread` to `*thread.OSThread` (or `thread.Thread`).

internal/thread/thread.go
```diff
@@ -58,7 +64,23 @@ func (t *Thread) Loop() {
...
 //
 // Call blocks if Loop is not called.
-func (t *Thread) Call(f func() error) error {
+func (t *OSThread) Call(f func() error) error {
 	t.funcs <- f
 	return <-t.results
 }
+
+// NoOpThread is used to disable threading.
+type NoOpThread struct{}
```
hajimehoshi, line 73: I prefer `Noop` to `NoOp`.

hajimehoshi (commented):
Please add an explanation about the build tag `ebitensinglethread` to `doc.go`.

internal/uidriver/glfw/run_notsinglethread.go
jakecoffman: I think I did this already or I am misunderstanding.

hajimehoshi: You are right. Thanks!

doc.go
```diff
@@ -70,4 +70,7 @@
 // number of graphics commands affects the performance of your game.
 //
 // `ebitengl` forces to use OpenGL in any environments.
+//
+// `ebitensinglethread` disables Ebiten's thread safety to unlock maximum performance. If you use this you will have
+// to manage threads yourself.
```
hajimehoshi, line 75: Please add a note that all the 'concurrent-safe' functions like IsKeyPressed will no longer be concurrent safe with ebitensinglethread.

hajimehoshi (approved):
LGTM, thanks!

PR #1446, opened 2020-12-21
Fix typos, and use panic instead of log.Fatal in doc.go

fuyuaki:
The use of panic is to match the image.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # Contributing to Ebiten
 
-Ebiten is an open source project and we would appricate your conributions!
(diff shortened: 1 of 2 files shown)


hajimehoshi:
Thank you for fixing the typos, but why do you prefer `panic` to `log.Fatal`?

If we want to replace it, let's create a separate PR for that.

hajimehoshi:
If I don't get your answer until tomorrow, I'll pick the typo fixes from your PR instead of merging it. Thanks.

fuyuaki:
I thought it would be good the code would match the overview image. Not about preference. I overthought about it.

hajimehoshi (approved):
LGTM, thanks!

PR #1543, opened 2021-03-14
internal/graphicscommand: fix missing composite mode

Humpheh:
Using `-tags=ebitendebug` would log the string: 
```
  %!s(PANIC=String method: graphicscommand: invalid composite mode: 13)
```
When using the multiply composite mode. This adds it to the string function.


hajimehoshi (approved):
LGTM, thanks!

PR #1547, opened 2021-03-23
Add a new Touch example

griffithsh:
I was looking for an example on how to use the Touch APIs in ebiten to implement complex touch gestures, but was unable to find any. This PR adds the examples I was looking for.

https://gophers.slack.com/archives/C19491JS2/p1615525929193900

### DIFF
--- a/examples/touch/main.go
+++ b/examples/touch/main.go
@@ -1,0 +1,52 @@
+// Copyright 2021 The Ebiten Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build example
+
+package main
+
+import (
+	"log"
+
+	"github.com/hajimehoshi/ebiten/v2"
+)
+
+const (
+	screenWidth  = 640
+	screenHeight = 480
+)
+
+type Game struct {
+}
+
+func (g *Game) Update() error {
+	return nil
+}
+
+func (g *Game) Draw(screen *ebiten.Image) {
+}
+
+func (g *Game) Layout(outsideWidth, outsideHeight int) (int, int) {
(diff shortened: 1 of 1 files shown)


examples/touch/main.go
```diff
@@ -0,0 +1,253 @@
...
+	// neither begun nor ended in this frame.
+	for _, id := range ebiten.TouchIDs() {
+		t := g.touches[id]
+		t.duration = inpututil.TouchPressDuration(id)
+		t.currX, t.currY = ebiten.TouchPosition(id)
+	}
+
+	// Interpret the raw touch data that's been collected into g.touches into
+	// gestures like two-finger pinch or single-finger pan.
+	if len(g.touches) == 2 {
```
hajimehoshi, line 128: Use `switch len(g.touches)` and `case`

examples/touch/main.go
```diff
@@ -0,0 +1,253 @@
...
+	zoom float64
+
+	touches map[ebiten.TouchID]*touch
+	pinch   *pinch
+	pan     *pan
+	taps    []tap
+}
+
+func (g *Game) Update() error {
+	g.taps = []tap{}
```
hajimehoshi, line 83: This initialization is not needed (appending works with nil)

hajimehoshi (commented):
Thank you very much!

examples/touch/main.go
griffithsh: The any taps from previous frames need to be removed here, but we can definitely assign `nil` instead of an empty slice.

examples/touch/main.go
```diff
@@ -0,0 +1,253 @@
...
+func (g *Game) Draw(screen *ebiten.Image) {
+	op := &ebiten.DrawImageOptions{}
+
+	// Apply zoom.
+	op.GeoM.Scale(g.zoom, g.zoom)
+
+	// Apply pan.
+	op.GeoM.Translate(g.x, g.y)
+
+	// Centre the image (corrected by the current zoom).
```
griffithsh, line 211:
```suggestion
	// Center the image (corrected by the current zoom).
```
Oops, I'm assuming we should stick to US English?

examples/touch/main.go
hajimehoshi: Ah, you're right. This should be initialized every tick.

Then, I'd use `g.taps = g.taps[:0]` to avoid allocations when appending. This should work even if `g.taps` is nil.

examples/touch/main.go
hajimehoshi: Yes, we use US English in this project.

hajimehoshi (approved):
LGTM, thanks!

PR #1600, opened 2021-04-19
Fallback to simpler logic when trying to get DPI from monitor and failing

corfe83:
This fixes a crash on at least one of my customer's machines. I don't have access to the machine, so I don't know why it's failing.

Instead of crashing, just falls back to other DPI.


hajimehoshi:
Thanks!

1. Could you file the bug first for the recording?
2. Could you add a comment about this bug with the filed issue?

hajimehoshi:
The comment will be like:

`// monitorFromRect can fail in some environments (#1601)`

hajimehoshi:
friendly ping

corfe83:
Thanks for feedback, I've filed a bug and added the comments.

hajimehoshi (approved):
LGTM, thanks!

Issue #1611, opened 2021-04-20
Fullscreen by `requestFullscreen` on browsers


hajimehoshi:
Now `SetFullscreen` doesn't work on browsers. I'm thinking if an Ebiten application in Wasm in an iframe could be fullscreen by `SetFullscreen(true)` by using Web API `requestFullscreen`.

shoebsd31:
> _No description provided._

Hi could you please add some description ?

hajimehoshi:
> Hi could you please add some description ?

Updated :)

tomlister:
Hi,

As requestFullscreen works on any element, simply calling requestFullscreen on the canvas element will trigger the browser to make ebiten fullscreen.
Caveats:
According to MDN
> Note: This method must be called while responding to a user interaction or a device orientation change; otherwise it will fail.
As a result of this during testing I noticed that sometimes I'd have to double tap "F" to trigger a successful fullscreen. Any quick input previous to the triggering key works too, i.e clicking or pressing WASD, and then pressing "F" to trigger fullscreen.
```go
func (u *UserInterface) SetFullscreen(fullscreen bool) {
	canvas.Call("requestFullscreen")
	return
}
```

tomlister:
Updating IsFullscreen and any documentation is also necessary.

hajimehoshi:
Yeah, good point! A user gesture is required like Pointer Lock API (https://github.com/hajimehoshi/ebiten/issues/1604#issuecomment-822946880).

So, we have to handle [`fullscreenerror`](https://developer.mozilla.org/en-US/docs/Web/API/Document/fullscreenerror_event) and show an error message on the console like what Ebiten has already done for Pointer Lock API.

> Updating IsFullscreen and any documentation is also necessary.

Yes!

tomlister:
SetFullscreen
- Should check if the canvas is initialised
- Enter fullscreen when fullscreen argument is true
- Exit fullscreen when fullscreen argument is false
```go
func (u *UserInterface) SetFullscreen(fullscreen bool) {
	if !canvas.Truthy() {
		return
	}
	if fullscreen {
		canvas.Call("requestFullscreen")
		return
	}
	canvas.Call("exitFullscreen")
	return
}
```

hajimehoshi:
Yes, the direction seems correct :-)

tomlister:
We can also check if the current ebiten canvas is fullscreen or not, but for now it just checks if there is any element in fullscreen.
```
[...]

PR #1623, opened 2021-04-27
Issue 1611

tomlister:
In relation to #1611 
- [x] All code has been implemented
- [x] Further enhancements have been made to the original proposals
- [x] Code documentation has been updated

### DIFF
--- a/internal/uidriver/js/ui_js.go
+++ b/internal/uidriver/js/ui_js.go
@@ -112,7 +112,6 @@ func (u *UserInterface) SetFullscreen(fullscreen bool) {
 		f = document.Get("webkitExitFullscreen")
 	}
 	f.Call("bind", document).Invoke()
-	return
 }
 
 func (u *UserInterface) IsFullscreen() bool {
--- a/internal/uidriver/js/ui_js.go
+++ b/internal/uidriver/js/ui_js.go
@@ -408,6 +408,10 @@ func init() {
 		js.Global().Get("console").Call("error", "pointerlockerror event is fired. 'sandbox=\"allow-pointer-lock\"' might be required. There is a known issue on Safari (hajimehoshi/ebiten#1604)")
 		return nil
 	}))
+	document.Call("addEventListener", "fullscreenerror", js.FuncOf(func(this js.Value, args []js.Value) interface{} {
+		js.Global().Get("console").Call("error", "fullscreenerror event is fired. This function on browsers must be called as a result of a gestural interaction or orientation change.")
+		return nil
+	}))
 }
 
 func setWindowEventHandlers(v js.Value) {
--- a/internal/uidriver/js/ui_js.go
+++ b/internal/uidriver/js/ui_js.go
@@ -110,7 +110,7 @@ func (u *UserInterface) IsFullscreen() bool {
 	if !document.Truthy() {
 		return false
 	}
-	if document.Get("fullscreenElement").IsNull() {
+	if !document.Get("fullscreenElement").Truthy() {
 		return false
 	}
 	return true
--- a/internal/uidriver/js/ui_js.go
+++ b/internal/uidriver/js/ui_js.go
@@ -99,10 +99,19 @@ func (u *UserInterface) SetFullscreen(fullscreen bool) {
 		return
 	}
 	if fullscreen {
-		canvas.Call("requestFullscreen")
+		f := canvas.Get("requestFullscreen")
+		if !f.Truthy() {
+			f = canvas.Get("webkitRequestFullscreen")
+			return
+		}
+		f.Call("bind", canvas).Invoke()
 		return
 	}
-	document.Call("exitFullscreen")
+	f := document.Get("exitFullscreen")
+	if !f.Truthy() {
+		f = document.Get("webkitExitFullscreen")
+	}
+	f.Call("bind", document).Invoke()
 	return
 }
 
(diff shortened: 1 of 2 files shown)


tomlister:
Well that build status is bothersome.

hajimehoshi:
The build bots are failing:

```
internal/uidriver/js/ui_js.go:113:38: document.Get("fullscreenElement").IsNull undefined (type js.Value has no field or method IsNull)
```

`IsNull` doesn't exist in Go 1.13 era. Let's use `Truthy`.

hajimehoshi:
Thank you for the PR!

Have you tested this with an actual application BTW?

tomlister:
> Thank you for the PR!
> 
> Have you tested this with an actual application BTW?

Yes, with examples/moire for toggling and examples/fullscreen for the error.

run.go
```diff
@@ -253,22 +253,22 @@ func SetCursorShape(shape CursorShapeType) {
...
 
-// SetFullscreen changes the current mode to fullscreen or not on desktops.
+// SetFullscreen changes the current mode to fullscreen or not on desktops and browsers.
 //
 // In fullscreen mode, the game screen is automatically enlarged
 // to fit with the monitor. The current scale value is ignored.
 //
 // On desktops, Ebiten uses 'windowed' fullscreen mode, which doesn't change
 // your monitor's resolution.
 //
```
hajimehoshi, line 270: Let's add an explanation that fullscreening on browsers require user gestures, and the behavior depends on browser implementations. (Probably this works in Safari in a more limited situation than Chrome, unfortunately)

tomlister: How about
```go
// On browsers, triggering fullscreen requires a user gesture otherwise an error is thrown.
// This behaviour varies across browser implementations.
```
Lmk if this sounds good and I'll commit it.

tomlister:
Actually on Safari, this doesn't work due a webkit prefix... 🤦🏻‍♂️
```
panic: syscall/js: Value.Call: property requestFullscreen is not a function, got undefined
```

hajimehoshi:
What about `webkitRequestFullscreen` for Safari?

run.go
hajimehoshi: > an error is thrown

This is a little ambiguous. Just `otherwise SetFullscreen does nothing but leave an error message on console`?

tomlister:
Ah shit, wish i saw that 10 mins ago.
Just made this abomination. 😂
```go
func setFullscreenWebkit(fullscreen bool) {
	if err := recover(); err != nil {
		if fullscreen {
			canvas.Call("webkitRequestFullscreen")
			return
		}
		document.Call("webkitExitFullscreen")
	}
}

func (u *UserInterface) SetFullscreen(fullscreen bool) {
	defer setFullscreenWebkit(fullscreen)
	if !canvas.Truthy() {
		return
	}
	if !document.Truthy() {
		return
	}
	if fullscreen {
		canvas.Call("requestFullscreen")
		return
	}
	document.Call("exitFullscreen")
	return
}
```

tomlister:
100% using your implementation.

tomlister:
Okay, weird. Seems to not want to go into fullscreen in safari :/ Works on Chrome.

tomlister:
I also had to modify IsFullscreen
```go
func (u *UserInterface) IsFullscreen() bool {
	if !document.Truthy() {
		return false
	}
	if !document.Get("fullscreenElement").Truthy() {
		if !document.Get("webkitFullscreenElement").Truthy() {
			return false
		}
		return false
	}
	return true
}
```

hajimehoshi:
OK I'm waiting for your latest update!

tomlister:
I can't get this to work on safari :/

tomlister:
Screen Shot 2021-04-27 at 8 27 10 pm
Manually triggering in console causes this graphical effect

hajimehoshi:
> I can't get this to work on safari :/

If you see an error message, that's fine. "User gesture" triggering varies among browsers.

> Manually triggering in console causes this graphical effect on moire example

Just black and white? That's pretty odd.

tomlister:
We can always patch this if a fix becomes available. Might be a good idea to open an issue on webkit regarding the bugs.

internal/uidriver/js/ui_js.go
```diff
@@ -92,11 +92,40 @@ func (u *UserInterface) ScreenSizeInFullscreen() (int, int) {
...
+		}
+		f.Call("bind", canvas).Invoke()
+		return
+	}
+	f := document.Get("exitFullscreen")
+	if !f.Truthy() {
+		f = document.Get("webkitExitFullscreen")
+	}
+	f.Call("bind", document).Invoke()
+	return
```
hajimehoshi, line 115: This return is not needed.

internal/uidriver/js/ui_js.go
```diff
@@ -92,11 +92,40 @@ func (u *UserInterface) ScreenSizeInFullscreen() (int, int) {
...
 func (u *UserInterface) IsFullscreen() bool {
-	return false
+	if !document.Truthy() {
+		return false
+	}
+	if !document.Get("fullscreenElement").Truthy() {
+		if !document.Get("webkitFullscreenElement").Truthy() {
+			return false
+		}
+		return false
```
hajimehoshi, line 126:
```
if !document.Get("fullscreenElement").Truthy() && if !document.Get("webkitFullscreenElement").Truthy() {
    return false
}
```
?

tomlister: Long past few days. My bad. Not on my A game at the moment.

hajimehoshi (approved):
LGTM, thanks!

Issue #1655, opened 2021-05-29
ebiten: Performance regression due to ColorM cache


jstaf:
I tried upgrading to 2.1.1 today and there is a massive performance regression caused by https://github.com/hajimehoshi/ebiten/issues/1474 - my game goes from a steady 60fps down to 25fps. (I'm doing a lot of ColorM scaling each frame to draw primitive shapes like lines, circles, etc.)

Can the changes to ColorM caching be reverted? I have pinpointed the performance regression to this commit: https://github.com/hajimehoshi/ebiten/commit/a6dd6196b460dfd1e576588fa9c6080deaa2a223 (using `go get github.com/hajimehoshi/ebiten/v2@14abc28d3ab97ee7d8976f11019fe5a38b150ea0` works great, `go get github.com/hajimehoshi/ebiten/v2@a6dd6196b460dfd1e576588fa9c6080deaa2a223` results in the massive FPS drop described above). I can supply profiles if that helps.

hajimehoshi:
Thanks, I'll take a look.

hajimehoshi:
I was wondering how many ColorM you created per frame on average. I guess this can be an issue only when you create extremely many ColorMs per frame...

jstaf:
I don't have an exact count, but its somewhere in the range of 10-20k ColorM per frame. 

The reason I'm doing this is that I need a fast way of drawing lines/shapes (https://pkg.go.dev/github.com/hajimehoshi/ebiten/v2@v2.1.1/vector is too slow and doesn't allow drawing color gradients or plain-old lines). I am basically doing a modified version of this to draw large numbers of lines/shapes quickly: https://github.com/hajimehoshi/ebiten/blob/d05bd7a58c324e5162406328f9a28595a4a1aa15/ebitenutil/shapes.go#L45

It would be *extremely* helpful to have a fast way of drawing lines and triangles that did not rely on DrawTriangles or DrawImage. I don't know too much about OpenGL, but I'm assuming there's a faster way of directly drawing these without using a 1px white image to draw from. I would be comfortable keeping the changes to ColorM if there was a better/faster way of drawing lines and triangles that didn't rely on ColorM to set a color (or use DrawTriangles/DrawImage at all).

hajimehoshi:
OK I reverted the problematic commit (and its related commit). I'll consider another way to reduce the memory allocation.

[...]

PR #1659, opened 2021-05-30
Reimplement DrawLine using DrawTriangles

jstaf:
This provides a DrawLine implementation that doesn't run into the performance issues from #1655 (because it doesn't use ColorM), and adds an extra utility function to draw a line with variable width. Though it looks a little dodgy if you draw a wide line exactly horizontally or vertically (it draws a parallelogram instead of a rectangle), it looks quite nice when used to draw paths and gradients. I think the only example that uses DrawLine is the raycasting one and you can't tell the difference too much unless you look really closely, so I attached a screenshot that demos the new line drawing from one of my projects. 

Screenshot from 2021-05-29 20-05-44

### DIFF
--- a/ebitenutil/shapes.go
+++ b/ebitenutil/shapes.go
@@ -17,7 +17,6 @@ package ebitenutil
 import (
 	"image"
 	"image/color"
-	"math"
 
 	"github.com/hajimehoshi/ebiten/v2"
 	"github.com/hajimehoshi/ebiten/v2/internal/colormcache"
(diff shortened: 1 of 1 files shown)


hajimehoshi:
Hmm, I fine to reimplement `DrawLine` with `DrawTriangles`, but I didn't want `DrawLineWidth` since functions `DrawLine` and `DrawRect` in `ebitendebug` are just for debugging, and I didn't intend these functions for production usages.

jstaf:
> but I didn't want DrawLineWidth since functions DrawLine and DrawRect in ebitendebug are just for debugging, and I didn't intend these functions for production usages.

Out of curiosity, why not? You say these functions aren't for production use, but there is no other way to draw a line with Ebiten. This is pretty important functionality for a game engine. For instance, Pixel has https://github.com/faiface/pixel/wiki/Drawing-shapes-with-IMDraw and LibGDX has https://github.com/libgdx/libgdx/wiki/Rendering-shapes

hajimehoshi:
> there is no other way to draw a line with Ebiten.
So I plan to polish the current vector package but now I'm pretty busy with other tasks.

> This is pretty important functionality for a game engine.
IMO rendering lines is less essential than the other features like rendering images, and actually there are some Ebiten games in production without using this feature. I understand, however, many people including you want to render vector graphics for various purposes rather than developing just a simple game.

hajimehoshi:
Let me close this. Feel free to reopen if you would like to update this. Thanks.

Issue #1688, opened 2021-06-28
SIGSEGV "rotate" example in Alpine Linux


Enrico204:
I tried to run the "rotate" example after installing Ebiten (`go run -tags=example github.com/hajimehoshi/ebiten/v2/examples/rotate`), and unfortunately there is a crash:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x58c87c]

goroutine 21 [running]:
github.com/hajimehoshi/ebiten/v2/internal/glfw.(*Window).GetAttrib(...)
        /root/go/pkg/mod/github.com/hajimehoshi/ebiten/v2@v2.1.2/internal/glfw/glfw_notwindows.go:109
github.com/hajimehoshi/ebiten/v2/internal/uidriver/glfw.(*UserInterface).loop(0x75e000, 0x0, 0x0)
        /root/go/pkg/mod/github.com/hajimehoshi/ebiten/v2@v2.1.2/internal/uidriver/glfw/ui.go:949 +0x7c
github.com/hajimehoshi/ebiten/v2/internal/uidriver/glfw.(*UserInterface).Run.func1(0x75e000, 0xc0000863c0)
        /root/go/pkg/mod/github.com/hajimehoshi/ebiten/v2@v2.1.2/internal/uidriver/glfw/run_notsinglethread.go:52 +0xd4
created by github.com/hajimehoshi/ebiten/v2/internal/uidriver/glfw.(*UserInterface).Run
        /root/go/pkg/mod/github.com/hajimehoshi/ebiten/v2@v2.1.2/internal/uidriver/glfw/run_notsinglethread.go:36 +0x185
exit status 2
```

I suppose that the `w.w` variable is `nil` here, as far as I understand by the stacktrace: https://github.com/hajimehoshi/ebiten/blob/8ab047853ef7ca7deaa9ef02570e9752c148badd/internal/glfw/glfw_notwindows.go#L109

My environment is the following:
* Alpine Linux 3.14 x86_64
* `dwm` as (tiled) window manager
* Go 1.16.5
* Ebiten v2.1.2
* I installed all dependencies listed in the installation page

The same code runs OK in Debian 10 (latest stable) using `dwm` and the same Go version. So apparently there is something about Alpine.

Let me know if you need any test :-)

hajimehoshi:
Thanks, I'll take a look!

hajimehoshi:
[...]

PR #1689, opened 2021-06-29
internal/uidriver/glfw: Bug fix: do not execute loop function if init failed

Enrico204:
In `internal/uidriver/glfw/run_notsinglethread.go`, if the `UserInterface.init()` function returns an error, the `loop` is executed regardless and the error is discarded. This behavior will hide the error returned by `init()` and might trigger some crashes (see #1688 ).

A partial fix was implemented in 6c4edf8605ea83a8c582cd228f230d1dbda213ce , however that commit alone is not enough: the code now is correctly returning the error via the `ch` channel, but it still executes the `loop()` function. This merge request skips `loop()` call if `init()` had an error.

### DIFF
--- a/internal/uidriver/glfw/run_notsinglethread.go
+++ b/internal/uidriver/glfw/run_notsinglethread.go
@@ -43,13 +43,16 @@ func (u *UserInterface) Run(uicontext driver.UIContext) error {
 
 		defer close(ch)
 
-		_ = u.t.Call(func() error {
+		err := u.t.Call(func() error {
 			if err := u.init(); err != nil {
 				ch <- err
-				return nil
+				return err
 			}
 			return nil
 		})
+		if err != nil {
+			return
+		}
 
 		if err := u.loop(); err != nil {
 			ch <- err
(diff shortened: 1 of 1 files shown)


internal/uidriver/glfw/run_notsinglethread.go
```diff
@@ -43,13 +43,16 @@ func (u *UserInterface) Run(uicontext driver.UIContext) error {
 
 		defer close(ch)
 
-		_ = u.t.Call(func() error {
+		err := u.t.Call(func() error {
```
hajimehoshi, line 46: Could you write this like this?:

```
if err := u.t.Call(func() error {
    //...
}); err != nil {
    ch <- err
    return
}
```

Enrico204: Done :-)

hajimehoshi: Thanks!

hajimehoshi (approved):
LGTM

PR #1732, opened 2021-07-27
text: Fix nil pointer dereference in drawGlyph

nanoslayer:
drawGlyph can panic when provided with a nil *ebiten.DrawImageOptions
reference. Check that the pointer is not nil before dereferencing.


hajimehoshi (approved):
LGTM, thanks!

PR #1735, opened 2021-07-28
ebitenutil: added NewImageFromReader(io.Reader)

TheMightyGit:
It's additionally useful if ebitenutil also supports an `io.Reader` method `NewImageFromReader(reader io.Reader)`.


ebitenutil/loadimage_go_1_16.go
```diff
@@ -0,0 +1,47 @@
+// Copyright 2014 Hajime Hoshi
```
hajimehoshi, line 1: Copyright 2021 The Ebiten Authors

ebitenutil/loadimage_go_1_16.go
```diff
@@ -0,0 +1,47 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build (darwin || freebsd || js || linux || windows) && !android && !ios && go1.16
+// +build darwin freebsd js linux windows
+// +build !android
+// +build !ios
```
hajimehoshi, line 18: Why are the mobiles opted out? I think this should work in every environment.

ebitenutil/loadimage_go_1_16.go
```diff
@@ -0,0 +1,47 @@
...
+
+// NewImageFromFS loads the file from a fs.FS filesystem with path and returns ebiten.Image and image.Image.
+//
+// go1.16+ only
+//
+// Image decoders must be imported when using NewImageFromFS. For example,
+// if you want to load a PNG image, you'd need to add `_ "image/png"` to the import section.
+//
+// This is useful if you've ebedded a filesystem with embed.FS.
+func NewImageFromFS(filesystem fs.FS, path string) (*ebiten.Image, image.Image, error) {
```
hajimehoshi, line 38: Hmm, while `NewImageFromReader` is fine, I tend not to want to have this.

```
f, err := filesystem.Open(path)
if err != nil {
    return err
}
defer f.Close()
img, _, err := ebitenutil.NewImageFromReader(f)
if err != nil {
    return err
}
// Use img
```

Is this so burden?

ebitenutil/loadimage_go_1_16.go
TheMightyGit: I've added ios and android - but should I instead remove all os tags?

ebitenutil/loadimage_go_1_16.go
TheMightyGit: I agree it has less value than `NewImageFromReader` (which only has a bad alternative). I'll remove it from this PR.

ebitenutil/loadimage_go_1_16.go
TheMightyGit: this file is removed now.

hajimehoshi: Ah yeah I intended to remove all the tags first. I acknowledged this file was removed.

hajimehoshi (approved):
LGTM, but is this still a draft?

PR #1775, opened 2021-08-20
Support updating the gamepad mapping.

### LINKED ISSUE #1723: Add an API to provide one's own game controller mapping
https://github.com/hajimehoshi/ebiten/commit/aa694be6f6b98106318bc5aac37d3a034a840cfa - first of all, this is awesome! Will allow me to delete a lot of code in my game.

However, for my own gamepad, I for quite a long time needed to provide my own mapping, which SDL supports by environment variable:

```
SDL_GAMECONTROLLERCONFIG=03000000d62000000228000001010000,PowerA Wired Controller for Xbox One,a:b0,b:b1,back:b6,dpdown:h0.4,dpleft:h0.8,dpright:h0.2,dpup:h0.1,guide:b8,leftshoulder:b4,leftstick:b9,lefttrigger:a2,leftx:a0,lefty:a1,rightshoulder:b5,rightstick:b10,righttrigger:a5,rightx:a3,righty:a4,start:b7,x:b2,y:b3,platform:Linux
```

Can the same mechanism be supported by ebiten? Not sure if it should be an EBITEN or a SDL variable, don't really care, but it may be nice to stick to this exact format.

Thanks!

divVerent:
This is only supported on desktops yet (on mobile standard layout isn't
implemented yet, and on the web this is the browser's responsibility).

Fixes #1723.

### DIFF
--- a/input.go
+++ b/input.go
@@ -234,6 +234,27 @@ func IsStandardGamepadLayoutAvailable(id GamepadID) bool {
 	return uiDriver().Input().IsStandardGamepadLayoutAvailable(id)
 }
 
+// UpdateStandardGamepadLayoutMappings updates the gamepad layout definitions using a set of definitions in SDL_GameControllerDB format.
+//
+// Multiple input definitions can be provided separated by newlines.
+// In particular, it is valid to pass an entire gamecontrollerdb.txt file.
+//
+// Note though that Ebiten already includes its own copy of this file,
+// so this call should only be necessary to add mappings for hardware not supported yet;
+// ideally games using the StandardGamepad functions should allow the user to provide mappings and then call this function if provided.
+// When using this facility to support new hardware, please also send a pull request to https://github.com/gabomdq/SDL_GameControllerDB to make your mapping available to everyone else.
+//
+// UpdateStandardGamepadLayoutMappings reports whether the mappings were applied, and returns an error in case any occurred while parsing the mappings.
+//
+// On platforms where gamepad mappings are not managed by Ebiten, this always returns false and nil.
+//
+// UpdateStandardGamepadLayoutMappings must be called on the main thread before ebiten.Run, and is concurrent-safe after ebiten.Run.
+//
+// Updated mappings take effect immediately even for already connected gamepads.
+func UpdateStandardGamepadLayoutMappings(mappings string) (bool, error) {
+	return uiDriver().Input().UpdateStandardGamepadLayoutMappings(mappings)
+}
+
 // TouchID represents a touch's identifier.
 type TouchID = driver.TouchID
 
--- a/internal/uidriver/glfw/input.go
+++ b/internal/uidriver/glfw/input.go
@@ -452,3 +452,26 @@ func standardButtonToGLFWButton(button driver.StandardGamepadButton) glfw.Gamepa
 		panic(fmt.Sprintf("glfw: invalid or inconvertible StandardGamepadButton: %d", button))
 	}
 }
+
+// UpdateStandardGamepadLayoutMappings can be used to provide new gamepad mappings to Ebiten.
+// The string must be in the format of SDL_GameControllerDB.
+func (i *Input) UpdateStandardGamepadLayoutMappings(mapping string) (bool, error) {
+	var err error
+	if i.ui.isRunning() {
+		err = i.ui.t.Call(func() error {
+			return i.updateStandardGamepadLayoutMappings(mapping)
+		})
+	} else {
+		err = i.updateStandardGamepadLayoutMappings(mapping)
+	}
+	return err == nil, err
+}
+
+func (i *Input) updateStandardGamepadLayoutMappings(mapping string) error {
+	if !glfw.UpdateGamepadMappings(mapping) {
+		// Would be nice if we could actually get the error string.
+		// However, go-gl currently does not provide it in any way.
+		return fmt.Errorf("glfw: could not parse or update gamepad mappings")
+	}
+	return nil
+}
--- a/internal/uidriver/js/input_js.go
+++ b/internal/uidriver/js/input_js.go
@@ -496,3 +496,9 @@ func (i *Input) IsStandardGamepadButtonPressed(id driver.GamepadID, button drive
 	}
 	return g.standardButtonPressed[button]
 }
+
+// UpdateStandardGamepadLayoutMappings is not supported for JS - the browser maintains the mappings.
+func (i *Input) UpdateStandardGamepadLayoutMappings(mapping string) (bool, error) {
+	// All OK - browser owns this though.
+	return false, nil
+}
--- a/internal/uidriver/mobile/input.go
+++ b/internal/uidriver/mobile/input.go
@@ -147,6 +147,13 @@ func (i *Input) StandardGamepadAxisValue(id driver.GamepadID, axis driver.Standa
 	return 0
 }
 
+// UpdateStandardGamepadLayoutMappings is not supported on mobile - this still needs to be implemented.
+func (i *Input) UpdateStandardGamepadLayoutMappings(mapping string) (bool, error) {
+	// TODO: Implement this (#1557)
+	// Note: NOT returning an error, as mappings also do not matter right now (all functions above return nothing is pressed anyway).
+	return false, nil
+}
+
 func (i *Input) AppendTouchIDs(touchIDs []driver.TouchID) []driver.TouchID {
 	i.ui.m.RLock()
 	defer i.ui.m.RUnlock()
--- a/internal/driver/input.go
+++ b/internal/driver/input.go
@@ -34,6 +34,7 @@ type Input interface {
 	IsStandardGamepadButtonPressed(id GamepadID, button StandardGamepadButton) bool
 	IsStandardGamepadLayoutAvailable(id GamepadID) bool
 	StandardGamepadAxisValue(id GamepadID, button StandardGamepadAxis) float64
+	UpdateStandardGamepadLayoutMappings(mapping string) (bool, error)
 	TouchPosition(id TouchID) (x, y int)
 	Wheel() (xoff, yoff float64)
 }


divVerent:
FYI - not tested yet; for now just sent for early review. Going to test early next week.

input.go
```diff
@@ -234,6 +234,15 @@ func IsStandardGamepadLayoutAvailable(id GamepadID) bool {
 	return uiDriver().Input().IsStandardGamepadLayoutAvailable(id)
 }
 
+// UpdateStandardGamepadLayoutMappings updates the gamepad layout definitions using a set of definitions in SDL_GameControllerDB format.
+//
+// This functionality is only supported on desktop systems and will just return an error elsewhere.
+//
+// UpdateStandardGamepadLayoutMappings must be called on the main thread before the main loop.
```
hajimehoshi, line 241: Is this true? We should be able to make this concurrent-safe.

input.go
```diff
@@ -234,6 +234,15 @@ func IsStandardGamepadLayoutAvailable(id GamepadID) bool {
...
+func UpdateStandardGamepadLayoutMappings(mappings string) error {
+	return uiDriver().Input().UpdateStandardGamepadLayoutMappings(mappings)
```
hajimehoshi, line 243: Just returning a boolean value should be fine. As some platforms (js) don't support it on purpose, this should not be an 'error'.

hajimehoshi (commented):
Thank you!

divVerent:
On Sat, Aug 21, 2021, 02:14 Hoshi, Hajime (星一) ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
>
> Thank you!
> ------------------------------
>
> In input.go
> <https://github.com/hajimehoshi/ebiten/pull/1775#discussion_r693310717>:
>
> > @@ -234,6 +234,15 @@ func IsStandardGamepadLayoutAvailable(id GamepadID) bool {
>  	return uiDriver().Input().IsStandardGamepadLayoutAvailable(id)
>  }
>
> +// UpdateStandardGamepadLayoutMappings updates the gamepad layout definitions using a set of definitions in SDL_GameControllerDB format.
> +//
> +// This functionality is only supported on desktop systems and will just return an error elsewhere.
> +//
> +// UpdateStandardGamepadLayoutMappings must be called on the main thread before the main loop.
>
> Is this true? We should be able to make this concurrent-safe.
>
GLFW docs say it must be run on the main thread. I could do the dance to
run it on the main thread from anywhere like many other functions do - did
not see it as necessary though.

What I do not yet know is when mapping apply when you do this when the game
is already running - e.g. does one need to reconnect the gamepad to
actually apply changes to it.

Will try, then do this change and document it properly.



------------------------------
>
> In input.go
> <https://github.com/hajimehoshi/ebiten/pull/1775#discussion_r693311425>:
>
> > @@ -234,6 +234,15 @@ func IsStandardGamepadLayoutAvailable(id GamepadID) bool {
>  	return uiDriver().Input().IsStandardGamepadLayoutAvailable(id)
>  }
>
> +// UpdateStandardGamepadLayoutMappings updates the gamepad layout definitions using a set of definitions in SDL_GameControllerDB format.
> +//
> +// This functionality is only supported on desktop systems and will just return an error elsewhere.
> +//
> +// UpdateStandardGamepadLayoutMappings must be called on the main thread before the main loop.
> +func UpdateStandardGamepadLayoutMappings(mappings string) error {
> +	return uiDriver().Input().UpdateStandardGamepadLayoutMappings(mappings)
>
> Just returning a boolean value should be fine. As some platforms (js)
> don't support it on purpose, this should not be an 'error'.
>
Will do; the reason why I went with the error is to distinguish the two
reasons "not needed here" and "failed to parse".

So to fix this, are you OK if I make it return false on mobile (not
implemeted yet), true on JS (not needed) and forward the value on desktop
(i.e. indicating parse error)?

Or would you prefer true on mobile too where standard layout is not
implemented at all yet?

hajimehoshi:
> What I do not yet know is when mapping apply when you do this when the game
> is already running - e.g. does one need to reconnect the gamepad to
> actually apply changes to it.
> Will try, then do this change and document it properly.

Ah good point. I appreciate your trial.

> So to fix this, are you OK if I make it return false on mobile (not
> implemeted yet), true on JS (not needed) and forward the value on desktop
> (i.e. indicating parse error)?

Hmm, I see. What about returning `(bool, error)`? The mobile will always return `false, nil`.

input.go
```diff
@@ -234,6 +234,18 @@ func IsStandardGamepadLayoutAvailable(id GamepadID) bool {
 	return uiDriver().Input().IsStandardGamepadLayoutAvailable(id)
 }
 
+// UpdateStandardGamepadLayoutMappings updates the gamepad layout definitions using a set of definitions in SDL_GameControllerDB format.
+//
+// Returns whether the mappings were applied, and whether an error occurred parsing the mappings.
```
hajimehoshi, line 239: * `This reports whether ...` or `UpdateStandardGamepadLayoutMappings reports whether ...`
* `and this returns an error occurred in parsing the mappings if it exists`

input.go
```diff
@@ -234,6 +234,18 @@ func IsStandardGamepadLayoutAvailable(id GamepadID) bool {
...
+//
+// On platforms where gamepad mappings are not managed by ebiten, this always returns false, nil.
```
hajimehoshi, line 241: `false and nil`

input.go
```diff
@@ -234,6 +234,18 @@ func IsStandardGamepadLayoutAvailable(id GamepadID) bool {
...
+//
+// UpdateStandardGamepadLayoutMappings must be called on the main thread before ebiten.Run, and is concurrent-safe after ebiten.Run.
```
hajimehoshi, line 243: Update this comment.

input.go
```diff
@@ -234,6 +234,18 @@ func IsStandardGamepadLayoutAvailable(id GamepadID) bool {
...
+// Updated mappings take effect immediately even for already connected gamepads.
```
hajimehoshi, line 244: Insert an empty line above this.

internal/uidriver/mobile/input.go
```diff
@@ -18,6 +18,8 @@
 package mobile
 
 import (
+	"errors"
```
hajimehoshi, line 21: This is no longer needed

internal/uidriver/js/input_js.go
```diff
@@ -16,6 +16,7 @@ package js
 
 import (
 	"encoding/hex"
+	"errors"
```
hajimehoshi, line 19: This is no longer needed.

internal/uidriver/glfw/input.go
```diff
@@ -452,3 +453,24 @@ func standardButtonToGLFWButton(button driver.StandardGamepadButton) glfw.Gamepa
...
+		})
+	} else {
+		err = i.updateStandardGamepadLayoutMappings(mapping)
+	}
+	return err == nil, err
+}
+
+func (i *Input) updateStandardGamepadLayoutMappings(mapping string) error {
+	if !glfw.UpdateGamepadMappings(mapping) {
+		return errors.New("glfw: could not parse or update gamepad mappings")
```
hajimehoshi, line 473: Optional: `fmt.Errorf`

input.go
divVerent: First applied. Sorry, I do not understand your second suggestion; rephrased another way, is this clearer?

input.go
divVerent: Done.

input.go
divVerent: Yes; now it's correct: before ebiten.Run it must be the main thread; after the game loop is running, it's concurrent-safe.

input.go
divVerent: Done.

internal/uidriver/js/input_js.go
divVerent: Done. Oops.

internal/uidriver/glfw/input.go
divVerent: Done. Also added a comment explaining why it's a constant error string for now (go-gl's glfw API does not provide the error, that's why).

internal/uidriver/mobile/input.go
divVerent: Oops, fixed.

input.go
divVerent: Done now (by moving it to the ui.t thread as many other places do).

This doesn't seem to be possible BEFORE the main loop is launched, so before the main loop it remains concurrent-unsafe.

divVerent:
BTW should I leave out the "take effect immediately" comment? It is true for GLFW according to its source, but I can neither tell if it's gonna stay this way in GLFW in the future, nor if the not yet written mobile support will have this property too.

Very likely semantics possible everywhere is applying it to all gamepads if run before the main loop, and any newly connected gamepads when run inside the main loop.

divVerent:
Another thing is - will the mobile support use the same format, or would you prefer a differently formatted string when implementing this for mobile? Could make the exact format of the string explained as platform dependent instead, and just say that on desktop platforms it's gonna be a SDL_GameControllerDB string.

input.go
hajimehoshi: The latest version seems good, thanks!

input.go
hajimehoshi: Thanks! Perhaps the other functions might be in the same situation, though their comments say just 'concurrent-safe'. I'll revisit them...

hajimehoshi:
Thank you for updating!

> BTW should I leave out the "take effect immediately" comment? It is true for GLFW according to its source, but I can neither tell if it's gonna stay this way in GLFW in the future (as the documentation nowhere states this), nor if the not yet written mobile support will have this property too.
> Very likely semantics possible everywhere is applying it to all gamepads if run before the main loop, and any newly connected gamepads when run inside the main loop.

Hmm, let's keep the comment and the current behavior for usability. We have to be careful about GLFW's update. And we should emulate exactly the same behavior for mobiles.

BTW, does "take effect immediately" mean that the mapping for an already-connected gamepad is also updated?

> Another thing is - will the mobile support use the same format, or would you prefer a differently formatted string when implementing this for mobile? Could make the exact format of the string explained as platform dependent instead, and just say that on desktop platforms it's gonna be a SDL_GameControllerDB string.

I plan to use exactly same format for mobiles. Actually the database already includes Android and iOS data. I don't want to bother Ebiten users with forcing different formats for other platforms.

divVerent:
> BTW, does "take effect immediately" mean that the mapping for an already-connected gamepad is also updated?

Yes; just confirmed this. Moved my code to update gamepad handling in the jump key handling, and provided a mapping with left/right directions swapped. Was hilarious, on the first jump the directions reversed.

So GLFW does properly apply this.

Does BTW mean that when wanting to add multiple mappings, one better adds them with one call than with many call - otherwise this becomes O(n^2) as every time this function is called, all mappings are evaluated for all connected gamepads.

divVerent:
> FYI - not tested yet; for now just sent for early review. Going to test early next week.

Change now tested properly and working. Ready for full review / merge now.

divVerent:
js build failure fixed, thanks for the automated workflows!

divVerent:
Sorry, there was another js failure. Fixed too now.

hajimehoshi:
Thanks!

> Does BTW mean that when wanting to add multiple mappings, one better adds them with one call than with many call - otherwise this becomes O(n^2) as every time this function is called, all mappings are evaluated for all connected gamepads.

Hm, I don't think this function is so often called. Then I think we don't have to care about the computational cost.

divVerent:
Yeah - just wondering if I should explicitly state in the comment that
multiple mappings separated by newlines are permitted.

hajimehoshi:
> Yeah - just wondering if I should explicitly state in the comment that
> multiple mappings separated by newlines are permitted.

We should permit this explicitly. Please fix the comment. Thanks!

input.go
```diff
@@ -234,6 +234,26 @@ func IsStandardGamepadLayoutAvailable(id GamepadID) bool {
 	return uiDriver().Input().IsStandardGamepadLayoutAvailable(id)
 }
 
+// UpdateStandardGamepadLayoutMappings updates the gamepad layout definitions using a set of definitions in SDL_GameControllerDB format.
+//
+// Multiple input definitions can be provided separated by newlines.
+// In particular, it is valid to pass an entire gamecontrollerdb.txt file.
+//
+// Note though that ebiten already includes its own copy of this file,
```
hajimehoshi, line 242: s/ebiten/Ebiten/

input.go
```diff
@@ -234,6 +234,26 @@ func IsStandardGamepadLayoutAvailable(id GamepadID) bool {
...
+// so this call should only be necessary to add mappings for hardware not supported yet.
+// In this case, please also send a pull request to https://github.com/gabomdq/SDL_GameControllerDB to make this mapping available to everyone else.
+//
+// UpdateStandardGamepadLayoutMappings reports whether the mappings were applied, and returns an error in case any occurred while parsing the mappings.
+//
+// On platforms where gamepad mappings are not managed by ebiten, this always returns false and nil.
```
hajimehoshi, line 248: ditto

divVerent:
I expanded upon the comments to describe this, and also now explain that it is valid but not necessary to pass an entire gamecontrollerdb.txt file, that this feature ideally should be exposed to the user in some form, and that people who have hardware needing this facility should send a pull request to have it included in the official database.

input.go
divVerent: Done.

input.go
divVerent: Done.

hajimehoshi (approved):
LGTM, thanks!

PR #1792, opened 2021-08-31
Minor style fixes

tslocum:


hajimehoshi (commented):
Could you separate this change into two PRs: fixing misspellings, and just style changes?

tslocum:
Sure, I've opened a separate PR.

internal/atlas/image.go
```diff
@@ -574,9 +574,9 @@ func (img *Image) Pixels(x, y, width, height int) ([]byte, error) {
 
 	bs := make([]byte, 4*width*height)
 	idx := 0
-	for j := y; j < y+height; j++ {
-		for i := x; i < x+width; i++ {
-			r, g, b, a, err := img.at(i, j)
+	for k := y; k < y+height; k++ {
+		for j := x; j < x+width; j++ {
+			r, g, b, a, err := i.at(j, k)
```
hajimehoshi, line 579: I think (i, j) for coordinates is more conventional.

hajimehoshi (commented):
I'd like to know the policy of these style changes. Could you explain why you thought these are needed? Thanks!

tslocum:
These changes are recommended by `go vet`.  This is the command I ran:

```
go vet -composites=false ./...
```

I focused only on inconsistent receiver names and unnecessary `else` blocks (because the prior condition returns anyway).

hajimehoshi:
Hmm, `go vet -composites=false ./...` doesn't say anything on my environment (macOS). What kind of messages did you see in what environment?

tslocum:
Thanks, I missed that I actually executed `golang.org/x/lint/golint` (which is now deprecated, so its output might not be entirely helpful/correct).  Leaving this open just in case these changes still seem useful.

hajimehoshi:
Thank you for confirming. IMO

* Using early returns are fine. The more early returns, the merrier.
* Renaming the variables could be controversial. I'd want to discuss more if you want to merge it.

tslocum:
That makes sense.  I am not particularly opinionated, so I have removed the rename changes.

hajimehoshi (commented):
LGTM, thanks!

hajimehoshi (approved):

PR #1797, opened 2021-09-04
Fix device scaling on Linux/X11 outside Cinnamon and GNOME

### LINKED ISSUE #1774: Fullscreen broken on X11 with HiDPI
Repro:

- Set display DPI to something other than 96:
  `echo Xft.dpi: 144 | xrdb -merge`
- Run `examples/windowsize`
- Press `F`

Observe that a large part of the screen is cut off (see screenshots).

Note how `Xft.dpi` does not influence the Device Scale Factor, but it does change the initial window size and causes this weird fullscreen scale up.

Reproduced on:
- X11/i3
- Wayland/Sway/Xwayland

Does NOT reproduce on:
- Cinnamon

Most likely because ebiten has its own support for querying device scale that is specific to Cinnamon and GNOME, while not supporting anything else; at the same time, glfw has its different support and these two disagree.

Maybe we should just always use glfw's content scale?

Note though:

```
// deviceScaleFactor() is a scale by desktop environment (e.g., Cinnamon), while GetContentScale() is X's scale.
```

I do not quite understand this - when do these two ever differ, except when there simply is no desktop environment managing this at all? In fact, do we ever handle the case of these two differing right?

divVerent:
Note: this changes a lot how device scaling works on Unix systems - we now use the deviceScale to map from window system pixels to device independent pixels (which always is what GLFW's content scale factor does), and an additional screenScale to map from physical resolutions to window system resolutions (may only be required on X11).

Works around https://github.com/glfw/glfw/issues/1961, fixes #1774.


divVerent:
NOTE: change is WIP at the moment and untested on "real Windows" (only tested through WINE) and OS X. The change to uidriver/input in particular may be breaking on OS X, although it is probably correct. In addition, OS X fromGLFWMonitorPixels may need changes, as I simply do not know how GLFW behaves on OS X.

internal/devicescale/impl_desktop.go
```diff
@@ -12,8 +12,13 @@
...
+//go:build !android && !ios && !js
+// +build !android,!ios,!js
+
 package devicescale
 
+import ()
+
 func impl(x, y int) float64 {
-	return 1
+	return 1.0
```
hajimehoshi, line 23: 1 is fine

hajimehoshi (commented):
Thank you!

Hmm, does the implementation consider the scale factor of Windows?

image


image


And, would it be possible to split the change into two: Reimplement devicescale package and reimplement uidriver/glfw package? The latter will depend on the former, but IIUC the former is independent.

divVerent:
As for splitting - this will be tricky, but I can try next week.

The main difficulty is that I changed what devicescale means on X11 - currently it returns the video mode scale on X11 (the thing I call screenScale in my patch), while after my changes it'll always return the content scale as GLFW calls it.

So most likely this will look as follows:

1. devicescale package will ALSO get a way to return the screenscale
2. existing code for X11 platforms moved to screenscale, while devicescale will be driven by GLFW, and uidriver for X11 platforms adjusted accordingly in same commit 
3. screenscale implementation replaced by my new XRandR based one
4. other fixes for likely pre-existing issues

However, commit 2 would still change uidriver and devicescale at once - but it'd be rather obviously a NOP

Sounds better that way? Can get to that next week.

hajimehoshi:
Well, my question so far is why your change replaces device scales on desktop with 1, so probably just changing device scales implementation would be complex enough. I appreciate you consider splitting this PR!

divVerent:
The reason why it is done this way is so the automated tests like go vet
pass - the devicescale module is not used at all on desktop anymore.

Nevertheless I can move the GLFW based code in there. I had not done that
as I cannot pass a glfw.Monitor* into devicescale or else it would depend
on GLFW on all platforms which would be bad. I could pass monitor X and Y
values, and do my own search for the monitor in the devicescale module.

That is still dirty though as GLFW does not return the actual width and
height of monitors, see my bug report with them, so I will have to do some
heuristics to find the most likely monitor given X and Y coordinates (but
of course this will not matter much as on desktop we will always pass a top
left corner).

I did not want to go through all this, as it has its own potential of bugs,
but if you insist I can do it.

divVerent:
Also, I plan to split into a logical sequence of commits, but probably not
into PRs - given how devicescale currently is sometimes one (window system
to device independent), sometimes another (physical to window system), it
may not be possible to have every intermediate state not be a regression.

I will try though.

hajimehoshi:
> The reason why it is done this way is so the automated tests like go vet
> pass - the devicescale module is not used at all on desktop anymore.
> ...
> I did not want to go through all this, as it has its own potential of bugs,
> but if you insist I can do it.

I see, so fixing uidriver/glfw and fixing devicescale might be coupled. It's ok if you cannot decouple them well and it is inevitable to send one PR.

> Also, I plan to split into a logical sequence of commits, but probably not
> into PRs - given how devicescale currently is sometimes one (window system
> to device independent), sometimes another (physical to window system), it
> may not be possible to have every intermediate state not be a regression.

Sure. Note that Ebiten adopts 'squash and rebase' way, then you have to send separate PRs for each commit if we want to leave commits as history.

I appreciate your dedicated work!

divVerent:
I've now split it up and am filing new PRs; first the "primary" parts that make most configurations sort of work, even though still wrong for windowed mode (wrong sizes), once these two PRs are in, I can send the rest of this which will split up the scaling concept into scaling from video mode to device dependent and from device dependent to device independent (as it should be so games can e.g. ensure integer pixel ratios for improved window mode quality).

PR #1801, opened 2021-09-07
Fix a wrong conversion function usage for converting mouse positions.

divVerent:
The function used was one for video mode to window system conversion -
however the intent was converting window system to device independent pixel,
as GLFW returns window system coordinates for mouse pointers.

This is part of fixing #1774.


divVerent:
I am taking this change back - it's necessary but can't be done independently from my other change to be submitted once my other PR is in.

divVerent:
The change is confirmed OK together with my other changes though.

It's gonna be part of this: https://github.com/divVerent/ebiten/commit/1449b68d0beb6adc4c2cb5da0973a134df9ca36d

PR #1811, opened 2021-09-14
Split the concept of device scale and screen scale

divVerent:
Fully fixes wrong dpi handling on X11 (#1774).

We now have two scale factors:

- Device scale: what GLFW calls content scale; maps window system device dependent pixels (as used for window sizes, mouse positions, frame buffers etc.) to device independent pixels (96dpi standard size).
- Screen scale: maps GLFW screen resolutions to window system device dependent pixels. Only actually is different from one on X11.

This also allows for unifying the OS dependent GLFW pixel scale conversion functions into OS independent ones in a single place.

TEST FOCUSES (will do my own round of it as well using a "weird" setup where both scales and their product aren't one):
- OS X - I am unsure about the new implementation there given the old one was very different and had a screen scale of 1. I strongly suspect it is the old one that was wrong, but if not, backing out 1a048cc will fix it.
- General - in examples/windowsize, verify that "Screen size in fullscreen" is the correct size in device _independent_ pixels (i.e. changes when changing display dpi, and at 96dpi it should be equal to the actual framebuffer resolution).
- General - in examples/windowsize, verify the initial window size is 480x360 window system pixels in device _independent_ pixels (i.e. scaled by display dpi).
- General - examples/fullscreen should always have the rotating thing in the center
- General - examples/paint should have cursor position match up with what it draws

### DIFF
--- a/internal/devicescale/devicescale.go
+++ b/internal/devicescale/devicescale.go
@@ -28,9 +28,8 @@ var (
 	screenScaleCache = map[pos]float64{}
 )
 
-// GetAt returns the device scale at (x, y).
+// GetAt returns the device scale at (x, y), i.e. the number of device-dependent pixels per device-independent pixel.
 // x and y are in device-dependent pixels and must be the top-left coordinate of a monitor, or 0,0 to request a "global scale".
-// The device scale maps device dependent pixels to device independent pixels.
 func GetAt(x, y int) float64 {
 	m.Lock()
 	defer m.Unlock()
@@ -48,9 +47,8 @@ func GetAt(x, y int) float64 {
 	return s
 }
 
-// ScreenScaleAt returns the screen scale at (x, y).
+// ScreenScaleAt returns the screen scale at (x, y), i.e. the number of video mode pixels per device-dependent pixel.
 // x and y are in device-dependent pixels and must be the top-left coordinate of a monitor, or 0,0 to request a "global scale".
-// The screen scale maps physical screen pixels to device dependent pixels.
 func ScreenScaleAt(x, y int) float64 {
 	m.Lock()
 	defer m.Unlock()
--- a/internal/devicescale/devicescale.go
+++ b/internal/devicescale/devicescale.go
@@ -23,9 +23,9 @@ type pos struct {
 }
 
 var (
-	m                sync.Mutex
-	cache            = map[pos]float64{}
-	screenScaleCache = map[pos]float64{}
+	m                   sync.Mutex
+	cache               = map[pos]float64{}
+	videoModeScaleCache = map[pos]float64{}
 )
 
 // GetAt returns the device scale at (x, y), i.e. the number of device-dependent pixels per device-independent pixel.
@@ -47,15 +47,15 @@ func GetAt(x, y int) float64 {
 	return s
 }
 
-// ScreenScaleAt returns the screen scale at (x, y), i.e. the number of video mode pixels per device-dependent pixel.
+// VideoModeScaleAt returns the video mode scale scale at (x, y), i.e. the number of video mode pixels per device-dependent pixel.
 // x and y are in device-dependent pixels and must be the top-left coordinate of a monitor, or 0,0 to request a "global scale".
-func ScreenScaleAt(x, y int) float64 {
+func VideoModeScaleAt(x, y int) float64 {
 	m.Lock()
 	defer m.Unlock()
-	if s, ok := screenScaleCache[pos{x, y}]; ok {
+	if s, ok := videoModeScaleCache[pos{x, y}]; ok {
 		return s
 	}
-	s := screenScaleImpl(x, y)
-	screenScaleCache[pos{x, y}] = s
+	s := videoModeScaleImpl(x, y)
+	videoModeScaleCache[pos{x, y}] = s
 	return s
 }
--- a/internal/devicescale/devicescale.go
+++ b/internal/devicescale/devicescale.go
@@ -23,8 +23,9 @@ type pos struct {
 }
 
 var (
-	m     sync.Mutex
-	cache = map[pos]float64{}
+	m                sync.Mutex
+	cache            = map[pos]float64{}
+	screenScaleCache = map[pos]float64{}
 )
 
 // GetAt returns the device scale at (x, y).
@@ -44,5 +45,19 @@ func GetAt(x, y int) float64 {
 	// The only known case is when the application works on macOS, with OpenGL, with a wider screen mode,
 	// and in the fullscreen mode (#1573).
 
+	return s
+}
+
+// ScreenScaleAt returns the screen scale at (x, y).
+// x and y are in device-dependent pixels and must be the top-left coordinate of a monitor, or 0,0 to request a "global scale".
+// The screen scale maps physical screen pixels to device dependent pixels.
+func ScreenScaleAt(x, y int) float64 {
+	m.Lock()
+	defer m.Unlock()
+	if s, ok := screenScaleCache[pos{x, y}]; ok {
+		return s
+	}
+	s := screenScaleImpl(x, y)
+	screenScaleCache[pos{x, y}] = s
 	return s
 }
--- a/internal/devicescale/impl_x.go
+++ b/internal/devicescale/impl_x.go
@@ -27,7 +27,7 @@ import (
 
 func screenScaleImpl(x, y int) float64 {
 	// TODO: if https://github.com/glfw/glfw/issues/1961 gets fixed, this function may need revising.
-	// In case GLFW decides to switch to returning logical pixels, we can just return 1.0.
+	// In case GLFW decides to switch to returning logical pixels, we can just return 1.
 
 	// Note: GLFW currently returns physical pixel sizes,
 	// but we need to predict the window system-side size of the fullscreen window
@@ -42,19 +42,19 @@ func screenScaleImpl(x, y int) float64 {
 		// No X11 connection?
 		// Assume we're on pure Wayland then.
 		// GLFW/Wayland shouldn't be having this issue.
-		return 1.0
+		return 1
 	}
 	if err = randr.Init(xconn); err != nil {
 		// No RANDR extension?
 		// No problem.
-		return 1.0
+		return 1
 	}
 	root := xproto.Setup(xconn).DefaultScreen(xconn).Root
 	res, err := randr.GetScreenResourcesCurrent(xconn, root).Reply()
 	if err != nil {
 		// Likely means RANDR is not working.
 		// No problem.
-		return 1.0
+		return 1
 	}
 	for _, crtc := range res.Crtcs[:res.NumCrtcs] {
 		info, err := randr.GetCrtcInfo(xconn, crtc, res.ConfigTimestamp).Reply()
@@ -78,5 +78,5 @@ func screenScaleImpl(x, y int) float64 {
 		}
 	}
 	// Monitor not known to XRandR. Weird.
-	return 1.0
+	return 1
 }
--- a/internal/devicescale/impl_x.go
+++ b/internal/devicescale/impl_x.go
@@ -25,7 +25,7 @@ import (
 	"github.com/jezek/xgb/xproto"
 )
 
-func screenScaleImpl(x, y int) float64 {
+func videoModeScaleImpl(x, y int) float64 {
 	// TODO: if https://github.com/glfw/glfw/issues/1961 gets fixed, this function may need revising.
 	// In case GLFW decides to switch to returning logical pixels, we can just return 1.
 
--- a/internal/devicescale/impl_x.go
+++ b/internal/devicescale/impl_x.go
@@ -25,7 +25,7 @@ import (
 	"github.com/jezek/xgb/xproto"
 )
 
-func impl(x, y int) float64 {
+func screenScaleImpl(x, y int) float64 {
 	// TODO: if https://github.com/glfw/glfw/issues/1961 gets fixed, this function may need revising.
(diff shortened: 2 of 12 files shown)


divVerent:
To try out on Linux with bare window manager:

```
xrandr --output DP-1 --scale 0.8x0.8
echo 'Xft.dpi: 120' | xrdb -merge
```

This should yield deviceScale of 1.5 and screenScale of 1.25.

Test results:

* General - in examples/windowsize, verify that "Screen size in fullscreen" is the correct size in device independent pixels (i.e. changes when changing display dpi, and at 96dpi it should be equal to the actual framebuffer resolution).
  * PASS: Prints 2048x1152, which is exactly my 3072x1728 framebuffer (due to the scale of 0.8, see above) divided by 1.5.
* General - in examples/windowsize, verify the initial window size is 480x360 window system pixels in device independent pixels (i.e. scaled by display dpi).
  * PASS: Window size is 720x540 according to xwininfo, which is consistent with a deviceScale of 1.5.
* General - examples/fullscreen should always have the rotating thing in the center
  * PASS
* General - examples/paint should have cursor position match up with what it draws
  * PASS

As these two scales are sufficiently screwed up, all not 1, all not equal, their product or quotient not 1 either, this should suffice to show correctness on this platform; please test the same on Windows and OS X. Thanks!

divVerent:
Doing the same with WINE on this setup now is a part-failure:

- Device scale is detected as 1.0. Video mode is the full mode, not the scaled one.
- Only the last test actually passes; basically, under WINE things are broken, but consistently so at least.

I'll investigate.

divVerent:
FYI SetProcessDPIAware is indeed no longer needed - _glfwPlatformInit(void) already calls it.

divVerent:
To make HiDPI work in WINE, I had to set the dpi to 144 manually in winecfg.

Fine. I'll assume on Windows things do not have this issue.

NOW:

* General - in examples/windowsize, verify that "Screen size in fullscreen" is the correct size in device independent pixels (i.e. changes when changing display dpi, and at 96dpi it should be equal to the actual framebuffer resolution).
  * Prints 1706x960 at a device scale factor of 1.5. Still seems wrong. Is 4/9 my physical res, or 5/9 my framebuffer.
  * Device scale factor is reported as 1.5, though.
  * WAIT, I GET IT: 4/9 is the square of 2/3. Seems like the scale is applied _twice_. Will fix on Ebiten end, but please verify on real Windows.
  * However, WINE still uses the video mode where it should scale it by the RANDR scale. WINE's bug, not ours (and native Windows doesn't have this issue in the first place).
* General - in examples/windowsize, verify the initial window size is 480x360 window system pixels in device independent pixels (i.e. scaled by display dpi).
  * PASS.
* General - examples/fullscreen should always have the rotating thing in the center
  * Double fail. Maybe same issue as fullscreen res.
* General - examples/paint should have cursor position match up with what it draws
  * PASS.

divVerent:
* General - in examples/windowsize, verify that "Screen size in fullscreen" is the correct size in device independent pixels (i.e. changes when changing display dpi, and at 96dpi it should be equal to the actual framebuffer resolution).
  * Mostly PASS, except for the screen scale part that is a WINE bug.
* General - in examples/windowsize, verify the initial window size is 480x360 window system pixels in device independent pixels (i.e. scaled by display dpi).
  * PASS.
* General - examples/fullscreen should always have the rotating thing in the center
  * Mostly PASS, except for the screen scale part that is a WINE bug.
* General - examples/paint should have cursor position match up with what it draws
  * PASS.

So, WINE-on-X11 now seems to be as good as it gets - non-support for xrandr --scale is to be expected as Windows doesn't have that feature.

divVerent:
Ready for you to test on real Windows and OS X now.

internal/devicescale/devicescale.go
```diff
@@ -46,3 +47,17 @@ func GetAt(x, y int) float64 {
 
 	return s
 }
+
+// ScreenScaleAt returns the screen scale at (x, y).
+// x and y are in device-dependent pixels and must be the top-left coordinate of a monitor, or 0,0 to request a "global scale".
```
hajimehoshi, line 52: When (0, 0) is given, this function just returns the scale for the monitor at (0, 0), right? What is the "global scale"?

(There is the same issue at `GetAt`).

Also, from this comment, I was not sure what's the difference between `GetAt` and `ScreenScaleAt`.

internal/devicescale/impl_x.go
```diff
@@ -35,27 +35,26 @@ func impl(x, y int) float64 {
...
 		// Assume we're on pure Wayland then.
 		// GLFW/Wayland shouldn't be having this issue.
-		return float64(sx)
+		return 1.0
 	}
 	if err = randr.Init(xconn); err != nil {
 		// No RANDR extension?
 		// No problem.
-		return float64(sx)
+		return 1.0
```
hajimehoshi, line 50: ditto

internal/devicescale/impl_x.go
```diff
@@ -35,27 +35,26 @@ func impl(x, y int) float64 {
...
 	}
 	root := xproto.Setup(xconn).DefaultScreen(xconn).Root
 	res, err := randr.GetScreenResourcesCurrent(xconn, root).Reply()
 	if err != nil {
 		// Likely means RANDR is not working.
 		// No problem.
-		return float64(sx)
+		return 1.0
```
hajimehoshi, line 57: ditto

internal/devicescale/impl_x.go
```diff
@@ -75,9 +74,9 @@ func impl(x, y int) float64 {
 			// Return one scale, even though there may be separate X and Y scales.
 			// Return the _larger_ scale, as this would yield a letterboxed display on mismatch, rather than a cut-off one.
 			scale := math.Max(float64(physWidth)/float64(xWidth), float64(physHeight)/float64(xHeight))
-			return float64(sx) * scale
+			return scale
 		}
 	}
 	// Monitor not known to XRandR. Weird.
-	return float64(sx)
+	return 1.0
```
hajimehoshi, line 81: ditto

hajimehoshi:
Question: Does `ebiten.DeviceScaleFactor` return `(device scale)` or `(device scale) * (screen scale)`? I think the former in your PR, but is this correct?

divVerent:
> Question: Does `ebiten.DeviceScaleFactor` return `(device scale)` or `(device scale) * (screen scale)`? I think the former in your PR, but is this correct?

Yes; screen scale is purely internal, and actually just an Ebiten-internal workaround for APIs returning "video mode" instead of "screen size".

internal/devicescale/devicescale.go
divVerent: This is hard to phrase, so let me explain:

- (0, 0) is specified in practice ONLY for APIs that have no per-screen scale support (e.g. js).
- Passing (0, 0) to APIs that do have per-screen scale support return the scale of the screen at (0, 0), if any exists; if none exists, they return the scale of the primary screen.

If you can phrase this better, please suggest something.

As for GetAt vs ScreenScaleAt, I'll explain this further.

divVerent:
BTW this is why I initially wanted screen scale stuff to live only in uidriver; it's actually just a workaround for GLFW and isn't even called on non-GLFW platforms. But I had to put it in devicescale to make a first change to upgrade it to be independent from desktop environment.

So maybe after this change series, I could actually move ScreenScaleAt back into uidriver in a otherwise-NOP change?

internal/devicescale/devicescale.go
hajimehoshi: Yeah probably this API is not good and we should define a 'dont-care' value rather than (0, 0). It's ok to leave this as it is.

internal/devicescale/devicescale.go
```diff
@@ -46,3 +46,16 @@ func GetAt(x, y int) float64 {
 
 	return s
 }
+
+// ScreenScaleAt returns the screen scale at (x, y), i.e. the number of video mode pixels per device-dependent pixel.
```
hajimehoshi, line 50: As this is used only for video mode, what about renaming this at `VideoModeScaleAt`? The term 'screen' is not clear here.

hajimehoshi (commented):
Almost good!

hajimehoshi:
> Yes; screen scale is purely internal, and actually just an Ebiten-internal workaround for APIs returning "video mode" instead of "screen size".

OK, thanks. `screen size` is just for adjusting GLFW's video mode values. Is this correct?

> BTW this is why I initially wanted screen scale stuff to live only in uidriver; it's actually just a workaround for GLFW and isn't even called on non-GLFW platforms. But I had to put it in devicescale to make a first change to upgrade it to be independent from desktop environment.
> So maybe after this change series, I could actually move ScreenScaleAt back into uidriver in a otherwise-NOP change?

This sounds a very good idea. As this is only for uidriver/glfw, we should move the logic to this place. Thank you for your dedicating work!

hajimehoshi:
> So, WINE-on-X11 now seems to be as good as it gets - non-support for xrandr --scale is to be expected as Windows doesn't have that feature.

Hmm, can we have workaround for wine?

hajimehoshi:
Another test case would be `examples/highdpi`. Does this work as expected? I'll test them later.

divVerent:
> > So, WINE-on-X11 now seems to be as good as it gets - non-support for xrandr --scale is to be expected as Windows doesn't have that feature.
> 
> Hmm, can we have workaround for wine?

Doubt we can; Windows doesn't have the problem, after all. Only way I could see would be using CreateProcess to spawn the "xrandr" binary, which is... AWFUL.

(Actually, the xgb way _might_ actually work too, but then this could break when run on real Windows with an X server running...)

divVerent:
> Another test case would be `examples/highdpi`. Does this work as expected? I'll test them later.

Yes, this one works fine, but I find it rather boring ;) the debug output of the other test cases already contains the one useful piece of info: the device scale.

Now, the highdpi example would be MUCH BETTER if it displayed e.g. a checkerboard or lines pattern that makes it trivial to see if HiDPI is actually working. As is, all I notice is that it looks slightly sharper than the others, but it would also if the scale factor were wrong as long as it's > 1.

internal/devicescale/devicescale.go
divVerent: Good idea, done.

hajimehoshi:
Hmm, on my macOS, `examples/highdpi` didn't work with your PR as expected.

 * The window size is doubled
 * The rendering was blurred

Could you take a look?

divVerent:
> Hmm, on my macOS, `examples/highdpi` didn't work with your PR as expected.
> 
>     * The window size is doubled
> 
>     * The rendering was blurred
> 
> 
> Could you take a look?

Sadly without a Mac there's not much I can do.

However, the deltas from main are that all the scales are 1.0 for the from/to pixel conversion functions; so can you try making all three of `fromGLFWPixel`, `toGLFWPixel` and `fromGLFWMonitorPixel` just do "return x" in your local copy? If tests then are all OK on OS X, I'll split them apart again and document the reasoning (which seems to then indicate that on OS X, GLFW uses device independent pixels for all coordinates, BUT Ebiten still needs the scaling factor internally so it can create a HiDPI surface).

hajimehoshi:
> can you try making all three of fromGLFWPixel, toGLFWPixel and fromGLFWMonitorPixel just do "return x" in your local copy?

Yes, the example worked as expected after all the function return `x`.

hajimehoshi:
> If tests then are all OK on OS X, I'll split them apart again and document the reasoning (which seems to then indicate that on OS X, GLFW uses device independent pixels for all coordinates, BUT Ebiten still needs the scaling factor internally so it can create a HiDPI surface).

This seems to match with my current understanding. MacOS's coordinate is always device-independent, while Windows and Linux are not.

divVerent:
Sounds good.

Of course, technically it'd be GLFW's job to handle that... but anyway, off
to fixing it.

divVerent:
Fix pushed.

hajimehoshi:
> Of course, technically it'd be GLFW's job to handle that... but anyway, off
> to fixing it.

I agree this should be handled in GLFW side. Now GLFW has inconsistent scalings unfortunately:

* https://github.com/glfw/glfw/issues/1800
* https://github.com/glfw/glfw/issues/1857

internal/uidriver/glfw/ui_darwin.go
```diff
@@ -81,22 +81,27 @@ import (
...
+// fromGLFWMonitorPixel must be called from the main thread.
+func (u *UserInterface) fromGLFWMonitorPixel(x float64, videoModeScale float64) float64 {
+	// videoModeScale is always 1 on OS X,
+	// however leaving the divison in place for consistency.
+	return x / videoModeScale
 }
 
+// fromGLFWPixel must be called from the main thread.
 func (u *UserInterface) fromGLFWPixel(x float64) float64 {
+	// NOTE: On OS X, GLFW exposes the device independent coordinate system.
```
hajimehoshi, line 93: s/OS X/macOS/

internal/uidriver/glfw/ui_darwin.go
```diff
@@ -81,22 +81,27 @@ import (
 	"github.com/hajimehoshi/ebiten/v2/internal/glfw"
 )
 
-func fromGLFWMonitorPixel(x float64, deviceScale float64) float64 {
-	return x
+// fromGLFWMonitorPixel must be called from the main thread.
+func (u *UserInterface) fromGLFWMonitorPixel(x float64, videoModeScale float64) float64 {
+	// videoModeScale is always 1 on OS X,
```
hajimehoshi, line 86: s/OS X/macOS/

hajimehoshi (commented):
I'm testing your PR now again...

hajimehoshi (approved):
LGTM % the spellings I pointed

PR #1812, opened 2021-09-14
Move VideoModeScaleAt into internal/uidriver/glfw.

divVerent:
It's only used by GLFW, and has been in devicescale only to ease migration
to separate video mode vs content scaling.


internal/uidriver/glfw/ui_unix.go
```diff
@@ -19,10 +19,89 @@
...
 	"github.com/hajimehoshi/ebiten/v2/internal/driver"
 	"github.com/hajimehoshi/ebiten/v2/internal/glfw"
+	"github.com/jezek/xgb"
+	"github.com/jezek/xgb/randr"
+	"github.com/jezek/xgb/xproto"
 )
 
+type videoModeScaleCacheKey struct{ X, Y int }
+
+var videoModeScaleCache map[videoModeScaleCacheKey]float64
```
hajimehoshi, line 33: Let's clear the cache at updateMonitors (a handler of Monitor event), or wrong values are cached when monitors are dynamically changed. (Actually there is a TODO for `devicescale` packages.)

hajimehoshi (commented):
Thank you!

internal/uidriver/glfw/ui_unix.go
divVerent: Doing that for now; in the future, maybe we should remove use of GetVideoMode() calls entirely and route all that through *monitor.

Leaving as is for now, though, as that's a separate refactor and also a bit unclear if a good idea (as that'd likely mean to query XRandR for _every_ monitor on monitor change, while usually we will only query for the one the window is on).

divVerent: Cache invalidation added.

internal/uidriver/glfw/ui_unix.go
```diff
@@ -30,7 +30,12 @@ import (
 
 type videoModeScaleCacheKey struct{ X, Y int }
 
-var videoModeScaleCache map[videoModeScaleCacheKey]float64
+var videoModeScaleCache = map[videoModeScaleCacheKey]float64{}
+
+// updateMonitorsByOS must be called from the main thread.
+func updateMonitorsByOS() {
+	videoModeScaleCache = map[videoModeScaleCacheKey]float64{}
```
hajimehoshi, line 37: In general, just a for-loop is more efficient to clear a map:

```
for k := range videModeScaleCache {
    delete(videModeScaleCache, k)
}
```

This should be optimized by Go compiler.

hajimehoshi:
Apparently the test bots fail. Please check the failures.

divVerent:
Fixed the one real error I found - but can you please fix the build bots? I admit I even stopped looking at them normally as they like 90% of cases run into a timeout during building, and another bunch of build bots fail because I have no access to them.

hajimehoshi:
I can see the error like this:

image

But perhaps other accounts cannot see this. I'm looking for the settings to adjust the visibility...

divVerent:
Oh, when I explicitly look into them, I can see errors - but most fail with "Cancelled after 2m". The one that fails because I have no permission is Sourcehut, permanently spamming my email with failures.

divVerent:
Anyway, can SourceHut be disabled for PRs, and the timeout bumped higher from 2 minutes? That should resolve everything regarding the tests.

hajimehoshi:
> Anyway, can SourceHut be disabled for PRs, and the timeout bumped higher from 2 minutes? That should resolve everything regarding the tests.

One of the 'test' bots fails, all the test bots are suspended. Then this is not 'time out'.

SourceHut builds should not happen on PRs, but did you see some emails? That's odd...

internal/uidriver/glfw/ui_unix.go
```diff
@@ -19,10 +19,96 @@
...
+	"github.com/jezek/xgb/randr"
+	"github.com/jezek/xgb/xproto"
 )
 
+type videoModeScaleCacheKey struct{ X, Y int }
+
+var videoModeScaleCache = map[videoModeScaleCacheKey]float64{}
+
+// updateMonitorsByOS must be called from the main thread.
+func updateMonitorsByOS() {
```
hajimehoshi, line 36: `clearVideoModeScaleCache` sounds better. This is not updating.

hajimehoshi (approved):
LGTM % my comment

hajimehoshi:
I'll go to bed soon and merge this tomorrow. Thank you for your dedicating work!

internal/uidriver/glfw/ui_unix.go
divVerent: Done; had considered that name too but went for the generic one because the calling code doesn't know about the cache (which also only exists on one OS), but yeah, naming by what it does is of course fine too.

divVerent:
Getting these emails on every commit: https://github.com/divVerent/ebiten/actions/runs/1235017943

hajimehoshi:
> Getting these emails on every commit: https://github.com/divVerent/ebiten/actions/runs/1235017943

Acknowledged. I'll forbid this action on a fork.

hajimehoshi:
https://github.com/hajimehoshi/ebiten/commit/f665abc8d41a0196b093f04df6baa141c85bfd1c should fix this issue.

PR #1840, opened 2021-10-04
support for gopherjs

visualfc:
update for support gopherjs ( https://github.com/gopherjs/gopehrjs )

### DIFF
--- a/internal/graphicsdriver/opengl/context_js.go
+++ b/internal/graphicsdriver/opengl/context_js.go
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build js
+// +build js
+
 package opengl
 
 import (
--- a/internal/graphicsdriver/opengl/context_desktop.go
+++ b/internal/graphicsdriver/opengl/context_desktop.go
@@ -12,8 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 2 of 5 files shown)


internal/graphicsdriver/opengl/context_js.go
```diff
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build js
+// +build js
```
hajimehoshi, line 16: I don't think we should care this. Wouldn't it be possible for GopherJS to recognize `*_js.go` files?

hajimehoshi (commented):
Hmm, I don't want to have special hacks for GopherJS. Could GopherJS have a compatibility with Wasm APIs (`syscall/js`) and so on?

visualfc:
> Hmm, I don't want to have special hacks for GopherJS. Could GopherJS have a compatibility with Wasm APIs (`syscall/js`) and so on?

gopherjs unsupport unsafe  sliceToByteSlice convert

hajimehoshi:
> gopherjs unsupport unsafe sliceToByteSlice convert

I don't think this is what Ebiten should fix. If it'd be possible to fix GopherJS, it'd be the best.

visualfc:
> > gopherjs unsupport unsafe sliceToByteSlice convert
> 
> I don't think this is what Ebiten should fix. If it'd be possible to fix GopherJS, it'd be the best.

yes, if fixed for gopherjs is best.

Issue #1848, opened 2021-10-12
internal/graphicsdriver/opengl,metal: Avoid uploading uniform variables when possible


hajimehoshi:
/CC @Zyko0

Zyko0:
I don't think this impacts performance on recent hardware : https://community.khronos.org/t/performance-hit-gluniform/61202/10

I couldn't make a difference happen by testing with the example in #1862 with 4 shaders, and a bunch of uniforms each (win10, rtx 2080):

3000~ fps
```
go run ./examples/issue_1848/main.go
```

3000~ fps
```
git checkout a3eb8933e5 ./internal/graphicsdriver/opengl/program.go
go run ./examples/issue_1848/main.go
```

Zyko0:
As a conclusion, caching by program seems correct on opengl (since there's already some sort of caching on Ebiten's internal).
We did prove that there's no relevant performance impact on different specs and OSes, so there's nothing immediately beneficial for implementing this.

Maybe the `performance` tag can go, for another one like `fix` or something

hajimehoshi:
> edit: also I don't know if opengl is a tag or if there's metal to remove from the title maybe

There is `ebitengl` tag to force to use GL.

hajimehoshi:
Let me deprioritize this task, but keep this issue open. When we find new things, let's revisit this.

Zyko0:
> There is ebitengl tag to force to use GL.

I meant a github tag* on this issue sorry

PR #1862, opened 2021-10-31
PoC: uniforms cached by programs

Zyko0:
Issue reference : https://github.com/hajimehoshi/ebiten/issues/1848

This is a dirty proof of concept, to check if uniforms have to be re-uploaded on any shader switch, or if they are kept within each program (which I think).
I added an `example/issue_1848` with two shaders having different uniforms to make sure it works on a machine.

Not sure if it proves anything yet, because I'm not fully aware of ebiten's backend implementation, also it might introduces bugs as is for the aforementioned reason 😅

### DIFF
--- a/internal/graphicsdriver/opengl/context_js.go
+++ b/internal/graphicsdriver/opengl/context_js.go
@@ -69,10 +69,6 @@ func (p program) equal(rhs program) bool {
 	return p.value.Equal(rhs.value) && p.id == rhs.id
 }
 
-func (p program) getID() int {
-	return int(p.id)
-}
-
 var InvalidTexture = textureNative(js.Null())
 
 var invalidUniform = uniformLocation(js.Null())
--- a/internal/graphicsdriver/opengl/context_js.go
+++ b/internal/graphicsdriver/opengl/context_js.go
@@ -69,6 +69,10 @@ func (p program) equal(rhs program) bool {
 	return p.value.Equal(rhs.value) && p.id == rhs.id
 }
 
+func (p program) getID() int {
+	return p.id
+}
+
 var InvalidTexture = textureNative(js.Null())
 
 var invalidUniform = uniformLocation(js.Null())
--- a/internal/graphicsdriver/opengl/context_js.go
+++ b/internal/graphicsdriver/opengl/context_js.go
@@ -70,7 +70,7 @@ func (p program) equal(rhs program) bool {
 }
 
 func (p program) getID() int {
-	return p.id
+	return int(p.id)
 }
 
 var InvalidTexture = textureNative(js.Null())
--- a/audio/internal/convert/float32.go
+++ b/audio/internal/convert/float32.go
@@ -27,10 +27,11 @@ func NewReaderFromFloat32Reader(r Float32Reader) io.Reader {
 }
 
 type f32Reader struct {
-	r    Float32Reader
-	eof  bool
-	buf  *byte
-	fbuf []float32
+	r         Float32Reader
+	eof       bool
+	hasRemain bool
+	remain    byte
+	fbuf      []float32
 }
 
 func max(a, b int) int {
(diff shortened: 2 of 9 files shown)


hajimehoshi:
Could you take a look at the build error? Thanks,

```
# github.com/hajimehoshi/ebiten/v2/internal/graphicsdriver/opengl
internal/graphicsdriver/opengl/program.go:133:15: invalid map key type program
internal/graphicsdriver/opengl/program.go:156:20: invalid map key type program
```

hajimehoshi:
Another failing:

```
# github.com/hajimehoshi/ebiten/v2/internal/graphicsdriver/opengl
internal/graphicsdriver/opengl/context_js.go:73:10: cannot use p.id (type programID) as type int in return argument
```

internal/graphicsdriver/opengl/context_js.go
```diff
@@ -69,6 +69,10 @@ func (p program) equal(rhs program) bool {
 	return p.value.Equal(rhs.value) && p.id == rhs.id
 }
 
+func (p program) getID() int {
```
hajimehoshi, line 72: We can use existing `getProgramID`

hajimehoshi (commented):
I'll take a look at the example later, thanks!

internal/graphicsdriver/opengl/context_js.go
Zyko0: Do you mean:
```go
func (p program) getID() int {
	return int(getProgramID(p))
}
```

Because depending on arch, `program` still has to implement the same method and this only exists in context_js
Or do you mean providing `type programID int` + `getProgramID` across all context_os.go files ?

Zyko0: Oh nvm, it does exist everywhere, I'll change

hajimehoshi:
@Zyko0 So I now checked out your PR (thanks!). How can I test this? What value (FPS) was changed in your environment?

Zyko0:
So, as stated on discord (should have shared here), I added the FPS in the example `issue_1848` (not pushed) and got 3600+ with my change while 1500 with yours.
The thing is, now I think about it, I was also debugging calls to `uniformFloats` to see if uploads were necessary, so the slow down might have been induced by all these debugs.

Anyways, the first point was to detect at which occurence uniform uploads are made using ebiten.
Since Ebiten has a default shader that is used every frame (I think ?), the previous `lastUniforms` value was kind of irrelevant if the end user wanted to preserve the same uniform values, since Ebiten would re-upload them every frame.

Now that they are recorded "by program" instead of the whole opengl state, we can see by debugging calls to `uniformFloats` that there are way less calls to uniform uploads, and that examples still work fine by switching program (shader from example).

Do you want me to add the additonal stuff I'm talking about in example+`uniformFloats` function ?

hajimehoshi:
As we discussed at Discord, we confirmed that the current example didn't show FPS improvements before and after the fix. I'm waiting for the example updated, or yet another benchmarks.

Even if there is no difference, I think I'm fine to cache uniform variables to reduce uploading uniform variables, as long as this doesn't degrade the performance. The less uploading, the happier we'll be in general. However, we need great care not to break compatibility and portability.

Thanks!

Zyko0:
Okay I think I tried my hardest to make it show a relevant difference, but I couldn't !
I guess my intuition has been really wrong or I don't know, maybe my gpu (rtx 2080 super) helps ?
Always thought uniform uploads was one of the things that cost a lot on opengl, maybe it's platform dependent, maybe it's nothing in the end haha

Anyway, I'd love if you could confirm this with another spec (maybe a lower one too ?)

Thanks for guiding me into asking myself the right questions btw

hajimehoshi:
> Always thought uniform uploads was one of the things that cost a lot on opengl (and ebiten does reupload at every frame), maybe it's platform dependent, maybe it's nothing in the end haha

Yeah I guess modern GPUs already do such optimization. I think it is fine to do efforts to reduce uploading, but my concern is portability and compatibility.

> Anyway, I'd love if you could confirm this with another spec (maybe a lower one too ?)

Is MacBook Pro 2020 fine?

Zyko0:
> Yeah I guess modern GPUs already do such optimization. I think it is fine to do efforts to reduce uploading, but my concern is portability and compatibility.

Agree !

> Is MacBook Pro 2020 fine?

It uses Metal right ? If so, it has no chance to be impacted since this only changes opengl unfortunately

But it's fine I think, I tried on my Linux laptop and there was no difference as well haha

hajimehoshi:
> It uses Metal right ? If so, it has no chance to be impacted since this only changes opengl unfortunately

I can test this with `ebitengl` build tag.

hajimehoshi:
> But it's fine I think, I tried on my Linux laptop and there was no difference as well haha

Hm ok. Probably the drivers already take care...

hajimehoshi:
On MacBook Pro 2020, the FPS was about 260 with and without the patch. (tested with `ebitengl`).

Zyko0:
Thanks ! Now we know 😄

hajimehoshi:
So, can we close this PR, or do you want to do experiments more?

Zyko0:
No it's fine, we can close !
Anyways I don't feel like uniforms abuse is very common to Ebiten users

PR #1866, opened 2021-11-05
Correction in go doc example

rikkuness:
Correction in go doc example to ensure example satisfies the `Game` interface.

Current example results in:
```
cannot use game (type *Game) as type ebiten.Game in argument to ebiten.RunGame:
	*Game does not implement ebiten.Game (wrong type for Update method)
		have Update() error
		want Update(*ebiten.Image) error (exit status 2)
```

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -21,7 +21,7 @@
 //
 //     // Update proceeds the game state.
 //     // Update is called every tick (1/60 [s] by default).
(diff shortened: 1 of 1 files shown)


rikkuness:
My bad, just spotted I was looking at the `/v2` doc and using a different version in the code.

PR #1867, opened 2021-11-08
Use const to represent bytePerSample

wasedaigo:
I did not understand what "4" meant in the example, so I decided to improve the readability of it.


hajimehoshi (approved):
LGTM, thanks!

PR #1877, opened 2021-11-15
Add getters for GeoM fields.

sailormoon:
`Element` exists but requires knowing which indices to look-up and can
fatal in the case of dev error. Adding getters let's the dev directly
access the field they care about and makes the callsite more readable.

e.g. `geoM.Tx()` over `geoM.Element(0, 2)`.

### DIFF
--- a/geom.go
+++ b/geom.go
@@ -80,6 +80,36 @@ func (g *GeoM) Element(i, j int) float64 {
 	}
 }
 
+// A returns the A value of the matrix.
+func (g *GeoM) A() float64 {
+	return g.a_1 + 1
+}
+
+// B returns the B value of the matrix.
+func (g *GeoM) B() float64 {
+	return g.b
+}
+
+// C returns the C value of the matrix.
+func (g *GeoM) C() float64 {
+	return g.c
+}
+
+// D returns the D value of the matrix.
+func (g *GeoM) D() float64 {
+	return g.d_1 + 1
+}
+
+// Tx returns the Tx value of the matrix.
+func (g *GeoM) Tx() float64 {
+	return g.tx
+}
+
+// Ty returns the Ty value of the matrix.
+func (g *GeoM) Ty() float64 {
+	return g.ty
+}
+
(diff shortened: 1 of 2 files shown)


hajimehoshi:
> Element exists but requires knowing which indices to look-up and can
> fatal in the case of dev error

Hmm, would `String` work for your debugging purpose?

sailormoon:
I wanted to use it in Draw to sort all my images by their GeoM.Ty before finally drawing them so they would draw in the right order :(

hajimehoshi:
> I wanted to use it in Draw to sort all my images by their GeoM.Ty before finally drawing them so they would draw in the right order :(

For this purpose, I'd record Y values for each object. Depending on a GeoM value is a little fragile (what if GeoM includes scaling or rotating?)

I understand the getters are useful, but my concern is that GeoM interface might be chaotic. I'm afraid I might be too nervous but let me think more... I'll revisit this tomorrow.

geom.go
```diff
@@ -80,6 +80,36 @@ func (g *GeoM) Element(i, j int) float64 {
 	}
 }
 
+// A returns the A value of the matrix.
```
hajimehoshi, line 83: `// A returns the element at the first row and the first column of the matrix` and so on.

sailormoon:
> I understand the getters are useful, but my concern is that GeoM interface might be chaotic. I'm afraid I might be too nervous but let me think more... I'll revisit this tomorrow.

I'm OK to decline this PR if you want, let me know. I don't have strong feelings and I'm OK storing the Y values separately as well or creating helper functions in my own project to access the right indices.

hajimehoshi:
> I'm OK to decline this PR if you want, let me know. I don't have strong feelings and I'm OK storing the Y values separately as well or creating helper functions in my own project to access the right indices.

Thanks. Until we find these are really needed, let me close this. My policy might be too conservative but it is pretty hard to remove APIs compared to add APIs in general.

I appreciate your trial to contribute Ebiten!

PR #1890, opened 2021-12-04
DrawRectShader add Filter

sunqirui1987:


hajimehoshi:
Filter doesn't work with custom shaders so far.

PR #1905, opened 2021-12-20
Add DrawLogo and splash screen example

tslocum:
DrawLogo draws the Ebiten logo on the image.

### DIFF
--- a/ebitenutil/logo.go
+++ b/ebitenutil/logo.go
@@ -1,0 +1,47 @@
+package ebitenutil
+
+import (
+	"image"
+	"image/color"
+	"math"
+
+	"github.com/hajimehoshi/ebiten/v2"
+)
+
+// DrawLogo draws the Ebiten logo on the image.
+func DrawLogo(dst *ebiten.Image) {
+	var (
+		bounds     = dst.Bounds()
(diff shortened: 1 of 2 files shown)


hajimehoshi:
* Does this use an (official) logo image? https://ebiten.org/images/splash_black.png
* I don't want to add a utility function to draw this as the image is pretty big.

tslocum:
Thanks for your quick feedback.    It draws the logo using `Fill` and `DebugPrint` only (so no assets are needed) and it scales according to the window size.

I would be glad to provide this in a third party library instead if it seems outside of Ebiten's scope.  Let me know what you think.  I believe a ready-made splash screen (this PR or even some other implementation) would help boost Ebiten's visibility in the games using it.

hajimehoshi:
> I would be glad to provide this in a third party library instead if it seems outside of Ebiten's scope. Let me know what you think. I believe a ready-made splash screen (this PR or even some other implementation) would help boost Ebiten's visibility in the games using it.

Thank you for suggestion. I understand a function rendering a splash logo would be useful. We need more discussion about the API and the logo (e.g., Should we target 4K resolutions? Is it enough to render it on an entire screen? What about the license?)

So can we have a new issue for this and more discussion before discussion on your PR?

PR #1924, opened 2021-12-30
closes #1923.  fix broken isometric example

### LINKED ISSUE #1923: $ go run examples/isometric/main.go # command-line-arguments examples/isometric/main.go:31:12: undefined: NewGame
`examples/isometric` does not run in it's current state.

```
$ go run examples/isometric/main.go
# command-line-arguments
examples/isometric/main.go:31:12: undefined: NewGame
```

jacobsalmela:
the Spritesheet_png was not included with go generate.  I also moved the
non-main files into their own folder to better match other examples in the repo.

Signed-off-by: Jacob Salmela <me@jacobsalmela.com>


hajimehoshi:
> the Spritesheet_png was not included with go generate.

Good point so we should fix only examples/resources/generate.go We don't have to touch other files.

jacobsalmela:
Wilco.  I'll adjust that.

examples/resources/generate.go
```diff
@@ -24,6 +24,7 @@
 //go:generate file2byteslice -package=images -input=./images/smoke.png -output=./images/smoke.go -var=Smoke_png
 //go:generate file2byteslice -package=images -input=./images/tile.png -output=./images/tile.go -var=Tile_png
```
hajimehoshi, line 25: Optional: you can move this so that the list is in alphabetical order

hajimehoshi (approved):
LGTM, thank you very much!

examples/resources/generate.go
jacobsalmela: It looks like there are other entries already not in alphabetical order... 🤷

hajimehoshi: Ack

PR #1927, opened 2022-01-02
ebiten: add RunOnMainThread(func())

### LINKED ISSUE #1926: proposal: provide public RunOnMainThread API
Running a function on the main thread is impossible and occupied by Ebiten internally. The reason is that some OS or graphical APIs require to be executed on the main thread.

However, it fundamentally limits other developers to extend unimplemented features in Ebiten, such as registering a hotkey.

It would be beneficial if "ebiten" package can provide a public API to execute functions on the main thread, the API signature can be:

```
package ebiten

func RunOnMainThread(f func() error) error
```

For a specific example, this would allow package "[hotkey](https://pkg.go.dev/golang.design/x/hotkey@v0.3.0-alpha)" to collaborate with Ebiten for an unimplemented feature.

Here is a proof-of-concept change that exposes `graphicscommand.RunOnMainThread`: https://github.com/changkun/ebiten/commit/8f368a5afb8f167c1b7ad9838f78d041cf7ee346

Here is a complete example that benefits the public version of `ebiten.RunOnMainThread`: https://github.com/changkun/demos/blob/main/hotkeys/ebitenapp/main.go

Any thoughts?

changkun:
Fixes #1926

### DIFF
--- a/internal/thread/thread.go
+++ b/internal/thread/thread.go
@@ -103,3 +103,6 @@ func (t *NoopThread) Loop() {}
 func (t *NoopThread) Call(f func() error) error {
 	return f()
 }
+
+// Stop does nothing
+func (t *NoopThread) Stop() {}
--- a/internal/thread/thread.go
+++ b/internal/thread/thread.go
@@ -14,8 +14,6 @@
 
 package thread
 
-import "sync"
-
 // Thread defines threading behavior in Ebiten.
 type Thread interface {
 	Call(func())
@@ -23,15 +21,11 @@ type Thread interface {
 	Stop()
 }
 
-type funcdata struct {
-	f    func()
-	done chan struct{}
-}
-
 // OSThread represents an OS thread.
 type OSThread struct {
-	funcs chan funcdata
-	done  chan struct{}
+	funcs     chan func()
+	done      chan struct{}
+	terminate chan struct{}
 }
 
 // NewOSThread creates a new thread.
@@ -39,8 +33,9 @@ type OSThread struct {
 // It is assumed that the OS thread is fixed by runtime.LockOSThread when NewOSThread is called.
 func NewOSThread() *OSThread {
 	return &OSThread{
-		funcs: make(chan funcdata),
-		done:  make(chan struct{}),
+		funcs:     make(chan func()),
+		done:      make(chan struct{}),
+		terminate: make(chan struct{}),
 	}
 }
 
@@ -51,11 +46,14 @@ func (t *OSThread) Loop() {
 	for {
 		select {
 		case fn := <-t.funcs:
-			fn.f()
-			if fn.done != nil {
-				fn.done <- struct{}{}
-			}
-		case <-t.done:
+			func() {
+				defer func() {
+					t.done <- struct{}{}
+				}()
+
+				fn()
+			}()
+		case <-t.terminate:
 			return
 		}
 	}
@@ -63,13 +61,7 @@ func (t *OSThread) Loop() {
 
 // Stop stops the thread loop.
 func (t *OSThread) Stop() {
-	close(t.done)
-}
-
-var donePool = sync.Pool{
-	New: func() interface{} {
-		return make(chan struct{})
-	},
+	close(t.terminate)
 }
 
 // Call calls f on the thread.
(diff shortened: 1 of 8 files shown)


hajimehoshi:
Please make `runOnMainThread` not to return error first. Probably we will have to fix `internal/thread`. Thanks,

changkun:
> Please make `runOnMainThread` not to return error first. Probably we will have to fix `internal/thread`. Thanks,

Done.

hajimehoshi:
```
panic: close of closed channel

goroutine 1 [running, locked to thread]:
github.com/hajimehoshi/ebiten/v2/internal/thread.(*OSThread).Loop(0xc0000248e0)
	/home/runner/work/ebiten/ebiten/internal/thread/thread.go:56 +0xfa
github.com/hajimehoshi/ebiten/v2/internal/uidriver/glfw.(*UserInterface).Run(0x94bf60, 0x7a8d00, 0x9c6840, 0x80, 0xc0004c1e78)
	/home/runner/work/ebiten/ebiten/internal/uidriver/glfw/run_notsinglethread.go:54 +0x1ab
github.com/hajimehoshi/ebiten/v2.RunGame(0x7a6880, 0xc0000248d0, 0x0, 0x0)
	/home/runner/work/ebiten/ebiten/run.go:164 +0xef
github.com/hajimehoshi/ebiten/v2/internal/testing.MainWithRunLoop(0xc000380180)
	/home/runner/work/ebiten/ebiten/internal/testing/testing.go:49 +0x68
github.com/hajimehoshi/ebiten/v2_test.TestMain(...)
	/home/runner/work/ebiten/ebiten/image_test.go:57
main.main()
	_testmain.go:275 +0x16c
FAIL	github.com/hajimehoshi/ebiten/v2	2.223s
```

Apparently the test fails...

changkun:
> Apparently the test fails...

Oppos... The panic should be resolved in the latest commit

run.go
```diff
@@ -168,6 +170,27 @@ func RunGame(game Game) error {
...
+//
+// This method panics if f panics.
+func RunOnMainThread(f func()) {
+	var err error
+	if graphicscommand.RunOnMainThread(func() {
+		// If f panics, capture the panic error and propagates to the
+		// calling goroutine.
+		defer func() {
+			if r := recover(); r != nil {
+				err = fmt.Errorf("%v", r)
```
hajimehoshi, line 183: Hmm, this loses the information about the stack trace.

IMO we don't have to `recover` a panic.

internal/thread/thread.go
```diff
@@ -14,59 +14,78 @@
...
 //
-// If f returns BreakLoop, Loop returns.
-//
 // Call blocks if Loop is not called.
-func (t *OSThread) Call(f func() error) error {
-	t.funcs <- f
-	return <-t.results
+func (t *OSThread) Call(f func()) {
+	done := donePool.Get().(chan struct{})
+	defer donePool.Put(done)
```
hajimehoshi, line 82: Do we need a pool for the `done` channel?

internal/thread/thread.go
```diff
@@ -14,59 +14,78 @@
...
 func (t *OSThread) Loop() {
-	for f := range t.funcs {
-		err := f()
-		if err == BreakLoop {
-			t.results <- nil
+	for {
+		select {
+		case fn := <-t.funcs:
+			fn.f()
+			if fn.done != nil {
```
hajimehoshi, line 55: When is `done` nil?

internal/thread/thread.go
```diff
@@ -14,59 +14,78 @@
...
+				fn.done <- struct{}{}
```
hajimehoshi, line 56: I prefer having one channel like `funcdone` at the thread rather than the function.

run.go
changkun: I thought about this, but I saw this one:
https://github.com/hajimehoshi/ebiten/pull/1927/files#diff-136576161d4f7b4c21043aa608a2b7e7d8acb542d5c7db83d17f57f2a6a6aa2aL617

If the given f panics on the main thread, then the main thread can block forever because the done channel will never be signaled.

Maybe we just send the done in a deferred call.

internal/thread/thread.go
changkun: It's only hypothetical. Maybe future main thread call could be interested in running on the main thread asynchronously, meaning we do not have to block on the Call until f returns. But since an alternative would be to spawn a new goroutine.

I can remove this.

internal/thread/thread.go
changkun: Not yet. I can remove this.

run.go
hajimehoshi: This panic should be moved out of the closure.

hajimehoshi (approved):
LGTM, thanks!

Issue #1918, opened 2021-12-27
Slow startup time


vladh:
Hello,

I've tried running the Hello World example from the website [0] on Windows, macOS and Linux, with `go run ebiten-hello.go`.

On Windows, it takes 3.7 seconds on average until the window is drawn. macOS is about the same, and Linux is slightly faster, but not by much.

I've compared this with a minimal example using https://github.com/go-gl/glfw, which takes 1.9s to display the window. Similarly, https://github.com/faiface/pixel also takes 1.9s.

Personally, I already find 1.9s to be quite long, but since go-gl/glfw doesn't run any faster, I assume this is the fastest time we can get currently.

However, 3.7 seconds is quite a lot, and it's a factor that discourages me from using Ebiten. Do you think this time could be improved? Can you think of any ways to run some profiling to figure out where specifically the issue is? I would also be happy to contribute to development if you had a few tips to help me along.

Thank you for developing Ebiten! 本当にありがとうございます！

[0]: https://ebiten.org/tour/hello_world.html

hajimehoshi:
Thank you for reporting the issue!

We need more data to know what is the culprit. Would it be possible to take a profile for this?

My rough guess is that Windows needs to extract a GLFW DLL file:

https://github.com/hajimehoshi/ebiten/blob/374871c03135d4a9912e6b12c898fec2fe655de5/internal/glfw/notexternaldll_windows.go#L49-L81

vladh:
Thank you for the quick reply! The only way I know to take a profile is with `pprof.StartCPUProfile(f)`. Please let me know if that's not what you need.

I've taken a profile with the above code, and you can find it here: [ebiten-hello.prof](https://vladh.net/things/ebiten-hello.prof].

hajimehoshi:
Thank you very much! Hmm, from the trace, the cost of calling C function is relatively high, but I don't get more insights than that. I need tracing (https://pkg.go.dev/runtime/trace). Let me try.

hajimehoshi:
![image](https://user-images.githubusercontent.com/16950/147593373-dc12838c-c75d-4f6e-9c5f-b728d6c57bfb.png)

Hmm, there is no useful information at tracing at all...

vladh:
Would it help if I did some tests on various operating systems and computers? I'm happy to help any way I can.

hajimehoshi:
[...]

PR #1931, opened 2022-01-04
Draft: #1918 Slow startup, debugging log and maybe fix

rolfrm:
This PR contains logging and a potential fix for #1918 . Maybe the fix is not the right thing, I am not sure what the function does or why. But this return fixes the problem for me, maybe you can use it as a hint.


hajimehoshi:
Thanks but I've already addressed on the culprit that slows the startup time.

Let me close this

Issue #1258, opened 2020-07-21
use //go:embed to embed files


hajimehoshi:
https://go.googlesource.com/proposal/+/master/design/draft-embed.md

gabstv:
TIL! Great!

hajimehoshi:
I'll use this when Go 1.17 is released (not Go 1.16). Then probably in August 2021.

dnnnvx:
My two cents: works fine using the new `embed` feature in `go v1.16`, like this

```go
import (
	"bytes"
	"embed"
	"image"
	"log"

	_ "image/png"

	"github.com/hajimehoshi/ebiten/v2"
)

//go:embed assets/*
var assetsFolder embed.FS

imgByte, err := assetsFolder.ReadFile("assets/main/idle.png")
if err != nil {
	log.Fatalln(err)
}
img, _, err := image.Decode(bytes.NewReader(imgByte))
if err != nil {
	log.Fatalln(err)
}
ebitenImg := ebiten.NewImageFromImage(img)
```

hajimehoshi:
Let's do this after Debian's default Go version is 1.16 or newer.

hajimehoshi:
Debian Bookworm (probably with Go 1.19) will be released in 2023-09, maybe.

hajimehoshi:
Note to myself:

We deterine the Go version by min(Debian's stable [1], Ubuntu's stable [2])

[1] https://packages.debian.org/stable/golang
[2] https://packages.ubuntu.com/jammy/golang (The URL indicates Ubuntu 22, rather than the stable)

PR #1946, opened 2022-01-10
Easiest usage of "embed"

neni:
Add new function ebitenutil.NewImageFromEmbedFile(embedFs embed.FS, path string) (*ebiten.Image, image.Image, error)

Change ebitenutil.NewImageFromURL(url string) (*ebiten.Image, error) to ebitenutil.NewImageFromURL(url string) (*ebiten.Image, image.Image, error)

Add "embed" example

Add "url" example


ebitenutil/loadembedimage.go
```diff
@@ -0,0 +1,40 @@
...
+	"image"
+
+	"github.com/hajimehoshi/ebiten/v2"
+)
+
+// NewImageFromEmbedFile loads the file from embed.FS with path and returns ebiten.Image and image.Image.
+//
+// Image decoders must be imported when using NewImageFromEmbedFile. For example,
+// if you want to load a PNG image, you'd need to add `_ "image/png"` to the import section.
+func NewImageFromEmbedFile(embedFs embed.FS, path string) (*ebiten.Image, image.Image, error) {
```
eliasdaler, line 31: This should probably be a `fs.FS`. No need to restrict to `embed.FS`
I'd also name this `NewImageFromFileSystem`.

examples/embed/main.go
```diff
@@ -0,0 +1,77 @@
...
+	"embed"
+	_ "image/jpeg"
+	"log"
+
+	"github.com/hajimehoshi/ebiten/v2"
+	"github.com/hajimehoshi/ebiten/v2/ebitenutil"
+)
+
+// Embedding files from assets directory
+//go:embed assets/*
```
eliasdaler, line 30: I feel like it'd be better to embed a single file here

examples/url/main.go
```diff
@@ -0,0 +1,73 @@
+// Copyright 2014 Hajime Hoshi
```
eliasdaler, line 1: Not sure if this is needed. Plus, it depends on external resources...

eliasdaler (changes requested):

ebitenutil/loadembedimage.go
neni: That's absolutely true.

examples/embed/main.go
neni: That's sure for a single file, but it is for show how to use embed.FS

examples/url/main.go
neni: The changes are cosmetic differences from original files.

neni (commented):
Use fs.FS in place of embed.FS 
Rename NewImageEmbedFile() to NewImageFromFileSystem()

examples/embed/main.go
eliasdaler: You can specify only one file and still use embed.FS
See the example here: https://pkg.go.dev/embed
```go
import "embed"

//go:embed hello.txt
var f embed.FS
data, _ := f.ReadFile("hello.txt")
print(string(data))
```

examples/url/main.go
eliasdaler: These are not changes, these are additions. And this is a code duplication which doesn't act as a useful example, in my opinion.
I'll let Hajime decide if this is needed or not.

neni (commented):
Fix "highdpi" example

examples/url/main.go
neni: I agree, "url" Example is not needed (I thought you were talking of "Copyright 2014 Hajime Hoshi").

hajimehoshi:
* `go:embed` is not allowed yet. See https://github.com/hajimehoshi/ebiten/issues/1258
* Breaking changes in existing APIs are not allowed.
* In general, please create an issue first if you want to suggest a new API.

PR #1952, opened 2022-01-12
internal/graphicscommand: pending calls if not on the mainthread

### LINKED ISSUE #1945: ebiten: forbid `RunOnMainThread` before the main loop
For example, there is no way to make this case work:

```go
func init() {
  go func() { ebiten.RunOnMainThread() }
}
```

/CC @changkun

changkun:
Fixes #1945

### DIFF
--- a/internal/graphicscommand/thread.go
+++ b/internal/graphicscommand/thread.go
@@ -14,7 +14,32 @@
 
 package graphicscommand
 
-var theThread Thread
+import (
+	"runtime"
+	"sync"
+)
+
+var (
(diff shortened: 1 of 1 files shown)


hajimehoshi:
* I don't want to rely on the gid hack. The thread struct is not only for the main thread anyway.
* When pending the given function, RunOnMainThread will behave in a different way than the user expects. RunOnMainThread should call the function synchronously.

hajimehoshi:
As there have been no updates recently, let me close this.

Issue #1921, opened 2021-12-28
docs: add "black hole" shader to shader examples


MrWormHole:
I am thinking to add a new shader example which is going to be a black hole. Do you think it is a good idea for Kage?

hajimehoshi:
Could you show me an actual example of the black hole shader? Thanks!

MrWormHole:
Here are the implementation details I have found [gravitational lens shader aka black hole](https://unionassets.com/blog/the-effect-of-the-gravitational-lens-195). It is in HLSL which is GLSL of the unity engine. 

I am quite optimistic about its implementation in Kage, hoping it will be ready as an example this weekend

hajimehoshi:
Let me close this. If you revisit this, feel free to reopen this.

PR #1958, opened 2022-01-15
example add blackhole shader

MrWormHole:
#1921 was the main issue that was opened, this adds an example. Let me know if this makes sense, had a bit of help from @Zyko0 but we still couldn't remove the bubbling behaviour which doesn't happen here in this example

https://www.shadertoy.com/view/Xt3GWj


examples/shader/blackhole.go
```diff
@@ -0,0 +1,40 @@
+// Copyright 2020 The Ebiten Authors
```
hajimehoshi, line 1: 2022

examples/shader/blackhole.go
```diff
@@ -0,0 +1,40 @@
...
+
+package main
+
+var Time float
+var ScreenSize vec2
+var Cursor vec2
+
+func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
+	uv := position.xy / ScreenSize.xy
+	texC := uv
```
hajimehoshi, line 26: This texC is not used anywhere.

examples/shader/blackhole.go
```diff
@@ -0,0 +1,40 @@
...
+	lpos := Cursor.xy / ScreenSize.x
+	texC2 := position.xy / ScreenSize.x
```
hajimehoshi, line 28: Dividing the position (texel) by the screen size (pixel) doesn't make sense...

Perhaps you might want:

```
        pos := position.xy / imageDstTextureSize()
        origin, size := imageDstRegionOnTexture()
        pos -= origin
        pos /= size
```

See texel.go

examples/shader/blackhole.go
Zyko0: It is used as texture coordinates in `imageSrcNAt` a few lines below

examples/shader/blackhole.go
Zyko0: @hajimehoshi `position vec4` isn't pixels not pixels

But it makes sense using these functions, as for now it just works "by chance" if I'm correct about the potential offsets of `dst` on the offscreen ?

I'd go for
```go
origin, size := imageDstRegionOnTexture()
texC := (position.xy/imageDstTextureSize() - origin) / size
texC2 := (position.xy/imageDstTextureSize().x - origin) / size.x
lpos := Cursor.xy / ScreenSize.x
```

examples/shader/blackhole.go
hajimehoshi: texC is reassigned before passing it to `imageSrcNAt`

examples/shader/blackhole.go
hajimehoshi: > position vec4 isn't texels but pixels

You are right, sorry

examples/shader/blackhole.go
Zyko0: Oh you meant like that, yeah uv could also go away I believe then (more like `uv` is only used with `texC`)
I think my block from below comment might cover this

hajimehoshi: > texC = mix(texC2, (texC*2.0-lpos*2.0)*0.2*0.5+lpos, (1.0 / (distance((texC2*2.0-lpos*2.0)*10.0*0.5+lpos, lpos) - 1.0)))

Ah, I found that `texC` is used in the right side, but is this intended?

hajimehoshi: Using the same variable name for multiple contexts is confusing. Let's use a different variable names.

Zyko0: Yeah it was intended because the whole shader is a distortion of texture coordinates, and there was no extra need for an additional variable (also it's translated from shadertoy)

I agree additional variables might help describe the different steps better, but I wouldn't know how to name it :/
Maybe keep uv for the first transformations, and declare `texC = mix(` later using `uv` you mean ?

examples/shader/blackhole.go
```diff
@@ -0,0 +1,40 @@
...
+package main
+
+var Time float
+var ScreenSize vec2
+var Cursor vec2
+
+func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
+	origin, size := imageDstRegionOnTexture()
+	texC := (position.xy/imageDstTextureSize() - origin) / size
+	texC2 := (position.xy/imageDstTextureSize().x - origin) / size.x
```
hajimehoshi, line 27: I'm not sure what this means. Why is the position divided by the width?

examples/shader/blackhole.go
```diff
@@ -0,0 +1,40 @@
...
+	lpos := Cursor.xy / ScreenSize.x
+	texC = mix(texC2, (texC*2.0-lpos*2.0)*0.4*0.5+lpos, (1.0 / (distance((texC2*2.0-lpos*2.0)*10*0.5+lpos, lpos) - 1.0)))
```
hajimehoshi, line 29: Can we use a different variable name to avoid confusion?

examples/shader/blackhole.go
MrWormHole: I am not sure either, it was adjusted from this shader toy example https://www.shadertoy.com/view/Xt3GWj

examples/shader/blackhole.go
MrWormHole: Only if I knew how to rename 😄 open to any naming here

examples/shader/blackhole.go
hajimehoshi: Please make the behavior clear to you before sending this PR... Thanks!

Zyko0: It is divided by the longest side (so the width usually on screens) so that the circle doesn't appear streched on X axis

Maybe we should try running this example with a resolution that has a height > width ?
That would mean we should divide it by the max of both

eliasdaler:
Here's the main question, though... Does this example show something that wasn't shown in previous ones?
I feel like the shaders in examples should be as simple as possible so that it's easy to understand what's going on and how the particular feature of Kage is used.
Otherwise there's many more shader examples that could be added, but they can be easily shared as GitHub gists, for example.

hajimehoshi:
> Here's the main question, though... Does this example show something that wasn't shown in previous ones?
> I feel like the shaders in examples should be as simple as possible so that it's easy to understand what's going on and how the particular feature of Kage is used.

I don't think the new shader must include an entirely new thing, but I agree that the example must include what is easy to understand to readers.

MrWormHole:
> > Here's the main question, though... Does this example show something that wasn't shown in previous ones?
> > I feel like the shaders in examples should be as simple as possible so that it's easy to understand what's going on and how the particular feature of Kage is used.
> 
> I don't think the new shader must include an entirely new thing, but I agree that the example must include what is easy to understand to readers.

I don't know about this much, aside from some built-in stuff like mix and src/dest functions it doesn't really show too much difference, I think cool examples is a must-have regardless of complexity but the logic here is too hard to formulate for me who is coming from frontend/backend job background. It would be cool to simplify this and show in the examples with its formula

eliasdaler:
Well... `mix` can be shown by drawing a rect/triangle and then shifting the colors on it. I feel like shaders in example should be as minimal as possible so it's clear what each function can be used for. If it's all in one place, it's harder to understand.

If you're showing how "+" and "*" work, you'd better show this:
```go
fmt.Println(a + b)
fmt.Println(a * b)
```
... then this:
```go
fmt.Println(a + b * c)
```

MrWormHole:
I agree on that one with you @eliasdaler , I wanted to add an example because existing ones were already complex. Maybe we should look for differences that can be done for galleries (complex ones) and examples(very minimal basic ones). If you look at the water shader, that is also super complex. 2 different sections is what I mean

eliasdaler:
I agree, the water shader is not that easy. I feel like there's merit in having both complex and simple shaders... But it'd be better to have more "simple" ones and less "complex" ones, because complex ones can usually be ported from ShaderToy's shaders or other resources.

What would be great to have is to show examples of porting GLSL shaders to Kage with GLSL code written on top and then corresponding Kage code being very similar.

hajimehoshi:
As there have been no changes recently, let me close this.

Issue #485, opened 2018-01-23
"Console Window Host" process is run when running a built app with Ebiten


hajimehoshi:
[![Screenshot from Gyazo](https://gyazo.com/d3c2e0cd975002e9bdee3614b31313db/raw)](https://gyazo.com/d3c2e0cd975002e9bdee3614b31313db)

1. `go build -o main.exe path/to/your/app`
2. Run the `main.exe`

hajimehoshi:
`go build -ldflags -H=windowsgui path/to/your/app` solved this. I'll note this at the Wiki.

hajimehoshi:
https://github.com/hajimehoshi/ebiten/wiki/FAQ#my-app-built-with-ebiten-creates-an-unknown-process-console-window-host-what-is-this done.

Ebiten might be able to kill the process, so I leave this issue open.

PR #1961, opened 2022-01-16
Use FreeConsole instead of hiding

Zyko0:
Issue: https://github.com/hajimehoshi/ebiten/issues/485

Using `FreeConsole` (https://docs.microsoft.com/en-us/windows/console/freeconsole) instead of hiding the console correctly works on my machine.

Calling it in every case, doesn't necessarily kill the console, but prevent logs from working, so it still needs to be done when we detect that current process created its own console.

From what is stated on `FreeConsole`'s page, the console will get destroyed automatically when no process refers to it anymore.
So it works in our case, since our process created its own console.

Also if this is adopted, `showWindowAsyncProc` reference could be dropped

### DIFF
--- a/internal/uidriver/glfw/hideconsole_windows.go
+++ b/internal/uidriver/glfw/hideconsole_windows.go
@@ -33,7 +33,6 @@ var (
 	getConsoleWindowProc         = kernel32.NewProc("GetConsoleWindow")
 	freeConsoleWindowProc        = kernel32.NewProc("FreeConsole")
 	getWindowThreadProcessIdProc = user32.NewProc("GetWindowThreadProcessId")
-	showWindowAsyncProc          = user32.NewProc("ShowWindowAsync")
 )
 
 func getCurrentProcessId() (uint32, error) {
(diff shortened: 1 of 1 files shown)


internal/uidriver/glfw/hideconsole_windows.go
```diff
@@ -60,6 +61,14 @@ func getConsoleWindow() (uintptr, error) {
...
 
+func freeConsole() error {
+	_, _, e := freeConsoleWindowProc.Call()
+	if e != nil && e.(windows.Errno) != 0 {
+		return fmt.Errorf("ui: FreeConsole failed: %d", e)
+	}
+	return nil
+}
+
 func showWindowAsync(hwnd uintptr, show int) error {
```
hajimehoshi, line 72: Do we still use this?

Zyko0: No, as mentioned !
Should I remove it ?

hajimehoshi:
Did you confirm the process Console Window Host was gone?

Zyko0:
Yes that's what I observed
Since the ebiten process creates its own console host, calling `FreeConsole` detached the only process attached to it, which gets it killed.
At least I cannot see it anymore in taskmgr under the main process

I'd be okay if someone can confirm this also, because I may have missed an edge case (I have 95% confidence)

internal/uidriver/glfw/hideconsole_windows.go
hajimehoshi: Yes, if this is not used anywhere.

Zyko0: Done!

hajimehoshi (approved):
LGTM, thanks!

hajimehoshi:
Confirmed this worked on my Windows machine!

Issue #1960, opened 2022-01-16
window: resize/float broken on i3 (and possibly other tiling WMs)


eliasdaler:
OS: Ubuntu 20.04
WM: i3 4.17

---

When I enable/disable float, I don't get new values of width/height in `Layout` function. 
Enabling/disabling float is similar to "maximize" button behaviour on Windows.

The behaviour works fine with SDL, SFML and glfw programs.
xev shows that the following event gets sent when I disable the float:

```sh
Expose event, serial 25, synthetic NO, window 0x9a00001,
    (0,68), width 1912, height 988, count 0
```

However, `Layout` doesn't send (1912, 988)

PR #1962, opened 2022-01-16
Fix window float enable/disable (#1960)

### LINKED ISSUE #1620: Glitches are shown when resizing the window (Linux)
This is forked from #1618 

Video: https://github.com/hajimehoshi/ebiten/issues/1618#issuecomment-825659074 (The window froze in the video, but freezing has already been fixed)

CC @sedyh

eliasdaler:
When resizing with i3 (and possibly other WMs), the only way to now new framebuffer/window size is to get event from glfw.
This also calls `setWindowSizeInDIP` to update window size afterwards (instead of calling `context.Layout` directly)

Framebuffer size callbacks are used for handling window resize in glfw examples [like here](https://github.com/glfw/glfw/blob/master/examples/gears.c#L248)

I'm not 100% sure that this is the best solution, so any suggestions welcome.
@sedyh this might also fix #1620 - can you please check (there are still glitches, but maybe it's somewhat better)?

### DIFF
--- a/internal/uidriver/glfw/ui.go
+++ b/internal/uidriver/glfw/ui.go
@@ -823,8 +823,9 @@ event:
 		}
 	}
 	if u.defaultFramebufferSizeCallback == 0 {
-		// When window gets resized (either by manual window resize or a window
-		// manager), glfw sends a framebuffer size callback which we need to handle
+		// When the window gets resized (either by manual window resize or a window
+		// manager), glfw sends a framebuffer size callback which we need to handle (#1960).
+		// This event is the only way to handle the size change at least on i3 window manager.
 		u.defaultFramebufferSizeCallback = glfw.ToFramebufferSizeCallback(func(_ *glfw.Window, w, h int) {
 			u.setWindowSizeInDIP(w, h, u.isFullscreen())
 		})
--- a/internal/uidriver/glfw/ui.go
+++ b/internal/uidriver/glfw/ui.go
@@ -823,6 +823,8 @@ event:
 		}
 	}
 	if u.defaultFramebufferSizeCallback == 0 {
+		// When window gets resized (either by manual window resize or a window
+		// manager), glfw sends a framebuffer size callback which we need to handle
 		u.defaultFramebufferSizeCallback = glfw.ToFramebufferSizeCallback(func(_ *glfw.Window, w, h int) {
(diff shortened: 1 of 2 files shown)


internal/uidriver/glfw/ui.go
```diff
@@ -821,7 +822,13 @@ event:
 			time.Sleep(time.Millisecond)
 		}
 	}
-	window.SetFramebufferSizeCallback(glfw.ToFramebufferSizeCallback(nil))
+	if u.defaultFramebufferSizeCallback == 0 {
+		u.defaultFramebufferSizeCallback = glfw.ToFramebufferSizeCallback(func(_ *glfw.Window, w, h int) {
```
hajimehoshi, line 826: Please leave a comment explaining why this handling is needed.

hajimehoshi:
Did you confirm this was not problematic on Windows and macOS?

eliasdaler:
> Did you confirm this was not problematic on Windows and macOS?

Tested on Windows, works fine. Can't test on macOS unfortunately (don't have a machine running it)

internal/uidriver/glfw/ui.go
```diff
@@ -821,7 +822,15 @@ event:
 			time.Sleep(time.Millisecond)
 		}
 	}
-	window.SetFramebufferSizeCallback(glfw.ToFramebufferSizeCallback(nil))
+	if u.defaultFramebufferSizeCallback == 0 {
+		// When window gets resized (either by manual window resize or a window
+		// manager), glfw sends a framebuffer size callback which we need to handle
```
hajimehoshi, line 827: * Add the issue number.
* This problem is confirmed by i3 window manager but not by other managers, right?

So the comment should be like this:

```
		// When the window gets resized (either by manual window resize or a window
		// manager), glfw sends a framebuffer size callback which we need to handle (#1960).
		// This event is the only way to handle the size change at least on i3 window manager.
```

eliasdaler: > This problem is confirmed by i3 window manager but not by other managers, right?

Yes, but I might try other tiling WMs, I'm just not used to other WMs. 
Also tested that xfwm (default XFCE window manager, not a tiling WM) works fine with this patch too

hajimehoshi (approved):
LGTM

PR #1973, opened 2022-01-20
examples/mandelbrot: Reduce CPU cost by not updating the window

hermunn:
When the Mandelbrot calculation is over, there is no reason to continuously paint the window contents.

### DIFF
--- a/examples/mandelbrot/main.go
+++ b/examples/mandelbrot/main.go
@@ -103,6 +103,8 @@ func (g *Game) Layout(outsideWidth, outsideHeight int) (int, int) {
 func main() {
 	ebiten.SetWindowSize(screenWidth, screenHeight)
 	ebiten.SetWindowTitle("Mandelbrot (Ebiten Demo)")
+	ebiten.SetFPSMode(ebiten.FPSModeVsyncOffMinimum)
+
 	if err := ebiten.RunGame(NewGame()); err != nil {
 		log.Fatal(err)
 	}


hermunn:
I have just started to learn Go and ebiten, so I might be completely off here. I think the docs on exactly what happens when

``` go
ebiten.SetFPSMode(ebiten.FPSModeVsyncOffMinimum)
```
is missing, but I might be looking at the wrong place? Can I improve this PR by adding text to some docs?

hermunn:
Sorry, the last commit is just a rebase to current `main`.

Zyko0:
> When the Mandelbrot calculation is over, there is no reason to continuously paint the window contents.

Actually this is the reason an offscreen image is used, not to recalculate the whole Mandelbrot every time.
But then, the offscreen image doesn't change anymore, it is just being "pasted" on screen image every frame.
The same way you would "paste" an image loaded from disk on screen every frame, this one has just been computed at runtime but shouldn't augment CPU usage.

> I think the docs is missing

https://github.com/hajimehoshi/ebiten/blob/main/run.go#L359-L367
But `ebiten.FPSModeVsyncOffMinimum` is to be used in very specific cases, I don't think it is suitable for this example 😅

hajimehoshi:
Thank you for the PR! Changing the FPS mode would reduce the CPU usage actually. In general, `ebiten.FPSModeVsyncOffMinimum` is for a GUI application. The example doesn't update the screen so `ebiten.FPSModeVsyncOffMinimum` might be fine, but I don't think this is a necessary fix.

hermunn:
> ... The example doesn't update the screen so `ebiten.FPSModeVsyncOffMinimum` might be fine, but I don't think this is a necessary fix.

I understand how the example works and the reason why `mandelbrot` has an offscreen buffer. On my laptop running Ubuntu 20.04 (Focal Fossa), `top` shows a 30% CPU usage when the process is running (ie. 30% of one core). It is not enough to make the fan spin up, but slower PCs might run a bit hotter. Clearing and painting the same image 60 frames a second is not very expensive, but it is not free, and I think the PR would be a nice improvement. If ebiten is supposed to be used also for causal games where the screen is mostly static, then at least one example with `FPSModeVsyncOffMinimum` seems like a good idea, at least.

Zyko0:
Yeah I think the CPU issue should be tackled elsewhere even with vsync on, because it wouldn't be viable for many games to be set in `FPSModeVsyncOffMinimum`

>  then at least one example with FPSModeVsyncOffMinimum seems like a good idea
Indeed ! But I think it's covered by this one:
https://github.com/hajimehoshi/ebiten/blob/12c144070f27a086cf80c6742e34c128cba5dbff/examples/windowsize/main.go#L208-L219

I believe `FPSModeVsyncOffMinimum` is a specific notion, and it would probably improve CPU in many other /examples but not setting this mode in the examples, make them simpler to understand and closer to typical-games I feel like.
It should be okay, since examples are not meant to be run 24/7, but mostly to show features or ebiten capabilities

hermunn:
Thanks for the kind words. I am a newcomer to Go and ebiten, but not to Mandelbrot programming. :)

We can close this PR, because I fill inspired to implement a Mandelbrot example with some user interface, like I often do when I try to learn a new language. This will be with goroutines (for calculation) and correct usage of `ScheduleFrame`, and I plan to make a new PR when I am done.

It will not be completely straightforward to read, but my goal will be to create something which uses the API in a reasonable way.

PR #1981, opened 2022-02-02
Allow GLES2 builds outside of iOS/Android

AndreRenaud:
This keeps the tags compatible with those use by https://github.com/go-gl/glfw


AndreRenaud:
I just realised this is incomplete. I will resubmit when it is finished.

PR #1994, opened 2022-02-19
Add Discard Builtin Function for GL and Metal

tomlister:
Added a builtin function named `discard()` to call `discard;` in GLSL and `discard_fragment();` in MSL.

#1968

### DIFF
--- a/internal/shader/expr.go
+++ b/internal/shader/expr.go
@@ -281,12 +281,20 @@ func (cs *compileState) parseExpr(block *block, expr ast.Expr, markLocalVariable
 		if !ok {
 			return nil, nil, nil, false
 		}
+
+		if len(ss) != 0 {
+			stmts = append(stmts, ss...)
+		}
+
+		if len(es) == 0 {
+			return nil, nil, stmts, true
+		}
+
 		if len(es) != 1 {
 			cs.addError(e.Pos(), fmt.Sprintf("multiple-value context is not available at a callee: %s", e.Fun))
 			return nil, nil, nil, false
 		}
 		callee = es[0]
-		stmts = append(stmts, ss...)
 
 		// For built-in functions, we can call this in this position. Return an expression for the function
 		// call.
@@ -500,6 +508,13 @@ func (cs *compileState) parseExpr(block *block, expr ast.Expr, markLocalVariable
 					Index: i,
 				},
 			}, []shaderir.Type{cs.ir.Uniforms[i]}, nil, true
+		}
+		if e.Name == "discard" {
+			return nil, nil, []shaderir.Stmt{
+				{
+					Type: shaderir.Discard,
+				},
+			}, true
 		}
 		if f, ok := shaderir.ParseBuiltinFunc(e.Name); ok {
 			return []shaderir.Expr{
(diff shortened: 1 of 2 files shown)


internal/shader/expr.go
```diff
@@ -281,12 +281,20 @@ func (cs *compileState) parseExpr(block *block, expr ast.Expr, markLocalVariable
...
 			return nil, nil, nil, false
 		}
+
+		if len(ss) != 0 {
+			stmts = append(stmts, ss...)
+		}
+
+		if len(es) == 0 {
+			return nil, nil, stmts, true
+		}
```
hajimehoshi, line 291: Why do you need this fix?

hajimehoshi (commented):
Let's add a test at internal/shader!

internal/shader/expr.go
tomlister: Because discard doesn't return a function, but a statement.
It avoids throwing this error
```go
if len(es) != 1 {
	cs.addError(e.Pos(), fmt.Sprintf("multiple-value context is not available at a callee: %s", e.Fun))
	return nil, nil, nil, false
}
```

internal/shader/expr.go
```diff
@@ -281,12 +281,20 @@ func (cs *compileState) parseExpr(block *block, expr ast.Expr, markLocalVariable
 		if !ok {
 			return nil, nil, nil, false
 		}
+
+		if len(ss) != 0 {
```
hajimehoshi, line 285: This `if` is not needed.

internal/shader/expr.go
```diff
@@ -281,12 +281,20 @@ func (cs *compileState) parseExpr(block *block, expr ast.Expr, markLocalVariable
...
+		if len(es) == 0 {
+			return nil, nil, stmts, true
+		}
+
 		if len(es) != 1 {
 			cs.addError(e.Pos(), fmt.Sprintf("multiple-value context is not available at a callee: %s", e.Fun))
 			return nil, nil, nil, false
 		}
 		callee = es[0]
-		stmts = append(stmts, ss...)
```
hajimehoshi, line 289: Do we really need to move this line?

tomlister: I believe so. All the changes make it work

tomlister: What happens is, that we don't have anything in `es`, but the existing code expects `len(es)` to equal 1 to avoid a multi return func, which blocks adding of our discard statement. Further, we do return something in `es` to avoid this check, then it will still error as whatever is in `es` does not have a call type.

internal/shader/expr.go
hajimehoshi: append works without this if.

internal/shader/expr.go
```diff
@@ -276,17 +276,23 @@ func (cs *compileState) parseExpr(block *block, expr ast.Expr, markLocalVariable
 			stmts = append(stmts, ss...)
 		}
 
-		// TODO: When len(ss) is not 0?
+		// TODO: When len(ss) is not 0? relevant to discard?
```
hajimehoshi, line 279: Yes, let's remove this comment.

hajimehoshi:
Any updates?

tomlister:
I'll spend time on this in coming days. Just busy with university and work!
😊

hajimehoshi:
Any progress?

Issue #1772, opened 2021-08-20
ebiten: `SetScreenFilterEnabled` to switch the screen filter


divVerent:
On my system here, the final blit from game output resolution to screen (both 3840x2160) takes 4.3ms.

However, I want to be able to customize this filtering to apply a CRT-style filter.

This however has a massive framerate cost as ebiten still performs a final blit from the output image I can write to to the screen. Is there any way how we can either skip or customize that blit?

As for how I measured the impact: I made a custom ebiten built with this delta:

```
diff --git a/uicontext.go b/uicontext.go
index 18ae973d..11096e81 100644
--- a/uicontext.go
+++ b/uicontext.go
@@ -190,6 +190,12 @@ func (c *uiContext) updateFrameImpl(updateCount int) error {
                uiDriver().ResetForFrame()
        }
 
+       if IsScreenClearedEveryFrame() {
+               c.screen.Clear()
+       }
+       c.game.Draw(c.screen)
+       return nil
+
        // Even though updateCount == 0, the offscreen is cleared and Draw is called.
        // Draw should not update the game state and then the screen should not be updated without Update, but
        // users might want to process something at Draw with the time intervals of FPS.
```

Clearly this can't seriously be applied that way, especially as it also flips the screen upside down. Other than that, everything worked perfectly and I got a massive fps boost with my changes:

- 88.7fps to 144fps with my CRT shader (and previously regularly dropping frames with vsync enabled, not with the change)
- 128fps to 263fps with a simpler CRT-ish shader
- 112fps to 222fps with a simple linear-filtered DrawImage call
- 141fps to 369fps with a simple nearest-filtered DrawImage call

GPU is: Mesa Intel(R) UHD Graphics (CML GT2)

So I suggest one of the following as a future API change:

- Optimize processing in case game-written screen image and output image are the same size. This is complicated due to flipping still being required - wonder though if the flipping could be handled while rendering to the image. Also due to the int/float conversions regarding Layout(), it is not always possible for a game to cause these two to be the same size, breaking this optimization on some odd HiDPI systems or window sizes.
- Provide a way to replace the current FilterLinear/filterScreen logic by a custom shader and a DrawRectShader call instead.
  - As flipping may be enabled or disabled at that point, and some shaders may not be symmetric regarding this (mine is), this may require providing an Uniform to the shader containing the flip status. Not sure, this may also be inferrable another way, although dfdy on the texcoord of the input texture sounds like a bad idea.

[...]

PR #1996, opened 2022-02-22
Optimization: send premultiplied alpha from vertex to fragment shader.

divVerent:
Note that this applies only to the builtin shaders - interface for Kage stays
unchanged for compatibility.

Minor compatibility delta: when interpolating alpha values, previous code has
created nonsense values, such as, when interpolating from
fully-transparent-black (0,0,0,0) to opaque-white (1,1,1,1), something like
half-transparent-grey (0.25,0.25,0.25,0.5) where half-transparent-white
(0.5,0.5,0.5,0.5) is used by the new code.

I assume this is a strict improvement, however this may warrant some testing.

Possible later improvement could be moving the premultiplication from fragment
shader to CPU. Did not do this as it makes the code rather inconsistent of Kage
vs built-in shader usage.

Helps with https://github.com/hajimehoshi/ebiten/issues/1772, achieved 23fps->27fps and 36fps->48fps on a Dell E6500.

### DIFF
--- a/internal/atlas/image.go
+++ b/internal/atlas/image.go
@@ -429,9 +429,18 @@ func (i *Image) drawTriangles(srcs [graphics.ShaderImageNum]*Image, vertices []f
 		g := colorm.At(1, 1)
 		b := colorm.At(2, 2)
 		a := colorm.At(3, 3)
-		if r <= a && g <= a && b <= a {
-			// Scale-only color matrix, and no color clamping needed either.
-			// (The identity color matrix case in the shader doesn't do that.)
+		if r <= 1 && g <= 1 && b <= 1 && a >= 0 {
+			// Color matrices work on non-premultiplied colors.
+			// This color matrix can only make colors darker or equal,
+			// and thus can never invoke color clamping.
+			// Thus the simpler vertex color scale based shader can be used.
+			//
+			// Color values being negative with positive alpha is harmless -
+			// will be clamped away when converting to integer texture format anyway.
+			//
+			// However, if both color values and alpha were negative
+			// (which can also be caused by vertex colors below, if negative alpha were allowed here),
+			// this can cause a difference in the min() call in the shader, yielding different alpha-zero outputs.
 			cr, cg, cb, ca = r, g, b, a
 			colorm = affine.ColorMIdentity{}
 		}
--- a/internal/atlas/image.go
+++ b/internal/atlas/image.go
@@ -425,11 +425,16 @@ func (i *Image) drawTriangles(srcs [graphics.ShaderImageNum]*Image, vertices []f
 	cb := float32(1)
 	ca := float32(1)
 	if !colorm.IsIdentity() && colorm.ScaleOnly() {
-		cr = colorm.At(0, 0)
-		cg = colorm.At(1, 1)
-		cb = colorm.At(2, 2)
-		ca = colorm.At(3, 3)
-		colorm = affine.ColorMIdentity{}
+		r := colorm.At(0, 0)
+		g := colorm.At(1, 1)
+		b := colorm.At(2, 2)
+		a := colorm.At(3, 3)
+		if r <= a && g <= a && b <= a {
+			// Scale-only color matrix, and no color clamping needed either.
+			// (The identity color matrix case in the shader doesn't do that.)
+			cr, cg, cb, ca = r, g, b, a
+			colorm = affine.ColorMIdentity{}
+		}
 	}
 
 	var dx, dy float32
--- a/internal/atlas/image.go
+++ b/internal/atlas/image.go
@@ -420,6 +420,15 @@ func (i *Image) drawTriangles(srcs [graphics.ShaderImageNum]*Image, vertices []f
 		i.processSrc(src)
 	}
 
+	// If a color matrix is used, but the matrix is merely a scaling matrix,
+	// and the scaling cannot cause out-of-range colors, do not use a color matrix
+	// when rendering but instead multiply all vertex colors by the scale.
+	// This speeds up rendering.
+	//
+	// NOTE: this is only safe when not using a custom Kage shader,
+	// as custom shaders may be using vertex colors for different purposes
+	// than colorization. However, currently there are no Ebiten APIs that
+	// support both shaders and color matrices.
 	cr := float32(1)
 	cg := float32(1)
 	cb := float32(1)
--- a/internal/atlas/image.go
+++ b/internal/atlas/image.go
@@ -429,18 +429,17 @@ func (i *Image) drawTriangles(srcs [graphics.ShaderImageNum]*Image, vertices []f
 		g := colorm.At(1, 1)
 		b := colorm.At(2, 2)
 		a := colorm.At(3, 3)
-		if r <= 1 && g <= 1 && b <= 1 && a >= 0 {
+		if r >= 0 && g >= 0 && b >= 0 && a >= 0 && r <= 1 && g <= 1 && b <= 1 {
 			// Color matrices work on non-premultiplied colors.
 			// This color matrix can only make colors darker or equal,
 			// and thus can never invoke color clamping.
 			// Thus the simpler vertex color scale based shader can be used.
 			//
-			// Color values being negative with positive alpha is harmless -
-			// will be clamped away when converting to integer texture format anyway.
+			// Negative color values can become positive and out-of-range
+			// after applying to vertex colors below, which can make the min() in the shader kick in.
 			//
-			// However, if both color values and alpha were negative
-			// (which can also be caused by vertex colors below, if negative alpha were allowed here),
-			// this can cause a difference in the min() call in the shader, yielding different alpha-zero outputs.
+			// Alpha values smaller than 0, combined with negative vertex colors,
+			// can also make the min() kick in, so that shall be ruled out too.
 			cr, cg, cb, ca = r, g, b, a
 			colorm = affine.ColorMIdentity{}
 		}
--- a/internal/graphicsdriver/metal/graphics_darwin.go
+++ b/internal/graphicsdriver/metal/graphics_darwin.go
@@ -83,7 +83,8 @@ vertex VertexOut VertexShader(
   VertexOut out = {
     .position = projectionMatrix * float4(in.position, 0, 1),
     .tex = in.tex,
-    .color = in.color,
+    // Fragment shader wants premultiplied alpha.
+    .color = float4(in.color.rgb, 1) * in.color.a,
   };
 
   return out;
@@ -230,13 +231,9 @@ struct FragmentShaderImpl {
     if (useColorM) {
       c.rgb /= c.a + (1.0 - sign(c.a));
       c = (color_matrix_body * c) + color_matrix_translation;
-      c *= v.color;
       c.rgb *= c.a;
-    } else {
-      float4 s = v.color;
-      c *= float4(s.r, s.g, s.b, 1.0) * s.a;
     }
-    c = min(c, c.a);
+    c *= v.color;
     return c;
   }
 };
--- a/internal/graphicsdriver/metal/graphics_darwin.go
+++ b/internal/graphicsdriver/metal/graphics_darwin.go
@@ -232,8 +232,11 @@ struct FragmentShaderImpl {
       c.rgb /= c.a + (1.0 - sign(c.a));
       c = (color_matrix_body * c) + color_matrix_translation;
       c.rgb *= c.a;
+      c *= v.color;
+      c.rgb = min(c.rgb, c.a);
+    } else {
+      c *= v.color;
     }
-    c *= v.color;
     return c;
   }
 };
--- a/internal/graphicsdriver/opengl/defaultshader.go
+++ b/internal/graphicsdriver/opengl/defaultshader.go
@@ -117,7 +117,9 @@ varying vec4 varying_color_scale;
 
 void main(void) {
   varying_tex = A1;
-  varying_color_scale = A2;
+
+  // Fragment shader wants premultiplied alpha.
+  varying_color_scale = vec4(A2.rgb, 1) * A2.a;
 
   mat4 projection_matrix = mat4(
     vec4(2.0 / viewport_size.x, 0, 0, 0),
@@ -267,15 +269,10 @@ void main(void) {
   color.rgb /= color.a + (1.0 - sign(color.a));
   // Apply the color matrix or scale.
   color = (color_matrix_body * color) + color_matrix_translation;
-  color *= varying_color_scale;
   // Premultiply alpha
   color.rgb *= color.a;
-# else
-  vec4 s = varying_color_scale;
-  color *= vec4(s.r, s.g, s.b, 1.0) * s.a;
 # endif  // defined(USE_COLOR_MATRIX)
-
-  color = min(color, color.a);
+  color *= varying_color_scale;
 
   gl_FragColor = color;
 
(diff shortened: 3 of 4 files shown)


internal/graphicsdriver/opengl/defaultshader.go
```diff
@@ -267,15 +269,10 @@ void main(void) {
   color.rgb /= color.a + (1.0 - sign(color.a));
   // Apply the color matrix or scale.
   color = (color_matrix_body * color) + color_matrix_translation;
-  color *= varying_color_scale;
```
hajimehoshi, line 270: Does this really work when a color matrix is used?

hajimehoshi (commented):
Thanks!

internal/graphicsdriver/opengl/defaultshader.go
hajimehoshi: NVM, this should work...

divVerent: Yes, I think this is correct with a color matrix too.

internal/graphicsdriver/metal/graphics_darwin.go
```diff
@@ -230,13 +231,9 @@ struct FragmentShaderImpl {
     if (useColorM) {
       c.rgb /= c.a + (1.0 - sign(c.a));
       c = (color_matrix_body * c) + color_matrix_translation;
-      c *= v.color;
       c.rgb *= c.a;
```
hajimehoshi, line 234: In theory it is still possible that RGB values exceed the alpha value after the calculation. So I think calling `min` is still required for the color-matrix case. Or this will break the compatibility.

hajimehoshi: On second thought, even without a color matrix, this might change the result when RGB values are extremely bigger than alpha. `min` might be required for such case.

Hmm, let me think more. I now started to think that we can break such a subtle backward compatibility...

divVerent: The min should only be needed for roundoff errors, malformed color inputs, or for malformed color matrices.

In theory none of these should ever happen, as linear combinations prefer wellformed premultiplied alpha.

If this cannot be done, though, I would still want to request this as an optional feature for games to opt into (i.e. "I promise my color matrices are all good, now skip this check".

divVerent: Another option of course is to only apply the min in the color matrix case, which has a lot of other complexity the min() can "hide behind" anyway.

divVerent: Here's the primary case where this would change things:

Use a Scale(2.0, 2.0, 2.0, 1.0) color matrix.

This can, on a color like "50% white" (0.5, 0.5, 0.5, 0.5), get you a "50% brighter-than-white" color (1.0, 1.0, 1.0, 0.5).

Now this would still kinda render "correctly" - it'd behave exactly like mixing with an overbright color - but it'd indeed look _different_ and could change games. This can be happening in real world code.

In DarkPlaces engine, we actually did this intentionally - used a premultiplied-alpha blending mode (GL_ONE, GL_ONE_MINUS_SRC_ALPHA), and in the texture we happily mixed normal premultiplied alpha data, and data where color values are set but alpha is all zero to get additive blending.

So, I suppose for a minor release, we probably should prefer keeping the min() there at least in the color matrix case. Do we still need it in the color multiply case though? In that case, I'll refactor this a little so my change only affects the non-color matrix case, and add a TODO to later reconsider or document the behavior of out-of-range colors with a color matrix.

divVerent: Hm... wait. The user doesn't even specify whether colormatrix is to be used or not.

So if I bring back the clamping only to the colormatrix case, one could now easily write code that looks different with and without this clamping (e.g. can be reproduced by using a mostly-scale color matrix vs a scale-only color matrix).

So personally I'd vote for removing this clamping entirely, but if this cannot be done safely, can we get an option to do so?

divVerent: Actually, I found an option: if any color multipliers are > 1, do not invoke the ScaleOnly optimization. I'll update this patch to do that.

divVerent:
Check failures were fun:

>   Could not connect to azure.archive.ubuntu.com:80 (52.250.76.244), connection timed out [IP: 52.250.76.244 80]

divVerent:
Benchmark results of 01-start-to-hub.dem on Dell E6500 (the underpowered GPU with the large screen, shader exec time bound):

- before 352fd76312fab3f11e369cb8ef26af333814c433: 171.7 seconds
- before 352fd76312fab3f11e369cb8ef26af333814c433 + this change: 171.6 seconds (yes, basically no change)
- main at 352fd76312fab3f11e369cb8ef26af333814c433: 119.3 seconds
- main at 352fd76312fab3f11e369cb8ef26af333814c433 + this change:  92.9 seconds (22% faster!)

Why does this change only really take effect with the nearest shader? Because it's really about fillrate - my game runs at 640x360 and has about 2x overdraw at the settings used here, so it renders 461k pixels - however this gets scaled up to 1920x1080 -> 2M pixels. This final render pass really seems to be what matters.

internal/atlas/image.go
```diff
@@ -425,11 +425,16 @@ func (i *Image) drawTriangles(srcs [graphics.ShaderImageNum]*Image, vertices []f
...
-		cr = colorm.At(0, 0)
-		cg = colorm.At(1, 1)
-		cb = colorm.At(2, 2)
-		ca = colorm.At(3, 3)
-		colorm = affine.ColorMIdentity{}
+		r := colorm.At(0, 0)
+		g := colorm.At(1, 1)
+		b := colorm.At(2, 2)
+		a := colorm.At(3, 3)
+		if r <= a && g <= a && b <= a {
```
hajimehoshi, line 432: and all the values must be >= 0, and a must be <= 1?

hajimehoshi (commented):
I feel like your suggestion makes sense, thanks!

internal/atlas/image.go
```diff
@@ -425,11 +425,25 @@ func (i *Image) drawTriangles(srcs [graphics.ShaderImageNum]*Image, vertices []f
...
+		b := colorm.At(2, 2)
+		a := colorm.At(3, 3)
+		if r <= 1 && g <= 1 && b <= 1 && a >= 0 {
+			// Color matrices work on non-premultiplied colors.
+			// This color matrix can only make colors darker or equal,
+			// and thus can never invoke color clamping.
+			// Thus the simpler vertex color scale based shader can be used.
+			//
+			// Color values being negative with positive alpha is harmless -
+			// will be clamped away when converting to integer texture format anyway.
```
hajimehoshi, line 439: > will be clamped away when converting to integer texture format anyway.

Is that true?

divVerent:
I just found the condition under which a < 0 can break things, and added a comment as well as adapted the check.

The part I missed was that these colors are unpremultiplied, so my arguments failed epically - a sign flips when premultiplying which can wreak havoc.

The proper condition is r >= 0 && g >= 0 && b >= 0 && r <= 1 && g <= 1 && b <= 1 && a >= 0, and the only possible condition that allows even more requires analyzing the vertex colors.

OTOH, if this condition holds, then multiplying this with the vertex colors does not make the vertex colors any more pathological than they already might be, essentially meaning that the clamping can never kick in due to the vertex colors themselves being unpremultiplied too.

Negative matrix colors can combine with negative vertex colors to something positive, which can cause clamping. It might be possible that I actually just need r >= -1, etc. - did not analyze close enough, probably not worthwile.

internal/atlas/image.go
```diff
@@ -429,9 +429,18 @@ func (i *Image) drawTriangles(srcs [graphics.ShaderImageNum]*Image, vertices []f
 		g := colorm.At(1, 1)
 		b := colorm.At(2, 2)
 		a := colorm.At(3, 3)
-		if r <= a && g <= a && b <= a {
-			// Scale-only color matrix, and no color clamping needed either.
-			// (The identity color matrix case in the shader doesn't do that.)
+		if r <= 1 && g <= 1 && b <= 1 && a >= 0 {
```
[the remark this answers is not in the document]
hajimehoshi: So if the negative number is clamped automatically,

```
r <= a && g <= a && b <= a && 0 <= a && a <= 1
```

should be enough?

hajimehoshi:
I'm confused. Why don't we need the condition like `r <= a`?

divVerent:
Because colors in this layer (atlas.Image) are all unpremultiplied, and they previously got premultiplied in the fragment shader (now will in the vertex shader - this is what I primarily am moving here, as well as making the min() optional).

hajimehoshi:
> So a >= 0 and r <= 1 here implies r <= a in the fragment shader.

What about the scale (r, g, b, a) = (0.5, 0.5, 0.5, 0) for the color (1, 1, 1, 1)? Will the result be (0.5, 0.5, 0.5, 0)?

(For `min`, I want to be flexible about the compatibility, but I'd like to know what is kept and what is not exactly.)

divVerent:
If the scale is (r, g, b, a) = (0.5, 0.5, 0.5, 0), this

- now: becomes (0, 0, 0, 0) in the vertex shader, and invariably makes the fully-transparent in the fragment shader
- previously: stayed (0.5, 0.5, 0.5, 0) in the vertex shader, got turned into (0, 0, 0, 0) by the fragment shader, and invariably makes the output fully-transparent

This is what is to be expected about the unpremultiplied color (0.5, 0.5, 0.5, 0), which is "fully transparent grey". The fact that it is grey is entirely invisible, and should be irrelevant, when alpha is zero - and premultiplying like Ebiten does ensures precisely that.

divVerent:
Note that there is one _real_ difference caused by changing the code to do color interpolation in domain of premultiplied rather than unpremultiplied colors - one can construct examples of vertex colorings that look rather different in the interpolated area.

However, in almost all cases the previous behavior would actually be considered a bug, as it makes the "color" of alpha-zero vertices visible.

As an example, change the "shapes" example as follows:

```
diff --git a/examples/shapes/main.go b/examples/shapes/main.go
index 57b0da4c..2854d07b 100644
--- a/examples/shapes/main.go
+++ b/examples/shapes/main.go
@@ -112,16 +112,21 @@ func rect(x, y, w, h float32, clr color.RGBA) ([]ebiten.Vertex, []uint16) {
        x1 := x + w
        y1 := y + h
 
+       r = 0
+       g = 0.25
+       b = 0
+       a = 1
+
        return []ebiten.Vertex{
                {
                        DstX:   x0,
                        DstY:   y0,
                        SrcX:   1,
                        SrcY:   1,
-                       ColorR: r,
-                       ColorG: g,
-                       ColorB: b,
-                       ColorA: a,
+                       ColorR: 1,
+                       ColorG: 0,
+                       ColorB: 0,
+                       ColorA: 0,
                },
                {
                        DstX:   x1,
@@ -157,8 +162,6 @@ func rect(x, y, w, h float32, clr color.RGBA) ([]ebiten.Vertex, []uint16) {
 }
 
-       g.count++
-       g.count %= 240
        return nil
 }
 
```

Note how I made one vertex of the quads "fully transparent bright red", and all other vertices "fully opaque dark green".

Before: http://www.rm.cloudns.org/img/uploaded/e4fcea908c3a22d0b32864343fbe42a0.png

After: http://www.rm.cloudns.org/img/uploaded/8c5d93a08ad1a36d1ad76d0f32668542.png

A human, asked to interpolate between these (a transparent piece of glass and a solid dark green piece of paper) would sure agree with then new rendering. That red tint that only occurs between vertices, but never at vertices, is the wellknown artifact that comes from interpolating unpremultiplied colors, and I assume it is why Ebiten usually thinks in premultiplied colors internally.

divVerent:
Should also mention that the _common_ cases of vertex interpolation are unchanged by this:

- Same color, different alpha - both ways of doing the math result in the same color tone at the same alpha slope.
- Different color, sample alpha - both ways of doing the math result in the same alpha level with the same color slope.

hajimehoshi:
> now: becomes (0, 0, 0, 0) in the vertex shader, and invariably makes the fully-transparent in the fragment shader

OK acknowledged.

hajimehoshi:
> A human, asked to interpolate between these (a transparent piece of glass and a solid dark green piece of paper) would sure agree with then new rendering. That red tint that only occurs between vertices, but never at vertices, is the wellknown artifact that comes from interpolating unpremultiplied colors, and I assume it is why Ebiten usually thinks in premultiplied colors internally.

I agree. The new version is much better and makes sense. My understanding is that this is a very similar issue of wrong interpolation due to straight-alpha (non-premultiplied-alpha) colors.

(To be exact, interpolation might have to consider gamma correction, but this is a different issue)

divVerent:
Yes, it is precisely the same interpolation issue you also get with images in straight-alpha notation when scaling them component-wise without regard for the alpha channel. This can be fixed in a scaling algorithm, of course, however the ones commonly used on GPUs do not.

The only difference is that we had the same issue here in vertex domain rather than texel domain. But indeed - in general, colors should be premultiplied before interpolation.

This underlying issue also affects Kage shaders which I am not fixing here as the chance is higher to break underlying code - this requires further analysis. I do wonder if we are lucky enough to find that nobody uses Fragment()'s "color vec4" argument in Kage with alpha != 1 anyway and thus we can just change it to be premultiplied. Maybe worth tracking this related item as an issue?

hajimehoshi:
> This underlying issue also affects Kage shaders which I am not fixing here as the chance is higher to break underlying code - this requires further analysis. I do wonder if we are lucky enough to find that nobody uses Fragment()'s "color vec4" argument in Kage with alpha != 1 anyway and thus we can just change it to be premultiplied. Maybe worth tracking this related item as an issue?

Kage users can determine how to use the color values. Surprisingly, Tetra3D by @SolarRune uses these components for different purposes, so modifying this value would break compatibility unfortunately.

So this reminds me that we have to skip the color component modifying when a custom shader is used... Does your change do so?

divVerent:
> So this reminds me that we have to skip the color component modifying when a custom shader is used... Does your change do so?

Yes, "accidentally" - no *Shader calls support a ColorM (unsurprisingly, given Kage also gives no access to it), so the ColorM is always identity when we get here.

hajimehoshi:
> Yes, "accidentally" - no *Shader calls support a ColorM (unsurprisingly, given Kage also gives no access to it), so the ColorM is always identity when we get here. In which case this code does nothing.

I see. I'd be happy if you could leave a comment about this (the color components must not be modified when a custom shader is used, and luckily when ColorM is used, a custom shader is not used)

hajimehoshi:
> As for Kage's use of colors, maybe it's worth stating explicitly in documentation that color values ColorR, ColorG, ColorB, ColorA are freely usable and get interpolated independently when using DrawTrianglesShader (as opposed to DrawTriangles, which treats them as a straight-alpha color)?

Yes, we need such comments at the APIs.

hajimehoshi:
Would it be fine to ask you to add API comments?

And, it is optional but a test for the interpolation behavior would be nice. If this is difficult to you, I can do this later.

divVerent:
Added API comments about Vertex colors. Especially important as this means the interpretations differ.

divVerent:
As for a regression test, sorry, I do not have experience with that in Ebiten - especially as nothing is really guaranteed there (all depends on how the GPU does it internally). Can do so later, but I'd prefer to not have this be blocked.

hajimehoshi:
> As for a regression test, sorry, I do not have experience with that in Ebiten - especially as nothing is really guaranteed there (all depends on how the GPU does it internally). Can do so later, but I'd prefer to not have this be blocked.

Acknowledged!

hajimehoshi (approved):
LGTM, thank you very much for the hard work!

divVerent:
Right now seeing if I can make a somewhat working test for this despite interpolation not clearly defined.

hajimehoshi:
I'll go to bed soon so I'm happy to take a look at your progress tomorrow if there is :-)

divVerent:
Yeah, sending PR. I decided to make it an approximate test, allowing for quite some rounding errors. Should be good enough.

PR #1997, opened 2022-02-22
Provide ebiten.SetScreenFilterEnabled, a way to disable the default screen filter

divVerent:
Using FilterNearest rather than filterScreen or FilterLinear speeds up a
Dell E6500 from 23fps->36fps and 27fps->48fps.

Helps with https://github.com/hajimehoshi/ebiten/issues/1772.

### DIFF
--- a/run.go
+++ b/run.go
@@ -18,7 +18,6 @@ import (
 	"sync/atomic"
 
 	"github.com/hajimehoshi/ebiten/v2/internal/clock"
-	"github.com/hajimehoshi/ebiten/v2/internal/graphicsdriver"
 	"github.com/hajimehoshi/ebiten/v2/internal/ui"
 )
 
@@ -95,15 +94,13 @@ func IsScreenClearedEveryFrame() bool {
 	return ui.IsScreenClearedEveryFrame()
 }
 
-// SetScreenFilter requests a specific Filter to be used for upscaling/downscaling.
+// SetScreenFilterEnabled enables/disables the use of the "screen" filter Ebiten uses.
 //
-// Pass nil to revert to Ebiten's default filter selection.
-func SetScreenFilter(filter *Filter) {
-	if filter == nil {
-		ui.SetScreenFilter(-1)
-	} else {
-		ui.SetScreenFilter(graphicsdriver.Filter(*filter))
-	}
+// The "screen" filter is a box filter from game to display resolution.
+//
+// If disabled, nearest-neighbor filtering will be used for scaling instead.
+func SetScreenFilterEnabled(enabled bool) {
+	ui.SetScreenFilterEnabled(enabled)
 }
 
 type imageDumperGame struct {
--- a/run.go
+++ b/run.go
@@ -18,6 +18,7 @@ import (
 	"sync/atomic"
 
 	"github.com/hajimehoshi/ebiten/v2/internal/clock"
+	"github.com/hajimehoshi/ebiten/v2/internal/graphicsdriver"
 	"github.com/hajimehoshi/ebiten/v2/internal/ui"
 )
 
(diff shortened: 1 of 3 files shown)


hajimehoshi:
I'll take a look tomorrow.

divVerent:
Note that I am not quite convinced about this API - having to pass a pointer to emulate an "optional" is weird. Alternate suggestions very welcome.

hajimehoshi:
`SetScreenFilterEnabled` taking a boolean value would be better. If the value is false, `FilterNearest` is adopted. Wdyt?

divVerent:
Can do; I personally would prefer also being able to set FilterLinear, but I don't strictly require that.

divVerent:
So... if you confirm that you still prefer the SetScreenFilterEnabled(false) approach (which turns off Ebiten's default filtering and switches to Nearest), I'll change this PR to that.

divVerent:
Benchmark results of 01-start-to-hub.dem on Dell E6500 (the underpowered GPU with the large screen, shader exec time bound):

* before https://github.com/hajimehoshi/ebiten/commit/352fd76312fab3f11e369cb8ef26af333814c433 and this change: 171.7 seconds
* main at https://github.com/hajimehoshi/ebiten/commit/352fd76312fab3f11e369cb8ef26af333814c433 with or without this change: 119.3 seconds (31% faster)
* before https://github.com/hajimehoshi/ebiten/commit/352fd76312fab3f11e369cb8ef26af333814c433 + https://github.com/hajimehoshi/ebiten/pull/1996: 149.5 seconds
* main at https://github.com/hajimehoshi/ebiten/commit/352fd76312fab3f11e369cb8ef26af333814c433 with or without this change + https://github.com/hajimehoshi/ebiten/pull/1996: 92.9 seconds (38% faster)

Wait, why do I say with or without this change? Why do I still want this change in?

On the E6500, this change is actually a NOP in fullscreen mode in my game because 640x360 fits at integer factor 3 into 1920x1200. If I were to use windowed mode, getting a scale-up factor of, say, 2.9x, I'd still get the same massive improvement from this change over main at the expense of lower video quality (which is why I am restricting use of this to lower quality settings in my game).

So I'd still like this in to have playable fullscreen performance on lowend GPUs for my game even on screens whose width is not a multiple of 640.

hajimehoshi:
The approach with boolean is better. The best thing is to use the framebuffer image directly, but as I've said, the framebuffer image is very special (e.g. `ReplacePixels` doesn't work) and the user would have to care about Y direction (e.g. Y direction depends on the graphics library. See `gameForUI`). So, using the nearest filter would be a good compromise so far.

run.go
```diff
@@ -94,6 +94,15 @@ func IsScreenClearedEveryFrame() bool {
 	return ui.IsScreenClearedEveryFrame()
 }
 
+// SetScreenFilterEnabled enables/disables the use of the "screen" filter Ebiten uses.
+//
+// The "screen" filter is a box filter from game to display resolution.
+//
+// If disabled, nearest-neighbor filtering will be used for scaling instead.
```
hajimehoshi, line 101: Add these:

* The default state is true.
* SetScreenFilterEnabled is concurrent-safe.

divVerent: Also added that it takes effect at next Draw call - so calling it during Draw is one frame delayed.

divVerent:
This should be ready now and is just the enabling/disabling, as requested.

run.go
```diff
@@ -94,6 +94,19 @@ func IsScreenClearedEveryFrame() bool {
...
+//
+// The default state is true.
+//
+// SetScreenFilterEnabled is concurrent-safe, but takes effect only at the next Draw call.
+func SetScreenFilterEnabled(enabled bool) {
```
hajimehoshi, line 106: Could you add `IsScreenFilterEnabled` for consistency?

divVerent: Done.

PR #2002, opened 2022-02-25
WIP: Add basic midi input support

adrianduke:
As title suggests a WIP. Adding midi input support only for now, output shouldn't be tricky for the inclined.

Looking for a first pass to ensure the direction is good? If so, I'll cleanup with appropriate file license headers, examples and documentation.

Theres a lot more midi keys that could be added, this currently just supports pressing a midi note (e.g. C4, G#1) for my personal project. It's also just dealing with midi keys in an `on`/`off` fashion, I'll likely add velocity as soon as I need it myself.

I've tried to ensure the `gomidi` dependency is separated off into a `midi` package so as not to muddy `internal`. Additionally ensured we aren't checking for midi key presses unless the user explicitly calls `ReadMidi()`.

An example usage to help understand the user API:

```
type Key uint8

const (
	KeyDominantTrigger Key = iota
	KeyEnter
	KeyEscape
)

type KeyInputter interface {
	IsKeyJustPressed(key Key) bool
	IsKeyJustReleased(key Key) bool
}

var midiKeyMapping = map[Key]ebiten.MidiKey{
	KeyDominantTrigger: ebiten.NoteC4,
}

type MidiInput struct{}

func (m MidiInput) IsKeyJustPressed(k Key) bool {
	if midiKey, found := midiKeyMapping[k]; !found {
		return false
	} else {
		return inpututil.IsMidiKeyJustPressed(midiKey)
	}
}

func (m MidiInput) IsKeyJustReleased(k Key) bool {
	if midiKey, found := midiKeyMapping[k]; !found {
		return false
	} else {
		return inpututil.IsMidiKeyJustReleased(midiKey)
	}
}
```

```
func init() {
	midiDriver, err := driver.New()
	if err != nil {
		panic(err)
	}
	defer midiDriver.Close()

	ctx := midi.NewMidiContext(g.midiDriver)
	midiIns, err := ctx.InputPorts()
	if err != nil {
		panic(err)
	}

	if err := ebiten.ReadMidi(midiIns[0]); err != nil {
		panic(err)
	}
}
```


hajimehoshi:
Before introducing a big thing, let's create an issue and discuss this there.

PR #2004, opened 2022-02-28
Expose the debug font character size

joewreschnig:
This can be helpful to lay out more complex debugging text.

### DIFF
--- a/ebitenutil/debugprint.go
+++ b/ebitenutil/debugprint.go
@@ -26,6 +26,14 @@ var (
 	debugPrintTextSubImages = map[rune]*ebiten.Image{}
 )
 
+const (
+	// DebugCharWidth is the fixed width of a character drawn by DebugPrint.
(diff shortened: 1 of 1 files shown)


hajimehoshi:
While I understand this is useful, I'm not sure finer controls for debugging message is required...

joewreschnig:
To be more concrete, my two use cases are:
- Drawing some perf meters aligned to the right side of the screen. which requires knowing the width
- Holding down a button to overlay the last few lines of logger output, which requires knowing the line height.

But I understand if these don't seen generally useful / worth the API surface.

joewreschnig:
Thanks, I wasn't aware `\n` worked.

The advantage of `DebugPrint` is that I can use it before/while loading my other assets but maybe I'm still coming at this from the wrong direction. I'll go back and see if I can find another way to achieve this without relying on the debug font.

Issue #2008, opened 2022-03-07
examples/shaders: CRT effect


hajimehoshi:
See https://github.com/libretro/ludo/blob/master/video/crt_frag_shader.go

PR #2012, opened 2022-03-09
Added CRT shader example

Zyko0:
This is the translation of the shader linked in #2008.
I tried to keep it as close to the source first to make sure it's working as expected, but we can iterate from here.
I picked the gopher with bricks background as source image since it contains more colors:

image


Zyko0:
Note that I didn't make a distinction between `gl_FragCoord` and `vTexCoord` variables here https://github.com/libretro/ludo/blob/master/video/crt_frag_shader.go#L42-L46

I used `pos` for both, but I'm not sure it's correct

hajimehoshi (approved):
LGTM, thanks!

divVerent:
Nice!

In my game I actually built a CRT shader too, that has a different feature set from this one:

https://github.com/divVerent/aaaaxy/blob/main/assets/shaders/linear2xcrt.kage
https://raw.githubusercontent.com/divVerent/aaaaxy/gh-pages/screenshots/shot4.png

Differences are:

- Mine has a CRT distortion effect
- Mine can be fully replicated using FFmpeg filters (meaning that for video capture, I feed the pre-shader picture into FFmpeg and do the rest there)
- Mine is based on top of a previous shader that simulates typical VGA-ish scan line doubling (https://github.com/divVerent/aaaaxy/blob/main/assets/shaders/linear2x.kage) and meant as an incremental switch ("now with CRT effect").
- This one has brightness dependent scanline strength (i.e. higher brightness has larger radius of cathode ray).
- This one has an added horizontal blur (not seeing much of that in the image though), which replicates typical VGA signal distortion at high resolutions

Generally, good work here! Can always have more of those, as they fit different kinds of games!

Issue #2027, opened 2022-03-25
Crash on Alpine Linux


divVerent:
Repro:

- Install Alpine Linux, X11 and Go
- Clone Ebiten
- Run: `go run examples/rotate/main.go`
- All works fine, until you press a key (any key will do). Then:

```
panic: runtime error: index out of range [65317] with length 512

goroutine 1 [running, locked to thread]:
github.com/hajimehoshi/ebiten/v2/internal/gamepad.(*nativeGamepad).update(0xc0000b4028, 0x41bf90)
        /home/rpolzer/src/ebiten/internal/gamepad/gamepad_linux.go:331 +0x21f
github.com/hajimehoshi/ebiten/v2/internal/gamepad.(*Gamepad).update(0x8afe48, 0x0)
        /home/rpolzer/src/ebiten/internal/gamepad/gamepad.go:166 +0x9b
github.com/hajimehoshi/ebiten/v2/internal/gamepad.(*gamepads).update(0x8afe20)
        /home/rpolzer/src/ebiten/internal/gamepad/gamepad.go:94 +0x152
github.com/hajimehoshi/ebiten/v2/internal/gamepad.Update(...)
        /home/rpolzer/src/ebiten/internal/gamepad/gamepad.go:55
github.com/hajimehoshi/ebiten/v2/internal/ui.(*Input).update(0x8b06b0, 0xc0004e2010, 0x0)
        /home/rpolzer/src/ebiten/internal/ui/input_glfw.go:206 +0x3cc
github.com/hajimehoshi/ebiten/v2/internal/ui.(*userInterfaceImpl).update(0x8b0580)
        /home/rpolzer/src/ebiten/internal/ui/ui_glfw.go:981 +0x170
github.com/hajimehoshi/ebiten/v2/internal/ui.(*userInterfaceImpl).loop.func1()
        /home/rpolzer/src/ebiten/internal/ui/ui_glfw.go:1023 +0x4a
github.com/hajimehoshi/ebiten/v2/internal/thread.(*OSThread).Loop.func1(0xc000445e10, 0xc000445e00)
        /home/rpolzer/src/ebiten/internal/thread/thread.go:54 +0x53
github.com/hajimehoshi/ebiten/v2/internal/thread.(*OSThread).Loop(0xc0004e4030)
        /home/rpolzer/src/ebiten/internal/thread/thread.go:55 +0x2e
[...]

PR #2030, opened 2022-03-25
Fix uint16 underflow handling in gamepads.

divVerent:
Fixes crash where Ebiten considers a keyboard a gamepad with way too many buttons.

For some reason, on one of my laptops, Ebiten tries to keyboard and touchpad devices as "gamepads" - but then when parsing its button data, it crashes. Apparently here, `EV_KEY` codes are returned with a code smaller than `_BTN_MISC` whenever pressing a key on the laptop.

https://github.com/hajimehoshi/ebiten/commit/b98568901086d92cf8a2565340a909cc04496822 attempted to fix this, but the subtraction was done as signed int data type, meaning it didn't actually detect underflow. My fix is doing the subtraction as uint16 just like in the following line where the same expression is used to access keyMap - an alternate fix would probably be adding an `e.code >= _BTN_MISC` check instead.

Helps with issue #2027.

### DIFF
--- a/internal/gamepad/gamepad_linux.go
+++ b/internal/gamepad/gamepad_linux.go
@@ -328,7 +328,7 @@ func (g *nativeGamepad) update(gamepad *gamepads) error {
 
 		switch e.typ {
 		case unix.EV_KEY:
-			if int(e.code)-_BTN_MISC < len(g.keyMap) {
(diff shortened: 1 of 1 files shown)


divVerent:
Checks failed due to random timeout.

hajimehoshi (approved):
LGTM, thanks!

PR #2031, opened 2022-03-25
inpututil: allow an arbitrary number of gamepad buttons.

divVerent:
Fixes crash in inpututil when a gamepad with more than "allowed" buttons is
connected.

For some reason, on one of my laptops, Ebiten tries to keyboard and touchpad devices as "gamepads". The keyboard in particular shows up with `GamepadButtonNum(id) > GamepadButtonMax+1`.

https://github.com/hajimehoshi/ebiten/commit/5bb22d2bcf45fe026b99820c90b8bcc9e02fc19a attempted to fix this, but failed as other inpututil functions (e.g. `IsGamepadButtonJustPressed`) would still crash when a too-high button is passed, which happens in `examples/gamepad/main.go` due to it looping from `0` to `GamepadButtonNum(id)-1`.

An alternate fix could be already hiding extraneous buttons in the gamepad driver itself so that `GamepadButtonNum` can never return values greater than `GamepadButtonMax+1`; I decided not go there as your previous patch indicated that supporting "extraneous" buttons might actually be intended.

Helps with issue #2027.

### DIFF
--- a/inpututil/inpututil.go
+++ b/inpututil/inpututil.go
@@ -138,7 +138,7 @@ func (i *inputState) update() {
 		i.gamepadIDs[id] = struct{}{}
 
 		n := ebiten.GamepadButtonNum(id)
-		if _, ok := i.gamepadButtonDurations[id]; !ok || n != len(i.gamepadButtonDurations) {
+		if n != len(i.gamepadButtonDurations[id]) {
 			i.gamepadButtonDurations[id] = make([]int, n)
 		}
 		for b := ebiten.GamepadButton(0); b < ebiten.GamepadButton(n); b++ {
--- a/inpututil/inpututil.go
+++ b/inpututil/inpututil.go
@@ -137,12 +137,9 @@ func (i *inputState) update() {
 	for _, id := range i.gamepadIDsBuf {
 		i.gamepadIDs[id] = struct{}{}
 
-		if _, ok := i.gamepadButtonDurations[id]; !ok {
-			i.gamepadButtonDurations[id] = make([]int, ebiten.GamepadButtonMax+1)
-		}
 		n := ebiten.GamepadButtonNum(id)
-		if n > int(ebiten.GamepadButtonMax)+1 {
-			n = int(ebiten.GamepadButtonMax) + 1
+		if _, ok := i.gamepadButtonDurations[id]; !ok || n != len(i.gamepadButtonDurations) {
+			i.gamepadButtonDurations[id] = make([]int, n)
 		}
 		for b := ebiten.GamepadButton(0); b < ebiten.GamepadButton(n); b++ {
 			if ebiten.IsGamepadButtonPressed(id, b) {


hajimehoshi (approved):
LGTM, thanks!

hajimehoshi (changes requested):
Sorry, but I came up with a better idea

hajimehoshi:
> https://github.com/hajimehoshi/ebiten/commit/5bb22d2bcf45fe026b99820c90b8bcc9e02fc19a attempted to fix this, but failed as other inpututil functions (e.g. IsGamepadButtonJustPressed) would still crash when a too-high button is passed, which happens in examples/gamepad/main.go due to it looping from 0 to GamepadButtonNum(id)-1.

That's true, but having a big slice for a big number of gamepad buttons might be inefficient.

Instead, can we change `GamepadButtonPressDuration` and similar things to return 0 when an out-of-range button is specified?

hajimehoshi:
> An alternate fix could be already hiding extraneous buttons in the gamepad driver itself so that GamepadButtonNum can never return values greater than GamepadButtonMax+1; I decided not go there as your previous patch indicated that supporting "extraneous" buttons might actually be intended.

Well, I didn't intend this and actually I was not so careful, sorry.

divVerent:
We can, but then inpututil would enforce the restriction while direct
ebiten gamepad functions would not. Is that inconsistency acceptable?

hajimehoshi:
> We can, but then inpututil would enforce the restriction while direct
> ebiten gamepad functions would not. Is that inconsistency acceptable?

Ah, that's true.

hajimehoshi (approved):
LGTM again

divVerent:
Thanks.

One issue that remains, the elephant in the room that I had chosen to
ignore at first as I thought we need to be resilient against gamepads with
way too many buttons anyway... after all, such monsters exist.

we still should find a way to skip keyboards and touchpads. Why does Ebiten
even try?

To exclude the keyboard, I guess we can just say, if it has no axes, skip
it. Although, that may break NES gamepads... so not quite good...

I need to check later and if possible, send a "final" change to skip
devices that are clearly not gamepads, joysticks etc., because as is this
will be annoying in games with an input settings dialog (e.g. just touching
the touchpad slightly counts as a button here, but the keyboard "gamepad
device" seems rather silent, I could not get it to emit anything that
Ebiten reports to a game).

hajimehoshi:
> we still should find a way to skip keyboards and touchpads. Why does Ebiten even try?

I'm not sure what you meant. In what case should we ignore keyboards and touchpads?

divVerent:
Ebiten should not report keyboards and touchpads as a gamepad, given it
already sends the same events to mouse and keyboard specific Ebiten APIs.
Otherwise one action would be seen as two things.

On one of my other devices, Ebiten sees volume buttons as a "gamepad" which
is weird too. Seems like there are lots of odd kinds of devices in
/dev/input/event* that we should skip.

hajimehoshi:
Very interesting. Could you make an another issue for this? Thanks,

PR #2058, opened 2022-04-06
Playground for editing and testing ebiten's examples

Yeicor:
First of all, thanks for building such an amazing tool.

I'll start with a little bit of background: I created the [Static Go Playground](https://github.com/Yeicor/static-go-playground) project, which is the Go Compiler running completely inside the web browser. It allows Go developers to easily publish their demos in such a way that they can be modified and tested by users without any Go installation, and from any device.

I used your project for my demo as it is a great fit and has nice visual examples (hope it's ok). This pull request just adds a quick workflow (of around 25 secs) that automates the only required operation for using the Static Go Playground: vendoring dependencies and hosting a zip file of all the sources.

After applying this, you can just share a link that will open the Static Go Playground and automatically download the sources, compile any of the examples and run it. The link that would do this (for my fork) is: https://yeicor.github.io/static-go-playground/?fs_dl_/src=https://yeicor.github.io/ebiten/sources-latest.zip&build=/src/examples/chipmunk. It always points to the latest sources, so it could also be useful for quickly testing pull requests. It takes a minute to do the download and initial compilation, but compiled files are cached and the next builds will be much faster.

In conclusion, just wanted to show you this related project and offer the possibility to integrate it, but if you think this falls outside the scope of this repository, feel free to close this request.

### DIFF
--- a/.github/workflows/playground.yml
+++ b/.github/workflows/playground.yml
@@ -1,0 +1,28 @@
+name: Steam
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    name: Prepare ebiten for testing it from the Static Go Playground
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.18
+      
+      - name: Download dependencies (vendor)
+        run: go mod vendor
+        
+      - name: Publish an artifact ready to use
+        uses: actions/upload-artifact@v3
+        with:
+          name: static-go-playground-sources
+          path: .
--- a/.github/workflows/playground.yml
+++ b/.github/workflows/playground.yml
@@ -21,10 +21,10 @@ jobs:
       - name: Download dependencies (vendor)
         run: go mod vendor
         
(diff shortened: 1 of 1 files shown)


hajimehoshi:
Hi,

Your "Static Go Playground" seems a very nice project! However, testing this is out of Ebiten's scope and I think the test proposed here seems a test for the Go Playground rather than for Ebiten.

So, let me close this. I appreciate you letting me know this awesome project and trying to add a test to make both project better.

hajimehoshi:
Just out of curiosity, is your project similar to my Asobiba? (https://github.com/hajimehoshi/asobiba,  https://storage.googleapis.com/soko-dot-hajimehoshi-dot-org/asobiba/index.html)

Yeicor:
Yes, it is very similar to your project. I didn't find your project on my initial search, thanks for the reference!

I can see that you opted for patching os/exec with your own implementation linked to JS code. I wish I had thought of that before writing my own go build replacement to generate the compilation commands (although it was a nice learning exercise 😅). I may implement this on the future for my project.

Other than that and just from a quick look, the main difference is just a nicer UI for my project that gives you more control over the file system and some automation on top of that.

hajimehoshi:
Nice!

> Can compile the standard library from source on the browser

My project also can do that :-)

PR #2070, opened 2022-04-13
Refactor example "stars"

JVMerkle:
Minor adjustments because of:
- Function abs(int) is unused.
- Color fields are assigned without explicit names
- Color variable shadows the color package name


hajimehoshi (approved):
LGTM, thanks!

PR #2072, opened 2022-04-15
Interpolation example

Zyko0:
Little example showing linear interpolation with differences between TPS and refresh rates, and a use of `time.Time`.
TBD:
* [ ] Add a `DebugPrint` statement with more details
* [ ] Replace stored `ebiten.GeoM`s with a custom struct so that GeoM are only manipulated in `Draw`
* [ ] Interpolate with next position if possible, for it to match the real sprite position (currently interpolating in the past, so there's a delay with the current logical position/angle of the sprite and what is displayed)

Feel free to tell me if I'm missing something, or if you have any inputs

### DIFF
--- a/examples/interpolation/main.go
+++ b/examples/interpolation/main.go
@@ -1,0 +1,162 @@
+// Copyright 2022 The Ebiten Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


hajimehoshi:
Any updates?

Zyko0:
Sorry, I have been motivated by a code contest last days
So you confirm it would be cool having this example with the others, after a few iterations/fixes?
Not only for the differences between refresh rates, but also to show it's possible to design a game logic running at 30TPS while keeping a cool 60FPS

hajimehoshi:
> So you confirm it would be cool having this example with the others, after a few iterations/fixes?

Well, I have a mixed feeling to have this example. Personally I don't use this method. However, if this is a common pattern as you have already explained to me, I'm OK to have this example.

Zyko0:
> Well, I have a mixed feeling to have this example. Personally I don't use this method

Okay, I think I prefer to close then, we can still revisit it later! 😄

Issue #2163, opened 2022-06-24
ebitenutil: add DrawCircle


tupini07:
We have `DrawLine` , we have `DrawRect` , but no `DrawCirlcle`. I think it would be really useful for quick prototyping (and general use).

Is there a specific reason why there isn't a `DrawCircle`?

hajimehoshi:
```
DrawCircle(dst *ebiten.Image, cx, cy, r float64, clr color.Color)
```

would be what we want. This should be easily implemented with the `vector` package.

PR #2175, opened 2022-07-02
Issue 2163 add circle

fpiwowarczyk:
PR for #2163

### DIFF
--- a/ebitenutil/shapes.go
+++ b/ebitenutil/shapes.go
@@ -69,7 +69,6 @@ func DrawCircle(dst *ebiten.Image, cx, cy, r float64, clr color.Color) {
 	rd, g, b, a := clr.RGBA()
 
 	path.Arc(float32(cx), float32(cy), float32(r), 0, 2*math.Pi, vector.Clockwise)
-	op := &ebiten.DrawTrianglesOptions{}
 
 	verticles, indices := path.AppendVerticesAndIndicesForFilling(nil, nil)
 	for i := range verticles {
@@ -80,5 +79,5 @@ func DrawCircle(dst *ebiten.Image, cx, cy, r float64, clr color.Color) {
 		verticles[i].ColorB = float32(b) / 0xffff
 		verticles[i].ColorA = float32(a) / 0xffff
 	}
-	dst.DrawTriangles(verticles, indices, emptySubImage, op)
+	dst.DrawTriangles(verticles, indices, emptySubImage, nil)
 }
--- a/ebitenutil/shapes.go
+++ b/ebitenutil/shapes.go
@@ -20,6 +20,7 @@ import (
 	"math"
 
 	"github.com/hajimehoshi/ebiten/v2"
+	"github.com/hajimehoshi/ebiten/v2/vector"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


ebitenutil/shapes.go
```diff
@@ -59,3 +60,25 @@ func DrawRect(dst *ebiten.Image, x, y, width, height float64, clr color.Color) {
...
+	op := &ebiten.DrawTrianglesOptions{
+		FillRule: ebiten.FillAll,
+	}
+
+	verticles, indices := path.AppendVerticesAndIndicesForFilling(nil, nil)
+	for i := range verticles {
+		verticles[i].ColorR = float32(rd) / 0xffff
+		verticles[i].ColorG = float32(g) / 0xffff
+		verticles[i].ColorB = float32(b) / 0xffff
+		verticles[i].ColorA = float32(a) / 0xffff
```
hajimehoshi, line 81: Shouldn't vertices' SrcX/SrcY be set?

ebitenutil/shapes.go
```diff
@@ -59,3 +60,25 @@ func DrawRect(dst *ebiten.Image, x, y, width, height float64, clr color.Color) {
 	// Linear filtering would make edges blurred.
 	dst.DrawImage(emptyImage.SubImage(image.Rect(1, 1, 2, 2)).(*ebiten.Image), op)
 }
+
+// DrawCircle draws a circle on given destination dst.
+//
+// DrawCircle is intended to be used mainly for debugging or prototyping puropose.
+func DrawCircle(dst *ebiten.Image, cx, cy, r float32, clr color.Color) {
```
hajimehoshi, line 67: Use float64 for consistency

ebitenutil/shapes.go
Zyko0: I don't think it has to
I think `vector.Path.AppendVerticesAndIndicesForFilling` does not do it by default it's still drawing
Here SrcX, SrcY will be fixed 0 in all interpolated points of the triangles, so it will map to 0,0 from the white filled image so the color will be white
I don't think it matters since we're drawing right after that (l.83) and controlling the input image

ebitenutil/shapes.go
fpiwowarczyk: Sure

ebitenutil/shapes.go
hajimehoshi: > Also it would probably not make sense if it was mapped to an asset image, if we imagine what kind of "triangles" compose the final shape

As the source image is the emtpySubImage whose bound is (1,1)-(2,2), we should set SrcX/SrcY to do so. I think setting all the SrcX/SrcY to (1, 1) would work at least.

Zyko0: > As the source image is the emtpySubImage whose bound is (1,1)-(2,2)

Oh, you're right I didn't realize the source image was a sub image, makes sense!

fpiwowarczyk: Yup, thanks for your explanation @hajimehoshi didn't know that :D, I will make fix. But I wonder what really difference it makes because it works either way ;)

hajimehoshi: Yeah, (0, 0) would work in 99% cases but this is unsafe unfortunately. I used a 1x1 sub image inside of a 3x3 white image to prevent bleeding edges. Bleeding edges are a common GPU issue that an outside pixel from the image region is picked. This is inevitable for any GPUs. By using a subimage whose bounds are (1,1)-(2,2), picking outside pixels should not be problematic.

hajimehoshi:
Thank you for the PR!

ebitenutil/shapes.go
fpiwowarczyk: Done ;)

ebitenutil/shapes.go
```diff
@@ -59,3 +60,27 @@ func DrawRect(dst *ebiten.Image, x, y, width, height float64, clr color.Color) {
...
+	verticles, indices := path.AppendVerticesAndIndicesForFilling(nil, nil)
+	for i := range verticles {
+		verticles[i].SrcX = 1
+		verticles[i].SrcY = 1
+		verticles[i].ColorR = float32(rd) / 0xffff
+		verticles[i].ColorG = float32(g) / 0xffff
+		verticles[i].ColorB = float32(b) / 0xffff
+		verticles[i].ColorA = float32(a) / 0xffff
+	}
+	dst.DrawTriangles(verticles, indices, emptySubImage, op)
```
hajimehoshi, line 85: `nil` instead of `op` is fine: `FillAll` is the default value.

hajimehoshi (commented):
Thanks! One request

ebitenutil/shapes.go
fpiwowarczyk: Sure, fixed

hajimehoshi (approved):
LGTM, thanks!

PR #2212, opened 2022-07-22
audio/mp3, audio/vorbis, audio/wav: add Resample

RusAD32:
updates (and probably closes) #2055

### DIFF
--- a/audio/audio.go
+++ b/audio/audio.go
@@ -35,6 +35,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"github.com/hajimehoshi/ebiten/v2/audio/internal/convert"
 	"io"
 	"runtime"
 	"sync"
@@ -434,4 +435,14 @@ func (h *hookImpl) OnResumeAudio(f func() error) {
 
 func (h *hookImpl) AppendHookOnBeforeUpdate(f func() error) {
 	hooks.AppendHookOnBeforeUpdate(f)
+}
+
+// Resample explicitly resamples a stream to fit a new sample rate
+//
+// if original sample rate matches the new one, nothing happens
+func Resample(source io.ReadSeeker, size int64, from, to int) io.ReadSeeker {
+	if from == to {
+		return source
+	}
+	return convert.NewResampling(source, size, from, to)
 }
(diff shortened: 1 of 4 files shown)


hajimehoshi:
Thanks, but before adding a public API, we should have a discussion.

For resampling, I want an API at the audio package rather than each package audio/mp3, audio/vorbis, and audio/wav.

hajimehoshi:
> should the function operate on a general io.Reader

Yes it is. We want this function:

```go
package audio

func Resample(source io.ReadSeeker, size int64, from, to int) io.ReadSeeker
```

The implementation is already in audio/internal/convert.

RusAD32:
Then it's basically a `convert.NewResampling` wrapper

audio/audio.go
```diff
@@ -435,3 +436,13 @@ func (h *hookImpl) OnResumeAudio(f func() error) {
 func (h *hookImpl) AppendHookOnBeforeUpdate(f func() error) {
 	hooks.AppendHookOnBeforeUpdate(f)
 }
+
+// Resample explicitly resamples a stream to fit a new sample rate
+//
+// if original sample rate matches the new one, nothing happens
```
hajimehoshi, line 442: `If`

hajimehoshi (commented):
Do we still need changes in audio/vorbis and audio/wav?

audio/audio.go
hajimehoshi: s/nothing happens/Reample returns the source as it is/

audio/audio.go
```diff
@@ -435,3 +436,13 @@ func (h *hookImpl) OnResumeAudio(f func() error) {
 func (h *hookImpl) AppendHookOnBeforeUpdate(f func() error) {
 	hooks.AppendHookOnBeforeUpdate(f)
 }
+
+// Resample explicitly resamples a stream to fit a new sample rate
```
hajimehoshi, line 440: Add `.` at the end of the sentence.

RusAD32:
> Do we still need changes in audio/vorbis and audio/wav?

No, reverted them

hajimehoshi (approved):
LGTM, thanks!

PR #2221, opened 2022-07-29
Do not require installing file2byteslice to update go generate'd files.

divVerent:
The tool is already referenced in `go.mod`, so it is trivial to just `go run` it - that way, `go generate` can use it without the user explicitly installing it first.

Tested: modifying `gobind.go` and running `go generate` now applies correctly to `gobind.src.go`.

Note that this means that `GOOS=windows go generate` no longer will work on Linux - if this is a problem, we could prepend the `go` command with `env GOOS= GOARCH=` every time. This same limitation also already exists for some of your `go run` commands in there, though, so I thought it's fine to keep it that way.

### DIFF
--- a/audio/vorbis/generate.go
+++ b/audio/vorbis/generate.go
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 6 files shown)


hajimehoshi (approved):
LGTM, thanks!

PR #2243, opened 2022-08-13
internal/graphicsdriver/metal: remove C for macOS

TotallyGamerJet:


internal/cocoasdk/api_coregraphics_darwin.go
```diff
@@ -0,0 +1,5 @@
+package cocoasdk
```
hajimehoshi, line 1: Add a licence comment.

internal/cocoasdk/nsinvocation_darwin.go
```diff
@@ -0,0 +1,46 @@
+package cocoasdk
```
hajimehoshi, line 1: Add a licence comment.

internal/cocoasdk/nsinvocation_darwin.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"github.com/ebitengine/purego/objc"
+	"unsafe"
```
hajimehoshi, line 5:
```
"unsafe"

"github.com/ebitengine/purego/objc"
```

internal/graphicsdriver/metal/ca/ca_darwin.go
```diff
@@ -21,21 +21,16 @@ package ca
...
 import (
 	"errors"
+	"fmt"
+	"runtime"
 	"unsafe"
 
+	"github.com/ebitengine/purego"
+	"github.com/ebitengine/purego/objc"
+	"github.com/hajimehoshi/ebiten/v2/internal/cocoasdk"
 	"github.com/hajimehoshi/ebiten/v2/internal/graphicsdriver/metal/mtl"
```
hajimehoshi, line 31: Separate the group for external dependencies and internal dependencies

```
	"github.com/ebitengine/purego"
	"github.com/ebitengine/purego/objc"

	"github.com/hajimehoshi/ebiten/v2/internal/cocoasdk"
	"github.com/hajimehoshi/ebiten/v2/internal/graphicsdriver/metal/mtl"
```

internal/cocoasdk/nsmethodsignature_darwin.go
```diff
@@ -0,0 +1,20 @@
+package cocoasdk
+
+import "github.com/ebitengine/purego/objc"
```
hajimehoshi, line 3: Use `()`

internal/cocoasdk/api_coregraphics_darwin.go
```diff
@@ -0,0 +1,5 @@
...
+
+type CGSize struct {
+	Width, Height float64
```
hajimehoshi, line 4: Isn't this float32?

internal/cocoasdk/nsmethodsignature_darwin.go
```diff
@@ -0,0 +1,20 @@
...
+var class_NSMethodSignature = objc.GetClass("NSMethodSignature\x00")
+
+var (
+	sel_instanceMethodSignatureForSelector = objc.RegisterName("instanceMethodSignatureForSelector:\x00")
+	sel_signatureWithObjCTypes             = objc.RegisterName("signatureWithObjCTypes:\x00")
+)
+
+type NSMethodSignature objc.ID
+
+func InstanceMethodSignatureForSelector(self objc.ID, _cmd objc.SEL) NSMethodSignature {
```
hajimehoshi, line 14: What about `cmd` instead of `_cmd`?

internal/cocoasdk/nsinvocation_darwin.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+var class_NSInvocation = objc.GetClass("NSInvocation\x00")
```
hajimehoshi, line 8: Actually we don't have to use `\x00` but we use `\x00` for performance, right?

If the performance really matters, it's OK, but I feel like this degrates the readability. I prefer not having `\x00`. What do you think?

internal/graphicsdriver/metal/ca/ca_darwin.go
```diff
@@ -106,94 +114,91 @@ func (ml MetalLayer) SetPixelFormat(pf mtl.PixelFormat) {
...
 // are synchronized with the display's refresh rate.
 //
 // Reference: https://developer.apple.com/documentation/quartzcore/cametallayer/2887087-displaysyncenabled.
 func (ml MetalLayer) SetDisplaySyncEnabled(enabled bool) {
-	switch enabled {
-	case true:
-		C.MetalLayer_SetDisplaySyncEnabled(ml.metalLayer, 1)
-	case false:
-		C.MetalLayer_SetDisplaySyncEnabled(ml.metalLayer, 0)
+	if runtime.GOOS == "ios" {
```
hajimehoshi, line 128: This would not work with Go 1.15 (`ios` is not one of GOOS yet at that time). Why did you change the logic here by the way?

internal/graphicsdriver/metal/ca/ca_darwin.go
```diff
@@ -106,94 +114,91 @@ func (ml MetalLayer) SetPixelFormat(pf mtl.PixelFormat) {
...
 	}
+	ml.metalLayer.Send(objc.RegisterName("setDisplaySyncEnabled:\x00"), enabled)
 }
 
 // SetDrawableSize sets the size, in pixels, of textures for rendering layer content.
 //
 // Reference: https://developer.apple.com/documentation/quartzcore/cametallayer/1478174-drawablesize.
 func (ml MetalLayer) SetDrawableSize(width, height int) {
-	C.MetalLayer_SetDrawableSize(ml.metalLayer, C.double(width), C.double(height))
+	//TODO: once objc supports calling functions with struct arguments replace this with just a ID.Send call
```
hajimehoshi, line 138: add a space after `//`

internal/graphicsdriver/metal/view_ios.go
```diff
@@ -17,6 +17,10 @@
 
 package metal
 
+// Suppress the warnings about availability guard with -Wno-unguarded-availability-new.
+// It is because old Xcode (8 or older?) does not accept @available syntax.
+
+// #cgo CFLAGS: -Wno-unguarded-availability-new
 // #cgo CFLAGS: -x objective-c
```
hajimehoshi, line 24: Unity these two `CFLAGS`.

hajimehoshi (commented):
Thank you!

internal/cocoasdk/api_coregraphics_darwin.go
hajimehoshi, line 1: I'm not sure, but what about just `cocoa`? Do you have a plan to use other similar packages? I'd like to know your plan.

internal/cocoasdk/api_coregraphics_darwin.go
TotallyGamerJet: CGSize is made up of CGFloat which is a double. [See Apple docs](https://developer.apple.com/documentation/corefoundation/cgfloat?language=objc)

internal/cocoasdk/nsmethodsignature_darwin.go
TotallyGamerJet: I did _cmd bc that is how it is done in Objc but if you prefer just cmd then I can do that

internal/cocoasdk/nsinvocation_darwin.go
TotallyGamerJet: I didn't notice any major differences. Ill remove them

internal/cocoasdk/nsmethodsignature_darwin.go
hajimehoshi: > I did _cmd bc that is how it is done in Objc

Oh really? Could you show me a link for the explanation?

internal/cocoasdk/api_coregraphics_darwin.go
hajimehoshi: Acknowledged, thanks!

hajimehoshi:
There seems an error in the testing bot:

```
# github.com/hajimehoshi/ebiten/v2/internal/graphicsdriver/metal
Undefined symbols for architecture arm64:
  "_objc_msgSend", referenced from:
      _addSublayer in _x002.o
      _setFrame in _x002.o
ld: symbol(s) not found for architecture arm64
```

internal/cocoasdk/api_coregraphics_darwin.go
TotallyGamerJet: Cocoa is really just a collection of most of the graphical frameworks like AppKit, Foundation etc. I think just cocoa is fine as the name. The goal is to port enough to make all of ebitengine including glfw into purego.

internal/graphicsdriver/metal/ca/ca_darwin.go
TotallyGamerJet: The logic should be the same. There was a check in the C code to only run it on IOS. I replaced all runtime.GOOS == "iOS" with just a check if the Object responds to a certain function.

internal/cocoasdk/nsmethodsignature_darwin.go
TotallyGamerJet: The reason is likely because they didn't want to take up the identifier `cmd` from use in a method. This [Apple Doc](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtHowMessagingWorks.html) says "The first two arguments passed to the procedure are the receiving object (self) and the method selector (_cmd). These arguments are hidden in method syntax but must be made explicit when the method is called as a function." Although it doesn't say _why_ it is this way.

hajimehoshi: OK I'm convinced.

internal/cocoasdk/api_coregraphics_darwin.go
hajimehoshi: I mean I wanted to know you would add other packages besides `cocoa` and I wanted to know the naming convention. Anway we can revisit this later. Let's go with `cocoa`.

internal/graphicsdriver/metal/ca/ca_darwin.go
hajimehoshi: The latest change (checking the existence of the method) seems fine.

hajimehoshi (commented):
Let me test your PR on my MacBook Pro and iOS. Thanks!

hajimehoshi:
I found https://github.com/hajimehoshi/ebiten/pull/2243#issuecomment-1213639418 still exists. Please take a look, thanks

TotallyGamerJet:
yeah it wasn't linking to QuartzCore since that got removed from metal and wasn't in the iOS specific code.

Should hopefully be fixed now.

hajimehoshi:
Other errors:

```
cannot use float64(width) (type float64) as type cocoa.CGFloat in field value
```

TotallyGamerJet:
Hmm I’m not sure why it’s complaining on iOS although I did just remember I was using a modified version of purego so I’ll create that PR tomorrow morning

hajimehoshi:
> Hmm I’m not sure why it’s complaining on iOS although I did just remember I was using a modified version of purego so I’ll create that PR tomorrow morning

So, we have to update purego in go.mod in this PR, right?

TotallyGamerJet:
Yes

hajimehoshi:
OK so are you updating go.mod in this PR? The test bots seem happy already.

TotallyGamerJet:
It should all be working now!

hajimehoshi:
Let me test with actual devices tomorrow. Thank you for your hard work!

internal/cocoa/nsmethodsignature_darwin.go
```diff
@@ -0,0 +1,36 @@
...
+	sel_signatureWithObjCTypes             = objc.RegisterName("signatureWithObjCTypes:")
+)
+
+type NSMethodSignature objc.ID
+
+func InstanceMethodSignatureForSelector(self objc.ID, _cmd objc.SEL) NSMethodSignature {
+	return NSMethodSignature(self.Send(sel_instanceMethodSignatureForSelector, _cmd))
+}
+
+func NSMethodSignature_SignatureWithObjCTypes(types string) NSMethodSignature {
```
hajimehoshi, line 34: I think both `InstanceMethodSignatureForSelector` and `SignatureWithObjCTypes` are static methods of NSMethodSignature. So, shouldn't they have consist names like `NSMethodSignature_InstanceMethodSignatureForSelector` and `NSMethodSignature_SignatureWithObjCTypes` (as it is)?

internal/graphicsdriver/metal/graphics_darwin.go
```diff
@@ -21,27 +21,27 @@ import (
...
-// }
-//
-// static void releaseAutoreleasePool(void* pool) {
-//   [(NSAutoreleasePool*)pool release];
-// }
-import "C"
+var (
+	class_NSAutoreleasePool = objc.GetClass("NSAutoreleasePool\x00")
+	sel_new                 = objc.RegisterName("new\x00")
+	sel_release             = objc.RegisterName("release\x00")
```
hajimehoshi, line 35: Remove `\x00`

hajimehoshi:
[go-inovation](https://github.com/hajimehoshi/go-inovation) crashed with this log:

```
2022-08-14 13:27:24.180272+0900 goinovation[407:9762] Metal API Validation Enabled
2022-08-14 13:27:24.363173+0900 goinovation[407:9762] -[AGXA13Device isHeadless]: unrecognized selector sent to instance 0x14a008800
2022-08-14 13:27:24.363780+0900 goinovation[407:9762] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[AGXA13Device isHeadless]: unrecognized selector sent to instance 0x14a008800'
*** First throw call stack:
(0x1e0cd8288 0x1f9a07744 0x1e0db5fc0 0x1e0c6cba0 0x1e0c6bf70 0x10468cee4 0x1045f188c)
libc++abi: terminating with uncaught exception of type NSException
dyld4 config: DYLD_LIBRARY_PATH=/usr/lib/system/introspection DYLD_INSERT_LIBRARIES=/Developer/usr/lib/libBacktraceRecording.dylib:/Developer/usr/lib/libMainThreadChecker.dylib:/Developer/Library/PrivateFrameworks/DTDDISupport.framework/libViewDebuggerSupport.dylib:/Developer/Library/PrivateFrameworks/GPUTools.framework/libglInterpose.dylib
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[AGXA13Device isHeadless]: unrecognized selector sent to instance 0x14a008800'
terminating with uncaught exception of type NSException
(
```

hajimehoshi:
On an iOS simulator, go-inovation crashed at 

```go
		colorspace, _, _ := purego.SyscallN(_CGColorSpaceCreateWithName, **(**uintptr)(unsafe.Pointer(&kCGColorSpaceDisplayP3))) // Dlsym returns pointer to symbol so dereference it
```

with EXC_BAD_ACCESS.

TotallyGamerJet:
Both of the crashes you experienced are from calling functions that in the .m files were marked as not to run on iOS Those are the ones that I had replaced with `if runtime.GOOS != "ios"`. I'm not sure of any other ways to not call them on iOS since they appear to be lying that they can be called. How do you want to proceed?

hajimehoshi:
If we just want to distinguish ios or not, what about making these files?

```go
//go:build ios
// +build ios

const isIos = true
```

```go
//go:build !ios
// +build !ios

const isIos = false
```

hajimehoshi:
Or we can move the method to these files.

TotallyGamerJet:
The build tag method I think is best

hajimehoshi:
Both (defining consts and defining functions) use build tags. You mean you prefer defining functions with build tags? That's fine :-)

TotallyGamerJet:
Uh no sorry. I meant I like the consts one. What I wrote was very unclear

hajimehoshi:
> Uh no sorry. I meant I like the consts one. What I wrote was very unclear

No problem, the const way is also fine!

internal/graphicsdriver/metal/ca/ca_darwin.go
```diff
@@ -49,40 +45,51 @@ type Layer interface {
...
+	kCGColorSpaceDisplayP3      = purego.Dlsym(coreGraphics, "kCGColorSpaceDisplayP3")
+)
+
 // MakeMetalLayer creates a new Core Animation Metal layer.
 //
 // Reference: https://developer.apple.com/documentation/quartzcore/cametallayer.
 func MakeMetalLayer() MetalLayer {
-	return MetalLayer{C.MakeMetalLayer()}
+	layer := objc.ID(objc.GetClass("CAMetalLayer")).Send(objc.RegisterName("new"))
+	if !cocoa.IsIOS {
```
hajimehoshi, line 63: I think this constant should be defined in ca as long as this is the only one usage, but what do you think?

internal/cocoa/isIOS.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 The Ebiten Authors
+// Copyright 2022 The Ebitengine Authors
```
hajimehoshi, line 1: Use a small cases for a file name.

hajimehoshi: I think `ios.go` is fine.

internal/cocoa/notIOS.go
```diff
@@ -0,0 +1,20 @@
+// Copyright 2022 The Ebitengine Authors
```
hajimehoshi, line 1: ditto: please use small cases. `notios.go` is fine. 

Unfortunately it might be hard to change only cases of a file name. You might have to remove this file once, commit the change, then add the file again.

internal/graphicsdriver/metal/ca/ca_darwin.go
TotallyGamerJet: It's also used in mtl package

hajimehoshi: NVM, I found it https://github.com/hajimehoshi/ebiten/pull/2243/files#diff-f436c80b858b5bc84325c6aa909731c966b62551cefefc37b6f6936ab48dccd5R584

hajimehoshi:
I'll test this with actual devices tomorrow, thanks!

hajimehoshi:
It seems like an iOS emulator and an actual iOS device worked with this PR. I'll double-check this PR as this is pretty big.

internal/cocoa/ios.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 The Ebiten Authors
+// Copyright 2022 The Ebitengine Authors
```
hajimehoshi, line 1: Sorry but I realized that `ios_darwin.go` is better so that non-Darwn environment cannot build this package.

internal/cocoa/notios.go
```diff
@@ -0,0 +1,20 @@
+// Copyright 2022 The Ebitengine Authors
```
hajimehoshi, line 1: ditto: `notios_darwin.go` is better.

internal/graphicsdriver/metal/ca/ca_darwin.go
```diff
@@ -106,94 +115,91 @@ func (ml MetalLayer) SetPixelFormat(pf mtl.PixelFormat) {
...
 // are synchronized with the display's refresh rate.
 //
 // Reference: https://developer.apple.com/documentation/quartzcore/cametallayer/2887087-displaysyncenabled.
 func (ml MetalLayer) SetDisplaySyncEnabled(enabled bool) {
-	switch enabled {
-	case true:
-		C.MetalLayer_SetDisplaySyncEnabled(ml.metalLayer, 1)
-	case false:
-		C.MetalLayer_SetDisplaySyncEnabled(ml.metalLayer, 0)
+	if runtime.GOOS == "ios" {
```
hajimehoshi, line 129: `cocoa.IsIOS`?

internal/graphicsdriver/metal/graphics_darwin.go
```diff
@@ -21,27 +21,27 @@ import (
 	"strings"
 	"unsafe"
 
+	"github.com/ebitengine/purego/objc"
 	"github.com/hajimehoshi/ebiten/v2/internal/graphics"
```
hajimehoshi, line 25: Separater external and internal dependencies i.e. add an empty line after `"github.com/ebitengine/purego/objc"`

internal/graphicsdriver/metal/mtl/mtl_darwin.go
```diff
@@ -338,6 +337,14 @@ const (
 	CommandBufferStatusError       CommandBufferStatus = 5 // Execution of the command buffer was aborted due to an error during execution.
 )
 
+var (
+	metal                         = purego.Dlopen("Metal.framework/Metal", purego.RTLD_GLOBAL)
+	_MTLCreateSystemDefaultDevice = purego.Dlsym(metal, "MTLCreateSystemDefaultDevice")
+
+	_ = purego.Dlopen("/System/Library/Frameworks/CoreGraphics.framework/Versions/Current/CoreGraphics", purego.RTLD_GLOBAL)
```
hajimehoshi, line 344: Why do we need `Dlopen` without using the returning value here? If we expect only side effects, what about moving this into `init` functions?

internal/graphicsdriver/metal/mtl/mtl_darwin.go
```diff
@@ -457,134 +464,257 @@ type Device struct {
...
+
+func _NSStringtoGoString(nsstring objc.ID) string {
+	// this will be nicer with unsafe.Slice once ebitengine requires 1.17
+	// reflect.SliceHeader is used because it will force Go to copy the string
+	// into Go memory when casted to a string
+	var s []byte
+	header := (*reflect.SliceHeader)(unsafe.Pointer(&s))
+	header.Data = uintptr(nsstring.Send(sel_UTF8String))
+	header.Len = int(nsstring.Send(sel_length))
+	return string(s)
```
hajimehoshi, line 479: Set `Cap` too.

internal/graphicsdriver/metal/mtl/mtl_darwin.go
```diff
@@ -814,33 +965,57 @@ type BlitCommandEncoder struct {
 //
 // Reference: https://developer.apple.com/documentation/metal/mtlblitcommandencoder/1400775-synchronize.
 func (bce BlitCommandEncoder) Synchronize(resource Resource) {
-	C.BlitCommandEncoder_Synchronize(bce.commandEncoder, resource.resource())
+	if runtime.GOOS == "ios" {
```
hajimehoshi, line 968: `cocoa.IsIOS`

internal/graphicsdriver/metal/mtl/mtl_darwin.go
```diff
@@ -814,33 +965,57 @@ type BlitCommandEncoder struct {
...
+		return
+	}
+	objc.ID(bce.commandEncoder).Send(sel_synchronizeResource, resource.resource())
 }
 
 func (bce BlitCommandEncoder) SynchronizeTexture(texture Texture, slice int, level int) {
-	C.BlitCommandEncoder_SynchronizeTexture(bce.commandEncoder, texture.texture, C.uint_t(slice), C.uint_t(level))
+	if runtime.GOOS == "ios" {
```
hajimehoshi, line 975: `cocoa.IsIOS`

internal/graphicsdriver/metal/mtl/mtl_darwin.go
```diff
@@ -849,103 +1024,130 @@ func (l Library) MakeFunction(name string) (Function, error) {
...
 }
 
 // GetBytes copies a block of pixels from the storage allocation of texture
 // slice zero into system memory at a specified address.
 //
 // Reference: https://developer.apple.com/documentation/metal/mtltexture/1515751-getbytes.
 func (t Texture) GetBytes(pixelBytes *byte, bytesPerRow uintptr, region Region, level int) {
-	r := region.c()
-	C.Texture_GetBytes(t.texture, unsafe.Pointer(pixelBytes), C.size_t(bytesPerRow), r, C.uint_t(level))
+	inv := cocoa.NSInvocation_invocationWithMethodSignature(cocoa.NSMethodSignature_SignatureWithObjCTypes("v@:^vQ{MTLRegion={MTLOrigin=qqq}{MTLSize=qqq}}Q"))
```
hajimehoshi, line 1047: Just out of curiosity, is there a document about this signature string rule?

TotallyGamerJet: Yeah it’s what’s returned from [@encode](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtTypeEncodings.html#//apple_ref/doc/uid/TP40008048-CH100). Obviously the best thing is to get the signature from the method itself but that doesn’t work with protocols bc the only way is with the protocol_getDescription function which returns a struct. We don’t have struct return support yet which is the reason I’m using NSInvocation.

hajimehoshi: Thanks!

hajimehoshi: > We don’t have struct return support yet which is the reason I’m using NSInvocation.

I think this is a good point to leave a comment.

TotallyGamerJet: Sure I leave it by use of nsinvocation or just at its definition?

hajimehoshi: just at its definition seems enough, thanks!

internal/graphicsdriver/metal/view_ios.go
```diff
@@ -17,8 +17,11 @@
 
 package metal
 
-// #cgo CFLAGS: -x objective-c
-// #cgo LDFLAGS: -framework UIKit
+// Suppress the warnings about availability guard with -Wno-unguarded-availability-new.
```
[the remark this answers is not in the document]
hajimehoshi: I'm fine with this change in another later PR. How big will the change be?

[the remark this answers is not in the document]
TotallyGamerJet: Not big but let’s just save it for another PR. This one is big enough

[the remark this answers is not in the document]
hajimehoshi: I agree. I'm looking forward to your next PR :-)

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/hajimehoshi/ebiten/v2
 go 1.15
 
 require (
-	github.com/ebitengine/purego v0.0.0-20220729024107-78cdc2949de6
+	github.com/ebitengine/purego v0.0.0-20220813165658-375146121d78
```
hajimehoshi, line 6: Update to 8dbe340b03f11115e843b9ee32252727985fb822

TotallyGamerJet:
I would test on actual device again since I did a pretty big refactor. I moved the common OBJC objects into the cocoa package because they will be needed when ports the rest of Ebitengine

internal/graphicsdriver/metal/mtl/example_darwin_test.go
```diff
@@ -17,6 +17,7 @@ package mtl_test
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/ebitengine/purego"
```
hajimehoshi, line 20: Move this to the same group as `"golang.org/x/image/math/f32"`

hajimehoshi (approved):
LGTM, thanks!

internal/graphicsdriver/metal/mtl/example_darwin_test.go
superloach: Would using something like `gofumpt` help here? IIRC it automatically groups imports like this. Not sure if it matches existing Ebitengine code style though.

internal/graphicsdriver/metal/mtl/mtl_darwin.go
```diff
@@ -457,134 +461,236 @@ type Device struct {
...
-		FrontFaceStencilDepthFailureOperation:     C.uint8_t(dsd.FrontFaceStencil.DepthFailureOperation),
-		FrontFaceStencilDepthStencilPassOperation: C.uint8_t(dsd.FrontFaceStencil.DepthStencilPassOperation),
-		FrontFaceStencilStencilCompareFunction:    C.uint8_t(dsd.FrontFaceStencil.StencilCompareFunction),
-	}
+	depthStencilDescriptor := objc.ID(class_MTLDepthStencilDescriptor).Send(sel_new)
+	backFaceStencil := depthStencilDescriptor.Send(sel_backFaceStencil)
+	backFaceStencil.Send(sel_setStencilFailureOperation, uintptr(dsd.BackFaceStencil.StencilFailureOperation))
+	backFaceStencil.Send(sel_setDepthFailureOperation, uintptr(dsd.BackFaceStencil.DepthFailureOperation))
+	backFaceStencil.Send(sel_setDepthStencilPassOperation, uintptr(dsd.BackFaceStencil.DepthStencilPassOperation))
+	backFaceStencil.Send(sel_setStencilCompareFunction, uintptr(dsd.BackFaceStencil.StencilCompareFunction)) //TODO
```
hajimehoshi, line 684: What is this TODO? @TotallyGamerJet

TotallyGamerJet: It can be removed. It was needed when I was working on it and just never got deleted.

hajimehoshi: Sure!

internal/graphicsdriver/metal/mtl/example_darwin_test.go
hajimehoshi: I'll take a look later, thanks

PR #2271, opened 2022-08-27
add Termination

superloach:
should fix #2266

PR #2272, opened 2022-08-27
regular termination (again)

### LINKED ISSUE #2266: ebiten: API to terminate an application explicitly
Now an Ebiten application can be terminated by returning an user-defined error at Update, but this is a little counterintuitive: even though the application exits in a regular way, the program treats 'error' values.

Instead of treating errors, what about having an explicit API to terminate an application, like `ebiten.Terminate`?

superloach:
- add RegularTermination, fix RunGame, correct references to Run
- ensure tests pass
- example tests also pass
- closes #2266
- second attempt at #2271, don't you love git

### DIFF
--- a/examples/fullscreen/main.go
+++ b/examples/fullscreen/main.go
@@ -41,7 +41,6 @@ import (
 var (
 	gophersImage       *ebiten.Image
 	mplusFont          font.Face
-	regularTermination = errors.New("regular termination")
 )
 
 func init() {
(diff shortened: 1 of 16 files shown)


superloach:
Now it's actually coffee time :coffee:

hajimehoshi:
Please update the comment of `RunGame`.

hajimehoshi (approved):
LGTM, thanks!

PR #2300, opened 2022-09-03
linux: Fix a typo that caused oddly cut off gamepad names.

divVerent:
This change allows gamepad names up to 256 characters, which is usually
enough. Before, names were cut off after 7 characters, matching the string
"Unknown" that is used if querying the name fails.


hajimehoshi:
Thanks. Would it be possible to file the issue first?

hajimehoshi (approved):
LGTM, thanks!

PR #2308, opened 2022-09-05
linux/gamepad: Do not fail on EPERM return values.

divVerent:
This appears to happen inside the Snap sandbox when accessing a non-gamepad /dev/input device.

### DIFF
--- a/internal/gamepad/gamepad_linux.go
+++ b/internal/gamepad/gamepad_linux.go
@@ -108,6 +108,10 @@ func (*nativeGamepadsImpl) openGamepad(gamepads *gamepads, path string) (err err
 		if err == unix.EACCES {
 			return nil
 		}
+		// This happens with the Snap sandbox.
+		if err == unix.EPERM {
+			return nil
+		}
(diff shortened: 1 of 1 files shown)


internal/gamepad/gamepad_linux.go
```diff
@@ -105,7 +105,7 @@ func (*nativeGamepadsImpl) openGamepad(gamepads *gamepads, path string) (err err
 
 	fd, err := unix.Open(path, unix.O_RDONLY|unix.O_NONBLOCK, 0)
 	if err != nil {
-		if err == unix.EACCES {
+		if err == unix.EACCES || err == unix.EPERM {
 			return nil
 		}
 		// This happens just after a disconnection.
```
hajimehoshi, line 111: Let's add a comment like this

```go
		if err == unix.EACCES {
			return nil
		}
		// This happens with the Snap sandbox
		if err == unix.EPERM {
			return nil
		}
```

hajimehoshi (approved):
LGTM, thanks!

PR #2330, opened 2022-09-15
remove Cgo on darwin from gamepad_darwin.go

TotallyGamerJet:


internal/gamepad/api_cf_darwin.go
```diff
@@ -0,0 +1,124 @@
+package gamepad
```
hajimehoshi, line 1: Add a license comment.

internal/gamepad/api_hid_darwin.go
```diff
@@ -0,0 +1,171 @@
+package gamepad
```
hajimehoshi, line 1: Add a license comment.

hajimehoshi (commented):
Would it be possible to suppress go-vet errors?

internal/gamepad/gamepad_darwin.go
```diff
@@ -73,82 +41,80 @@ func newNativeGamepadsImpl() nativeGamepads {
...
 			pageRef,
 			usageRef,
 		}
 
-		dict := C.CFDictionaryCreate(C.kCFAllocatorDefault,
+		dict := _CFDictionaryCreate(kCFAllocatorDefault,
 			(*unsafe.Pointer)(unsafe.Pointer(&keys[0])),
 			(*unsafe.Pointer)(unsafe.Pointer(&values[0])),
-			C.CFIndex(len(keys)), &C.kCFTypeDictionaryKeyCallBacks, &C.kCFTypeDictionaryValueCallBacks)
+			_CFIndex(len(keys)), *(**_CFDictionaryKeyCallBacks)(unsafe.Pointer(&kCFTypeDictionaryKeyCallBacks)), *(**_CFDictionaryValueCallBacks)(unsafe.Pointer(&kCFTypeDictionaryValueCallBacks)))
```
hajimehoshi, line 76: Hmm, the pointer indirection seems a dirty hack to trick go-vet. I'd skip go-vet for this package if we need. Is there a option to skip a specific package for go-vet by the way?

TotallyGamerJet: I don't think there is a way. I feel it's better to use the hack for the 6 or so needs of it and still have go get for the rest of the package then to try and not have go get at all

hajimehoshi: You mean go vet?

TotallyGamerJet: Yes

hajimehoshi: Sure, there are pros and cons, and keeping go-vet enabled seems slightly better.

internal/gamepad/api_cf_darwin.go
```diff
@@ -1,3 +1,17 @@
+// Copyright 2022 The Ebiten Authors
```
hajimehoshi, line 1: Ebitengine authors

internal/gamepad/api_hid_darwin.go
```diff
@@ -1,3 +1,17 @@
+// Copyright 2022 The Ebiten Authors
```
hajimehoshi, line 1: Ditto

internal/gamepad/api_cf_darwin.go
```diff
@@ -0,0 +1,138 @@
...
+	purego.SyscallN(procCFRelease, uintptr(cf))
+}
+
+func _CFRunLoopGetMain() _CFRunLoopRef {
+	ret, _, _ := purego.SyscallN(procCFRunLoopGetMain)
+	return _CFRunLoopRef(ret)
+}
+
+func _CFRunLoopRunInMode(mode _CFRunLoopMode, seconds _CFTimeInterval, returnAfterSourceHandled bool) _CFRunLoopRunResult {
+	var b uintptr = 0
```
hajimehoshi, line 113: (I'll update this but) `= 0` was not needed.

internal/gamepad/api_hid_darwin.go
```diff
@@ -0,0 +1,185 @@
...
+	kIOHIDDeviceUsagePageKey = []byte("DeviceUsagePage\x00")
+	kIOHIDDeviceUsageKey     = []byte("DeviceUsage\x00")
+)
+
+type _IOOptionBits uint32
+type _IOHIDManagerRef uintptr
+type _IOHIDDeviceRef uintptr
+type _IOHIDElementRef uintptr
+type _IOHIDValueRef uintptr
+type _IOReturn int
```
hajimehoshi, line 82: Isn't this `int32` rather than `int`?

```c
typedef	kern_return_t		IOReturn;
```

```c
typedef int kern_return_t;
```

internal/gamepad/api_hid_darwin.go
```diff
@@ -0,0 +1,185 @@
...
+type _IOHIDElementType uint
```
hajimehoshi, line 83: Is this really `uint`?

internal/gamepad/api_hid_darwin.go
TotallyGamerJet: Yes that should be int32

internal/gamepad/api_hid_darwin.go
TotallyGamerJet: It should be uint32 actually. see [Apple](https://developer.apple.com/documentation/iokit/iohidelementtype)

internal/gamepad/api_hid_darwin.go
hajimehoshi: Sure, I'll make that fix, thanks

internal/gamepad/api_hid_darwin.go
hajimehoshi: Thanks!

PR #2335, opened 2022-09-18
internal/gamepaddb: fix mapping support for e.g. `dpleft:-a0`.

### LINKED ISSUE #2334: internal/gamepaddb: interpret "-a0" correctly
SDL source has:

```
        if (half_axis_input == '+') {
            bind.input.axis.axis_min = 0;
            bind.input.axis.axis_max = SDL_JOYSTICK_AXIS_MAX;
        } else if (half_axis_input == '-') {
            bind.input.axis.axis_min = 0;
            bind.input.axis.axis_max = SDL_JOYSTICK_AXIS_MIN;
        } else {
            bind.input.axis.axis_min = SDL_JOYSTICK_AXIS_MIN;
            bind.input.axis.axis_max = SDL_JOYSTICK_AXIS_MAX;
        }
```

Ebitengine has:

```
                min := -1
                max := 1
                numstr := str[1:]

                if str[0] == '+' {
                        numstr = str[2:]
                        min = 0
                } else if str[0] == '-' {
                        numstr = str[2:]
                        max = 0
                }
```

Observe that for the case of a lone '-', SDL maps [0, -1] to [-1, 1], while Ebitengine maps [-1, 0] to [-1, 1].

Or in other words, if SDL needs "-a0" for a certain gamepad, Ebitengine would currently need "-a0~" which is wrong.

Reference: https://github.com/divVerent/aaaaxy/issues/28

divVerent:
SDL interprets this as "map -1 to 1, map 0 to -1", so we should do the same.

This fix contains two parts:

- Fix the intended output range.
- Also fix the formula to map a range to a range.

The fix does not change behavior if a "-a" mapping isn't used, as in any other case max-min == 1 or max+min == 0.

Fixes #2334

### DIFF
--- a/internal/gamepaddb/gamepaddb.go
+++ b/internal/gamepaddb/gamepaddb.go
@@ -228,14 +228,30 @@ func parseMappingElement(str string) (*mapping, error) {
 
 		if str[0] == '+' {
 			numstr = str[2:]
+			// Only use the positive half, i.e. 0..1.
 			min = 0
 		} else if str[0] == '-' {
 			numstr = str[2:]
-			max = 0
+			// Only use the negative half, i.e. -1..0,
+			// but invert the sense so 0 does not "press" buttons.
+			//
+			// In other words, this is the same as '+' but with the input axis
+			// value reversed.
+			//
+			// See SDL's source:
+			// https://github.com/libsdl-org/SDL/blob/f398d8a42422c049d77c744658f1cd2bb011ed4a/src/joystick/SDL_gamecontroller.c#L960
+			min, max = 0, min
 		}
 
+		// Map min..max to -1..+1.
+		//
+		// See SDL's source:
+		// https://github.com/libsdl-org/SDL/blob/f398d8a42422c049d77c744658f1cd2bb011ed4a/src/joystick/SDL_gamecontroller.c#L276
+		// then simplify assuming output range -1..+1.
+		//
+		// Yields:
 		scale := 2 / (max - min)
-		offset := -(max + min)
+		offset := -(max + min) / (max - min)
 		if tilda {
 			scale = -scale
 			offset = -offset


internal/gamepaddb/gamepaddb.go
```diff
@@ -228,14 +228,30 @@ func parseMappingElement(str string) (*mapping, error) {
...
+		// Map min..max to -1..+1.
+		//
+		// See SDL's source:
+		// https://github.com/libsdl-org/SDL/blob/f398d8a42422c049d77c744658f1cd2bb011ed4a/src/joystick/SDL_gamecontroller.c#L276
+		// then simplify assuming output range -1..+1.
+		//
+		// Yields:
 		scale := 2 / (max - min)
-		offset := -(max + min)
+		offset := -(max + min) / (max - min)
```
hajimehoshi, line 254: Doesn't this change the result when `max` is 1 and `min` is -1?

divVerent: No, because in that case max + min == 0 and 0 / something is still 0.

internal/gamepaddb/gamepaddb.go
```diff
@@ -228,14 +228,30 @@ func parseMappingElement(str string) (*mapping, error) {
...
+		}
+
+		// Map min..max to -1..+1.
+		//
+		// See SDL's source:
+		// https://github.com/libsdl-org/SDL/blob/f398d8a42422c049d77c744658f1cd2bb011ed4a/src/joystick/SDL_gamecontroller.c#L276
+		// then simplify assuming output range -1..+1.
+		//
+		// Yields:
 		scale := 2 / (max - min)
```
hajimehoshi, line 253: Is it OK for scale to be negative?

internal/gamepaddb/gamepaddb.go
hajimehoshi: Hmm, this seems very tricky.. but yes.

internal/gamepaddb/gamepaddb.go
divVerent: Yes, it is; note that tilda also causes this. In fact, my fix is kinda equivalent to keeping the code as is and just adding "tilda = !tilda" to the "-" case, but that'd be dirty and matches neither GLFW's nor SDL's implementation.

hajimehoshi: OK so let's revisit when GLFW is updated.

hajimehoshi (approved):
LGTM, thanks!

internal/gamepaddb/gamepaddb.go
hajimehoshi: Out of curiosity, is it possible to use `~` and `-` at the same time?

divVerent: Absolutely; both in SDL and Ebitengine this then inverts the output signal after all other processing. The implementation of this is correct.

However no current gamecontroller mapping uses this combination.

There are also a few extra peculiar entries in gamecontrollerdb that synthesize an axis from two buttons like this:

`+rightx:b9,+righty:b3,-rightx:b4,-righty:b8`

This seems to indicate a gamepad with multiple D-pads, where the only way to achieve standard mapping is a combination like this. Ebitengine and GLFW both don't support this, and at least Ebitengine will simply ignore these mappings. It seems like this applies primarily to N64 controllers.

hajimehoshi: I see. We don't plan to support such extra dpad buttons since the current standard layout cannot have them. If you think we have to fix this, please file this...

divVerent: Yeah; we maybe should ignore mapping to unknown buttons/axes though. Right now Ebitengine considers them unknown axes with names like "-leftx", so the definition still parses and the N64 controller would be usable apart from these four buttons (meaning it'd basically still be a NES-ish controller with an extra stick). I guess that's fine.

PR #2350, opened 2022-09-25
misc: added more logo examples

erexo:
I had those logos on my disc for quite a while now, so I thought that it may be useful for someone.


hajimehoshi:
Logos are very important for branding strategies, so I cannot merge this without discussions, sorry

Issue #2351, opened 2022-09-26
ebiten: deprecate `(*ebiten.Image).Size()` and recommend `(*ebiten.Image).Bounds()`


hajimehoshi:
Not determined yet.

PR #2353, opened 2022-09-26
Update image.go #2351

JupiterRider:
#2351


hajimehoshi:
I've not determined whether I'd do it or not.

PR #2358, opened 2022-09-28
examples/audio: reduce noticeable delay for the sound effect

xyproto:
There was a noticeable delay each time the sound effect was played (when pressing the `p` button).

With the proposed changes, there is only a slight audible delay the first time the button is pressed, but not for the following key-presses.


hajimehoshi:
This seems a breaking change as a SE cannot be played when another SE is playing.

xyproto:
Ah, thanks for reviewing.

PR #2359, opened 2022-09-28
image: reorder fields as recommended by the fieldalignment utility

xyproto:
This reduces the struct size from 64 to 32 bytes.


hajimehoshi (approved):
LGTM

Issue #2286, opened 2022-08-31
audio/internal/convert: lazy initialization of `cosTable`


hajimehoshi:
`cosTable` takes 65536 float64s = 512KB. This is not neeeded when you don't use resampling. Let's use sync.Once to initialize this lazily.

ReCodee:
Is this issue relevant?
I'd like to work on this issue

hajimehoshi:
Yes, go ahead

PR #2398, opened 2022-10-20
lazy initialization of cosTable

ReCodee:
solves #2286 
initialized cosTable lazily with the help of sync.once to improve performance
@hajimehoshi please take a look

### DIFF
--- a/audio/internal/convert/resampling.go
+++ b/audio/internal/convert/resampling.go
@@ -17,14 +17,19 @@ package convert
 import (
 	"io"
 	"math"
+	"sync"
 )
 
-var cosTable = [65536]float64{}
+var lazyInit sync.Once = sync.Once{}
+var cosTable []float64
 
 func init() {
-	for i := range cosTable {
-		cosTable[i] = math.Cos(float64(i) * math.Pi / 2 / float64(len(cosTable)))
-	}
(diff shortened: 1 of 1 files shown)


audio/internal/convert/resampling.go
```diff
@@ -17,14 +17,19 @@ package convert
...
 
-var cosTable = [65536]float64{}
+var lazyInit sync.Once = sync.Once{}
+var cosTable []float64
 
 func init() {
-	for i := range cosTable {
-		cosTable[i] = math.Cos(float64(i) * math.Pi / 2 / float64(len(cosTable)))
-	}
+	lazyInit.Do(func() {
```
hajimehoshi, line 27: I don't think this is a lazy initialization.

ReCodee: Thanks for pointing this out, Actually I took a reference from this resource [Link](https://cloud.google.com/functions/docs/bestpractices/tips#functions-tips-lazy-globals-go)
I'd be grateful if you can give suggestions on how should I go ahead to implement lazy initialization.

hajimehoshi: So you used a code even you didn't fully understand? That's not good.

And I don't think you understand what #2286 is yet. Please come back when you understand this.

PR #2433, opened 2022-11-03
examples: remove `example` build tag

### LINKED ISSUE #1462: examples: remove `examples` build tags when we can drop Go 1.16
The `examples` build tag is to prevent users from install example binaries unexpectedly. However, as installing with a version will be must outside a module, and `...` like `go install foo/bar/...` will no longer be recommended (not sure, let me double-check), we might not need the tag any more.

https://tip.golang.org/doc/go1.16#go-command

> Module-aware mode is enabled by default, regardless of whether a go.mod file is present in the current working directory or a parent directory. More precisely, the GO111MODULE environment variable now defaults to on. To switch to the previous behavior, set GO111MODULE to auto.
>
> [...]
>
> go install now accepts arguments with version suffixes (for example, go install example.com/cmd@v1.0.0). This causes go install to build and install packages in module-aware mode, ignoring the go.mod file in the current directory or any parent directory, if there is one. This is useful for installing executables without affecting the dependencies of the main module.
>
> go install, with or without a version suffix (as described above), is now the recommended way to build and install packages in module mode. go get should be used with the -d flag to adjust the current module's dependencies without building packages, and use of go get to build and install packages is deprecated. In a future release, the -d flag will always be enabled.

mymmrac:
Closes #1462 

Refactoring

## What this PR does | solves
Removed `example` build tag as no needed any more


.builds/alpine.yml
```diff
@@ -14,4 +14,4 @@ sources:
 tasks:
   - build: |
       cd ebiten
-      go build -tags=example ./...
+      go build ./examples/...
```
hajimehoshi, line 17: This changes the build targets

mymmrac:
Hi @hajimehoshi, could you please check if something else needs to be updated? I am not very familiar with `gomobile` and does it need `@latest` at the end (in README.md)

.builds/alpine.yml
mymmrac: Should it be then: `go build -o . ./examples/...` ?

hajimehoshi: No.

```
go build ./...
```

is correct

examples/README.md
```diff
@@ -3,21 +3,21 @@
...
+go run github.com/hajimehoshi/ebiten/v2/examples/rotate@latest
 ```
 
 ## Android
 
 Install [gomobile](https://pkg.go.dev/golang.org/x/mobile/cmd/gomobile) first.
 
 ```sh
-gomobile install -tags=example github.com/hajimehoshi/ebiten/v2/examples/rotate
+gomobile install github.com/hajimehoshi/ebiten/v2/examples/rotate
```
hajimehoshi, line 14: I think `@latest` is needed

examples/README.md
```diff
@@ -3,21 +3,21 @@
...
 ```
 
 ## iOS
 
 ```sh
-gomobile build -target=ios -tags=example -work github.com/hajimehoshi/ebiten/v2/examples/rotate
+gomobile build -target=ios -work github.com/hajimehoshi/ebiten/v2/examples/rotate
```
hajimehoshi, line 20: ditto

.builds/alpine.yml
mymmrac: Hmm, but it will build everything and not only examples, is it what we want?

hajimehoshi: Yes, and currently

```
go build -tags=example ./...
```

builds everything including examples and non-examples, right?

mymmrac: Oh, sorry, messed up that, correct it should be just `./...`

hajimehoshi (approved):
LGTM, thanks!

Issue #1162, opened 2020-05-19
Remove cgo on macos


hajimehoshi:
https://groups.google.com/forum/m/#!topic/golang-nuts/Wt6P0BYndvs

mortenson:
Would this enable cross compilation, i.e. compiling mac binaries from linux?

hajimehoshi:
Yes, theoretically :-)

hajimehoshi:
Another nice reference CL: https://go-review.googlesource.com/c/go/+/227037/

hajimehoshi:
As M1 appears, might this be in higher demand than before?

hajimehoshi:
Hmm, GLFW is the toughest blocker so far.

hajimehoshi:
https://github.com/progrium/macdriver sounds reasonable!

hajimehoshi:
I want to prioritize this, as it is now impossible to build amd64 version from M1 macOS.

hajimehoshi:
Hm, we should be able to use environment variables `CC`, `CFLAGS` or something for other architectures.

hajimehoshi:
https://developer.apple.com/documentation/objectivec/objective-c_runtime?language=objc

TotallyGamerJet:
https://gist.github.com/vaiorabbit/6204931#file-building-libglfw-dylib-on-macos-apple-silicon

This explains how to build glfw as a dylib which can then by dynamically linked to using dlopen and dlsym.

hajimehoshi:
Oops, I used a wrong issue number...

aprchen:
> I want to prioritize this, as it is now impossible to build amd64 version from M1 macOS.

Is it can be compiled in version 2.4 ?

hajimehoshi:
> Is it can be compiled in version 2.4 ?

Hopefully yes

TotallyGamerJet:
[...]

PR #2453, opened 2022-11-13
all: use RTLD_LAZY in dlopen for darwin

TotallyGamerJet:
Updates #1162

bug 

## What this PR does | solves
dlopen requires either RTLD_LAZY or RTLD_NOW but there was neither.

### DIFF
--- a/internal/gamepad/api_cf_darwin.go
+++ b/internal/gamepad/api_cf_darwin.go
@@ -50,7 +50,7 @@ const (
 )
 
 var (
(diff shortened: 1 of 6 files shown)


hajimehoshi (approved):
LGTM, but which is preferralble RTLD_NOW and RTLD_LAZY?

hajimehoshi:
I thought RTLD_LAZY seemed preferrable to load symbols lazily, so that's fine.

TotallyGamerJet:
I believe RTLD_LAZY is the preferred way. RTLD_NOW will resolve ALL symbols while RTLD_LAZY resolves only the ones that are used. Since we aren't using the full functionality of any one framework I think RTLD_LAZY is best.

PR #2464, opened 2022-11-20
internal/cocoa & internal/graphicsdriver/mtl: remove usages of NSInvocations to directly call ID.Send

TotallyGamerJet:
Updates #1162 

feature

## What this PR does | solves
purego now supports using floats as arguments. We can remove these instances of NSInvocation that only existed to circumvent that feature.


hajimehoshi (approved):
LGTM, thanks!

Issue #2469, opened 2022-11-23
text: redesign APIs to considering sub-pixels


hajimehoshi:
### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] Xbox
- [ ] Web Browsers

### What feature would you like to be added?

Now one glyph image is cached for one glyph, and they don't consider 'sub-pixel' offsets. Then, whatever the offset is, the same glyph image is always used. The rendering quality is not good.

@tinne26 suggested 'horizontal non-quantized rastering' to use 64 variations of a glyph image. As a glyph image only has alpha channel information, we can use 4 channels for one texture image. Then, as the overhead, the number of images will be increased by 64/4 = 16 times. I think this is reasonable.

See also https://github.com/tinne26/etxt/blob/main/docs/quantization.md

### Why is this needed?

For high-quality text rendering.

hajimehoshi:
OK, updated the title.

hajimehoshi:
> This is more important for text animation and high-resolution horizontal interspacing than overall quality

I meant these are about qualities actually.

hajimehoshi:
Hm, the font [`goregular`](https://pkg.go.dev/golang.org/x/image/font/gofont/goregular) has integer advances and kerns, so subpixels are not available (as long as the start position is an integer)

hajimehoshi:
Same as "M+ font". I'm not sure but `github.com/benoitkugler/textlayout/fonts/truetype` might have a different result.

hajimehoshi:
The current APIs

```
func Draw(dst *ebiten.Image, text string, face font.Face, x, y int, clr color.Color)
func DrawWithOptions(dst *ebiten.Image, text string, face font.Face, ...)
func AppendGlyphs(glyphs []Glyph, face font.Face, text string) []Glyph
```

don't take float positions. Thus, there is nothing we can do with the current API so far...

tinne26:
[...]

PR #2529, opened 2023-01-07
text: adjust the glyph cache size

### LINKED ISSUE #2528: text: request a variant of CacheGlyphs that caches glyphs _entirely_
### Operating System

- [X] Windows
- [X] macOS
- [X] Linux
- [X] FreeBSD
- [X] OpenBSD
- [X] Android
- [X] iOS
- [X] Nintendo Switch
- [X] Xbox
- [X] Web Browsers

For performance reasons, I need to pin all glyphs to cache that my font may be using.

I am aware of CacheGlyphs(), which until recently has done that; however, now it only caches the glyphs necessary to render the given string.

The difference is that since recently, each glyph can have four variants depending on its x offset.

I do think CacheGlyphs() is within spec, which is why I am considering this a FR, not a bug. Maybe what I need is something like CacheAllGlyphs which then takes a []rune and not a string, and caches all that is needed to render any string consisting of these runes, not just one specific string?

(Obviously I could implement my own workaround: manipulate a font so a character - say, "\001" - I am not using has a width of 1/4 pixel, and call CacheGlyphs with "charset\n\001charset\n\001\001charset\n\001\001\001charset" to cover all cases - but I do not think this would be a great idea, and relies heavily on how Ebitengine does glyph coordinate rounding)

Edit: scratch that about cache size - any glyph used or cached within the last 60 frames is not freed, so if I call CacheGlyphs once per frame, I can already have any cache size I want.

### Why is this needed?

Otherwise, frames where a new text appears occasionally have lag spikes in my game AAAAXY because font rendering can be expensive. It is just a single spike as subsequent frames have all they need in glyph cache, but still is annoying and jarring.

hajimehoshi:
Now, if the font doesn't have hinting, there are at most 4 glyph variations for one rune. This means we need bigger size of the glyph cache.

This change changes the algorithm of the cache size. Now the cache size is calculated based on the size of the font.

Updates #2469
Closes #2528

### DIFF
--- a/text/text.go
+++ b/text/text.go
@@ -272,7 +272,13 @@ func DrawWithOptions(dst *ebiten.Image, text string, face font.Face, options *eb
 	// If the number of glyphs exceeds this soft limits, old glyphs are removed.
 	// Even after clearning up the cache, the number of glyphs might still exceeds the soft limit, but
 	// this is fine.
-	const cacheSoftLimit = 512
+	//
+	// Note that at most there are 4 variations for one rune if the font doesn't have hintings (#2469, #2528).
+	// So, if the size is 64px, the size is 4000000 / (64*64) = 976, and 976 / 4 = 244 runes can be cached.
+	// The constants in the algorithm is arbitrary, but this should be enough to cover ASCII glyphs.
+	m := face.Metrics()
+	size := (m.Ascent + m.Descent).Ceil()
+	cacheSoftLimit := min(4000000/(size*size), 2048)
 
 	// Clean up the cache.
 	if len(glyphImageCache[face]) > cacheSoftLimit {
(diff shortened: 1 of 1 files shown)


hajimehoshi:
@divVerent Does this change work to you?

divVerent:
This is half of what I'd like - the other half - providing a way to request that all images for a glyph are cached - remains open.

Although I could stick with my workaround at https://github.com/divVerent/aaaaxy/commit/19dba865b79f5ead294bcb7e3f275c8250f0ddf9 which is making one unused glyph have a width of 1/4 pixel, and then caching "charset", "\001charset", "\001\001charset" and "\001\001\001charset". Not quite nice.

hajimehoshi:
I see, now I understood the issue...

PR #2530, opened 2023-01-07
text: make CacheGlyphs create all the sub-pixel variations

hajimehoshi:
Before this change, CacheGlyphs created only necessary glyphs wihtout considering variations by sub-pixels in X direction. This was sometimes unexpected when a user might want to prepare glyph cache with a string including all the letters, instead of actual texts.

This change makes CacheGlyphs create all the sub-pixel variations of the given runes.

Updates #2469
Updates #2528

### DIFF
--- a/text/text.go
+++ b/text/text.go
@@ -362,6 +362,9 @@ func BoundString(face font.Face, text string) image.Rectangle {
 // merged into one draw call regardless of the size of the text.
 //
 // If a rune's glyph is already cached, CacheGlyphs does nothing for the rune.
+//
+// One rune can have multiple varitations of glyphs due to sub-pixels in X direction.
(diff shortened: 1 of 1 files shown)


hajimehoshi:
@divVerent Please take a look.

divVerent:
This looks like it will work. Let me quickly test - with my other PR (texture unit bug) still open this is really easy to verify :)

hajimehoshi:
> with my other PR (texture unit bug) still open this is really easy to verify :)

I'll merge your PR as soon as possible after all the tests pass.

divVerent:
Yes, confirmed the glyph cache is complete with this - with this commit, my game also works fine despite the texture unit bug.

That bug was a nice test case for the glyph cache working, as the precaching happened at frame start where no shader calls took place yet :)

PR #2561, opened 2023-02-05
Fixed BlendDestinationOut behaviour #2560

sedyh:
# What issue is this addressing?
Closes #2560

image
image

## What _type_ of issue is this addressing?
Bug

## What this PR does | solves
Fixes BlendDestinationOut acting like BlendSourceOut.

PR #2563, opened 2023-02-06
Added example/blend

sedyh:
I don't know if it's worth adding this to the examples. But it seems to me that the output of each available blend mode specifically in ebitengine can be clearer than examples/masking (also it uses blend modes and not composite modes), when you need to quickly see how to assemble the appropriate mode. And it's also easy to add new modes without changing the rendering.

image

Updates #2561

feature

## What this PR does | solves
Adding the example of how to use blend modes fast.


examples/blendmodes/main.go
```diff
@@ -0,0 +1,192 @@
+// Copyright 2020 The Ebiten Authors
```
hajimehoshi, line 1: 2023 The Ebitengine Authors

examples/resources/images/blendmodes/embed.go
```diff
@@ -0,0 +1,26 @@
+// Copyright 2022 The Ebitengine Authors
```
hajimehoshi, line 1: 2023

examples/blendmodes/main.go
```diff
@@ -0,0 +1,192 @@
...
+	_ "image/png"
+	"log"
+	"math"
+
+	"github.com/hajimehoshi/ebiten/v2"
+	"github.com/hajimehoshi/ebiten/v2/examples/resources/images/blendmodes"
+	"github.com/hajimehoshi/ebiten/v2/text"
+	"golang.org/x/image/colornames"
+	"golang.org/x/image/font/inconsolata"
+)
```
hajimehoshi, line 29:
```
	"golang.org/x/image/colornames"
	"golang.org/x/image/font/inconsolata"

	"github.com/hajimehoshi/ebiten/v2/examples/resources/images/blendmodes"
	"github.com/hajimehoshi/ebiten/v2/text"
```

hajimehoshi (commented):
* I prefer 'blend' for the example name.
* What are the images' licenses?

sedyh:
>  I prefer 'blend' for the example name.
Ok, I'll fix it.

> What are the images' licenses?
They're drawn in inkscape, but the ampersand is made from sans-serif font.
I can choose any other font with free license.

hajimehoshi:
> I can choose any other font with free license.

Thanks, so can this be CC0? :-) Please edit https://github.com/hajimehoshi/ebiten/blob/main/examples/resources/images/license.md

sedyh:
> Thanks, so can this be CC0? :-)
Yeah, sure. This one is CC0. Btw, is SIL free and open soucre friendly?
[Pixel Operator](https://fontlibrary.org/en/font/pixel-operator#Pixel Operator 8-Bold)

examples/blendmodes/main.go
sedyh: Fixed

examples/resources/images/blendmodes/embed.go
sedyh: Fixed

examples/blendmodes/main.go
sedyh: Fixed

sedyh:
> Please edit https://github.com/hajimehoshi/ebiten/blob/main/examples/resources/images/license.md

Added.

examples/blend/main.go
```diff
@@ -0,0 +1,193 @@
+// Copyright 2023 The Ebiten Authors
```
hajimehoshi, line 1: Ebitengine

examples/resources/images/license.md
```diff
@@ -102,6 +102,14 @@ https://opengameart.org/content/super-mario-world-style-tiles
 CC0 1.0
 ```
 
+## blend/dest.png
```
hajimehoshi, line 105: What about source.png?

sedyh:
Is that related?
https://github.com/hajimehoshi/ebiten/pull/2563/checks#step:16:688

examples/resources/images/license.md
hajimehoshi, line 105: Put this in between `audio/pause.png` and `flappy/gopher.png` for alphabetical order

examples/blend/main.go
sedyh: Hm, I was referring to the other examples where the old name was kept.

examples/resources/images/license.md
sedyh: It's fully made by me.

Here is sources with both examples:
blendmodes

examples/blend/main.go
sedyh: Fixed it here rn.

examples/resources/images/blend/embed.go
```diff
@@ -0,0 +1,26 @@
...
+
+package blend
+
+import (
+	_ "embed"
+)
+
+var (
+	//go:embed source.png
+	Source []byte
```
hajimehoshi, line 23: Source_png

examples/resources/images/blend/embed.go
```diff
@@ -0,0 +1,26 @@
...
+	//go:embed dest.png
+	Dest []byte
```
hajimehoshi, line 25: Dest_png

examples/resources/images/license.md
```diff
@@ -86,6 +86,14 @@ https://fonts.google.com/icons
...
+
+```
+https://fontlibrary.org/en/font/seshat-regular
+
+CC0 1.0
+```
+
```
hajimehoshi, line 96: What about source.png?

sedyh: https://github.com/hajimehoshi/ebiten/pull/2563#discussion_r1097064119

examples/blend/main.go
hajimehoshi: > Should we update it everywhere in the examples?

You should not update the existing license comments.

examples/resources/images/blend/embed.go
sedyh: Hm, why we're using snake case for the var name? Isn't it wrong?

examples/resources/images/blend/embed.go
sedyh: Hm, why we're using snake case for the var name? Isn't it wrong?

examples/resources/images/blend/embed.go
hajimehoshi: Also please add an empty line between `Source []byte` and `//go:embed dest.png`

examples/resources/images/license.md
hajimehoshi: I mean adding source.png here too.

examples/blend/main.go
sedyh: > You should not update the existing license comments.
I mean not here, in the separate issue.
Or there's a special reason why we can't never change them?

examples/resources/images/license.md
sedyh: Oh, ok.

examples/resources/images/blend/embed.go
hajimehoshi: > Hm, why we're using snake case for the var name? Isn't it wrong?

This is just for consistency with other variable names.

hajimehoshi:
> Is that related? There was a fail due to timeout on go1.18+ubuntu/latest
> https://github.com/hajimehoshi/ebiten/pull/2563/checks#step:16:688

I don't think so.

examples/blend/main.go
hajimehoshi: > Or there's a special reason why we can't never change them?

I just follow other OSS's rules like Chromium: the copyright comment (year) is determined when the file is created, and is not updated basically.

examples/resources/images/blend/embed.go
sedyh: Fixed

examples/resources/images/blend/embed.go
sedyh: Fixed

examples/resources/images/license.md
sedyh: Added

examples/resources/images/blend/embed.go
hajimehoshi: > Hm, why we're using snake case for the var name? Isn't it wrong?

This is just for consistency with other variable names of other resources.

(Sorry I left this comment in a wrong thread)

examples/blend/main.go
```diff
@@ -0,0 +1,193 @@
...
+	g.offscreen.DrawImage(g.source, op)
+
+	// Draw the result on the passed coordinates.
+	op = &ebiten.DrawImageOptions{}
+	op.GeoM.Translate(x, y)
+	screen.DrawImage(g.offscreen, op)
+}
+
+// loadImage is a util function for loading embedded images.
+func loadImage(data []byte) *ebiten.Image {
```
hajimehoshi, line 139: Let's return an error instead of logging in this function

examples/blend/main.go
```diff
@@ -0,0 +1,193 @@
...
+	textSpacing  float64 // Additional vertical margin for text.
+	modes        []mode
+}
+
+func NewGame() *Game {
+	g := &Game{
+		source: loadImage(blend.Source_png),
+		dest:   loadImage(blend.Dest_png),
+		gridW:  4, gridH: 3,
+		tileSpacing: 64, textSpacing: 16,
```
hajimehoshi, line 60: This is just my preference but put one member for one line

```
		source: loadImage(blend.Source_png),
		dest:   loadImage(blend.Dest_png),
		gridW:  4,
		gridH: 3,
		tileSpacing: 64,
		textSpacing: 16,
```

examples/blend/main.go
```diff
@@ -0,0 +1,193 @@
...
+package main
+
+import (
+	"bytes"
+	"image"
+	_ "image/png"
+	"log"
+	"math"
+
+	"golang.org/x/image/colornames"
```
hajimehoshi, line 24: You don't have to use golang.org/x for white and black. Use `image/color`'s white and black.

examples/blend/main.go
```diff
@@ -0,0 +1,193 @@
...
+	for y := 0.; y < g.gridH; y++ {
+		for x := 0.; x < g.gridW; x++ {
+			px, py := x*(g.tileSize+g.tileSpacing), y*(g.tileSize+g.tileSpacing)
+			if y > 0 {
+				// Making a place for the text.
+				py += g.textSpacing * y
+			}
+
+			// Drawing the blend mode and it's name.
+			mode := getMode(g.modes, int(x), int(y), int(g.gridW), int(g.gridH))
```
hajimehoshi, line 111: Instead of rendering `empty`, what about letting `getMode` return mode and a boolean and skipping rendering when getMode returns false?

examples/blend/main.go
sedyh: Is there a sense to return an error here?
Where then to handle it? In main and then pass into the constructor?

hajimehoshi: Yeah. I think this makes sense. Handling error is not loadImage's responsibility.

hajimehoshi: So yes, NewGame will return an error.

examples/blend/main.go
sedyh: Ok

examples/blend/main.go
sedyh: Ok

examples/blend/main.go
sedyh: Ok

sedyh:
azure.archive.ubuntu.com don't feel so good :)

examples/blend/main.go
sedyh: Fixed

examples/blend/main.go
sedyh: Fixed

examples/blend/main.go
sedyh: Fixed

examples/blend/main.go
sedyh: Fixed

examples/blend/main.go
```diff
@@ -0,0 +1,218 @@
...
+// Game is a canvas for drawing blend mode tiles.
+type Game struct {
+	source      *ebiten.Image
+	dest        *ebiten.Image
+	offscreen   *ebiten.Image
+	tileSize    float64
+	gridW       float64
+	gridH       float64
+	tileSpacing float64 // Horizontal and vertical margin between tiles.
+	textSpacing float64 // Additional vertical margin for text.
```
hajimehoshi, line 53: Use ints instead of float64s for all the members. (float64 is necessary only when passing them to GeoM)

examples/blend/main.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+	return ebiten.NewImageFromImage(m), nil
+}
+
+// maxSide is a util function for adjusting the tile size for the available images.
+func maxSide(a, b *ebiten.Image) float64 {
+	xMax := math.Max(float64(a.Bounds().Dx()), float64(a.Bounds().Dx()))
+	yMax := math.Max(float64(b.Bounds().Dx()), float64(b.Bounds().Dx()))
+	return math.Max(xMax, yMax)
+}
```
hajimehoshi, line 171: After changing the member from float64 to int, you would need to implement your own max function for integers.

As this function is called only once, I think this doesn't have to be a separate function by the way.

examples/blend/main.go
```diff
@@ -0,0 +1,218 @@
...
+
+// inRange is a util function to check if the index is in range.
+func inRange(x, y, w, h int) bool {
+	xOut := x < 0 || x >= w
+	yOut := y < 0 || y >= h
+	if xOut || yOut {
+		return false
+	}
+	return true
```
hajimehoshi, line 185: `return 0 <= x && x < w && 0 <= y && y < h`

As this function is called only once, I think this doesn't have to be a separate function.

examples/blend/main.go
hajimehoshi: In general, we should avoid float values in logics as float values can include numerical errors.

examples/blend/main.go
sedyh: Only to improve readability with early return

examples/blend/main.go
sedyh: After all, they will still have to be cast in float64 for a smooth resize. Am I wrong?

hajimehoshi: Oh you are right. At least `tileSpacing` and `textSpacing` can be float64. I still think the others should be integers.

I realized `tileSpacing` and `textSpacing` can be consts (= untyped numeric values) instead of Game's members. What do you think?

examples/blend/main.go
sedyh: I mean, that I will need additional "if" for this in getMode func.

examples/blend/main.go
sedyh: Yeah, actually we can bring every setting to the package level and keep only resources and calculated values in the Game.

sedyh: > Oh you are right. At least tileSpacing and textSpacing can be float64. I still think the others should be integers.

Idk it makes too many casts from scratch.

`float64(g.gridW)*(float64(g.tileSize)+g.tileSpacing)-g.tileSpacing, float64(g.gridH)*(float64(g.tileSize)+g.tileSpacing)`

And we can't keep everything int too cause it will cause jitter.

examples/blend/main.go
hajimehoshi: > I mean, that I will need additional "if" for this in getMode func that I previously had in separate function.

I'm afraid I still don't understand what you said, but if we remove this function, getMode will be like this:

```go
func getMode(modes []mode, x, y, w, h int) (m mode, ok bool) {
	if x < 0 || x >= w || y < 0 || y >= h {
		return mode{}, false
	}
	if index(x, y, w) > len(modes)-1 {
		return mode{}, false
	}
	return modes[idx], true
}
```

> I'm talking about the first return in this func.

I prefer one-line implementation as I said, if we keep this function

```
return 0 <= x && x < w && 0 <= y && y < h
```

hajimehoshi: So I'm fine with an additional `if`.

examples/blend/main.go
hajimehoshi: Why not

```go
gw, gh := g.gridW*(g.tileSize+g.tileSpacing)-g.tileSpacing, g.gridH*(g.tileSize+g.tileSpacing)
```

hajimehoshi: > Yeah, actually we can bring every setting to the package level

You don't have to define the consts in the package level. Defining them in Draw is fine.

examples/blend/main.go
sedyh: Ok, I'll inline InRange.

examples/blend/main.go
sedyh: > Why not
> 
> ```go
> gw, gh := g.gridW*(g.tileSize+g.tileSpacing)-g.tileSpacing, g.gridH*(g.tileSize+g.tileSpacing)
> ```

Because you said, that g.gridW, g.gridH and g.tileSize should be integers.
If we keep them all float64 it will be the same as you wrote (but removing floats from util function makes sense).

examples/blend/main.go
sedyh: Ok

examples/blend/main.go
hajimehoshi: Why do gw and gh have to be float64?

sedyh: Hmm, they don't have division, right...
I need to think about it a bit.

examples/blend/main.go
sedyh: Fixed.

examples/blend/main.go
sedyh: Fixed.

examples/blend/main.go
sedyh: Fixed. I removed most floats I had.

hajimehoshi:
> azure.archive.ubuntu.com don't feel so good :)

Today's GitHub Actions seems a little sick unfortunately.

sedyh:
Btw, how the uploading prcoess work for wasm files?
All I found is this command at the site repo:

`GOOGLE_APPLICATION_CREDENTIALS=/path/to/credentials.json go run ./cmd/uploadwasm/ -ebitenpath=../path/to/ebiten -upload`

hajimehoshi:
I'll try GitHub actions later again

> Are they uploaded only by you, or there are some tools for uploading?

This is done by me so you don't have to care

examples/blend/main.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+// index returns 1D index from 2D.
+func index(x, y, w int) int {
+	return y*w + x
+}
+
+// getMode returns a blend mode by index and found status.
+func getMode(modes []mode, x, y, w, h int) (m mode, ok bool) {
+	idx := index(x, y, w)
```
hajimehoshi, line 196: What about inlining this function?

```go
idx := y*w + x
```

sedyh: Added

examples/blend/main.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/hajimehoshi/ebiten/v2/text"
+)
+
+const (
+	screenWidth  = 780
+	screenHeight = 600
+	tileSpacing  = 64
+	textSpacing  = 16
+	gridW        = 4
+	gridH        = 3
```
hajimehoshi, line 38:
```go
	tileSpacing  = 64
	textSpacing  = 16
	gridW        = 4
	gridH        = 3
```

I don't think we need these consts for Game's member variables.

I suggested to remove Game's members and replace them with the consts, but this is optional.

sedyh: I don't know, the Game will have a tighter coupling then, no?

sedyh:
Can we restart just one job btw?

hajimehoshi:
> Can we restart just one job btw?

I did 30 minutes ago

examples/blend/main.go
hajimehoshi: So you mean you prefer keeping member variables? Then remove these consts.

sedyh: Idk, the consts are more convenient. But can't we kept both? Then the user can just grab the Game implementation and use it in he's own code.

hajimehoshi: I don't think so since they are redundant. If we keep constants, 1) move the four constsnts to Draw and 2) remove the member variables. 

> Then the user will be able to just grab and use the Game implementation in his code.

In terms of reusability, I think consts in Draw is enough.

sedyh: Removed constructor args

sedyh: Oh, I didn't see that...
Shove them right inside the draw? Do you mean like this?
```go
	screen.Fill(color.White)
	const (
		tileSpacing = 64
		textSpacing = 16
		gridW       = 4
		gridH       = 3
	)
	...
}	
```
Hm, it looks strange.
Is it makes sense to separate them, why just not let user edit them in order to extend the grid?

hajimehoshi: > Do you mean like this?

Yes

> Is it makes sense to separate them,

I think this makes sense since the smaller scoper are the better they are in general.

> why just not let user edit them in order to extend the grid?

I don't understand. The users still can edit the consts whererever they are.

sedyh: Fixed

hajimehoshi (approved):
LGTM, thanks!

Issue #2546, opened 2023-01-17
internal/glfw: rewrite GLFW from C to Go for macOS and Linux


hajimehoshi:
### Operating System

- [ ] Windows
- [X] macOS
- [X] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] Xbox
- [ ] Web Browsers

### What feature would you like to be added?

We have already ported from GLFW to Go for Windows. We want to do the same thing for macOS and Linux.

Eliminating Cgo is not the goal of this issue (#1162, #2284)

### Why is this needed?

This is nice to remove Cgo from macOS: #remove
This is a blocker for some features: #2476, #2511

hajimehoshi:
We can have a better fix for #2518 if we can modify GLFW.

hajimehoshi:
> This is a blocker for some features that requires to modify GLFW: https://github.com/hajimehoshi/ebiten/issues/2476, https://github.com/hajimehoshi/ebiten/issues/2511, https://github.com/hajimehoshi/ebiten/issues/2258, https://github.com/hajimehoshi/ebiten/issues/2260, https://github.com/hajimehoshi/ebiten/issues/2289, https://github.com/hajimehoshi/ebiten/issues/2475

In order to resolve them, just moving the GLFW source to the current internal/goglfw should be fine. As the next step, we can rewrite them into Go.

PR #2567, opened 2023-02-06
internal/goglfw: separate windows specific struct from common structs

TotallyGamerJet:
Updates #2546

feature

## What this PR does | solves
Many of the structs found in internal.h (internal_windows.go) contain defines such as GLFW_WIN32_WINDOW_STATE which gets replaced with a struct defined in win32platform_windows.go (win32_platform.h).

Originally, these structs where directly placed inside of internal_windows.go. However, to make it easier to add macOS and Linux these cannot be in this file.

This commit separates the windows specific structs into the respective windows file and updates the field to be named state.

### DIFF
--- a/internal/goglfw/internal_windows.go
+++ b/internal/goglfw/internal_windows.go
@@ -7,8 +7,6 @@ package goglfw
 
 import (
 	"unsafe"
-
-	"golang.org/x/sys/windows"
 )
 
 const (
@@ -93,11 +91,7 @@ type context struct {
 	getProcAddress     func(string) uintptr
 	destroy            func(*Window) error
 
-	wgl struct {
-		dc       _HDC
-		handle   _HGLRC
-		interval int
-	}
+	state platformContextState
 }
 
 type (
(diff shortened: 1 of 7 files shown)


TotallyGamerJet:
@hajimehoshi this test failure doesn't look related to this commit?

internal/goglfw/internal_windows.go
```diff
@@ -93,11 +91,7 @@ type context struct {
 	getProcAddress     func(string) uintptr
 	destroy            func(*Window) error
 
-	wgl struct {
-		dc       _HDC
-		handle   _HGLRC
-		interval int
-	}
+	state platformContextState
```
hajimehoshi, line 94: `platform` sounds better than `state` as this represents a platform-specific state. What do you think?

internal/goglfw/internal_windows.go
```diff
@@ -253,42 +212,8 @@ type library struct {
...
-		ARB_pixel_format               bool
-		ARB_create_context             bool
-		ARB_create_context_profile     bool
-		EXT_create_context_es2_profile bool
-		ARB_create_context_robustness  bool
-		ARB_create_context_no_error    bool
-		ARB_context_flush_control      bool
-	}
+	windowState  platformLibraryWindowState
+	contextState platformLibraryContextState
```
hajimehoshi, line 216: `platformWindow` and `platformContext` (if we replace `state` with `platform`)

hajimehoshi:
> @hajimehoshi this test failure doesn't look related to this commit?

I believe so.

internal/goglfw/internal_windows.go
TotallyGamerJet: The thing is a state though which just happens to run on a specific platform. I think state is more clear to what it is while platform describes how is ran. I prefer state but if you want platform that's fine by me.

hajimehoshi: Hmm, IMO `state` doesn't represent anything since most member variables are 'states' in general... So, I would like to go with `platform`

hajimehoshi (approved):
LGTM, thanks!

PR #2587, opened 2023-03-03
gamepad: native standard layout for Linux

divVerent:
# What issue is this addressing?
Closes #2052

## What _type_ of issue is this addressing?
feature

## What this PR does | solves
Implements native standard layout for Linux gamepads by using the kernel-provided button IDs, thereby expanding support to gamepads not listed in `gamecontrollerdb.txt`.

Requires https://github.com/hajimehoshi/ebiten/pull/2588

Linux's docs: https://www.kernel.org/doc/Documentation/input/gamepad.txt

SDL2's source: https://fossies.org/linux/SDL2/src/joystick/linux/SDL_sysjoystick.c#l_1740

Note that I am NOT 100% convinced about the X/Y swap between Xbox and PlayStation controllers - the Xbox compatible pad I have however does have `BTN_NORTH` and `BTN_WEST` swapped (and thus `BTN_X` and `BTN_Y` assigned right), which confirms SDL's logic and opposes the kernel docs.

Tested with this gamepad: "20d6:2802 BDA Xbox ONE Core controller", label says "PowerA Model 1508491-02" - even after clearing out `gamecontrollerdb.txt`, `examples/gamepad` shows a 100% correct mapping.

Issue #2052, opened 2022-04-04
[linux] support gamepad standard layout without mapping when mapping is provided by kernel


divVerent:
This Xbox gamepad has the following buttons and axes:

```
#     Event code 304 (BTN_SOUTH) = A
#     Event code 305 (BTN_EAST) = B
#     Event code 307 (BTN_NORTH) = Y
#     Event code 308 (BTN_WEST) = X
#     Event code 310 (BTN_TL) = left shoulder button
#     Event code 311 (BTN_TR) = right shoulder button
#     Event code 314 (BTN_SELECT) = left center button
#     Event code 315 (BTN_START) = right center button
#     Event code 316 (BTN_MODE) = middle center button
#     Event code 317 (BTN_THUMBL) = pushing down on left joystick
#     Event code 318 (BTN_THUMBR) = pushing down on right joystick
#     Event code 0 (ABS_X) = left joystick
#     Event code 1 (ABS_Y) = left joystick
#     Event code 2 (ABS_Z) = left trigger
#     Event code 3 (ABS_RX) = right joystick
#     Event code 4 (ABS_RY) = right joystick
#     Event code 5 (ABS_RZ) = right trigger
#     Event code 16 (ABS_HAT0X) = gamepad hat
#     Event code 17 (ABS_HAT0Y) = gamepad hat
```

This smells a lot like a mapping to standard layout - meaning this should be supportable even without a gamecontrollerdb mapping, which would make new devices instantly supported.

divVerent:
To be 100% clear - if both these buttons and axes exist AND a gamepaddb entry exists, Ebiten should prefer the gamepaddb entry. After all, there may be gamepads where this is assigned wrong by the manufacturer.

However, if there is no gamepaddb entry, and a "critical mass" of these is defined (it seems like ABS_X and ABS_Y also have other uses, e.g. for touchpads), Ebiten should offer standard layout.

As for what that "critical mass" could be: I suggest ANY of the BTN_* codes here is enough, and if there are no BTN_* codes above, existence of ABS_HAT0X and ABS_HAT0Y is a strong indicator for a gamepad-ish device too. The other ABS_ axes are also used for non-gamepads, which is kinda sad as this means one can't really autodetect a good old PC joystick and map it to standard layout.

[...]

PR #2588, opened 2023-03-03
gamepad: enhanced native standard layout support

divVerent:
Updates #2052

feature

## What this PR does | solves
Refactors native standard layout so standard axes and buttons can be implemented using any of axes, buttons or hats.

This is required to be able to implement a native standard layout mapping for Linux, as e.g. shoulder buttons can be backed either by an analog or digital input.

Precedes https://github.com/hajimehoshi/ebiten/pull/2587

### DIFF
--- a/internal/gamepad/gamepad.go
+++ b/internal/gamepad/gamepad.go
@@ -187,11 +187,59 @@ type Gamepad struct {
 	native nativeGamepad
 }
 
+type mappingInput interface {
+	Pressed() bool
+	Value() float64 // Normalized to range: 0..1.
+}
+
+type axisMappingInput struct {
+	g    nativeGamepad
+	axis int
+}
+
+func (a axisMappingInput) Pressed() bool {
+	return a.g.axisValue(a.axis) > gamepaddb.ButtonPressedThreshold
+}
+
+func (a axisMappingInput) Value() float64 {
+	return a.g.axisValue(a.axis)*0.5 + 0.5
+}
+
+type buttonMappingInput struct {
+	g      nativeGamepad
+	button int
+}
+
+func (b buttonMappingInput) Pressed() bool {
+	return b.g.isButtonPressed(b.button)
+}
+
+func (b buttonMappingInput) Value() float64 {
+	return b.g.buttonValue(b.button)
+}
+
+type hatMappingInput struct {
+	g         nativeGamepad
+	hat       int
+	direction int
+}
+
+func (h hatMappingInput) Pressed() bool {
+	return h.g.hatState(h.hat)&h.direction != 0
+}
+
+func (h hatMappingInput) Value() float64 {
+	if h.Pressed() {
+		return 1
+	}
+	return 0
+}
+
 type nativeGamepad interface {
 	update(gamepads *gamepads) error
 	hasOwnStandardLayoutMapping() bool
-	isStandardAxisAvailableInOwnMapping(axis gamepaddb.StandardAxis) bool
-	isStandardButtonAvailableInOwnMapping(button gamepaddb.StandardButton) bool
+	standardAxisInOwnMapping(axis gamepaddb.StandardAxis) mappingInput
+	standardButtonInOwnMapping(button gamepaddb.StandardButton) mappingInput
 	axisCount() int
 	buttonCount() int
 	hatCount() int
@@ -291,7 +339,7 @@ func (g *Gamepad) IsStandardAxisAvailable(axis gamepaddb.StandardAxis) bool {
 	if gamepaddb.HasStandardLayoutMapping(g.sdlID) {
 		return gamepaddb.HasStandardAxis(g.sdlID, axis)
 	}
-	return g.native.isStandardAxisAvailableInOwnMapping(axis)
+	return g.native.standardAxisInOwnMapping(axis) != nil
 }
 
 // IsStandardButtonAvailable is concurrent safe.
@@ -302,7 +350,7 @@ func (g *Gamepad) IsStandardButtonAvailable(button gamepaddb.StandardButton) boo
 	if gamepaddb.HasStandardLayoutMapping(g.sdlID) {
 		return gamepaddb.HasStandardButton(g.sdlID, button)
 	}
-	return g.native.isStandardButtonAvailableInOwnMapping(button)
+	return g.native.standardButtonInOwnMapping(button) != nil
 }
 
 // StandardAxisValue is concurrent-safe.
@@ -310,8 +358,8 @@ func (g *Gamepad) StandardAxisValue(axis gamepaddb.StandardAxis) float64 {
 	if gamepaddb.HasStandardLayoutMapping(g.sdlID) {
 		return gamepaddb.AxisValue(g.sdlID, axis, g)
 	}
-	if g.native.hasOwnStandardLayoutMapping() {
-		return g.native.axisValue(int(axis))
+	if m := g.native.standardAxisInOwnMapping(axis); m != nil {
+		return m.Value()*2 - 1
 	}
 	return 0
 }
@@ -321,8 +369,8 @@ func (g *Gamepad) StandardButtonValue(button gamepaddb.StandardButton) float64 {
 	if gamepaddb.HasStandardLayoutMapping(g.sdlID) {
 		return gamepaddb.ButtonValue(g.sdlID, button, g)
 	}
-	if g.native.hasOwnStandardLayoutMapping() {
-		return g.native.buttonValue(int(button))
+	if m := g.native.standardButtonInOwnMapping(button); m != nil {
+		return m.Value()
 	}
 	return 0
 }
@@ -332,8 +380,8 @@ func (g *Gamepad) IsStandardButtonPressed(button gamepaddb.StandardButton) bool 
 	if gamepaddb.HasStandardLayoutMapping(g.sdlID) {
 		return gamepaddb.IsButtonPressed(g.sdlID, button, g)
 	}
-	if g.native.hasOwnStandardLayoutMapping() {
-		return g.native.isButtonPressed(int(button))
+	if m := g.native.standardButtonInOwnMapping(button); m != nil {
+		return m.Pressed()
 	}
 	return false
 }
--- a/internal/gamepad/gamepad_xbox_windows.go
+++ b/internal/gamepad/gamepad_xbox_windows.go
@@ -157,25 +157,27 @@ func (n *nativeGamepadXbox) hasOwnStandardLayoutMapping() bool {
 	return true
 }
 
-func (n *nativeGamepadXbox) isStandardAxisAvailableInOwnMapping(axis gamepaddb.StandardAxis) bool {
+func (n *nativeGamepadXbox) standardAxisInOwnMapping(axis gamepaddb.StandardAxis) mappingInput {
 	switch axis {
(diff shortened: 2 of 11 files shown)


internal/gamepad/gamepad.go
```diff
@@ -187,11 +187,30 @@ type Gamepad struct {
...
+const (
+	mappingKindNone mappingKind = iota
+	mappingKindAxis
+	mappingKindButtonAnalog
+	mappingKindButtonDigital
+	mappingKindButtonMixed
+	mappingKindHatUp
+	mappingKindHatRight
+	mappingKindHatDown
+	mappingKindHatLeft
```
hajimehoshi, line 201: For a standard map, there are only axes and buttons. Do we need some types of buttons? Also, do we need hats here?

divVerent: See https://github.com/hajimehoshi/ebiten/pull/2587 - the mappingKind is what the _source_ is.

E.g. shoulder buttons can be represented by any of an axis, hat or button.

divVerent: Similarly, the split of analog/digital source buttons is needed to support as Linux only supports digital buttons but some other platforms (and Ebitengine's own API) support analog ones.

divVerent: (Could ditch THAT split though by deprecating internal isButtonPressed APIs and only using buttonValue - which I didn't do, as on some platforms they both go through different OS APIs, letting the OS implement its own threshold)

hajimehoshi: So this is for Linux and is an implementation details.

I think we should return an interface (or a struct with methods) to hide the logic, rather than exposing some various key types.

divVerent: Not just Linux - every implementation has its own issues there. I do not think it makes sense to have each OS's implementation define its own code for this, as all standard layout items should really be backed by a native button, axis or hat.

I could however change the enum into a globally defined interface + globally defined structs that are based on a native button, axis or hat.

divVerent: Also, it _seems_ like mappingKindButtonAnalog is never needed - as long as all OSes that implement buttonValue also implement isButtonPressed, I can ditch that one.

divVerent: Similarly, if you are OK with having all OSes that implement isButtonPressed also implement buttonValue (even if that only ever returns 0.0 or 1.0), we can move to only one button type.

divVerent: To elaborate - I think it'd be a good policy decision that any standard layout item MUST be backed by a native button, axis or hat, so that the StandardGamepad* APIs can't return anything that GamepadAxisValue/IsGamepadButtonPressed cannot return as well. That is why I really would like to require that this is not an OS-specific interface but there's really a requirement that the input shall be reachable by existing APIs.

hajimehoshi: > Not just Linux - every implementation has its own issues there. I do not think it makes sense to have each OS's implementation define its own code for this, as all standard layout items should really be backed by a native button, axis or hat.
> I could however change the enum into a globally defined interface + globally defined structs that are based on a native button, axis or hat.

Hmm I see. As long as the interface has the notion of hats, we can go with your PR. We can revisit this (how to refactor this) later.

> Also, it seems like mappingKindButtonAnalog is never needed - as long as all OSes that implement buttonValue also implement isButtonPressed, I can ditch that one.
> Similarly, if you are OK with having all OSes that implement isButtonPressed also implement buttonValue (even if that only ever returns 0.0 or 1.0), we can move to only one button type.

I think I'm fine to do this. Let's remove isButtonPressed, which might be redundant.

hajimehoshi: Isn't it inevitable to have types for all four directions for hats?

hajimehoshi: > Hmm I see. As long as the nativeGamepad interface has the notion of hats, we can go with your PR. We can revisit this (how to refactor this) later.

If you have already done with interfaces, that's also fine.

divVerent: Did the global interface way now (which also unified hats into one type). I am not sure about isButtonPressed, though.

gamepad_js.go, gamepad_nintendosdk.go and  implements buttonValue and isButtonPressed separately. As I do not have access to the Nintendo SDK, I can't judge whether that is important.

internal/gamepad/gamepad.go
```diff
@@ -187,11 +187,76 @@ type Gamepad struct {
...
+	Value() float64 // Normalized to range: 0..1.
+}
+
+type axisMappingInput struct {
+	g    nativeGamepad
+	axis int
+}
+
+func (a axisMappingInput) Pressed() bool {
+	const axisToButtonThreshold = 30 / 255.0
```
hajimehoshi, line 201: Let's integrate `threshold` in `gamepad_xbox_windows.go`

internal/gamepad/gamepad.go
hajimehoshi: > Did the global interface way now (which also unified hats into one type). I am not sure about isButtonPressed, though.
> gamepad_js.go, gamepad_nintendosdk.go implement buttonValue and isButtonPressed separately. As I do not have access to the Nintendo SDK, I can't judge whether that is important.

Ah I see. As at least *_js has a separate API for is-pressed and value, let's keep the is-pressed part. Sorry for going back and forth.

divVerent: For gamepad_js, I'd slightly tend towards keeping the separation, as that means the thresholding is implemented in the web browser. Although truth is, browsers are most likely rather dumb about this...

So I guess the decision boils down to gamepad_nintendosdk. If anyone may have better thresholding than just a simple > x comparison (e.g. dynamic thresholding based on the gamepad's specs or past performance), then that'd be Nintendo.

divVerent: Actually, let's do this: I'll ONLY keep mixed buttons, and rather replace all the panics in .buttonValue by if-pressed-then-1-else-0 logic.

divVerent: Done that now. All the buttonValue panics are gone, so we no longer need more than one button input type.

hajimehoshi: FYI: For NintendoSwitch, there are no analog button and a button value only takes 0 and 1

internal/gamepad/gamepad.go
divVerent: Done.

internal/gamepad/gamepad_xbox_windows.go
```diff
@@ -225,12 +227,11 @@ func (n *nativeGamepadXbox) isButtonPressed(button int) bool {
 	// Use XInput's trigger dead zone.
 	// See https://source.chromium.org/chromium/chromium/src/+/main:device/gamepad/public/cpp/gamepad.h;l=22-23;drc=6997f8a177359bb99598988ed5e900841984d242
 	// TODO: Integrate this value with the same one in the package gamepaddb.
```
hajimehoshi, line 229: Move this comment to an appropriate place.

divVerent:
Note that I had decided to only represent analog inputs in 0..1 range in the interface - which means that for an axis-to-axis mapping, we first map -1..1 to 0..1, and then back to -1..1.

Not TOTALLY nice, but do note that this loses at most one bit of accuracy as the multiplications by 0.5 or 2 are just exponent inc/dec.

internal/gamepad/gamepad.go
```diff
@@ -302,40 +352,43 @@ func (g *Gamepad) IsStandardButtonAvailable(button gamepaddb.StandardButton) boo
...
 		return gamepaddb.ButtonValue(g.sdlID, button, g)
 	}
-	if g.native.hasOwnStandardLayoutMapping() {
-		return g.native.buttonValue(int(button))
+	m := g.native.standardButtonInOwnMapping(button)
+	if m == nil {
+		return 0
 	}
-	return 0
+	return m.Value()
```
hajimehoshi, line 379: ditto

internal/gamepad/gamepad.go
```diff
@@ -302,40 +352,43 @@ func (g *Gamepad) IsStandardButtonAvailable(button gamepaddb.StandardButton) boo
...
 		return gamepaddb.IsButtonPressed(g.sdlID, button, g)
 	}
-	if g.native.hasOwnStandardLayoutMapping() {
-		return g.native.isButtonPressed(int(button))
+	m := g.native.standardButtonInOwnMapping(button)
+	if m == nil {
+		return false
 	}
-	return false
+	return m.Pressed()
```
hajimehoshi, line 391: ditto

internal/gamepad/gamepad_xbox_windows.go
divVerent: Done, and also did that integrating.

internal/gamepad/gamepad.go
```diff
@@ -302,40 +352,43 @@ func (g *Gamepad) IsStandardButtonAvailable(button gamepaddb.StandardButton) boo
...
 		return gamepaddb.AxisValue(g.sdlID, axis, g)
 	}
-	if g.native.hasOwnStandardLayoutMapping() {
-		return g.native.axisValue(int(axis))
+	m := g.native.standardAxisInOwnMapping(axis)
+	if m == nil {
+		return 0
 	}
-	return 0
+	return m.Value()*2 - 1
```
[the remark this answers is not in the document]
divVerent: Sure, if you prefer so - normally I am used to indenting the error flow for consistency (is also part of Google's internal style guide).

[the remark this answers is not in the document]
hajimehoshi: Just out of curiosity, does Google prefer early-returns with errors, rather than just early returns?

I prefer narrower scopes of variables in general. Both cases are not errors, and I think it is natural that the last item indicates a 'fallback'.

hajimehoshi:
Please fix the test failures.

divVerent:
> Please fix the test failures.

Sure, am on it - have to wait for a build for each iteration, as I don't have many of these platforms easily available locally.

internal/gamepaddb/gamepaddb.go
```diff
@@ -519,11 +519,12 @@ func buttonValue(id string, button StandardButton, state GamepadState) float64 {
 	return 0
 }
 
-func IsButtonPressed(id string, button StandardButton, state GamepadState) bool {
-	// Use XInput's trigger dead zone.
-	// See https://source.chromium.org/chromium/chromium/src/+/main:device/gamepad/public/cpp/gamepad.h;l=22-23;drc=6997f8a177359bb99598988ed5e900841984d242
-	const threshold = 30.0 / 255.0
+// Use XInput's trigger dead zone.
```
hajimehoshi, line 522: Start the comment with `ButtonPressedThreshold represents ...` for the Go convention.

hajimehoshi (commented):
The implementation seems very good. Let's wait for the test results.

hajimehoshi:
> If you want, I can change this around and make the range of Value() always -1..1 - but that makes the threshold comparisons more ugly.

> It would gain 1 bit in float64 accuracy, which I strongly doubt matters for gamepads ;)

I don't think we should do this. As you said, we should not have to care the 1bit accuracy.

hajimehoshi (approved):
LGTM, thanks! (let's wait for the test results)

PR #2596, opened 2023-03-11
android: fix Xbox-ish gamepad axis assignment

### LINKED ISSUE #2557: Support standard layout on Android
### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [X] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] Xbox
- [ ] Web Browsers

IsStandardGamepadLayoutSupported should be implemented for Android

### Why is this needed?

SDL_GameControllerDB is far from complete for Android, and also Android already provides its own mostly correct mappings. Buttons and axes already are named similar to Linux standard layout ones.

This issue is similar to https://github.com/hajimehoshi/ebiten/issues/2052 in that the solution is very similar. It appears like Android even uses the same axis/button constant naming as the Linux kernel itself, even though the documentation seems to swap ABS_RX/ABS_Z and ABS_RY/ABS_RZ between the two (but based on my sample size of ONE game pad, it's not actually swapped in the implementation).

divVerent:
Closes #2557

bug

## What this PR does | solves
Fixes wrong automatic gamepad mapping for Xbox-ish game controllers. The main visible symptom would be axis flips (X/Y) and the right stick controlling trigger or triggers controlling right stick movement.

This commit mirrors https://github.com/libsdl-org/SDL/pull/7405 (https://github.com/libsdl-org/SDL/commit/6f1f586086a18ab32ddccc05e4616aac02774a1b).

Note that for SDL, this code is used a lot less than for Ebitengine, as SDL mostly migrated to HIDAPI and direct USB device access rather than using Android's APIs. For Bluetooth devices, however, the Java APIs are used the same way.

This was the remaining problem to be solved to automatically support standard layout on most standard gamepads (this should cover most Xbox-ish and PS-ish gamepads on the market).

In particular this covers gamepads with the following assignment:

- Left stick = X/Y, right stick = Z/RZ, triggers = LEFTTRIGGER/RIGHTTRIGGER (which basically is what Android docs say and some PS gamepads do)

- Left stick = X/Y, right stick = RX/RY, triggers = Z/RZ (Xbox gamepad style, apparently)

- Left stick = X/Y, right stick = RX/RY, triggers = LEFTTRIGGER/RIGHTTRIGGER (Not sure if this exists, but it's conceivable)

As we found on the SDL pull request discussion, gamepads that offer flight controls (e.g. THROTTLE and RUDDER) will likely not work well, before and after this change.

NOTE: this change is slightly incompatible - a few SDL_GameControllerDB mappings became invalid (or even unnecessary) with this. SDL's built-in mappings have been updated in https://github.com/libsdl-org/SDL/commit/6f1f586086a18ab32ddccc05e4616aac02774a1b, but the SDL_GameControllerDB project has not been updated yet.


cmd/ebitenmobile/_files/EbitenView.java
```diff
@@ -58,6 +58,33 @@ public int compare(InputDevice.MotionRange arg0, InputDevice.MotionRange arg1) {
...
+            // Gamepads that don't have AXIS_Z/AXIS_RZ but use
+            // AXIS_LTRIGGER/AXIS_RTRIGGER are unaffected by this.
+            //
+            // References:
+            // - https://developer.android.com/develop/ui/views/touch-and-input/game-controllers/controller-input
+            // - https://www.kernel.org/doc/html/latest/input/gamepad.html
+            if (arg0Axis == MotionEvent.AXIS_Z) {
+                arg0Axis = MotionEvent.AXIS_RZ - 1;
+            } else if (arg0Axis > MotionEvent.AXIS_Z && arg0Axis < MotionEvent.AXIS_RZ) {
+                --arg0Axis;
```
hajimehoshi, line 80: I slightly prefer `arg0Axis--`

cmd/ebitenmobile/_files/EbitenView.java
```diff
@@ -58,6 +58,33 @@ public int compare(InputDevice.MotionRange arg0, InputDevice.MotionRange arg1) {
...
+            }
+            if (arg1Axis == MotionEvent.AXIS_Z) {
+                arg1Axis = MotionEvent.AXIS_RZ - 1;
+            } else if (arg1Axis > MotionEvent.AXIS_Z && arg1Axis < MotionEvent.AXIS_RZ) {
+                --arg1Axis;
```
hajimehoshi, line 85: ditto

hajimehoshi (commented):
Thanks!

cmd/ebitenmobile/_files/EbitenView.java
```diff
@@ -58,6 +58,33 @@ public int compare(InputDevice.MotionRange arg0, InputDevice.MotionRange arg1) {
             } else if (arg1Axis == MotionEvent.AXIS_BRAKE) {
                 arg1Axis = MotionEvent.AXIS_GAS;
             }
+
+            // Make sure the AXIS_Z is sorted between AXIS_RY and AXIS_RZ.
+            // This is because the usual pairing are:
+            // - AXIS_X + AXIS_Y (left stick).
+            // - AXIS_RX, AXIS_RY (sometimes the right stick, sometimes triggers).
+            // - AXIS_Z, AXIS_RZ (sometimes the right stick, sometimes triggers).
```
hajimehoshi, line 66: Let's comment the original order.

divVerent: Done.

cmd/ebitenmobile/_files/EbitenView.java
divVerent: Done.

cmd/ebitenmobile/_files/EbitenView.java
divVerent: Done.

divVerent:
Note: if you want, we can wait with this commit until SDL_GameControllerDB is updated for it. IIRC there are eight game pads in there that will need adjustment.

The impacted ones are rather easy to recognize by having rightx:a2,lefttrigger:a3,righttrigger:a4,righty:a5 or lefttrigger:a2,rightx:a3,righty:a4,righttrigger:a5 and need to be remapped in a matching way: a2 becomes a4, a4 becomes a3 and a3 becomes a2.

hajimehoshi:
Have you committed your change?

divVerent:
Oops, there it is. Had forgotten the push part.

hajimehoshi:
> Note: if you want, we can wait with this commit until SDL_GameControllerDB is updated for it. IIRC there are eight game pads in there that will need adjustment.

Sure, let's wait for the database adjusted.

divVerent:
Sounds good; if that takes too long I can always add code into Ebitengine to skip the affected entries (as the auto mapping should be rather good for those already).

divVerent:
Sent https://github.com/gabomdq/SDL_GameControllerDB/pull/671 - note that it might be smart for them to rather _remove_ impacted mappings, and thus we'd have to add them into Ebitengine as an additional mapping file (or mark them in a way so SDL will skip them but Ebitengine will use them).

divVerent:
I.e. it is possible that the SDL_GameControllerDB PR will have to be revised to rather comment out or remove those mappings, and I can then add them into Ebitengine similar to additionalGLFWGamepads.

hajimehoshi:
> I.e. it is possible that the SDL_GameControllerDB PR will have to be revised to rather comment out or remove those mappings, and I can then add them into Ebitengine similar to additionalGLFWGamepads.

This might make sense, but I hope your change to the DB is merged soon. Let's wait for a few days.

If the change is merged, please update the database in this PR too by executing:

```
curl --location --remote-name https://raw.githubusercontent.com/gabomdq/SDL_GameControllerDB/master/gamecontrollerdb.txt
```

divVerent:
Ultimately depends on how high they judge the impact of possibly breaking those three pads in released SDL versions (i.e. having them only work in SDL git versions) - especially as it seems like they're working on SDL3, so it's possible my change will not make it into SDL2 at all.

hajimehoshi:
I see. I hope they will explicitly respond whether they will accept your PR or refuse it soon.

divVerent:
We will likely end up with another SDL change:
https://github.com/libsdl-org/SDL/pull/7444

which I will then also port into this PR.

divVerent:
SDL change is in - now waiting for SDL_GameControllerDB change - once that's in, I can finish this PR by curling its current version in.

divVerent:
Note that the new code is snake_case, which is unusual for Java, but I had done it this way to be consistent with SDL's nearby code (e.g. button_mask). If you want, I can of course convert it to lowerCamelCase. Not sure what is more important here - closeness with SDL, or standard Java syntax.

divVerent:
Tried AAAAXY with this change and all buttons/axes re-enabled (in internal/input/gamepad_ignored_buttons_android.go). Only one minor issue: the triggers both appear as activated right at start (probably because they start out at a buttonValue of 0.5, not 0, which is above the usual 30/255 threshold). This bug is shared with SDL's gamepad demo.

Still means I can safely unignore the right stick, as triggers appearing as the right stick on Android is rather unlikely after these changes - already great!

divVerent:
Added code to my game to simply ignore any gamepad axis or button that's never been in center or non-pressed state since it was connected. This matches event based processing of keyboard, thus seems reasonable, and works around many mapping issues as long as the item in question is at least never touched.

cmd/ebitenmobile/_files/EbitenView.java
```diff
@@ -329,11 +363,19 @@ private static int getAxisMask(InputDevice joystickDevice) {
         final int SDL_CONTROLLER_AXIS_TRIGGERRIGHT = 5;
 
         int naxes = 0;
+        boolean have_z = false;
+        boolean have_past_z_before_rz = false;
```
hajimehoshi, line 367: Can we use `fooBar` instead of `foo_bar` for consistency?

hajimehoshi:
> Tried AAAAXY with this change and all buttons/axes re-enabled (in internal/input/gamepad_ignored_buttons_android.go). Only one minor issue: the triggers both appear as activated right at start (probably because they start out at a buttonValue of 0.5, not 0, which is above the usual 30/255 threshold). This bug is shared with SDL's gamepad demo.

Do you plan to fix the issue in this PR, or next PR, or just ignore this?

divVerent:
> > Tried AAAAXY with this change and all buttons/axes re-enabled (in internal/input/gamepad_ignored_buttons_android.go). Only one minor issue: the triggers both appear as activated right at start (probably because they start out at a buttonValue of 0.5, not 0, which is above the usual 30/255 threshold). This bug is shared with SDL's gamepad demo.
> 
> Do you plan to fix the issue in this PR, or next PR, or just ignore this?

For now ignore, until I know the real root cause.

I did observe some SDL mappings map triggers not to "a4" but to "+a4", i.e. the positive half of the range, probably specifically to counter that issue, as it effectively moves the threshold above half the axis range. But this isn't something I would want to apply to existing game and gamepad combinations, as trigger threshold can change the feel of a game a lot.

Another approach could be doing the same hack as in my game in Ebitengine itself - i.e. ignore any gamepad control until it's been released at least once. But as that means the first frame of a game can't have actions anymore, I'd be worried it would be breaking for some games too (or slowing down some speedruns by one frame) - and of course, automated tests can easily be impacted by such a change.

And of course, it might even be a bug that can be fixed. I'll have to do further debugging to see if it's even a bug in Ebitengine or in Android (like, what if it's actually Android reporting them as 50% pressed until receiving the first event)? If it's a bug in Ebitengine, I'll sure want to fix it in a separate PR, but the issue being shared with SDL makes this somewhat unlikely.

What speaks for this being actually an Android bug is that if I press and release the left trigger, the right trigger stays in pressed state, and it looks like Ebitengine's code just polls _all_ axes on any GenericMotionEvent. If it were a pure Ebitengine bug, then pressing and releasing the left trigger should fix the right trigger.

cmd/ebitenmobile/_files/EbitenView.java
divVerent: Done.

cmd/ebitenmobile/_files/EbitenView.java
```diff
@@ -329,11 +363,19 @@ private static int getAxisMask(InputDevice joystickDevice) {
...
+        boolean haveZ = false;
+        boolean havePastZBeforeRZ = false;
         for (InputDevice.MotionRange range : joystickDevice.getMotionRanges()) {
             if ((range.getSource() & InputDevice.SOURCE_CLASS_JOYSTICK) != 0) {
-                if (range.getAxis() != MotionEvent.AXIS_HAT_X && range.getAxis() != MotionEvent.AXIS_HAT_Y) {
+                int axis = range.getAxis();
+                if (axis != MotionEvent.AXIS_HAT_X && axis != MotionEvent.AXIS_HAT_Y) {
                     naxes++;
                 }
+                if (axis == MotionEvent.AXIS_Z) {
```
divVerent, line 374: Note: this code depends on AXIS value ordering (in particular on AXIS_HAT_X and AXIS_HAT_Y not lying between AXIS_Z and AXIS_RZ). SDL's code implicitly has this in the above != block, one indent deeper.

If you want me to move it inside the if() block above that increments naxes, I'll do that - however I find it more readable without likely excessive indent, as I find it unlikely these axis values will ever change.

hajimehoshi: I prefer less indentations so let's keep this as it is.

hajimehoshi:
> For now ignore, until I know the real root cause.

Sure. Let's revisit that issue later.

cmd/ebitenmobile/_files/EbitenView.java
divVerent: Let's keep it as is rather - that way adding something below won't get confusing.

Of course the two blocks are mutually exclusive anyway, even without `continue` or `else` - I wrote them this way to show a form of logical grouping, that's all.

hajimehoshi: Acknowledged.

cmd/ebitenmobile/_files/EbitenView.java
```diff
@@ -348,6 +390,11 @@ private static int getAxisMask(InputDevice joystickDevice) {
         if (naxes >= 6) {
             axisMask |= ((1 << SDL_CONTROLLER_AXIS_TRIGGERLEFT) | (1 << SDL_CONTROLLER_AXIS_TRIGGERRIGHT));
         }
+        // Also add an indicator bit for whether the sorting order has changed.
+        // This serves to disable outdated gamecontrollerdb.txt mappings.
+        if (haveZ && havePastZBeforeRZ) {
+            axisMask |= 0x8000;
```
hajimehoshi, line 396: This is also a magic number in the SDL source, right?

hajimehoshi (approved):
LGTM, thanks!

divVerent:
> > For now ignore, until I know the real root cause.
> 
> Sure. Let's revisit that issue later.

Turns out Ebitengine already has a hack against this:

```
        // Initialize the trigger axes values explicitly, or the initial button values would be 0.5 instead of 0.
        if (gamepad.axes.size() >= 6) {                                          
            Ebitenmobileview.onGamepadAxisChanged(deviceId, 4, -1);              
            Ebitenmobileview.onGamepadAxisChanged(deviceId, 5, -1);              
        }                                                                        
```

It doesn't work though, as the first GenericMotionEvent - even involving other axes - will reset them to zero, until they get moved themselves. But even worse, this means the hack in my game won't always work either (but it happens to work right now, but now I have no idea why), as this hack here will make them non-pressed in the first frame. That breaks my workaround as well, so I am reverting that for now.

I guess one approach could be not polling _all_ axes, but just the ones that changed - but that's still dirty, as there's no guarantee that axes 4 and 5 are actually the triggers (this is only the case if gamecontrollerdb does no further mapping). Also right now I don't see how GenericMotionEvent even tells one which axes changed and which did not.

divVerent:
Moved that discussion to https://github.com/hajimehoshi/ebiten/issues/2598

cmd/ebitenmobile/_files/EbitenView.java
divVerent: Yes; it only exists in the Java code actually, as it's only used for looking up SDL game controller DB mappings, and C code doesn't use it in any way.

It might get a name in the future, but I doubt it.

PR #2621, opened 2023-04-01
WIP: Port GLFW to Go

TotallyGamerJet:
This is a draft for porting glfw to Go. It is just for reference because the code is extremely messy as quite a few changes have happened to ebitengine and purego since it was started. Work will be moving to @hajimehoshi's branch. The only example tested that works is flappy with metal on amd64 (see example). Also, there is some C because purego cannot return a struct type yet. So I had to resort to using C calls for the few functions that return structs.
flappy


TotallyGamerJet:
Closing since this branch is now really far from main. Especially since internal/glfw and internal/goglfw have merged.

PR #2627, opened 2023-04-06
removeDuplicatedRegions: change from O((n+m)^2) to O(n*m + m^2).

### LINKED ISSUE #2626: Android: removeDuplicatedRegions is very slow
### Ebitengine Version

2.5.0 (also occurs in main)

### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [X] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] Xbox
- [ ] Web Browsers

### Go Version (`go version`)

go version go1.19.5 linux/amd64

Run https://github.com/divVerent/aaaaxy with the `profile-the-android-loading-screen` branch.

Download the pprof file (attached as initial.svg).

Loading should be a wild mix of PNG/Vorbis decoding, font rendering and texture uploading.

### What happens instead?

Observe that most loading screen time is in removeDuplicatedRegions.

Going to provide a fix.

divVerent:
Closes #2626

bug

## What this PR does | solves

This speeds up AAAAXY loading on a Moto G7 Play from 52.27 seconds to 8.06 seconds.

The usual case of removeDuplicatedRegions is number of previous rectangles being big and number of new rectangles being small. As the function is always called after adding more rectangles, one can assume that the set of pre-existing rectangles has no rectangles containing other rectangles, and thus that one only needs to search for duplicates when one of the two rectangles is a newly added one.

When adding m rectangles to an already existing list of n rectangles, this PR changes the complexity from O((n+m)^2) to O(m * (n + m)) by assuming that the already existing rectangles in the list have no duplicates.

In particular, when adding a small amount of rectangles to a large amount, this changes complexity from quadratic to linear.

That seems especially to be the case for the call from makeStale which is called when uploading textures, which happens at AAAAXY's loading screen for image and font loading/precaching.

Tested on my Moto G7 Play.

Loading screen profile before:

initial

Loading screen profile after:

fixed

### DIFF
--- a/internal/restorable/image.go
+++ b/internal/restorable/image.go
@@ -257,7 +257,7 @@ func (i *Image) makeStale(rect image.Rectangle) {
 	}
 
 	// Remove duplicated regions to avoid unnecessary reading pixels from GPU.
-	n := removeDuplicatedRegions(i.staleRegions)
+	n := removeDuplicatedRegions(i.staleRegions, origNum)
 	i.staleRegions = i.staleRegions[:n]
 
 	// Don't have to call makeStale recursively here.
@@ -483,11 +483,12 @@ func (i *Image) readPixelsFromGPU(graphicsDriver graphicsdriver.Graphics) error 
 	if i.stale {
 		rs = i.staleRegions
 	} else {
+		origNum := len(i.regionsCache)
 		i.regionsCache = i.appendRegionsForDrawTriangles(i.regionsCache)
 		defer func() {
 			i.regionsCache = i.regionsCache[:0]
 		}()
-		n := removeDuplicatedRegions(i.regionsCache)
+		n := removeDuplicatedRegions(i.regionsCache, origNum)
 		rs = i.regionsCache[:n]
 	}
 
@@ -625,11 +626,12 @@ func (i *Image) restore(graphicsDriver graphicsdriver.Graphics) error {
 
 	// In order to clear the draw-triangles history, read pixels from GPU.
 	if len(i.drawTrianglesHistory) > 0 {
+		origNum := len(i.regionsCache)
 		i.regionsCache = i.appendRegionsForDrawTriangles(i.regionsCache)
 		defer func() {
 			i.regionsCache = i.regionsCache[:0]
 		}()
-		n := removeDuplicatedRegions(i.regionsCache)
+		n := removeDuplicatedRegions(i.regionsCache, origNum)
 		rs := i.regionsCache[:n]
 
 		for _, r := range rs {
@@ -708,7 +710,7 @@ func (i *Image) appendRegionsForDrawTriangles(regions []image.Rectangle) []image
 		regions = append(regions, r)
 	}
 
-	nn := removeDuplicatedRegions(regions[n:])
+	nn := removeDuplicatedRegions(regions[n:], 0)
 	return regions[:n+nn]
 }
 
@@ -722,15 +724,24 @@ func regionToRectangle(region graphicsdriver.Region) image.Rectangle {
 
 // removeDuplicatedRegions removes duplicated regions and returns the new size of the slice.
 // If a region covers other regions, the covered regions are removed.
-func removeDuplicatedRegions(regions []image.Rectangle) int {
+// This assumes for efficiency that no duplicated regions exist in regions[0:alreadyCheckedUntil].
+func removeDuplicatedRegions(regions []image.Rectangle, alreadyCheckedUntil int) int {
+	m := len(regions)
 	for i, r := range regions {
 		if r.Empty() {
 			continue
 		}
-		for j, rr := range regions {
+		jStart := 0
+		if i < alreadyCheckedUntil {
+			// Skip any rectangle pair i, j
+			// where i < alreadyCheckedUntil && j < alreadyCheckedUntil.
+			jStart = alreadyCheckedUntil
+		}
+		for j := jStart; j < m; j++ {
 			if i == j {
 				continue
 			}
+			rr := regions[j]
 			if rr.Empty() {
 				continue
 			}
--- a/internal/restorable/export_test.go
+++ b/internal/restorable/export_test.go
@@ -18,6 +18,6 @@ import (
 	"image"
 )
 
-func RemoveDuplicatedRegions(regions []image.Rectangle) int {
-	return removeDuplicatedRegions(regions)
+func RemoveDuplicatedRegions(regions []image.Rectangle, alreadyCheckedUntil int) int {
+	return removeDuplicatedRegions(regions, alreadyCheckedUntil)
 }
--- a/internal/restorable/rect_test.go
+++ b/internal/restorable/rect_test.go
@@ -37,8 +37,9 @@ func areEqualRectangles(a, b []image.Rectangle) bool {
 
 func TestRemoveDuplicatedRegions(t *testing.T) {
 	cases := []struct {
-		In  []image.Rectangle
-		Out []image.Rectangle
+		In                  []image.Rectangle
+		AlreadyCheckedUntil int
+		Out                 []image.Rectangle
 	}{
 		{
 			In:  nil,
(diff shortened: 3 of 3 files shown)


divVerent:
Should note: the issue in AAAAXY is much smaller when NOT profiling the loading screen, but drawing it normally, as then only one font is loaded per display frame. That seems to reduce, but not entirely remove, the impact of this issue.

hajimehoshi:
Please addrss on the test failures. Thanks,

divVerent:
Sure, fixed; didn't notice them when running "go test" in the root, now I know it doesn't work that way :)

divVerent:
Hm... in AAAAXY with the "real" loading screen (which "flushes" this list after every font loaded by doing a render frame), profiled again, there this function is nowhere near as pathological. With this change this function goes from 0.58s to zero.

Still half a second saved, but nowhere near as big as it is when doing all loading at startup without render frames.

divVerent:
Oh, hadn't repushed it right. Fixed.

divVerent:
Stopwatch results:

* Initial: 90.52s (AAAAXY v1.3.393)
* Load fonts in second frame: 78.86s
* Load each font in its own frame: 19.98s
* Faster font outlines: 18.00s (AAAAXY v1.3.421)
* This fix: 17.22s
* https://github.com/hajimehoshi/ebiten/pull/2629 on top: 14.94s

However, the 78.86s and 90.52s times were that high because of the issue this PR fixes; the time reduction by loading each font in its own frame was accidental, but did indeed help.

internal/restorable/image.go
```diff
@@ -257,7 +257,7 @@ func (i *Image) makeStale(rect image.Rectangle) {
 	}
 
 	// Remove duplicated regions to avoid unnecessary reading pixels from GPU.
-	n := removeDuplicatedRegions(i.staleRegions)
+	n := removeDuplicatedRegions(i.staleRegions, origNum)
```
hajimehoshi, line 260: Would this work instead of modifying `removeDuplicatedRegions`?

```go
n := removeDuplicatedRegions(i.staleRegions[origNum:])
i.staleRegions = i.staleRegions[:len(origNum)+n]
```

hajimehoshi: I realized my suggestion is different from the fix. I'll take a look at your PR tomorrow...

divVerent: Yeah - basically, the idea is that in the entire list, there is no pair of rectangles for which each is contained in the other.

My change turns this into an invariant, i.e. assumes that before adding more, this property is already the case, so we only need to check for these duplicates if EITHER the inner OR the outer rectangle is among the new ones.

internal/restorable/image.go
```diff
@@ -722,15 +724,23 @@ func regionToRectangle(region graphicsdriver.Region) image.Rectangle {
 
 // removeDuplicatedRegions removes duplicated regions and returns the new size of the slice.
 // If a region covers other regions, the covered regions are removed.
-func removeDuplicatedRegions(regions []image.Rectangle) int {
+// This assumes for efficiency that no duplicated regions exist in regions[0:firstAdded].
+func removeDuplicatedRegions(regions []image.Rectangle, firstAdded int) int {
```
hajimehoshi, line 728: I feel like the name `firstAdded` seems confusing. Is there a better name?

hajimehoshi: What about `alreadyCheckedUntil`? (Disclaimer: I am not an English native speaker)

hajimehoshi:
What about doing this duplication check at `appendRegionsForDrawTriangles` like this?

```go
func (i *Image) appendRegionsForDrawTriangles(regions []image.Rectangle) []image.Rectangle {
        origNum := len(regions)
        for _, d := range i.drawTrianglesHistory {
                r := regionToRectangle(d.dstRegion)
                if r.Empty() {
                        continue
                }
                if d-is-not-duplicated-with-other-regions {
                        regions = append(regions, r)
                }
        }
}
```

It's because 1removeDuplicatedRegions` is always called with 1appendRegionsForDrawTriangles` and they can be integrated into one.

internal/restorable/image.go
divVerent: Renamed.

divVerent:
> What about doing this duplication check at appendRegionsForDrawTriangles like this?

That won't actually work - the current code is able to remove pre-existing rectangles if a newly added one covers them (e.g. if you first draw a small triangle, and then a big one whose bounding box fully covers it). Your suggestion would not do that.

Of course, personally I wouldn't care - in case of the performance critical case of AAAAXY, those rectangles never overlap at all, so it would have little performance implications to remove the entire function simply as I never read back the texture in the first place (except when Ebitengine might do this internally to resize or save off the atlas, which only happens a few times).

However it seems like this is meant to optimize the case of multiple DrawTriangles cases that partially overlap, and then a texture readback. In worst case, those triangles could be forming a 3D-ish scene, in which case any weird patterns of overlap could be occurring, and thus it is useful to reduce the amount of data read back.

Having said that, I do cast some doubt on this algorithm - it only detects one rect _fully_ containing another. I can imagine scenes where this will still read back way more than the entire texture, if the existing rectangles are in a really "weird" pattern - although that should actually be rather common in 3D scenes. Personally I probably would rather have made it detect _any_ overlap (even partial), and replace the two affected rectangles by the "range union" of the two - but that's a more complex algorithm (as the union could have overlaps that previously didn't exist), and will read more pixels but with fewer GL calls (which usually is the better tradeoff).

Still, this duplicate removal is better than none, and I wouldn't feel comfortable replacing it by an even weaker algorithm like you're suggesting here, one where whether duplicates are detected depends on the order of adding them.

hajimehoshi:
> That won't actually work - the current code is able to remove pre-existing rectangles if a newly added one covers them (e.g. if you first draw a small triangle, and then a big one whose bounding box fully covers it). Your suggestion would not do that.

What about this?

```go
func (img *Image) appendRegionsForDrawTriangles(regions []image.Rectangle) []image.Rectangle {
        origNum := len(regions)
loop:
        for _, d := range img.drawTrianglesHistory {
                r := regionToRectangle(d.dstRegion)
                if r.Empty() {
                        continue
                }
                for i, rr := range regions {
                        if d.In(r) {
                                continue loop
                        }
                        if r.In(d) {
                                regions[i] = image.Rectangle{}
                        }
                }
                regions = append(regions, r)
        }

        // Remove empty rectangles from `rectangles`
}
```

divVerent:
> > That won't actually work - the current code is able to remove pre-existing rectangles if a newly added one covers them (e.g. if you first draw a small triangle, and then a big one whose bounding box fully covers it). Your suggestion would not do that.
> 
> What about this?
> 
> ```go
> func (img *Image) appendRegionsForDrawTriangles(regions []image.Rectangle) []image.Rectangle {
>         origNum := len(regions)
> loop:
>         for _, d := range img.drawTrianglesHistory {
>                 r := regionToRectangle(d.dstRegion)
>                 if r.Empty() {
>                         continue
>                 }
>                 for i, rr := range regions {
>                         if r.In(rr) {
>                                 continue loop
>                         }
>                         if rr.In(r) {
>                                 regions[i] = image.Rectangle{}
>                         }
>                 }
>                 regions = append(regions, r)
>         }
> 
>         // Remove empty rectangles from `rectangles`
> }
> ```

That will work, but there are more places that add rectangles and this would have to be repeated for every one of those places (e.g. makeStale has another append to staleRegions if rect is not Empty, and why does the call inside appendRegionsForDrawTriangles even exist? It's redundant with what the other function does.

One thing I could offer though would be changing the algorithm to only add one rectangle at a time, i.e. a

```
func appendRegionRemovingDuplicates(regions []image.Rectangle, added image.Rectangle) []image.Rectangle
```

However, there's still danger of O(n^2) behavior with this, as the process of removing regions may cause a need to shift up to n regions around.

This could be fixed though by not preserving the order of regions in the list.

However, I feel this will be a much bigger and more dangerous change, so I'm reluctant to do that in the same PR - I can however commit to creating this change as a followup once this - which should have no behavioral change at all - is in.

divVerent:
Also, I'd feel very tempted that if doing a major change like this anyway, to try making `appendRegionRemovingDuplicates` do the union thing I proposed instead (i.e. not detect rects that fully contain others, but already such that overlap with another in at least one pixel, and replacing the pair by the "union" in both x and y range). This will yield a shorter list of rectangles, but fewer ReadPixels calls, which will often perform better as ReadPixels is a GPU round trip and latency tends to be dominating in performance, not bandwidth.

hajimehoshi:
My concern of this PR is that `alreadyCheckedUntil` exists only just for optimization. This argument is not essential for the algorithm of `removeDuplicatedRegions`.

>  (e.g. makeStale has another append to staleRegions if rect is not Empty, and why does the call inside appendRegionsForDrawTriangles even exist?

This is a good point and this is only one exception that `removeDuplicatedRegions` is called without `appendRegionsForDrawTriangles`. Your suggested `appendRegionRemovingDuplicates` seems nice.

> However, there's still danger of O(n^2) behavior with this, as the process of removing regions may cause a need to shift up to n regions around.
> This could be fixed though by not preserving the order of regions in the list (i.e. removing an element not by shifting all following elements by one left, but by moving the last element into the newly freed up spot).

I see. Reordering elements in `staleRegions` should be OK. The regions are used to keep pixel data at the region and the order should not matter.

> However, I feel this will be a much bigger and more dangerous change, so I'm reluctant to do that in the same PR - I can however commit to creating this change as a followup once this - which should have no behavioral change at all - is in.

What do you think of making a PR with `appendRegionRemovingDuplicates` instead of merging this PR?

> (i.e. not detect rects that fully contain others, but already such that overlap with another in at least one pixel, and replacing the pair by the "union" in both x and y range)

Interesting, but this has a risk to enlarge a rectangle unnecessarily.

divVerent:
Filed as https://github.com/hajimehoshi/ebiten/pull/2631

hajimehoshi:
Will this PR be abandoned for #2631?

divVerent:
> Will this PR be abandoned for #2631?

Yes, once #2631 is merged. I see there's still some minor things to fix there, will do.

divVerent:
Obsoleted by https://github.com/hajimehoshi/ebiten/pull/2631

PR #2629, opened 2023-04-07
packing: remove redundant canAlloc checking and alloc allocating.

### LINKED ISSUE #2628: performance: packing.canAlloc is redundant with packing.alloc in extendFor
### Ebitengine Version

v2.5.0

### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [X] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] Xbox
- [ ] Web Browsers

### Go Version (`go version`)

go version go1.19.5 linux/amd64

Profiling AAAAXY's loading screen reproduces the issue: https://user-images.githubusercontent.com/251568/230496805-c5e32c19-9258-49c8-800b-a3f0bc3b072d.svg

Not much time spent allocating atlas regions.

### What happens instead?

A lot of time, over half the total loading time, is spent allocating atlas regions. Furthermore, it looks extremely suspicious that on that profile, canAlloc and alloc take basically the same amount of CPU time.

Looking in code, they indeed seem redundant: whenever canAlloc succeeds, and it almost always does (except when the atlas needs resizing, which should only happen a small amount of times), alloc is called soon after to actually allocate the node.

Sending PR for this.

divVerent:
Closes #2628

bug

## What this PR does | solves
By removing redundant work done in canAlloc and just calling alloc right away, this removes 35% contribution to AAAAXY loading CPU time on the profile on https://user-images.githubusercontent.com/251568/230496805-c5e32c19-9258-49c8-800b-a3f0bc3b072d.svg, or - as measured via stopwatch - brings loading time on Moto G7 Play from 17.22s to 14.94s after already applying https://github.com/hajimehoshi/ebiten/pull/2627.

This should be safe as there is no case in which alloc succeeds and the allocated region isn't used; also, there is no case in which alloc mutates the tree when it doesn't actually succeed (comment added in one place to justify this).

### DIFF
--- a/internal/packing/packing.go
+++ b/internal/packing/packing.go
@@ -107,32 +107,6 @@ func square(width, height int) float64 {
 	return float64(height) / float64(width)
 }
 
-func (p *Page) canAlloc(n *Node, width, height int) bool {
-	if p.root == nil {
-		return p.width >= width && p.height >= height
-	}
-	return canAlloc(p.root, width, height)
-}
-
-func canAlloc(n *Node, width, height int) bool {
-	if n.width < width || n.height < height {
-		return false
-	}
-	if n.used {
-		return false
-	}
-	if n.child0 == nil && n.child1 == nil {
-		return true
-	}
-	if canAlloc(n.child0, width, height) {
-		return true
-	}
-	if canAlloc(n.child1, width, height) {
-		return true
-	}
-	return false
-}
-
 func alloc(n *Node, width, height int) *Node {
 	if n.width < width || n.height < height {
 		return nil
@@ -178,7 +152,17 @@ func alloc(n *Node, width, height int) *Node {
 				parent: n,
 			}
 		}
-		return alloc(n.child0, width, height)
+		// Note: it now MUST fit, due to above preconditions (repeated here).
+		if n.child0.width < width || n.child0.height < height {
+			panic(fmt.Sprintf("packing: the newly created child node (%d, %d) unexpectedly does not contain the requested size (%d, %d)", n.child0.width, n.child0.height, width, height))
+		}
+		// Thus, alloc can't return nil, but it may do another split along the other dimension
+		// to get a node with the exact size (width, height).
+		node := alloc(n.child0, width, height)
+		if node == nil {
+			panic(fmt.Sprintf("packing: could not allocate the requested size (%d, %d) in the newly created child node (%d, %d)", width, height, n.child0.width, n.child0.height))
+		}
+		return node
 	}
 	if n.child0 == nil || n.child1 == nil {
 		panic("packing: both two children must not be nil at alloc")
@@ -201,10 +185,6 @@ func (p *Page) Alloc(width, height int) *Node {
 		panic("packing: width and height must > 0")
 	}
 
-	if !p.extendFor(width, height) {
-		return nil
-	}
-
 	if p.root == nil {
 		p.root = &Node{
 			width:  p.width,
@@ -217,7 +197,7 @@ func (p *Page) Alloc(width, height int) *Node {
 	if height < minSize {
 		height = minSize
 	}
-	return alloc(p.root, width, height)
+	return p.extendForAndAlloc(width, height)
 }
 
 func (p *Page) Free(node *Node) {
@@ -255,13 +235,13 @@ func walk(n *Node, f func(n *Node) error) error {
 	return nil
 }
 
-func (p *Page) extendFor(width, height int) bool {
-	if p.canAlloc(p.root, width, height) {
-		return true
+func (p *Page) extendForAndAlloc(width, height int) *Node {
+	if n := alloc(p.root, width, height); n != nil {
+		return n
 	}
 
 	if p.width >= p.maxSize && p.height >= p.maxSize {
-		return false
+		return nil
 	}
 
 	// (1, 0), (0, 1), (2, 0), (1, 1), (0, 2), (3, 0), (2, 1), (1, 2), (0, 3), ...
@@ -284,14 +264,14 @@ func (p *Page) extendFor(width, height int) bool {
 			}
 
 			rollback := p.extend(newWidth, newHeight)
-			if p.canAlloc(p.root, width, height) {
-				return true
+			if n := alloc(p.root, width, height); n != nil {
+				return n
 			}
 			rollback()
 
 			// If the allocation failed even with a maximized page, give up the allocation.
 			if newWidth >= p.maxSize && newHeight >= p.maxSize {
-				return false
+				return nil
 			}
 		}
 	}


divVerent:
Tested: AAAAXY builds the same atlas in internalimages*/7.png with and without this change.

divVerent:
Stopwatch results:

* Initial: 18.00s (AAAAXY v1.3.421)
* https://github.com/hajimehoshi/ebiten/pull/2627: 17.22s
* This fix: 14.94s

internal/packing/packing.go
```diff
@@ -201,8 +177,11 @@ func (p *Page) Alloc(width, height int) *Node {
 		panic("packing: width and height must > 0")
 	}
 
-	if !p.extendFor(width, height) {
-		return nil
+	if width < minSize {
+		width = minSize
+	}
+	if height < minSize {
+		height = minSize
```
hajimehoshi, line 184: Do we need to move this?

hajimehoshi (commented):
Seems good but I'll take a look carefully later.

internal/packing/packing.go
divVerent: I do think so - but also I believe the previous code had a bug, even though I could find no way to trigger it.

Previous code first extended for the initial width and height, then added the root node if missing, then expanded the size to at least match the minimum, then allocated.

New code first expands the size to at least match the mininum, then creates the root node if missing (these two steps can freely be switched as they're independent from each other), then allocates and expands in one step.

So the previous code had a bug: the checked and allocated sizes differ. However, as minSize is 1, I guess this simply doesn't matter at all. If minSize were, say, 16, and one were trying to allocate 2x2, it'd check for a free 2x2 spot (and will not expand if it finds a spot for it), but then actually it could try to allocate 16x16 and fail - so it could fail even though extending would still be possible. That is why I decided to move the size extending first, so it's obviously just parameter preprocessing and one get it out of the way mentally.

If you want, though, I can do the p.root == nil block first, as those two are fully independent from each other anyway; I just found it cleaner to see the size change as parameter preprocessing and thus closer to the top of the function, but I now do see I could make the diff slightly smaller by swapping them back.

There is one other behavioral difference though, which is technically a bug but I believe it does not matter: if the requested size is bigger than the maximum allowed size by the hardware, then extendFor would have failed and it wouldn't have created the root node. With my code, it then WILL allocate the root node, then fail. My basis for considering this harmless is that on the next call with a smaller size, this root node will be used and split as normally, so the "unused" root node that could be a leftover from a failed allocation doesn't matter in practice, beyond being an unnecessarily allocated object that takes up memory.

hajimehoshi: Hmm? I still don't understand why this doesn't work:

```go
	if p.root == nil {
		p.root = &Node{
			width:  p.width,
			height: p.height,
		}
	}

	if width < minSize {
		width = minSize
	}
	if height < minSize {
		height = minSize
	}

	return p.extendForAndAlloc(width, height)
```

> I do think so - but also I believe the previous code had a bug, even though I could find no way to trigger it.

If there is a bug, please send a PR to fix the bug.

hajimehoshi: > With my code, it then WILL allocate the root node, then fail. My basis for considering this harmless is that on the next call with a smaller size, this root node will be used and split as normally, so the "unused" root node that could be a leftover from a failed allocation doesn't matter in practice, beyond being an unnecessarily allocated object that takes up memory.

My undertanding is that this is a side-effect of your optimization PR. Is this correct?

divVerent: Is a side effect of this PR. But as said, I do not think it matters whether this root node exists or not (and in fact, it could safely be created right at construction time).

divVerent: Order reversed as requested.

divVerent: > If there is a bug, please send a PR to fix the bug first, before doing this optimization. But I still don't understand what is the bug...

BTW, that "bug" I referenced is VERY theoretical, as it requires first changing the minSize compile time constant. Right now minSize does nothing beyond preventing to add a x * 0 rectangle which would add useless nodes, but shouldn't be happening in the first place. Thus, as there's no way to reproduce it, there's no bug right now, unless I file a PR of the sort "if you patch this line of code, this other part breaks", which doesn't seem useful.

internal/packing/packing.go
```diff
@@ -178,6 +152,8 @@ func alloc(n *Node, width, height int) *Node {
 				parent: n,
 			}
 		}
+		// Note: it now MUST fit, as child0 is bigger than width * height when getting here.
+		// Thus, alloc can't return nil, but it may do another split along the other dimension.
```
hajimehoshi, line 156: I don't understand this comment. What does width * height mean? Can we replace this with an if condition with a panic as an assertion?

divVerent: It means width x height. But yes, I can add a panic here to document this better.

The reason why this is important is that if this alloc call fails, we'll have mutated the tree WITHOUT actually allocating anything, which could be a mess during the extend process.

hajimehoshi: OK, let's add assertions here.

We can also add an assertion to check that alloc returns a non-nil value here.

divVerent: Added a panic and more explanation.

internal/packing/packing.go
```diff
@@ -178,7 +152,15 @@ func alloc(n *Node, width, height int) *Node {
 				parent: n,
 			}
 		}
-		return alloc(n.child0, width, height)
+		// Note: it now MUST fit, as child0.width >= width and child0.height >= height.
+		// Otherwise, this function would have already failed right at the top.
+		// Thus, alloc can't return nil, but it may do another split along the other dimension
+		// to get an exactly width x height node.
```
hajimehoshi, line 158: > it now MUST fit

Does `fit` mean that `child0` has enough capability to allocate a new node (width, height)?

> child0.width >= width and child0.height >= height

I suggested we could write this as an assertion in Go.

> width x height node

Does this mean "a node with a size (width, height)"?

divVerent: Applied.

internal/packing/packing.go
```diff
@@ -178,7 +152,17 @@ func alloc(n *Node, width, height int) *Node {
 				parent: n,
 			}
 		}
-		return alloc(n.child0, width, height)
+		// Note: it now MUST fit, due to above preconditions (repeated here).
+		if n.child0.width < width || n.child0.height < height {
+			panic("packing: the newly created child node unexpectedly does not contain the requested size")
```
hajimehoshi, line 157:
```go
panic(fmt.Sprintf("packing: the newly created child node (%d, %d) unexpectedly does not contain the requested size (%d, %d)", n.child0.width, n.child0.height, width, height))
```

internal/packing/packing.go
```diff
@@ -178,7 +152,17 @@ func alloc(n *Node, width, height int) *Node {
...
+		}
+		// Thus, alloc can't return nil, but it may do another split along the other dimension
+		// to get a node with the exact size (width, height).
+		node := alloc(n.child0, width, height)
+		if node == nil {
+			panic("packing: could not allocate the requested size in the newly created child node")
```
hajimehoshi, line 163:
```go
panic("packing: could not allocate the requested size (%d, %d) in the newly created child node (%d, %d)", width, height, n.child0.width, n.child0.height)
```

divVerent:
(interesting, I've never seen `panic(fmt.Sprintf(` before - where I work, we use `log.Fatalf` :)

hajimehoshi (approved):
LGTM, thanks!

Issue #2634, opened 2023-04-09
Mouse CursorModeCapture broken in WASM


harbdog:
### Ebitengine Version

2.5

### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] Xbox
- [X] Web Browsers

### Go Version (`go version`)

1.19.6

### What steps will reproduce the problem?

Created an example and gh-pages to host the wasm example in my fork:

https://harbdog.github.io/ebiten/

Branch: `wasm_mouse_capture`
https://github.com/harbdog/ebiten/tree/wasm_mouse_capture

```bash
git checkout wasm_mouse_capture
wasmserve.exe ./examples/capture
```

It works when not using WASM:
```bash
git checkout wasm_mouse_capture
go run ./examples/capture
```

### What is the expected result?

Mouse capture should work in browser the same as it does in desktop application mode.

### What happens instead?

The mouse tends to move only in a single direction or jumps around unexpectedly.

harbdog:
It seems that using the Escape key to toggle capture may conflict a bit with the browser also using Escape to release capture, that might be the source of some of the weird behavior.

However I also tried changing to use Space to toggle capture, which did at least address some of the weirdness with re-capturing the cursor. However, what I noticed was that it still showed only the X mouse movement to register when using Chrome.

However in Edge, it only has Y mouse movement registering.

In Firefox it registers both X and Y mouse movement, but the movement is not smooth, almost "drunk" like.

harbdog:
Thanks @hajimehoshi, confirmed that change fixes the mouse movement as expected!

PR #2635, opened 2023-04-09
create mouse capture example

harbdog:
Example used for #2634

bug

## What this PR does | solves
Provides an example demo for mouse capture

### DIFF
--- a/examples/capture/main.go
+++ b/examples/capture/main.go
@@ -1,0 +1,125 @@
+// Copyright 2018 The Ebiten Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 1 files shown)


examples/capture/main.go
```diff
@@ -0,0 +1,131 @@
+// Copyright 2018 The Ebiten Authors
```
TotallyGamerJet, line 1: Should be 2023

examples/capture/main.go
```diff
@@ -0,0 +1,131 @@
+// Copyright 2023 The Ebiten Authors
```
hajimehoshi, line 1: Ebitengine

hajimehoshi:
I'm not sure the purpose of this PR. This PR does not fix the bug, right?

hajimehoshi:
Let me close this as #2634 was resolved.

harbdog:
Sorry it was orignally to provide an example for that bug, so I've opened instead as feature enhancement now: #2636

PR #2636, opened 2023-04-10
create mouse capture example

harbdog:
n/a

feature

## What this PR does | solves
Provides an example of mouse capture, using similar style as the mouse wheel example.


harbdog:
I probably should have made this PR against main instead so I'll close this one. Also based on recent findings about browser mouse capture I'll need to change a couple things, browsers (especially Chrome) seem to only allow capture when it's done very close in timing to a key/button press. This means it cannot start initialized as captured when run as wasm.

Issue #2641, opened 2023-04-14
Provide example of mouse capture using relative mouse movement


harbdog:
### Operating System

- [x] Windows
- [x] macOS
- [x] Linux
- [x] FreeBSD
- [x] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] Xbox
- [x] Web Browsers

### What feature would you like to be added?

An example of mouse capture, similar to the provided mouse wheel example.

### Why is this needed?

Originally used to show a regression bug, now it can be provided as an example to others!

PR #2642, opened 2023-04-14
mouse capture example

harbdog:
#2641

feature

## What this PR does | solves
Provides an example of mouse capture, using similar style as the mouse wheel example.

### DIFF
--- a/examples/capture/main.go
+++ b/examples/capture/main.go
@@ -18,8 +18,8 @@ import (
 	"fmt"
 	"image/color"
 	"log"
-	"math"
 	"runtime"
+	"sync"
 
 	"github.com/hajimehoshi/ebiten/v2"
 	"github.com/hajimehoshi/ebiten/v2/ebitenutil"
@@ -38,32 +38,30 @@ const (
 )
 
 type Game struct {
-	mouseX int
-	mouseY int
-	x      int
-	y      int
+	initOnce sync.Once
+	mouseX   int
+	mouseY   int
+	x        int
+	y        int
 }
 
 func (g *Game) Update() error {
+	g.initOnce.Do(func() {
+		// initialize cursor position on first update
+		g.mouseX, g.mouseY = ebiten.CursorPosition()
+	})
+
 	if ebiten.CursorMode() == ebiten.CursorModeCaptured {
 		cursorX, cursorY := ebiten.CursorPosition()
+		deltaX, deltaY := cursorX-g.mouseX, cursorY-g.mouseY
+		g.mouseX, g.mouseY = cursorX, cursorY
 
-		if g.mouseX == math.MinInt32 && g.mouseY == math.MinInt32 {
-			// Initialize first position to establish delta.
-			if cursorX != 0 && cursorY != 0 {
-				g.mouseX, g.mouseY = cursorX, cursorY
-			}
-		} else {
-			deltaX, deltaY := cursorX-g.mouseX, cursorY-g.mouseY
-			g.mouseX, g.mouseY = cursorX, cursorY
+		if deltaX != 0 {
+			g.x += deltaX
+		}
 
-			if deltaX != 0 {
-				g.x += deltaX
-			}
-
-			if deltaY != 0 {
-				g.y += deltaY
-			}
+		if deltaY != 0 {
+			g.y += deltaY
 		}
 
 		// Constrain red dot within screen view.
@@ -89,7 +87,7 @@ func (g *Game) Update() error {
 			ebiten.SetCursorMode(ebiten.CursorModeCaptured)
 
 			// Reset mouse cursor position for its return.
-			g.mouseX, g.mouseY = math.MinInt32, math.MinInt32
+			g.mouseX, g.mouseY = ebiten.CursorPosition()
 		}
 	}
 
@@ -125,10 +123,8 @@ func main() {
 	}
 
 	g := &Game{
-		mouseX: math.MinInt32,
-		mouseY: math.MinInt32,
-		x:      screenWidth / 2,
-		y:      screenHeight / 2,
+		x: screenWidth / 2,
+		y: screenHeight / 2,
 	}
 	if err := ebiten.RunGame(g); err != nil {
 		log.Fatal(err)
--- a/examples/capture/main.go
+++ b/examples/capture/main.go
@@ -69,14 +69,14 @@ func (g *Game) Update() error {
 		// Constrain red dot within screen view.
 		if g.x < 0 {
 			g.x = 0
-		} else if g.x > screenWidth-8 {
-			g.x = screenWidth - 8
+		} else if g.x > screenWidth-pointerImage.Bounds().Dx() {
+			g.x = screenWidth - pointerImage.Bounds().Dx()
 		}
 
 		if g.y < 0 {
 			g.y = 0
-		} else if g.y > screenHeight-8 {
-			g.y = screenHeight - 8
+		} else if g.y > screenHeight-pointerImage.Bounds().Dy() {
+			g.y = screenHeight - pointerImage.Bounds().Dy()
 		}
 	}
 
@@ -118,10 +118,9 @@ func main() {
 	ebiten.SetWindowTitle("Mouse Capture (Ebitengine Demo)")
 	ebiten.SetWindowSize(screenWidth, screenHeight)
 
-	if runtime.GOOS == "js" {
-		// Web browsers allow cursor mode capture only with user interaction.
-		// Start without cursor captured with message indicating how to capture.
-	} else {
+	// Web browsers allow cursor mode capture only with user interaction.
+	// Start without cursor captured with message indicating how to capture.
+	if runtime.GOOS != "js" {
 		ebiten.SetCursorMode(ebiten.CursorModeCaptured)
 	}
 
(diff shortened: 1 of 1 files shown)


examples/capture/main.go
```diff
@@ -0,0 +1,137 @@
...
+		x: float64(screenWidth / 2), y: float64(screenHeight / 2),
+	}
+
+	ebiten.SetWindowSize(screenWidth, screenHeight)
+	ebiten.SetWindowTitle("Mouse Capture (Ebitengine Demo)")
+
+	if runtime.GOOS == "js" {
+		// web browsers only allow cursor mode capture with user interaction
+		// start without cursor captured with message indicating how to capture
+	} else {
```
hajimehoshi, line 130:
```go
// Web browsers only allow cursor mode capture with user interaction.
// Start without cursor captured with message indicating how to capture.
if runtime.GOOS != "js" {
```

examples/capture/main.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func (g *Game) Layout(outsideWidth, outsideHeight int) (int, int) {
+	return screenWidth, screenHeight
+}
+
+func main() {
+	g := &Game{
+		mouseX: math.MinInt32, mouseY: math.MinInt32,
+		x: float64(screenWidth / 2), y: float64(screenHeight / 2),
```
hajimehoshi, line 121: Separate members to lines:

```go
mouseX: math.MinInt32,
mouseY: math.MinInt32,
x:      screenWidth / 2.0,
y:      screenHeight / 2.0,
```

examples/capture/main.go
```diff
@@ -0,0 +1,137 @@
...
+	return nil
+}
+
+func (g *Game) Draw(screen *ebiten.Image) {
+	op := &ebiten.DrawImageOptions{}
+	op.GeoM.Translate(g.x, g.y)
+	screen.DrawImage(pointerImage, op)
+
+	var message string
+
```
hajimehoshi, line 105: Remove this empty line.

examples/capture/main.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	ebitenutil.DebugPrint(screen, message)
+}
+
+func (g *Game) Layout(outsideWidth, outsideHeight int) (int, int) {
+	return screenWidth, screenHeight
+}
+
+func main() {
+	g := &Game{
```
hajimehoshi, line 119: Move this just before `RunGame`.

examples/capture/main.go
```diff
@@ -0,0 +1,137 @@
...
+	x      float64
+	y      float64
+}
+
+func (g *Game) Update() error {
+	if ebiten.CursorMode() == ebiten.CursorModeCaptured {
+		cursorX, cursorY := ebiten.CursorPosition()
+
+		if g.mouseX == math.MinInt32 && g.mouseY == math.MinInt32 {
+			// initialize first position to establish delta
```
hajimehoshi, line 52: * Start with an upper case.
* Use a period.

```
Initialize the first position to establish delta.
```

examples/capture/main.go
hajimehoshi: By the way,

> Web browsers only allow cursor mode capture with user interaction.

Does this sentence mean "Web browsers allow cursor mode 'capture' only with user interaction"?

harbdog: yes same meaning, i'll rearrange the wording as you have it though

examples/capture/main.go
```diff
@@ -0,0 +1,137 @@
...
+		// Start without cursor captured with message indicating how to capture.
+	} else {
+		ebiten.SetCursorMode(ebiten.CursorModeCaptured)
+	}
+
+	g := &Game{
+		mouseX: math.MinInt32,
+		mouseY: math.MinInt32,
+		x:      float64(screenWidth / 2),
+		y:      float64(screenHeight / 2),
```
hajimehoshi, line 132: I prefer

```
screenWidth / 2.0
```

examples/capture/main.go
```diff
@@ -0,0 +1,137 @@
...
+		// Start without cursor captured with message indicating how to capture.
+	} else {
+		ebiten.SetCursorMode(ebiten.CursorModeCaptured)
+	}
+
+	g := &Game{
+		mouseX: math.MinInt32,
+		mouseY: math.MinInt32,
+		x:      float64(screenWidth / 2.0),
+		y:      float64(screenHeight / 2.0),
```
hajimehoshi, line 132: We don't need `float64` casting

harbdog: good point, it had just been copied over when I started it from the wheel.go example. I'll go ahead and change over to int

hajimehoshi: I meant `screenWidth / 2.0` is already float64

harbdog: oh duh sorry I misread that, however it seems int should be fine anyway since cursor position returns as int. Or would you prefer it to go back to float?

hajimehoshi: I'm fine with both.

examples/capture/main.go
```diff
@@ -0,0 +1,137 @@
...
+
+			if deltaY != 0 {
+				g.y += deltaY
+			}
+		}
+
+		// Constrain red dot within screen view.
+		if g.x < 0 {
+			g.x = 0
+		} else if g.x > screenWidth-8 {
```
hajimehoshi, line 72: Replace `8` with `pointerImage.Bounds().Dx()` or `pointerImage.Bounds().Dy()`

examples/capture/main.go
hajimehoshi: This is not fixed yet.

harbdog: the new wording has already been rearranged as you said:

```
// Web browsers allow cursor mode capture only with user interaction.
// Start without cursor captured with message indicating how to capture.
```

hajimehoshi: The `if` statement is not fixed yet...

```go
// Web browsers allow cursor mode capture only with user interaction.
// Start without cursor captured with message indicating how to capture.
if runtime.GOOS != "js" {
    ebiten.SetCursorMode(ebiten.CursorModeCaptured)
}
```

harbdog: thanks for clarifying I see now what you mean

examples/capture/main.go
```diff
@@ -0,0 +1,136 @@
...
+		if ebiten.CursorMode() == ebiten.CursorModeCaptured {
+			// Release mouse cursor capture.
+			ebiten.SetCursorMode(ebiten.CursorModeVisible)
+		} else {
+			// Recapture mouse cursor.
+			ebiten.SetCursorMode(ebiten.CursorModeCaptured)
+
+			// Reset mouse cursor position for its return.
+			g.mouseX, g.mouseY = math.MinInt32, math.MinInt32
+		}
```
hajimehoshi, line 93: Let's refactor this logic out to a new function like `updateCursorMode`

```go
func (g *Game) updateCursorMode(cursorMode ebiten.CursorModeType) {
	if ebiten.CursorMode() == cursorMode {
		return
	}
	ebiten.SetCursorMode(mode)
	if cursorMode == ebiten.CursorModeCaptured {
		// Initialize first position to establish delta.
		g.mouseX, g.mouseY = ebiten.CursorPosition()
	}
}
```

```go
	if inpututil.IsKeyJustPressed(ebiten.KeySpace) {
		if ebiten.CursorMode() == ebiten.CursorModeVisible {
			g.updateCursorMode(ebiten.CursorModeCaptured)
		} else {
			g.updateCursorMode(ebiten.CursorModeVisible)
		}
	}
```

Also, let's call this from `main`

```go
	if runtime.GOOS != "js" {
		g.updateCursorMode(ebiten.CursorModeCaptured)
	}
```

Then, we don't have to have `MinInt32` as a special value.

harbdog: I like the idea of the function, the only problem is that by not initializing the special value `MinInt32` in `main` the cursor start position ends up being thrown to the bottom right corner `632, 472` instead of the center as desired `(screenWidth/2, screenHeight/2)`.

harbdog: Actually it seems to appear the the start position of the red dot depends on the position where the mouse happens to be on the desktop when the program first starts up. My preference would be the always have the dot start in the center, however I'll defer to you. If the start position doesn't bother you I'll go ahead with your idea.

hajimehoshi: Oh right, I realized `ebiten.CursorPosition` is not valid before `ebiten.RunGame`.

What about using `sync.Once`?:

```go
type Game struct {
	// ...
	initOnce sync.Once
}

func (g*Game) Update() error {
	g.initOnce.Do(func() {
		g.mouseX, g.mouseY = ebiten.CursorPosition()
	})
	// ...
	if inpututil.IsKeyJustPressed(ebiten.KeySpace) {
		if ebiten.CursorMode() == ebiten.CursorModeVisible {
			ebiten.SetCursorMode(ebiten.CursorModeCaptured)
		} else {
			ebiten.SetCursorMode(ebiten.CursorModeVisible)
			g.mouseX, g.mouseY = ebiten.CursorPosition()
		}
	}
}
```

`main` remains not to use `updateCursorMode`:

```go
	if runtime.GOOS != "js" {
		ebiten.SetCursorMode(ebiten.CursorModeCaptured)
	}
```

hajimehoshi (approved):
LGTM, thanks!

PR #2665, opened 2023-05-09
Fix _GET_X_LPARAM and _GET_Y_LPARAM from api_windows.go

the-goodies:
Negative values should not disappear. For example, if _HIWORD returns 63000, then _GET_Y_LPARAM should return -2536, but now it returns 63000. This fixes that.

Here's how it's defined in windowsx.h:

`#define GET_X_LPARAM(lp)                        ((int)(short)LOWORD(lp))`
`#define GET_Y_LPARAM(lp)                        ((int)(short)HIWORD(lp))`

### DIFF
--- a/internal/goglfw/api_windows.go
+++ b/internal/goglfw/api_windows.go
@@ -367,7 +367,7 @@ type (
 )
 
 func _GET_X_LPARAM(lp _LPARAM) int {
-	return int(_LOWORD(uint32(lp)))
+	return int(int16(_LOWORD(uint32(lp))))
 }
 
 func _GET_XBUTTON_WPARAM(wParam _WPARAM) uint16 {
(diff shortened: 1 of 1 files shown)


hajimehoshi:
Did the original code cause an actual issue? I'd like to know how serious the issue is. If this is serious, I'll back-port the fix to the stable branch. Thanks,

hajimehoshi:
These functions (macros) are used at `WM_MOUSEMOVE` (a mouse move) and `WM_MOVE` (a window move). For a mouse move, the cursor position is a relative position in a window so these cannot be a negative value. For a window move, I have confirmed a too-big value can be passed to `inputWindowPos`. However, Ebitengine doesn't use any callbacks for window position changes. Therefore, I think there is no serious issue, which means we don't have to backport the fix to the stable branch.

hajimehoshi (approved):
LGTM, thanks!

the-goodies:
> Did the original code cause an actual issue?

Yes, but not while using Ebiten engine. I'm using yours goglfw (together with OpenGL (no CGO!) with my own ImGui Go bindings, and encountered an issue where cursorPosCallback is called with wrong cursor position parameters (huge values) when moving or resizing secondary viewports up or to the left.

PR #2678, opened 2023-06-23
iOS Keyboard support.

### LINKED ISSUE #1090: cmd/ebitenmobile: support keybords on iOS
https://stackoverflow.com/questions/3717141/how-to-detect-keyboard-events-on-hardware-keyboard-on-iphone-ios

This might be a little hard.

divVerent:
Closes #1090

feature

## What this PR does | solves
It creates a keyboard event mapping for physical keyboards on iOS.

Minimum supported iOS version is 13.4 - before iOS 13.4, this API only can return four arrow keys and five extra keys, which is usually not enough. So I didn't bother implementing the pre-13.4 case; instead, pre-13.4, this code compiles but does nothing.

Tested in AAAAXY - the game is now fully playable in the iOS Simulator, which is one step for potentially releasing the iOS version on the Mac App Store, and otherwise a major improvement for local testing.

Caveat: I am unsure about the "aliased" key mappings, commented in genkeys.go. In particular, I do not know what the NonUSBackslash and NonUSPound keys are and what Ebitengine key they SHOULD map to. I ASSUME they may be the extra key many international keyboards have near the enter key, and thus mapped to Backslash as that's what pressing the key does on DOS with US layout, but why have two codes for that then?


hajimehoshi:
The change seems fine, thank you!

>  I ASSUME they may be the extra key many international keyboards have near the enter key, and thus mapped to Backslash as that's what pressing the key does on DOS with US layout, but why have two codes for that then?

https://www.w3.org/TR/uievents-code/#keyboard-102
> A second key is also added (labelled #~ on a UK keyboard) which is partially tucked under the Enter key. This key is encoded as "[Backslash](https://www.w3.org/TR/uievents-code/#code-Backslash)", using the same code as the \| key found on the "101" keyboard layout. According to [[USB-HID]](https://www.w3.org/TR/uievents-code/#biblio-usb-hid), the US \| and UK #~ are actually two separate keys (named "Keyboard \ and |" and "Keyboard Non-US # and ~"), but since these two keys never co-occur on the same keyboard most platforms use the same scancode for both keys, making them difficult to distinguish. It is for this reason that the code "[Backslash](https://www.w3.org/TR/uievents-code/#code-Backslash)" is used for both of these keys.

So assigning two different code to the same key name "backslash" seems OK. What do you think?

divVerent:
Thanks for the link. Yes, both of these key codes probably are best mapped to Backslash.

Keyboard Non-US # and ~ is the #' key on German layout, and does emit \ on US keyboard layout settings. It's close to the Enter key and emits the same scancode as backslash. Also, TIL that # is a pound sign, I thought £ is the pound sign, but fine.

The other one the \ key next to shift - is the <>| key on German layout, and too acts just like \ when on US layout.

So yeah, let's map all three to the "Backslash" Ebitengine key code for now. This will likely have one minor fun bug that holding one, then the other, and releasing first makes Ebitengine think no key is held. If we ever want to fix this, we probably should fix it by assigning a separate keycode in Ebitengine, but it's probably not a high priority.

hajimehoshi:
Thanks. Could you add a comment to explain this background (why we assign the two keys to 'backslash')

divVerent:
Thanks - more explanation added.

genkeys.go
```diff
@@ -196,6 +197,88 @@ func init() {
...
+		0x50: "ArrowLeft",
+		0x4F: "ArrowRight",
+		0x52: "ArrowUp",
+		0x35: "Backquote",
+
+		// These three keys are:
+		// - US backslash-pipe key (above return),
+		// - non-US backslash key (next to left shift; on German layout this is the <>| key), and
+		// - non-US hashmark key (bottom left of return; on German layout, this is the #' key).
+		// On US layout configurations, they all map to the same characters - the backslash.
```
hajimehoshi, line 214: Let's add a URL https://www.w3.org/TR/uievents-code/#keyboard-102

mobile/ebitenmobileview/input_ios.go
```diff
@@ -58,3 +58,22 @@ func UpdateTouchesOnIOS(phase int, ptr int64, x, y int) {
...
 	}
 }
+
+func UpdatePressesOnIOS(phase int, keyCode int, keyString string) {
+	switch phase {
+	case C.UITouchPhaseBegan, C.UITouchPhaseMoved, C.UITouchPhaseStationary:
+		if key, ok := iosKeyToUIKey[keyCode]; ok {
+			keys[key] = struct{}{}
+		}
+		runes = []rune(keyString)
```
hajimehoshi, line 68:
```go
if phase == C.UITouchPhaseBegan {
    for _, r := range keyString {
        if !unicode.IsPrint(r) {
            continue
        }
        runes = append(runes, r)
    }
}
```

divVerent: Why only in UITouchPhaseBegan?

Trying out if this is necessary, or even correct, by checking what phase is used for key repeat.

divVerent: Here in the Simulator, I get no events at all for key repeat. Let's actually do that instead.

hajimehoshi: Hmm, I see. I think we should consider repeating (so the current input_android.go has the same issue), but is no event is depetected by repeating, there is nothing we can do. Let's handle only `UITouchPhaseBegan` to avoid unexpected behaviors.

divVerent: Done that now.

Also moved clearing of runes to after UpdateInput(). This is possibly an existing bug with Android - it might be possible that touch events can repeat the last key event. Don't know for sure though.

MIGHT be related to my not yet tracked down https://github.com/divVerent/aaaaxy/issues/222

mobile/ebitenmobileview/input.go
```diff
@@ -46,4 +46,5 @@ func updateInput() {
 	}
 
 	ui.Get().UpdateInput(keys, runes, touchSlice)
+	runes = nil
```
hajimehoshi, line 49: Why is this needed?

mobile/ebitenmobileview/input_ios.go
```diff
@@ -58,3 +59,28 @@ func UpdateTouchesOnIOS(phase int, ptr int64, x, y int) {
...
+				}
+				runes = append(runes, r)
+			}
+		}
+		updateInput()
+	case C.UITouchPhaseEnded, C.UITouchPhaseCancelled:
+		if key, ok := iosKeyToUIKey[keyCode]; ok {
+			delete(keys, key)
+		}
+		updateInput()
```
hajimehoshi, line 82:
```go
runes = nil
updateInput()
```

mobile/ebitenmobileview/input.go
hajimehoshi: This affects the Android side. Please do not this (if needed, please make another PR)

divVerent: It's probably an existing bug, but I have no code that uses the runes feature. Moved it to iOS-only now.

mobile/ebitenmobileview/input_ios.go
divVerent: Done that, but had to add this to EVERY callk site then.

hajimehoshi (approved):
LGTM, thanks!

I found an issue in Android: `runes = nil` might be needed in some places. I'll take a look later.

divVerent:
Just checked, the runes = nil is necessary indeed. Verified by removing the `runes = nil` from the touch processing - if I now press the a key and click the Simulator window, that counts as extra a key presses.

divVerent:
I think the shared runes = nil, plus making all code append, is probably ideal. I'll send a separate PR for that once this is merged.

mobile/ebitenmobileview/input_ios.go
hajimehoshi: > Also moved clearing of runes to after UpdateInput(). This is possibly an existing bug with Android - it might be possible that touch events can repeat the last key event. Don't know for sure though.

Sorry but I realized this comment now. I'll take a look at Android later anyway.

hajimehoshi:
> I think the shared runes = nil, plus making all code append, is probably ideal. I'll send a separate PR for that once this is merged.

As `runes` doesn't have to be a global state, adding an argument `runes` to `updateInput` sounds better.

divVerent:
Absolutely. I don't even know why it's currently a global.

hajimehoshi:
I'm gonna sleep soon. If you could confirm and fix the Android issue, I'd be very happy!

PR #2681, opened 2023-06-23
Fix antialias region size

### LINKED ISSUE #2679: failed assertion `Set Scissor Rect Validation
### Ebitengine Version

v2.5.4

### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [X] iOS
- [ ] Nintendo Switch
- [ ] Xbox
- [ ] Web Browsers

### Go Version (`go version`)

Run AAAAXY in the iOS Simulator. Issue doesn't happen on real iPhone 6S or real iPad Air 1st Gen. Walk right and jump until you touch the first checkpoint, then click the back button at the top left and click "Play".

A map screen shows up.

### What happens instead?

Game halts, and debugger shows:

```
-[MTLDebugRenderCommandEncoder setScissorRect:]:3814: failed assertion `Set Scissor Rect Validation
(rect.x(0) + rect.width(800))(800) must be <= render pass width(128)
(rect.y(0) + rect.height(384))(384) must be <= render pass height(64)
```

Issue already exists in v2.5.0. I cannot go back further. Do not yet know what exact render call triggers it.

Size 800x384 surprises me - there's no object in my game of that size. Also don't know yet which render pass this is.

divVerent:
Closes #2679 

bug

## What this PR does | solves
Fixes coordinate transformation math on dstRegion.

Previous code had x1 and y1 shifted by exactly (i.region.X, i.region.Y), which allowed it to exceed the width and height of the destination image. iOS MTLDebugRenderCommandEncoder detects this issue and crashes.

No known test case for OpenGL or similar, as the rectangle is always strictly larger or equal to the correct rectangle.

It is POSSIBLE, but not reproduced, this this can in OpenGL cause DrawTriangles to draw on a different image than the destination image if they share an atlas. However, as this depends a lot on how the atlas is built, I am not sure if it can actually be triggered reliably, or even at all.

### DIFF
--- a/internal/ui/image.go
+++ b/internal/ui/image.go
@@ -382,14 +382,16 @@ func (i *bigOffscreenImage) drawTriangles(srcs [graphics.ShaderImageCount]*Image
 		vertices[idx+1] = (vertices[idx+1] - i.region.Y) * bigOffscreenScale
 	}
 
+	// Compute corners in dst coordinate space.
 	x0 := dstRegion.X
 	y0 := dstRegion.Y
 	x1 := dstRegion.X + dstRegion.Width
 	y1 := dstRegion.Y + dstRegion.Height
+	// Translate to i.region coordinate space, and clamp against region size.
 	x0 = max(x0-i.region.X, 0)
 	y0 = max(y0-i.region.Y, 0)
-	x1 = min(x1, i.region.X+i.region.Width)
-	y1 = min(y1, i.region.Y+i.region.Height)
+	x1 = min(x1-i.region.X, i.region.Width)
+	y1 = min(y1-i.region.Y, i.region.Height)
 	dstRegion = graphicsdriver.Region{
 		X:      x0 * bigOffscreenScale,
 		Y:      y0 * bigOffscreenScale,


internal/ui/image.go
```diff
@@ -382,14 +382,16 @@ func (i *bigOffscreenImage) drawTriangles(srcs [graphics.ShaderImageCount]*Image
...
 	y0 := dstRegion.Y
 	x1 := dstRegion.X + dstRegion.Width
 	y1 := dstRegion.Y + dstRegion.Height
+	// Translate to i.region coordinate space, and clamp against region size.
 	x0 = max(x0-i.region.X, 0)
 	y0 = max(y0-i.region.Y, 0)
-	x1 = min(x1, i.region.X+i.region.Width)
-	y1 = min(y1, i.region.Y+i.region.Height)
+	x1 = min(x1-i.region.X, i.region.Width)
+	y1 = min(y1-i.region.Y, i.region.Height)
```
hajimehoshi, line 394: I don't understand this fix. Why didn't the previous code work?

hajimehoshi: Ah, now I think I understand. Thanks,

hajimehoshi:
LGTM.

If we find it hard to add a test, it's OK, let's merge this without tests. What do you think?

divVerent:
I tried finding a test case for this by detecting overwriting pixels on other textures, however it appears that the texture antialiasing uses doesn't end up on the atlas of the NewImage textures. Don't know why.

If this can't be done, then yeah, let's merge this without test cases. At the very least we have test cases to ensure this isn't a regression.

hajimehoshi:
A destination image is often a different texture from the source, and I understand it pretty hard to make a reliable test. OK let's merge this without a test.

hajimehoshi (approved):
LGTM, thanks!

divVerent:
Found the reason: https://github.com/hajimehoshi/ebiten/blob/main/internal/ui/image.go#L91

So the backing offscreen is only ever unmanaged or volatile. And only regular images ever end up on an atlas.

So there's never any texture sharing, and in OpenGL, writing out of bounds of a texture is harmless - so yeah, no test case possible, except if we can produce the Metal crash by forcing the same debug backend. All god then - thanks!

divVerent:
The other fact that makes this "untestable" is - region.X and region.Y are elsewhere clamped to be at least zero. Thus, the wrong region is only ever bigger, never smaller, than the correct region. Making the bug 100% harmless on OpenGL backends, except maybe a bit more wasteful regarding dirtying of texture space or similar.

PR #2683, opened 2023-06-23
Make runes a parameter.

### LINKED ISSUE #2682: Android: touch events can cause key "repeat"
### Ebitengine Version

v2.5.4

### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [X] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] Xbox
- [ ] Web Browsers

### Go Version (`go version`)

Add this patch to AAAAXY (or a similar patch to virtually anything using Ebitengine but not using key characters):

```
diff --git a/internal/aaaaxy/game.go b/internal/aaaaxy/game.go
index 2e524a8a..4645bf4f 100644
--- a/internal/aaaaxy/game.go
+++ b/internal/aaaaxy/game.go
@@ -173,6 +173,12 @@ func (g *Game) updateFrame() error {
 }
 
+       var r []rune
+       r = ebiten.AppendInputChars(r)
+       if len(r) != 0 {
+               log.Infof("received runes: %q", r)
+       }
+
        if !g.canUpdate {
                return nil
        }
```

Launch the game on Android.

Hold a key, and right after pressing the key down, touch the screen.

Log message with key event should show up once.

### What happens instead?

Log message with key event should shows up twice.

Found while making https://github.com/hajimehoshi/ebiten/pull/2678 for iOS.

Best solution is probably changing the `runes` argument from a global to a parameter of `updateInput`.

divVerent:
This prevents duplicate input characters due to concurrent touch events.

Closes #2682

bug

## What this PR does | solves
`updateInput` currently receives `runes` in a global variable as a hidden parameter. Some calls do not clear it though, but leave the previous value in it - in particular calls on behalf of touch or key release events.

### DIFF
--- a/mobile/ebitenmobileview/input.go
+++ b/mobile/ebitenmobileview/input.go
@@ -27,7 +27,6 @@ type position struct {
 
 var (
 	keys    = map[ui.Key]struct{}{}
-	runes   []rune
 	touches = map[ui.TouchID]position{}
 )
 
@@ -35,7 +34,7 @@ var (
 	touchSlice []ui.TouchForInput
 )
 
-func updateInput() {
+func updateInput(runes []rune) {
 	touchSlice = touchSlice[:0]
 	for id, position := range touches {
 		touchSlice = append(touchSlice, ui.TouchForInput{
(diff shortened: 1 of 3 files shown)


hajimehoshi (approved):
LGTM, thanks!

PR #2685, opened 2023-06-23
Android: support keypress runes even for unmapped keys

### LINKED ISSUE #2684: mobile/ebitenmobileview: Android: pressing unknown keys is ignored, even if Android assigns a text to them
### Ebitengine Version

v2.5.4

### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [X] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] Xbox
- [ ] Web Browsers

### Go Version (`go version`)

Assuming you can't buy a keyboard with an unsupported key, do this:

1. Edit `genkeys.go` to start the loop of `ASCII: 0 - 9` at `'1'`.
2. Patch AAAAXY, or any other game, with the same change as https://github.com/hajimehoshi/ebiten/issues/2682.
3. Compile, deploy, connect external keyboard.
4. Press the `0` key.

Expected result is that although `ebiten.IsKeyPressed` will not be able to return this unknown key, `ebiten.AppendInputChars` should still return it.

### What happens instead?

`ebiten.AppendInputChars` does not return the key.

This is rather theoretical in most cases, but practical implication exists when using a keyboard with extra keys not known to the engine (e.g. some "weird" international keyboards). Whether a key is supported by Ebitengine or not should not impact text entry, as the mapping from key to character is already being done by the OS.

divVerent:
Closes #2684

bug

## What this PR does | solves
It makes it possible to still receives characters from a key that Ebitengine does not know about. This can be helpful with international text entry.

### DIFF
--- a/mobile/ebitenmobileview/input_android.go
+++ b/mobile/ebitenmobileview/input_android.go
@@ -142,12 +142,12 @@ func OnKeyDownOnAndroid(keyCode int, unicodeChar int, source int, deviceID int) 
 	case source&sourceKeyboard == sourceKeyboard:
 		if key, ok := androidKeyToUIKey[keyCode]; ok {
 			keys[key] = struct{}{}
-			var runes []rune
-			if r := rune(unicodeChar); r != 0 && unicode.IsPrint(r) {
-				runes = []rune{r}
-			}
-			updateInput(runes)
 		}
+		var runes []rune
+		if r := rune(unicodeChar); r != 0 && unicode.IsPrint(r) {
+			runes = []rune{r}
+		}
+		updateInput(runes)
 	}
 }
 
@@ -163,8 +163,8 @@ func OnKeyUpOnAndroid(keyCode int, source int, deviceID int) {
 	case source&sourceKeyboard == sourceKeyboard:
 		if key, ok := androidKeyToUIKey[keyCode]; ok {
 			delete(keys, key)
-			updateInput(nil)
 		}
+		updateInput(nil)
 	}
 }
 


hajimehoshi (commented):
This change seems fine, but is this still a draft?

divVerent:
Un-drafted it now, as the parent change is in. In this case I went with marking as draft at first as I may have had to make adjustments to the preceding commit, and rebased.

Now it's properly rebased on main and thus OK to merge. I did test them together - but probably should check if any other platforms have a similar issue.

divVerent:
glfw: runes are handled entirely independently from key mapping, and thus don't depend on Ebitengine's key mapping.

js: runes are handled entirely independently from key mapping, and thus don't depend on Ebitengine's key mapping.

nintendosdk: no keyboard support at all.

So looks like other platforms are all OK - provided glfw and chrome/firefox do it right on their own.

divVerent:
glfw cocoa: insertText handler is separate event and always called through keyDown, good

glfw win32: has nasty hack for the AltGr key, is that REALLY how it's done? But there too, WM_*CHAR events are entirely separate from WM_KEYDOWN, so it should be fine.

glfw x11: both is in KeyPress event, but glfw correctly calls Xutf8LookupString etc. independently from whether glfw recognizes the keycode in _glfwInputKey.

iOS: I made sure to get this right when implementing iOS keyboard support, and that's how I found the issue on the Android side.

So seems all else are fine.

hajimehoshi (approved):
LGTM, thanks!

PR #2730, opened 2023-08-24
Correct iteration start button in gamepad example

david-poirier:

Start iterating at zero instead of gamepad id (which is usually 0).

bug

## What this PR does | solves

fixes small bug ;)


hajimehoshi (approved):
Good catch! LGTM, thanks!

PR #2816, opened 2023-10-19
example commit

yuuuuuuan:
an example about animation&keyboardcontrol&collision

## What this PR does | solves

an example about a character who can act like example-animation but it can be controlled by keyboard and collision with obstacle


hajimehoshi:
* Read https://github.com/hajimehoshi/ebiten/blob/main/CONTRIBUTING.md
* Do not add `go.mod` and `go.sum`
* Clarify the copyright of the image

Thanks,

PR #2862, opened 2023-12-16
Update main.go

AhmetAffan0:

## What this PR does | solves
It's just modified platformer for understanding jumping logic better

### DIFF
--- a/examples/platformer/main.go
+++ b/examples/platformer/main.go
@@ -79,12 +79,13 @@ type char struct {
 }
 
 func (c *char) tryJump() {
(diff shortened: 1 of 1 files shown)


AhmetAffan0:
In the platformer example, I made the jumping logic so that it would jump only once and wrote something else above, I think it turned out better.

hajimehoshi:
Hmm, I think the comment is already enough so I don't have to modify the code.

AhmetAffan0:
OK👌

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #2870, opened 2023-12-21
Polygons comments are added

lidldev:

## What this PR does | solves
I add polygons message for basic thing after I learn geometry better I will add geometry terms


hajimehoshi:
I don't think we need these comments.

Issue #2725, opened 2023-08-16
Setting a background color for the window when opening it, before starting the rendering


mat007:
### Operating System

- [X] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] Xbox
- [ ] Web Browsers

### What feature would you like to be added?

A way to control the initialization of the window when the game opens up, e.g. at least be able to set a color.

### Why is this needed?

My game intro screen has a black background, and when launching it, there is a not very pleasing white background that flashes in the window, before the first draw.

This can be reproduced by running the `paint` example with the following changes:
```
diff --git a/examples/paint/main.go b/examples/paint/main.go
index de3b5df1..362977b0 100644
--- a/examples/paint/main.go
+++ b/examples/paint/main.go
@@ -27,8 +27,8 @@ import (
 )

 const (
-       screenWidth  = 640
-       screenHeight = 480
+       screenWidth  = 1024
+       screenHeight = 768
 )

 var (
@@ -76,7 +76,7 @@ func NewGame() *Game {
        g := &Game{
                canvasImage: ebiten.NewImage(screenWidth, screenHeight),
        }
-       g.canvasImage.Fill(color.White)
+       g.canvasImage.Fill(color.Black)
        return g
 }
```
The bigger the window, the more visible it seems to be.

hajimehoshi:
I am not sure this is controllable. Let me investigate this

mat007:
Thanks @hajimehoshi !

On Windows I believe it could just be doing
```
wc.hbrBackground = …
```
in https://github.com/hajimehoshi/ebiten/blob/3b67b91bb2c4c44af7e187670e54895ede6e4b30/internal/goglfw/win32window_windows.go#L1351 passing it a brush.

But I have no clue for other platforms.

I’m also thinking I could perhaps delay the showing of the window after it’s been created somehow, to make sure there is at least of rendering that’s been done?

PR #2875, opened 2023-12-29
internal/ui: show window after first draw on glfw

theinternetftw:

#2725

Issue #2725 is written as a feature request but it looks like that request was only made because the feature would mitigate a bug it describes (white flash of window background before content is shown on Windows). This PR removes the need for the feature by resolving that bug.

## What this PR does | solves

The PR delays window.Show until after the first buffer swap, which was already the way it was done under macOS. This avoids the window appearing before its contents have been initialized by the user.

### DIFF
--- a/internal/ui/ui_glfw.go
+++ b/internal/ui/ui_glfw.go
@@ -104,6 +104,7 @@ type userInterfaceImpl struct {
 	framebufferSizeCallbackCh      chan struct{}
 
 	darwinInitOnce        sync.Once
+	showWindowOnce        sync.Once
 	bufferOnceSwappedOnce sync.Once
 
 	m sync.RWMutex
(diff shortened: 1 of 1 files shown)


internal/ui/ui_glfw.go
```diff
@@ -1321,8 +1320,17 @@ func (u *UserInterface) update() (float64, float64, error) {
...
+			return 0, 0, err
+		}
+	}
 
-			// The window is not shown at the initialization on macOS. Show the window here.
+	if u.bufferOnceSwapped {
+		var err error
+		u.showWindowOnce.Do(func() {
+			// Show the window after first buffer swap to avoid flash of white on Windows.
+			// On macOS this delay was already the behavior; it now the default for all glfw platforms.
```
hajimehoshi, line 1333:
```
// Show the window after first buffer swap to avoid flash of white especially on Windows.
```

And remove the second line comment.

hajimehoshi (commented):
Thank you!

internal/ui/ui_glfw.go
theinternetftw: Updated, cheers.

hajimehoshi (approved):
LGTM, thanks!

PR #2895, opened 2024-01-25
feat(kage): support variable shader image count

pedro-tiple:
Follow up on discussion #2882, found that uniform variables are hard-coded to four images.

Addresses a feature request.

## What this PR does | solves
These changes allow increasing `ShaderImageCount` without breaking the image display.
Simply made the uniform processing dynamic based on `ShaderImageCount`.

It doesn't actually increase the image count, as the value used there likely needs to depend on the system the code is running on. Would be happy to discuss that and make the necessary changes in a future PR.


hajimehoshi (approved):
LGTM, thanks!

PR #2903, opened 2024-02-04
add Image.SubImageInto method

### LINKED ISSUE #2902: Alloc-free way of using SubImage
### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

A way to render/draw a subimage of an image without extra allocations.

Some ways to do it:

* A new `DrawImageOptions` field that can take a subimage bounds (zero value - no subimage is needed)
* A new `DrawSubImage` function that takes the bounds in addition to the image and options

There could be other ways to solve it.

I don't really need a generic solution, but I want there to be a way to avoid that extra allocation. Then I can wrap the Image type with my own code and make this allocation go away using any means possible.

Right now I'm using an unsafe approach.

https://github.com/quasilyte/ge/blob/master/ebiten_image.go
https://github.com/quasilyte/ge/blob/7696bd01a11f683b4067c8fdf14c63ab29dee920/image_cache.go#L20
https://github.com/quasilyte/ge/blob/7696bd01a11f683b4067c8fdf14c63ab29dee920/sprite.go#L299-L309

### Why is this needed?

Let's assume that there is a user-defined Sprite type. It uses the ebitengine Image to do stuff.

Some of the use-cases of this Sprite may involve SubImage per Draw call: an underlying image represents a sprite-sheet with multiple frames (we only want to render single "frame"). This happens with animations, tilesets, user-defined atlases, etc.

Let's suppose that this Sprite calls an `Image.SubImage` for every `Draw` call (every sprite's Draw is called for every Game's Draw call). This will cause the game to potentially do hundreds of unintended heap allocations per every Draw call.

This is especially painful on **mobiles** and **wasm** targets.

Given this benchmark:

```go
package mytest

import (
    "image"
    "testing"

)

func BenchmarkSubImage(b *testing.B) {
    img := ebiten.NewImage(1, 1)
    rect := image.Rect(0, 0, 1, 1)

    b.ResetTimer() // Resets allocs as well
    for i := 0; i < b.N; i++ {
        img.SubImage(rect)
    }
}
```
We'll get these results:
```
$ go test -bench=. -benchmem .
goos: linux
goarch: amd64
pkg: github.com/quasilyte/mytest
cpu: 12th Gen Intel(R) Core(TM) i5-1235U
BenchmarkSubImage-12         8730554           128.8 ns/op         112 B/op           1 allocs/op
PASS
ok      github.com/quasilyte/mytest    1.314s
```
This allocation comes from here:
```
   38.63GB    38.63GB (flat, cum)   100% of Total
         .          .    823:func (i *Image) SubImage(r image.Rectangle) image.Image {
         .          .    824:	i.copyCheck()
         .          .    825:	if i.isDisposed() {
         .          .    826:		return nil
         .          .    827:	}
         .          .    828:
         .          .    829:	r = r.Intersect(i.Bounds())
         .          .    830:	// Need to check Empty explicitly. See the standard image package implementations.
         .          .    831:	if r.Empty() {
         .          .    832:		r = image.ZR
         .          .    833:	}
         .          .    834:
         .          .    835:	var orig = i
         .          .    836:	if i.isSubImage() {
         .          .    837:		orig = i.original
         .          .    838:	}
         .          .    839:
   38.63GB    38.63GB    840:	img := &Image{
         .          .    841:		image:    i.image,
         .          .    842:		bounds:   r,
         .          .    843:		original: orig,
         .          .    844:	}
```
And even if we would return `Image` instead of `*Image`, it would still allocate as we return it as an interface value.

Every allocation is of `sizeof(ebiten.Image)`.

quasilyte:
Instead of allocating the result on heap for the user, allow the caller side to decide whether it should be heap-allocated or stack-allocated.

This new method allows us to have a zero-alloc SubImage routine for the use cases where we don't want to make an extra allocation.

The performance difference is measurable (comparison is done with benchstat):

	name             old time/op    new time/op    delta
	SubImage-12     203ns ± 2%      33ns ± 2%   -83.91%  (p=0.000 n=20+19)

	name             old alloc/op   new alloc/op   delta
	SubImage-12      112B ± 0%        0B       -100.00%  (p=0.000 n=20+20)

Using SubImage is `~203ns`, SubImageInto is `~33ns`. SubImage does 1 allocation, SubImageInto does none.

SubImage is rewritten in a way to avoid the code duplication. It should work identically to the previous contract (it returns nil if `i` is disposed, etc.)

Fixes #2902

https://github.com/hajimehoshi/ebiten/issues/2902

Feature.

## What this PR does | solves

It introduces a more efficient way of making subimages.

### DIFF
--- a/image.go
+++ b/image.go
@@ -827,9 +827,37 @@ func (i *Image) DrawRectShader(width, height int, shader *Shader, options *DrawR
 // Successive uses of multiple various regions as rendering destination is still efficient
 // when all the underlying images are the same, but some platforms like browsers might not work efficiently.
 func (i *Image) SubImage(r image.Rectangle) image.Image {
+	var s Image
+	if !i.SubImageInto(&s, r) {
+		return nil
+	}
+	return &s
+}
+
+// SubImageInto is like SubImage, but it returns the result through the s parameter.
+// The s image will be initialized to a result of taking the SubImage from i.
+//
+// This code is semantically equivalent to:
+//
+//	s := i.SubImage(r)
+//
+// You might want to use this method to avoid a new Image object heap allocations.
+// Use it like this:
+//
+//	var s ebiten.Image
+//	i.SubImageInto(&s, r)
+//
+// And then use s as your sub image.
+//
+// This method returns false if i is disposed.
+func (i *Image) SubImageInto(s *Image, r image.Rectangle) bool {
+	// Try to keep this method zero alloc.
+	// There is a test for that (TestSubImageIntoZeroAlloc).
+	// There is also a benchmark for the comparison.
+
 	i.copyCheck()
 	if i.isDisposed() {
-		return nil
+		return false
 	}
 
 	r = r.Intersect(i.Bounds())
@@ -843,14 +871,14 @@ func (i *Image) SubImage(r image.Rectangle) image.Image {
 		orig = i.original
 	}
 
-	img := &Image{
+	*s = Image{
 		image:    i.image,
 		bounds:   r,
 		original: orig,
 	}
-	img.addr = img
+	s.addr = s
 
-	return img
+	return true
 }
 
 // Bounds returns the bounds of the image.
--- a/image_test.go
+++ b/image_test.go
@@ -4582,3 +4582,29 @@ func TestImageDrawImageAfterDeallocation(t *testing.T) {
 		}
 	}
 }
+
+func TestSubImageIntoZeroAlloc(t *testing.T) {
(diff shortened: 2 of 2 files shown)


hajimehoshi (changes requested):
SubImageInto breaks an assumption that an image's size never changes. So unfortunately I cannot accept this change.

We need more discussion.

PR #2906, opened 2024-02-08
add CursorPositionF()

setanarut:

`CursorPositionF()` Provides float64 Cursor position values directly without needing the `float64 -> int -> float64` conversion.

```Go
func CursorPositionF() (x, y float64) {
	return theInputState.cursorPosition()
}
```

### DIFF
--- a/input.go
+++ b/input.go
@@ -87,6 +87,18 @@ func CursorPosition() (x, y int) {
 	return int(cx), int(cy)
 }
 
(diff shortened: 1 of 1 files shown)


hajimehoshi:
Before adding a public API, please file an issue and discuss it.

setanarut:
there are no other participants to discuss the issue except you. You are the only moderator and you know the subject now.

hajimehoshi:
In order to track what happens in a milestone, a filed issue is needed.

setanarut:
oh sorry. I understand

Issue #2755, opened 2023-09-11
internal/shader: support shift operators for ints and int vectors


hajimehoshi:
### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] Xbox
- [ ] Web Browsers

### What feature would you like to be added?

`<<` is available at e.g. GL ES 3.0.

RiXelanya:
I am interested in doing this.

hajimehoshi:
PRs are always welcome!

PR #2916, opened 2024-02-26
Kage update support for shift operators

aoyako:
Updates #2755

feature

## What this PR does | solves
1. Add type validation for left and right shifts.
2. Add support for the shift assignments operation.
According to the docs, it should work everywhere:
[HLSL](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-operators#bitwise-operators)
[GLSL](https://registry.khronos.org/OpenGL/specs/es/3.0/GLSL_ES_Specification_3.00.pdf#page=67)
[MSL](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf#page=75)


internal/shader/syntax_test.go
```diff
@@ -1314,6 +1314,149 @@ func Fragment(dstPos vec4, srcPos vec2, color vec4) vec4 {
 	}
 }
 
+// Issue: #2755
+func TestSyntaxOperatorShift(t *testing.T) {
+	cases := []struct {
+		stmt string
+		err  bool
+	}{
+		{stmt: "a := 1 << 2; _ = a", err: false},
```
hajimehoshi, line 1323: Does this test work?

```go
{stmt: "a := 1.0 << 2.0; _ = a", err: false},
```

hajimehoshi (commented):
Thanks!

internal/shader/syntax_test.go
aoyako: No, it doesn't. In `ResolveUntypedConstsForBinaryOp` it is resolved to `(float << float)`.
The same behavior could be observed with other binary operators, such as `^`, `&`, and  `|`.
However, op + assignment can truncate to integer, meaning `a := 1; a <<= 2.0; _ = a` is valid.

I will fix that if it is expected not to alert an error in `a := 1 << 2.0` case.

hajimehoshi: These are expected to work without errors. Please fix them. Thanks!

https://go.dev/play/p/ZBrITUbJjlN

hajimehoshi: The interesting thing is, a strict type checking for constants is valid for `%`, but not for `<<` or `>>`.

https://go.dev/play/p/kF_j7sWFfCU

hajimehoshi: Today I found even `go vet` failed to treat this correctly! https://github.com/golang/go/issues/65939

internal/shader/syntax_test.go
```diff
@@ -1320,12 +1320,30 @@ func TestSyntaxOperatorShift(t *testing.T) {
...
+		// {stmt: "s := 1; var a float = float(1.0 << s); _ = a", err: true},
+		// {stmt: "s := 1; var a float = 1.0 << s; _ = a", err: true},
+		// {stmt: "s := 1; var a float = 1 << s; _ = a", err: true},
+		// {stmt: "s := 1; var a int = 1.0 << s; _ = a", err: false},
+		// {stmt: "s := 1; var a int = 1 << s; _ = a", err: false},
+
+		{stmt: "var a float = 1.0 << 2.0; _ = a", err: false},
+		{stmt: "var a int = 1.0 << 2; _ = a", err: false},
+		{stmt: "var a float = 1.0 << 2; _ = a", err: false},
+		{stmt: "var a = 1.0 << 2; _ = a", err: false},
```
hajimehoshi, line 1339: Isn't this the same as `a := 1.0 << 2`?

internal/shader/syntax_test.go
```diff
@@ -1320,12 +1320,30 @@ func TestSyntaxOperatorShift(t *testing.T) {
...
+		// {stmt: "s := 1; var a int = int(1 << s); _ = a", err: false},
+		// {stmt: "s := 1; var a int = int(1.0 << s); _ = a", err: false},
+		// {stmt: "s := 1; a := 1 << s; _ = a", err: false},
+		// {stmt: "s := 1; a := 1.0 << s; _ = a", err: true},
+		// {stmt: "s := 1; a := int(1.0 << s); _ = a", err: false},
+		// {stmt: "s := 1; var a float = float(1.0 << s); _ = a", err: true},
+		// {stmt: "s := 1; var a float = 1.0 << s; _ = a", err: true},
+		// {stmt: "s := 1; var a float = 1 << s; _ = a", err: true},
+		// {stmt: "s := 1; var a int = 1.0 << s; _ = a", err: false},
+		// {stmt: "s := 1; var a int = 1 << s; _ = a", err: false},
```
hajimehoshi, line 1334: I assume this is still a work in progress. Please let me know if this is ready to review.

hajimehoshi (commented):
Another tricky case:

```go
{stmt: "var a = 1; b := a << 2.0; _ = b", err: false},
{stmt: "var a = 1; b := 2.0 << a; _ = b", err: true},
```

https://go.dev/play/p/qNT1jMrNgCe

internal/shader/syntax_test.go
aoyako: Yes, if used inside a function body.
https://go.dev/play/p/I5OIJe7WjOD

hajimehoshi: So this is redundant. We already have `{stmt: "a := 1.0 << 2; _ = a", err: false}`.

hajimehoshi:
Any updates? I'll take over this if you are not available. Thanks!

aoyako:
I am sorry, it took more time than I expected. I will upload updates today.

internal/shaderir/program.go
```diff
@@ -109,6 +111,7 @@ type Expr struct {
 	Swizzling   string
 	Index       int
 	Op          Op
+	Ast         ast.Expr
```
aoyako, line 114: `Expr` are stored by value. Is there a better way to search for a specific statement without pointers?

internal/shaderir/program.go
```diff
@@ -74,16 +75,17 @@ type Block struct {
...
+	Exprs         []Expr
+	Blocks        []*Block
+	ForVarType    Type
+	ForVarIndex   int
+	ForInit       constant.Value
+	ForEnd        constant.Value
+	ForOp         Op
+	ForDelta      constant.Value
+	InitIndex     int
+	IsTypeGuessed bool
```
aoyako, line 88: If it is a short declaration `:=`, type is implicit.

hajimehoshi: shaderir.Stmt itself doesn't have a type information, so having `IsTypeGuessed` here is odd.

hajimehoshi:
I'll review this later, but I feel like we don't have to introduce complex things such as `delayedTypeValidator`. We should be able to do with much smaller modifications.

hajimehoshi:
I think I found an incosistent behavior in the Go compiler (https://go.dev/play/p/pCHHBoblbAY)

```go
package main

import "fmt"

func main() {
	s := 1
	a := 2.0 << s
	b := 2.0<<s + s<<3.0
	fmt.Println(a, b)
}
```

I'll report this. TBH I don't know how to treat this.

hajimehoshi:
I reported the above issue https://github.com/golang/go/issues/66071. Let's wait for the Go team's answer. @aoyako, your test catched this corner case. Great job!

aoyako:
> I think I found an incosistent behavior in the Go compiler (https://go.dev/play/p/pCHHBoblbAY)
> 
> ```go
> package main
> 
> import "fmt"
> 
> func main() {
> 	s := 1
> 	a := 2.0 << s
> 	b := 2.0<<s + s<<3.0
> 	fmt.Println(a, b)
> }
> ```
> 
> I'll report this. TBH I don't know how to treat this.

I think it is an expected behaviour.
- In the first case:
2.0 would be a float because it would be so in `a := 2.0`. Therefore, the left operand for shift is float for `a := 2.0 << s` and thus it will fail.
- In the second case:
The result of `s << 3.0` is int, so when guessing a type for 2.0 compiler should try `b := 2.0 + int` implying 2.0 to be converted to an integer type. Therefore, a constant 2.0 is treated as an integer.
- However, I don't know if s << 3.0 should give a panic for `go vet`.

hajimehoshi:
> The result of s << 3.0 is int, so when guessing a type for 2.0 compiler should try b := 2.0 + int implying 2.0 to be converted to an integer type. 

You meant `2.0<<s + int`?

hajimehoshi:
> However, I don't know if s << 3.0 should give a panic for go vet.

This was another bug https://github.com/golang/go/issues/65939

hajimehoshi:
I tested a + operator with an integer, and apparently you are right... 🤔 

```go
package main

import "fmt"

func main() {
	a := 1
	b := 1.0<<a + 0
	c := 1.0<<a + int(0)
	fmt.Println(b, c)
}
```

aoyako:
> > The result of s << 3.0 is int, so when guessing a type for 2.0 compiler should try b := 2.0 + int implying 2.0 to be converted to an integer type.
> 
> You meant `2.0<<s + int`?

Yes, it is a final expression.
According to the docs:
> If the left operand of a non-constant shift expression is an untyped constant, it is first implicitly converted to the type it would assume if the shift expression were replaced by its left operand alone.

The compiler will try to assume a type in the expression `b := 2.0 + int` first.

aoyako:
> I tested a + operator with an integer, and apparently you are right... 🤔
> 
> https://go.dev/play/p/OHnUA3weL8w
> 
> ```go
> package main
> 
> import "fmt"
> 
> func main() {
> 	a := 1
> 	b := 1.0<<a + 0
> 	c := 1.0<<a + int(0)
> 	fmt.Println(b, c)
> }
> ```

It is funny you can't easily define a float variable equal to a power of 2 using shifts: https://go.dev/play/p/qvk1sHyIOhK
```
package main
import "fmt"

func main() {
	s := 1
	floatPowerOfTwo1 := float32(1<<s + int(0))
	floatPowerOfTwo2 := float32(1 << s) // Fail
	fmt.Println(floatPowerOfTwo1, floatPowerOfTwo2)
}

```

hajimehoshi:
> `floatPowerOfTwo2 := float32(1 << s) // Fail`

This blowed my mind! 🤯 

Hmm, ok. Even if this is an expected behavior, I now doubt it'd be worth emulating the authentic Go behavior faithfully: I feel like it is OK to do a simpler type resolusion. Let me think more. I'd like to wait for the Go team answer anyway.

hajimehoshi:
Sorry for keeping you waiting. Now I don't think we have to follow the Go's current specification in a very strict way. So, we can change the Kage behavior like

```go
{stmt: "var a = 1; b := a << 2.0; _ = b", err: false},
{stmt: "var a = 1; b := 2.0 << a; _ = b", err: false},
```

so that we no longer need the delayed type determination. We can treat shift operators as the other binary operators especiallyl like a mod operator which takes only integers. Thoughts?

hajimehoshi:
https://github.com/golang/go/issues/66071#issuecomment-1974744117

One possible reason why the shift operators behave like this is a natural type deduction for constant values, but as Kage doesn't have int32 and int64, we don't need this.

```go
var n = 32
var y = int64(1 << n) // y should be int64 whichever the archtecture is 32bit or 64bit
```

aoyako:
> Sorry for keeping you waiting. Now I don't think we have to follow the Go's current specification in a very strict way. So, we can change the Kage behavior like
> 
> ```go
> {stmt: "var a = 1; b := a << 2.0; _ = b", err: false},
> {stmt: "var a = 1; b := 2.0 << a; _ = b", err: false},
> ```
> 
> so that we no longer need the delayed type determination. We can treat shift operators as the other binary operators especially like a mod operator which takes only integers. Thoughts?

Thanks for the confirmation. It makes sense to use simpler rules since Kage uses the same integer type.
I also wondered whether allowing untyped float constants in shift operators is a good idea in the first place. However, keeping rules the same for all operators will be much simpler.

I will update the PR with these new rules then. It might also be worth adding a compile error when shifting by a constant causes overflow.

hajimehoshi:
> I also wondered whether allowing untyped float constants in shift operators is a good idea in the first place. However, keeping rules the same for all operators will be much simpler.

I confirmed that % works as following:

```
	println(1.0 % a) // works
	println(a % 1.0) // works
	println(1.0 % 1.0) // doesn't work
```

https://go.dev/play/p/R2MFM9f0Cxe

I think doing the same thing with % would make sense.

> It might also be worth adding a compile error when shifting by a constant causes overflow.

As the integer size is not determined, can we detect overflow?

aoyako:
> As the integer size is not determined, can we detect overflow?

In [HLSL](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-scalar) and [MSL](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf#page=22) the max size for ints is 64 bits. 32 bits are not guaranteed in [GLSL](https://registry.khronos.org/OpenGL/specs/es/3.0/GLSL_ES_Specification_3.00.pdf#page=31) for non-high-precision numbers, and it won't apply here.
Also, it might be unnecessary since the API might change and cause outdated restrictions on 64 bits for integers.

hajimehoshi:
> Also, it might be unnecessary since the API might change and cause outdated restrictions on 64 bits for integers.

IMO the Kage compiler doesn't have to detect the overflow and we can let the drivers how to treat them.

hajimehoshi:
I realized `1.0 << 1.0` worked (https://go.dev/play/p/wTSIXouoc-9), so this is different from the mod operation. I think all these cases should work

```go
	a := 1
	println(1.0 << a)
	println(a << 1.0)
	println(1.0 << 1.0)
```

Sorry for going back and forth

aoyako:
These cases are included:
```go
{stmt: "a := 1.0 << 2.0; _ = a", err: false},
{stmt: "var a = 1; b := a << 2.0; _ = b", err: false},
{stmt: "var a = 1; b := 2.0 << a; _ = b", err: false}, // PR: #2916
```

hajimehoshi:
Thanks, I'll take a look later.

internal/shader/expr.go
```diff
@@ -149,6 +149,13 @@ func (cs *compileState) parseExpr(block *block, fname string, expr ast.Expr, mar
 				v = gconstant.MakeBool(b)
 			case token.EQL, token.NEQ, token.LSS, token.LEQ, token.GTR, token.GEQ:
 				v = gconstant.MakeBool(gconstant.Compare(lhs[0].Const, op, rhs[0].Const))
+			case token.SHL, token.SHR:
+				shift, ok := gconstant.Int64Val(rhs[0].Const)
+				if !ok {
+					cs.addError(e.Pos(), fmt.Sprintf("unexpected %s type for: %s", rhs[0].Const.String(), e.Op))
+				} else {
+					v = gconstant.Shift(lhs[0].Const, op, uint(shift))
+				}
```
hajimehoshi, line 158: Let's use an early return for consistency.

```go
				if !ok {
					cs.addError(e.Pos(), fmt.Sprintf("unexpected %s type for: %s", rhs[0].Const.String(), e.Op))
					return nil, nil, nil, false
				}
				v = gconstant.Shift(lhs[0].Const, op, uint(shift))
```

internal/shaderir/check.go
```diff
@@ -98,6 +111,13 @@ func TypeFromBinaryOp(op Op, lhst, rhst Type, lhsConst, rhsConst constant.Value)
 			return Type{}, false
 		}
 
+		if op == LeftShift || op == RightShift {
+			if lhsConst.Kind() == constant.Int && rhsConst.Kind() == constant.Int {
+				return Type{Main: Int}, true
+			}
+			return Type{}, false
```
hajimehoshi, line 118: Even if one of the consts are typed, if both can be truncated to integers, this should be fine (https://go.dev/play/p/o5hJUTxFbYK). Does this test pass?

```go
{stmt: "1.0 << int(1)", err: false}
{stmt: "int(1) << 1.0", err: false}
```

aoyako: Yes, if one of the constants is typed, the second will be deduced similarly to other operators. The only difference is when both of them are untyped const, operands will be treated as integers.
```go
{stmt: "a := 1.0 << int(2); _ = a", err: false}, // 1.0 becomes int
{stmt: "a := int(1) << 2.0; _ = a", err: false}, // 2.0 becomes int
{stmt: "a := int(1) << 2.0; _ = a", err: }, // 2.0 become sint
{stmt: "a := 1.0 << 2.0; _ = a", err: false}, // Both are ints
{stmt: "a := 1.0 << float(2.0); _ = a", err: true}, // 1.0 becomes float, TypeFromBinaryOp will fail for float << float
```

hajimehoshi (approved):
LGTM, thanks!

PR #2943, opened 2024-03-26
internal/gamepaddb: generate respective controller dbs for each platform

### LINKED ISSUE #2936: internal/gamepaddb: split the text file for each platform
### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

Now the text file is about 525 kb. Splitting this file and keeping a necessary part of it for each platform should reduce the executable binary sizes.

### Why is this needed?

The smaller executable size, the happier we are.

kungfukennyg:
Closes #2936 

feature

## What this PR does | solves
Replace singular embedded controller database file with separate generated files for each supported platform. The controller database file is some ~500KB, so this change should reduce the overall binary size per platform.

Manual validation was performed to ensure that `gen_dbs.go#splitControllerDb` outputs the same controllers per platform as listed in the text file. I didn't see an easy way to bake this into an automated test without rewriting the file parsing logic but I am happy to take another stab at it if preferred. In the interest of keeping the surface impact of this change small I did not modify any of the existing code in `gamepaddb.go` that intakes and parses the controller database.


kungfukennyg:
Got it. For point 2, there is a go generate tag right above the main function. I can move it higher in the file for visibility if that helps.

hajimehoshi:
> For point 2, there is a go generate tag right above the main function.

Unfortunately this is not called by `go generate` as this file is marked as `ignored`.

Thanks!

kungfukennyg:
Ah I see. I've made those changes!

hajimehoshi:
Thanks, I'll take a look tomorrow!

internal/gamepaddb/gamepaddb.go
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:generate go run gen.go
 package gamepaddb
```
hajimehoshi, line 16: Add `go fmt -s -w .` and add an empty line above `package gamepaddb`.

```go
//go:generate go run gen.go
//go:generate go fmt -s -w .

package gamepaddb
```

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,183 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build ignore
+
+// This program generates platform-specific controller dbs. It can be invoked by running
+// go generate
```
hajimehoshi, line 18: We don't need this comments.

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,183 @@
...
+
+// To update the database file, run:
+//
+//     curl --location --remote-name https://raw.githubusercontent.com/mdqinc/SDL_GameControllerDB/master/gamecontrollerdb.txt
+
+//go:embed gamecontrollerdb.txt
+var gamecontrollerdb_txt []byte
+
+const db_template string = `
+{{.License}}
```
hajimehoshi, line 56: Snake cases are not a convention in Ebitengine. Rename `db_template` to `dbTemplate`. Also, remove the first new line.

```
const dbTemplate = `{{.License}}
```

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,183 @@
...
+	if err := Update(controllerBytes); err != nil {
+		panic(err)
+	}
+{{if eq .Platform "windows" }} 
+    if err := Update(glfwControllerBytes); err != nil {
+        panic(err)
+    }
+{{end}}}`
+
+func main() {
```
hajimehoshi, line 82:
```go
func main() {
    if err := run(); err != nil {
        log.Fatal(err)
    }
}

func run() error {
```

and let run return error.

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,183 @@
...
+const db_template string = `
+{{.License}}
+
+{{.DoNotEdit}}
+
+{{.BuildTag}}
+
+package gamepaddb
+
+import _ "embed"
```
hajimehoshi, line 64: I prefer

```go
import (
    _ "embed"
)
```

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,183 @@
...
+var glfwControllerBytes []byte
+{{end}}
+func init() {
+	if err := Update(controllerBytes); err != nil {
+		panic(err)
+	}
+{{if eq .Platform "windows" }} 
+    if err := Update(glfwControllerBytes); err != nil {
+        panic(err)
+    }
```
hajimehoshi, line 79: Use tab characters

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,183 @@
...
+var controllerBytes []byte
+{{if eq .Platform "windows" }} 
+//go:embed gamecontrollerdb_glfw.txt
+var glfwControllerBytes []byte
+{{end}}
+func init() {
+	if err := Update(controllerBytes); err != nil {
+		panic(err)
+	}
+{{if eq .Platform "windows" }}
```
hajimehoshi, line 76: Why is this only for Windows?

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+	}
+}
+
+// splitControllerDb maps the game controller db into respective controller definition byte slices for
+// each platform
+func splitControllerDb(controllerDb []byte) map[string][]byte {
+	dbs := map[string][]byte{}
+	// split at #, skip first 3 header chunks
+	chunks := bytes.Split(controllerDb, []byte{'#'})[3:]
```
hajimehoshi, line 168: Just skipping certain amount of lines seems fragile.

Let's use `bytes.Reader` and `bufio.Scanner` to process each line.

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,183 @@
...
+			buildTag: "//go:build ios",
+		},
+		"Android": {
+			name:     "android",
+			buildTag: "//go:build android",
+		},
+	}
+
+	// divide controller dbs into chunks by platform
+	controllerDbs := splitControllerDb(gamecontrollerdb_txt)
```
hajimehoshi, line 115: Use `DBs` instead of `Dbs` to follow the Go convention.

https://google.github.io/styleguide/go/decisions.html#initialisms

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,183 @@
...
+        panic(err)
+    }
+{{end}}}`
+
+func main() {
+	// Follow the standard comment rule (https://pkg.go.dev/cmd/go#hdr-Generate_Go_files_by_processing_source).
+	doNotEdit := "// Code generated by gen.go using 'go generate'. DO NOT EDIT."
+
+	type gamePadPlatform struct {
+		name     string
```
hajimehoshi, line 87: `filenameSuffix` sounds better

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,183 @@
...
+		buildTag string
```
hajimehoshi, line 88: `buildConstraints` sounds better

https://pkg.go.dev/cmd/go#hdr-Build_constraints

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,183 @@
...
+
+//go:embed gamecontrollerdb.txt
+var gamecontrollerdb_txt []byte
+
+const db_template string = `
+{{.License}}
+
+{{.DoNotEdit}}
+
+{{.BuildTag}}
```
hajimehoshi, line 60: `{{.BuildConstraints}}`

internal/gamepaddb/gen.go
kungfukennyg: I misunderstood what platforms GLFW was supported on, will fix.

kungfukennyg:
I've made the requested changes. I've been writing too much rust, snake case keeps sneaking through 😅

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,218 @@
...
+	for sdlName, platform := range supported {
+		controllerDB, ok := controllerDBs[sdlName]
+		if !ok {
+			return fmt.Errorf("failed to find controller db for platform %s in gamecontrollerdb_txt", sdlName)
+		}
+
+		// write each chunk into separate text file for embedding into respective generated files
+		txtFile, err := os.Create(fmt.Sprintf("gamecontrollerdb_%s.txt", platform.filenameSuffix))
+		if err != nil {
+			return fmt.Errorf("failed to open file: %v", err)
```
hajimehoshi, line 133: Just returning err is fine

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+		defer txtFile.Close()
+		written, err := txtFile.Write(controllerDB)
+		if err != nil {
+			return fmt.Errorf("failed to write controller db for %s, expected to write %d bytes, wrote %d: %v", sdlName, len(controllerDB), written, err)
```
hajimehoshi, line 138: ditto

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+
+		path := fmt.Sprintf("db_%s.go", platform.filenameSuffix)
+		tmpl, err := template.New(path).Parse(dbTemplate)
+		if err != nil {
+			return fmt.Errorf("failed to parse template: %v", err)
```
hajimehoshi, line 144: ditto

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,218 @@
...
+			buildConstraints: "//go:build android",
+		},
+	}
+
+	controllerDBs, err := splitControllersByPlatform(gameControllerDB)
+	if err != nil {
+		return err
+	}
+
+	for sdlName, platform := range supported {
```
hajimehoshi, line 124: s/sdlName/sdlPlatformName/g

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,218 @@
...
+func run() error {
+	// Follow the standard comment rule (https://pkg.go.dev/cmd/go#hdr-Generate_Go_files_by_processing_source).
+	doNotEdit := "// Code generated by gen.go using 'go generate'. DO NOT EDIT."
+
+	type gamePadPlatform struct {
+		filenameSuffix   string
+		buildConstraints string
+	}
+
+	supported := map[string]gamePadPlatform{
```
hajimehoshi, line 96: s/supported/platforms/g

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	for sdlName, platform := range supported {
+		controllerDB, ok := controllerDBs[sdlName]
+		if !ok {
+			return fmt.Errorf("failed to find controller db for platform %s in gamecontrollerdb_txt", sdlName)
+		}
+
+		// write each chunk into separate text file for embedding into respective generated files
+		txtFile, err := os.Create(fmt.Sprintf("gamecontrollerdb_%s.txt", platform.filenameSuffix))
```
hajimehoshi, line 131: Use `os.WriteFile`

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,218 @@
...
+				currentPlatform = "Mac OS X"
+			case "iOS":
+				currentPlatform = "iOS"
+			case "Android":
+				currentPlatform = "Android"
+			case "Linux":
+				currentPlatform = "Linux"
+			default:
+				continue
+			}
```
hajimehoshi, line 204: We don't need this switch-cases. Accept the written platform as it is.

kungfukennyg:
Done!

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,190 @@
...
+			filenameSuffix:   "macos_darwin",
+			buildConstraints: "//go:build darwin && !ios",
+		},
+		"Linux": {
+			filenameSuffix:   "linbsd",
+			buildConstraints: "//go:build (freebsd || (linux && !android) || netbsd || openbsd) && !nintendosdk && !playstation5",
+		},
+		"iOS": {
+			filenameSuffix:   "ios",
+			buildConstraints: "//go:build ios",
```
hajimehoshi, line 111: We don't need this. The filename suffix is already the constraint.

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	platforms := map[string]gamePadPlatform{
+		"Windows": {
+			filenameSuffix:   "windows",
+			buildConstraints: "//go:build windows && !microsoftgdk",
+		},
+		"Mac OS X": {
+			filenameSuffix:   "macos_darwin",
+			buildConstraints: "//go:build darwin && !ios",
```
hajimehoshi, line 103:
```go
//go:build !ios
```

is enough.

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,190 @@
...
+			filenameSuffix:   "linbsd",
+			buildConstraints: "//go:build (freebsd || (linux && !android) || netbsd || openbsd) && !nintendosdk && !playstation5",
+		},
+		"iOS": {
+			filenameSuffix:   "ios",
+			buildConstraints: "//go:build ios",
+		},
+		"Android": {
+			filenameSuffix:   "android",
+			buildConstraints: "//go:build android",
```
hajimehoshi, line 115: We don't need this. The filename suffix is already the constraint.

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,190 @@
...
+package gamepaddb
+
+import (
+	_ "embed"
+)
+
+//go:embed gamecontrollerdb_{{.FileNameSuffix}}.txt
+var controllerBytes []byte
+//go:embed gamecontrollerdb_glfw.txt
+var glfwControllerBytes []byte
```
hajimehoshi, line 69: This file is only for GLFW environments, right? We don't have to 'generate' this part. We should have `db_glfw.go` with the build constraint `//go:build !android && !ios && !js && !microsoftgdk && !nintendosdk && !playstation5`. `db_glfw.go` would be very similar to `db.go`. The current `db.go` doesn't have the constaints for android and ios, but I think we need this constraints as GLFW is only for desktops. We don't have to generate `db_glfw.go` by gen.go.

hajimehoshi (commented):
Good job!

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,190 @@
...
+	for s.Scan() {
+		line := s.Text()
+		if len(line) == 0 {
+			continue
+		}
+
+		if line[0] == '#' {
+			currentPlatform = strings.Replace(line, "# ", "", 1)
+			continue
+		}
```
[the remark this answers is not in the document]
kungfukennyg: Ah yeah that's a better approach, I didn't realize the platform string was on every controller line. Will do.

internal/gamepaddb/gen.go
kungfukennyg: Are you suggesting I remove `buildConstraints` from the iOS and Android defintions here, and then modify the template to set the build directive to e.g. `//go:build {{.FileNameSuffix}}` in the case it's not set?

hajimehoshi: I mean

```go
"iOS": {
			filenameSuffix:   "ios",
}
```

is fine. We don't have to modify the template. `db_ios.go` doesn't need the constaraints since the file name already indicates the constraint.

> If a file's name, after stripping the extension and a possible _test suffix, matches any of the following patterns:
>
> *_GOOS
> *_GOARCH
> *_GOOS_GOARCH

https://pkg.go.dev/cmd/go#hdr-Build_constraints

internal/gamepaddb/gen.go
```diff
@@ -0,0 +1,184 @@
...
+
+		parts := strings.Split(line, ",")
+		var platform string
+		for _, part := range parts {
+			if strings.HasPrefix(part, "platform:") {
+				platform = strings.Replace(part, "platform:", "", 1)
+				dbs[platform] += line + "\n"
+				break
+			}
```
hajimehoshi, line 176:
```go
for _, part := range strings.Split(line, ",") {
    if platform, found := strings.CutPrefix(part, "platform:"); found {
        dbs[platform] += line + "\n"
        break
    }
}
```

hajimehoshi (approved):
Thank you!

hajimehoshi (changes requested):
Sorry, I'll approve after the above one is fixed

kungfukennyg:
Done!

hajimehoshi (approved):
LGTM, thanks!

hajimehoshi:
FYI: I committed a change that I missed in this PR review https://github.com/hajimehoshi/ebiten/commit/24238e16af65be8f5813fda2fa4f1c6a5e62d905

Thank you for your great work, @kungfukennyg!

kungfukennyg:
Sweet, thanks!

Issue #2930, opened 2024-03-16
Support Multiple render target (MRT)


Zyko0:
### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

### What feature would you like to be added?

I believe it could be a great addition if ebitengine could support [MRT](https://en.wikipedia.org/wiki/Multiple_Render_Targets).

At the moment we can write to a single `dst` image, and pass multiple `src` image (to a `DrawTriangles`/`DrawTrianglesShader`) function, writing to multiple `dst` images with the same function call (and with the same internal draw call) would be nice for some specific usecases.

On software side, something like:
```go
DrawTrianglesShadersMRT(dst [4]*ebiten.Image, vertices []ebiten.Vertex, indices []uint16, opts *DrawTrianglesShaderOptions
```

On kage side:
```go
func Fragment(dst vec4, src vec2, color vec4) (vec4, vec4, vec4) {
    var common float
    // heavy calculations on common
    mask0 := Mask0(common)
    colorOut := ColorOut(common)
    dataOut := GetData(common)
    
    return mask0, colorOut, dataOut
}
```

### Why is this needed?

I've had many usecases where 80% of the heavy computations made within a shader invocation are needed for multiple destination images.
Having the possibility to re-use the same vertices, the same shader draw call, the same 80% initial work that is common to all destination images could give new possibilities.

So far, in order to do so, we need:
- To write multiple shaders with repeated code, or a single one with uniforms branches
- To make distinct calls for each of the destination images (and each of these shaders/new uniforms => no batching), even though we use the same vertices and most of the same shader code (and results of the same calculations) => this obviously forces the re-invocation of a fragment shader for the same pixels 2-3 times, and the computations that come with those

One of the most common use nowadays, is for 3D pipelines where you process some geometry or any kind of maths for meshes / triangles, and you want to write to multiple offscreens at once (https://learnopengl.com/Advanced-Lighting/Deferred-Shading):
- Diffuse (albedo), Normal, Depth, Specular, UVs, etc..

If this can be supported, it would certainly unlock new rendering potential for Ebitengine, even for 2D/2.5D workflows I believe.
[...]

PR #2953, opened 2024-04-06
WIP: MRT Proof of concept

Zyko0:
## What issue is this addressing?

https://github.com/hajimehoshi/ebiten/issues/2930

## What this PR does | solves

This is a proof of concept of MRT (multiple render target), to investigate if the concept could be integrated to ebitengine, this is not a merge candidate for now.

`go run ./examples/mrt`
image

```go
dsts = [4]*ebiten.Image{
		ebiten.NewImageWithOptions(image.Rect(0, 0, 128, 128), &ebiten.NewImageOptions{Unmanaged: true}),
		ebiten.NewImageWithOptions(image.Rect(0, 0, 128, 128), &ebiten.NewImageOptions{Unmanaged: true}),
		ebiten.NewImageWithOptions(image.Rect(0, 0, 128, 128), &ebiten.NewImageOptions{Unmanaged: true}),
		ebiten.NewImageWithOptions(image.Rect(0, 0, 128, 128), &ebiten.NewImageOptions{Unmanaged: true}),
}
shaderSrc = []byte(`
//kage:units pixels

package main

func Fragment(dst vec4, src vec2, color vec4) (vec4, vec4, vec4, vec4) {
	return vec4(1,0,0,1), vec4(0,1,0,1), vec4(0,0,1,1), vec4(1,0,1,1)
}
`

{
      ebiten.DrawTrianglesShaderMRT(dsts, vertices, indices, s, nil)
      // screen.Draw all 4 images to assess that they've been written to with the same geometry / draw call
}
```

- [x] OpenGL
- [x] DirectX 11
- [x] DirectX 12
- [x] WebGL => A bit more problematic due to the glsl compatibility between desktop
- [ ] Metal

Notes:
- It only seems to make sense with unmanaged `*ebiten.Image` (as in different internal textures).
  - dst images could be moved to separate atlases if they're not ebiten.unmanaged: https://github.com/hajimehoshi/ebiten/issues/2930#issuecomment-2041157459
  - dst images, if moved to new textures or existing (available) ones should share the same region (rectangle) as they originally do, because of scissor, viewport definition but mostly because the triangles define the logical destination position, moving to an arbitrary region would make the triangles target the wrong rectangle
  - dst images should have the same sizes probably, since the triangles, scissor and viewport should target the same region for all destination textures
- I set the max dst images count to 8, since 8 seems to be the guaranteed minimum by all drivers
- The example found in `examples/mrt` is probably to be removed and changed for a more relevant example, I put it there so that we can see a minimal reproduction of the feature temporarily


examples/mrt/main.go
```diff
@@ -0,0 +1,134 @@
+// Copyright 2024 The Ebiten Authors
```
hajimehoshi, line 1: Ebitengine

hajimehoshi:
Please rebase your PR, thanks!

internal/atlas/image.go
```diff
@@ -440,6 +440,27 @@ func (i *Image) DrawTriangles(srcs [graphics.ShaderSrcImageCount]*Image, vertice
 	i.drawTriangles(srcs, vertices, indices, blend, dstRegion, srcRegions, shader, uniforms, fillRule)
 }
 
+func DrawTrianglesMRT(dsts [graphics.ShaderDstImageCount]*Image, srcs [graphics.ShaderSrcImageCount]*Image, vertices []float32, indices []uint32, blend graphicsdriver.Blend, dstRegion image.Rectangle, srcRegions [graphics.ShaderSrcImageCount]image.Rectangle, shader *Shader, uniforms []uint32, fillRule graphicsdriver.FillRule) {
```
hajimehoshi, line 443: Why not unifying DrawTriangles with this?

image.go
```diff
@@ -708,6 +708,139 @@ func (i *Image) DrawTrianglesShader(vertices []Vertex, indices []uint16, shader
...
+//
+// When a specified image is non-nil and is disposed, DrawTrianglesShader panics.
+//
+// If a specified uniform variable's length or type doesn't match with an expected one, DrawTrianglesShader panics.
+//
+// Even if a result is an invalid color as a premultiplied-alpha color, i.e. an alpha value exceeds other color values,
+// the value is kept and is not clamped.
+//
+// When the image i is disposed, DrawTrianglesShader does nothing.
+func DrawTrianglesShaderMRT(dsts [graphics.ShaderDstImageCount]*Image, vertices []Vertex, indices []uint16, shader *Shader, options *DrawTrianglesShaderOptions) {
```
hajimehoshi, line 735: We should be able to share most of implementations with DrawTriangles

internal/buffered/image.go
```diff
@@ -211,6 +211,51 @@ func (i *Image) DrawTriangles(srcs [graphics.ShaderSrcImageCount]*Image, vertice
 	i.pixels = nil
 }
 
+func DrawTrianglesMRT(dsts [graphics.ShaderDstImageCount]*Image, srcs [graphics.ShaderSrcImageCount]*Image, vertices []float32, indices []uint32, blend graphicsdriver.Blend, dstRegion image.Rectangle, srcRegions [graphics.ShaderSrcImageCount]image.Rectangle, shader *atlas.Shader, uniforms []uint32, fillRule graphicsdriver.FillRule) {
```
hajimehoshi, line 214: Why not unifying DrawTriangles with this?

internal/graphicscommand/image.go
```diff
@@ -145,6 +145,47 @@ func (i *Image) DrawTriangles(srcs [graphics.ShaderSrcImageCount]*Image, vertice
...
+//
+// src and shader are exclusive and only either is non-nil.
+//
+// The elements that index is in between 2 and 7 are used for the source images.
+// The source image is 1) src argument if non-nil, or 2) an image value in the uniform variables if it exists.
+// If there are multiple images in the uniform variables, the smallest ID's value is adopted.
+//
+// If the source image is not specified, i.e., src is nil and there is no image in the uniform variables, the
+// elements for the source image are not used.
+func DrawTrianglesMRT(dsts [graphics.ShaderDstImageCount]*Image, srcs [graphics.ShaderSrcImageCount]*Image, vertices []float32, indices []uint32, blend graphicsdriver.Blend, dstRegion image.Rectangle, srcRegions [graphics.ShaderSrcImageCount]image.Rectangle, shader *Shader, uniforms []uint32, fillRule graphicsdriver.FillRule) {
```
hajimehoshi, line 169: Why not unifying DrawTriangles with this?

internal/atlas/image.go
Zyko0: I originally thought I might have a lot of additional things to write and wanted to make it distinct, but I can merge them now!
Another point is that I was afraid of the cost of the copy of an extra array of 8 pointers in all paths with (5-6x layer depth, ebiten,ui,buffered,mipmap,atlas, etc..) which would apply to all other standard two-triangles `DrawImage` call.

hajimehoshi: > Another point is that I was afraid of the cost of the copy of an extra array of 8 pointers in all paths with (5-6x layer depth, ebiten,ui,buffered,mipmap,atlas, etc..) which would apply to all other standard two-triangles DrawImage call.

I don't think this would be problematic in terms of performance. Also, I'm not sure we really want 8 images (4 would be enough for example?).

Zyko0:
Let's close this PR, since we require more 2D use cases and user requests regarding this feature: https://github.com/hajimehoshi/ebiten/issues/2930#issuecomment-2214822950

PR #2966, opened 2024-04-21
chore: use errors.New to replace fmt.Errorf with no parameters will much better

ChengenH:
use errors.New to replace fmt.Errorf with no parameters will much better


hajimehoshi:
> use errors.New to replace fmt.Errorf with no parameters will much better

I was wondering what this change will make better.

PR #3008, opened 2024-06-03
fix memory leak

thechampagne:


hajimehoshi (approved):
LGTM, thanks!

I've confirmed other `C.CString` usages and these are the only places where the memory was not freed.

This leak should not be a serious issue since this happens only when an error occurs at POSIX APIs, and in this case, the application cannot recover. Thus, I don't backport this fix to the stable branch.

PR #3012, opened 2024-06-09
Renamed shader image count to specify src

Zyko0:
This just specifies that the constant refers to the number of source images passed to a shader (as discussed on discord).
This makes a distinction with the number of dst images, that could potentially be more than 1 in the future


hajimehoshi (approved):
LGTM, thanks!

PR #3024, opened 2024-06-20
audio/wav: accept more audio formats [24|32]bit

### LINKED ISSUE #2215: audio/wav: accept more formats like 24bit

jeremyfaller:
Closes #2215 

Bug/Feature request

## What this PR does | solves
Adds support for 24 and 32 bit audio files.


audio/internal/convert/stereo.go
```diff
@@ -0,0 +1,116 @@
...
+// limitations under the License.
+
+package convert
+
+import (
+	"fmt"
+	"io"
+)
+
+// Setero objects convert little-endian audio mono or stereo audio streams into
```
hajimehoshi, line 22: Stereo?

audio/wav/decode.go
```diff
@@ -215,16 +215,14 @@ chunks:
...
+			case 1, 2:
+				mono = channelCount == 1
 			default:
 				return nil, fmt.Errorf("wav: number of channels must be 1 or 2 but was %d", channelCount)
 			}
 			bitsPerSample = int(buf[14]) | int(buf[15])<<8
-			if bitsPerSample != 8 && bitsPerSample != 16 {
-				return nil, fmt.Errorf("wav: bits per sample must be 8 or 16 but was %d", bitsPerSample)
+			if !convert.IsValidResolution(bitsPerSample) {
+				return nil, fmt.Errorf("wav: invalid bits per sample must be [8,16,24,32] but was %d", bitsPerSample)
```
themowski, line 225: Adding "invalid" here makes this error message sound unnatural. Better message: "wav: bits per sample must be 8, 16, 24, or 32, but was %d"

audio/wav/decode.go
```diff
@@ -215,16 +215,14 @@ chunks:
 			}
 			channelCount := int(buf[2]) | int(buf[3])<<8
 			switch channelCount {
-			case 1:
-				mono = true
-			case 2:
-				mono = false
+			case 1, 2:
+				mono = channelCount == 1
```
themowski, line 219: IMO this new code is less clear and less explicit than before.

audio/internal/convert/stereo.go
```diff
@@ -0,0 +1,116 @@
...
+	source io.ReadSeeker
+	mono   bool
+	bytes  int
+	buf    []byte
+}
+
+// IsValidResolution returns true if the given bit resolution is supported.
+func IsValidResolution(r int) bool {
+	switch r {
+	case 8, 16, 32, 24:
```
hajimehoshi, line 34: 32bit is not the target. (I think 32bit can be float values https://www.sounddevices.com/32-bit-float-files-explained/)

audio/internal/convert/stereo.go
```diff
@@ -0,0 +1,116 @@
...
+			b[4*i+2] = s.buf[m*(i+1)-2]
+			b[4*i+3] = s.buf[m*(i+1)-1]
+		}
+	case !s.mono:
+		m := s.bytes
+		for i := 0; i < n/m; i++ {
+			b[2*i+0] = s.buf[m*(i+1)-2]
+			b[2*i+1] = s.buf[m*(i+1)-1]
+		}
+	}
```
hajimehoshi, line 100: Does this really work with 24bits?

audio/internal/convert/stereo_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+	}
+
+	cases := []testCase{
+		{8, true, nil, res8},
+		{8, false, nil, res8},
+		{16, true, nil, res16},
+		{16, false, nil, res16},
+		{24, true, nil, res16},
+		{24, false, nil, res16},
```
hajimehoshi, line 73: Why don't we have res24?

jeremyfaller: AFAICT, the whole point of `Stereo` is that it converts 8/16[/24/32] bit samples into 16 bit stereo samples. I was just keeping the same functionality as previous. If that understanding is correct, then 16bit values are the gold standard. Now that doesn't work for 8bit samples, as we just don't have the low-bits of resolution, but I'm just trying to create the same functionality.

Incidentally, 2 small notes:

- 16 bits of resolution is 96db of SNR (at full scale). That should be plenty, and using 16bits throughout the sound system is probably adequate.

- The original conversion code looks incorrect to me. For example, the [8bit code](https://github.com/hajimehoshi/ebiten/blob/main/audio/internal/convert/stereo16.go#L56) duplicated the byte high and low (constant `0x101`). While that's a solution, I think it's less surprising than just calling the low-bits 0.

audio/internal/convert/stereo.go
jeremyfaller: If you look at my test case, it's just throwing away the low 8 bits of resolution. See comments in the test file to follow up, please.

audio/internal/convert/stereo.go
jeremyfaller: I see references to both PCM and float. It seems there's lots of abuse of the WAV file format. In any case, I'll eliminate it from this change.

audio/internal/convert/stereo.go
jeremyfaller: Tahnks. :)

audio/wav/decode.go
jeremyfaller: Your codebase. Your call. :)

audio/wav/decode.go
jeremyfaller: okey dokey.

audio/internal/convert/stereo_test.go
hajimehoshi: > AFAICT, the whole point of Stereo is that it converts 8/16[/24/32] bit samples into 16 bit stereo samples.

Yes that's correct. Stereo16 (the original name) indicates that the stereo 16bit (little endian) PCM source, whatever the original source is. So, 8bit code is multiplied by 0x101 to convert 8bit to 16bit values.

Please revert your renaming (Stereo16 to Stereo).

> Now that doesn't work for 8bit samples, as we just don't have the low-bits of resolution, but I'm just trying to create the same functionality.

Do you mean your current test res8 doesn't work, or the current implementation (before your PR) doesn't work?

If we think the current implementation doesn't work with 8bit wav, we should focus on this problem first.

hajimehoshi: I've confiemd 8bits worked correctly with the current implementation. https://github.com/hajimehoshi/ebiten/commit/1cd6a1f10ce6ec5251bc0834f135ded0b92c4269 It's because samples in 8bit wav file are unsigned, not signed.

Then, please revert your implementation for 8bits. Thanks,

PR #3026, opened 2024-06-24
Update image.DrawTriangles to re-use indices

ernestrc:
Reduces allocations and GC overhead for programs that call `ebiten.Image.DrawTriangles` hundreds of times per rendered frame.

### DIFF
--- a/image.go
+++ b/image.go
@@ -43,6 +43,9 @@ type Image struct {
 	// tmpVertices must not be reused until ui.Image.Draw* is called.
 	tmpVertices []float32
 
+	// tmpIndices must not be reused until ui.Image.Draw* is called.
+	tmpIndices []uint32
+
 	// tmpUniforms must not be reused until ui.Image.Draw* is called.
 	tmpUniforms []uint32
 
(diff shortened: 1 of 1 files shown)


hajimehoshi (commented):
Looks good. Let's wait for the test results. Thanks!

image.go
```diff
@@ -42,6 +42,7 @@ type Image struct {
 
 	// tmpVertices must not be reused until ui.Image.Draw* is called.
 	tmpVertices []float32
+	tmpIndices  []uint32
```
hajimehoshi, line 45: Add a comment like

```
// tmpIndices must not be reused until ui.Image.Draw* is called.
```

image.go
```diff
@@ -514,7 +515,7 @@ func (i *Image) DrawTriangles(vertices []Vertex, indices []uint16, img *Image, o
 			vs[i*graphics.VertexFloatCount+7] = v.ColorA * ca
 		}
 	}
-	is := make([]uint32, len(indices))
+	is := i.ensureTmpIndices(len(indices))
```
hajimehoshi, line 518: We can do the same thing at `DrawTrianglesShader`

ernestrc: done

image.go
```diff
@@ -42,6 +42,8 @@ type Image struct {
 
 	// tmpVertices must not be reused until ui.Image.Draw* is called.
 	tmpVertices []float32
+	// tmpIndices must not be reused until ui.Image.Draw* is called.
```
hajimehoshi, line 45: Add an empty line above this.

PR #3027, opened 2024-06-24
Update builtinShader to not use mutex when single thread build tag is set

ernestrc:
Reduces mutex overhead for programs that invoke `ebiten.Image.DrawTriangles` hundreds of times per frame and run in single thread mode.

I'm not sure we want to go in this direction or even that this is the preferred approach for making use of the `ebitensinglethread` build tag, but this improved performance on my program so I thought I'd contribute it back.


shader_multithread.go
```diff
@@ -0,0 +1,32 @@
+// Copyright 2020 The Ebiten Authors
```
hajimehoshi, line 1: Copyright 2024 The Ebitengine Authors

shader_singlethread.go
```diff
@@ -0,0 +1,23 @@
+// Copyright 2020 The Ebiten Authors
```
hajimehoshi, line 1: Ditto

shader_multithread.go
hajimehoshi, line 1: Rename this file to `shader_notsinglethread.go` for the consistency

hajimehoshi:
Assuming 99% use cases are reading, can we implement this function with a better way like what `sync.Once` does, instead of splitting files?

https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/sync/once.go;l=48-67

hajimehoshi:
My vague idea is 

* Define atomic values for all the possible combinations of the arguments
* Use an array instead of a map. This requires to encode arguments into an integer index.
* If an atomic value is 0, a shader for the arguments is not created. Create a shader for it.
* If an atomic value is 1, a shader for the arguments is already created. Return it without a mutex lock.

What do you think?

ernestrc:
> My vague idea is
> 
> * Define atomic values for all the possible combinations of the arguments
> * Use an array instead of a map. This requires to encode arguments into an integer index.
> * If an atomic value is 0, a shader for the arguments is not created. Create a shader for it.
> * If an atomic value is 1, a shader for the arguments is already created. Return it without a mutex lock.
> 
> What do you think?

Atomics would speed up this but my main concern still remains: if an application runs in single thread mode and I'm not accessing the ebiten library from multiple goroutines, why should I pay the price of synchronizing hundreds of times per frame, even if via atomics? I don't know what your plans are for v3 but I suspect there might be a more elegant way to do this than splitting implementations into files with different build tags, that also doesn't impose synchronization overhead for single-threaded applications.

hajimehoshi:
> Atomics would speed this up but my main concern still remains: if an application runs in single thread mode and I'm not accessing the ebiten library from multiple goroutines, why should I pay the price of synchronizing hundreds of times per frame, even if via atomics?

Does accessing atomic variables really matter compared to mutex cost? I don't think so. We should measure performance if we really want to split the file.

Issue #3043, opened 2024-07-17
internal/ui: crash by dropping an HTML element onto the canvas


cowboymacke:
### Ebitengine Version

v2.7.5

### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [X] Web Browsers

### Go Version (`go version`)

go1.22.3 darwin/arm64

### What steps will reproduce the problem?

Problem happens you object are drag from the web view into the canvas/iframe,

Follow example on [https://ebitengine.org/en/documents/webassembly.html](https://ebitengine.org/en/documents/webassembly.html])

1. add image into your simple index.html

```
<!DOCTYPE html>
<iframe src="main.html" width="640" height="480"></iframe>
<img src="https://ebitengine.org/images/logo.png">
```
2. Run using favorite web-server
3.  drag and drop image into the iframe

https://github.com/user-attachments/assets/d7a97c11-13f0-4364-8d83-bbc3278e19d1

### What is the expected result?

Not to crash the engine

### What happens instead?

Engine crash with following console output

``` 
goroutine 630 [running]:
wasm_exec.js:22 syscall/js.Value.Get({{}, 0x7ff8000000000002, 0x0}, {0x131fad, 0xa})
wasm_exec.js:22     /opt/homebrew/Cellar/go/1.22.3/libexec/src/syscall/js/js.go:290 +0xc
wasm_exec.js:22 github.com/hajimehoshi/ebiten/v2/internal/ui.(*UserInterface).appendDroppedFiles(0x1cb0248, {{}, 0x7ff8000100002292, 0x284a878})
wasm_exec.js:22     /Users/ling/go/pkg/mod/github.com/hajimehoshi/ebiten/v2@v2.7.5/internal/ui/ui_js.go:713 +0xa
wasm_exec.js:22 created by github.com/hajimehoshi/ebiten/v2/internal/ui.(*UserInterface).setCanvasEventHandlers.func13 in goroutine 12
wasm_exec.js:22     /Users/ling/go/pkg/mod/github.com/hajimehoshi/ebiten/v2@v2.7.5/internal/ui/ui_js.go:693 +0xa
```

### Anything else you feel useful to add?

Only been verified on safari and chrome

hajimehoshi:
The point is

```
[...]

PR #3044, opened 2024-07-17
internal/ui: crash by dropping an HTML element onto the canvas

cowboymacke:
[internal/ui: crash by dropping an HTML element onto the canvas](https://github.com/hajimehoshi/ebiten/issues/3043)

The application crashes when an HTML element is dropped onto the canvas.

## What this PR does | solves
This pull request addresses the crash by implementing the following:
- Loops through all dropped items
- Only handles items with the "file" kind
- Ignores other types of dropped elements, preventing the crash

## Suggesting for future improvements
   - Convert `DroppedFiles` to an array to handle multiple file drops
   - Implement handling for all types of dropped items under the switch case

### DIFF
--- a/internal/ui/ui_js.go
+++ b/internal/ui/ui_js.go
@@ -702,7 +702,6 @@ func (u *UserInterface) setCanvasEventHandlers(v js.Value) {
 	}))
 }
 
-// appendDroppedFiles will only take the first file, all other files that was drop will get thrown
 func (u *UserInterface) appendDroppedFiles(data js.Value) {
 	u.dropFileM.Lock()
 	defer u.dropFileM.Unlock()
@@ -712,21 +711,16 @@ func (u *UserInterface) appendDroppedFiles(data js.Value) {
 		return
 	}
 
-	var files []*file.FileEntryFS
+loop:
 	for i := 0; i < items.Length(); i++ {
 		kind := items.Index(i).Get("kind").String()
 		switch kind {
 		case "file":
 			fs := items.Index(i).Call("webkitGetAsEntry").Get("filesystem").Get("root")
-			files = append(files, file.NewFileEntryFS(fs))
+			u.inputState.DroppedFiles = file.NewFileEntryFS(fs)
+			break loop
 		}
 	}
-
-	if len(files) <= 0 {
-		return
-	}
-
-	u.inputState.DroppedFiles = files[0]
 }
 
 func (u *UserInterface) forceUpdateOnMinimumFPSMode() {
--- a/internal/ui/ui_js.go
+++ b/internal/ui/ui_js.go
@@ -705,11 +705,7 @@ func (u *UserInterface) setCanvasEventHandlers(v js.Value) {
 func (u *UserInterface) appendDroppedFiles(data js.Value) {
 	u.dropFileM.Lock()
 	defer u.dropFileM.Unlock()
-
 	items := data.Get("items")
-	if items.Length() <= 0 {
-		return
-	}
 
 	for i := 0; i < items.Length(); i++ {
 		kind := items.Index(i).Get("kind").String()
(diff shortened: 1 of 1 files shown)


internal/ui/ui_js.go
```diff
@@ -711,8 +712,21 @@ func (u *UserInterface) appendDroppedFiles(data js.Value) {
...
 
-	fs := items.Index(0).Call("webkitGetAsEntry").Get("filesystem").Get("root")
-	u.inputState.DroppedFiles = file.NewFileEntryFS(fs)
+	var files []*file.FileEntryFS
+	for i := 0; i < items.Length(); i++ {
+		kind := items.Index(i).Get("kind").String()
+		switch kind {
+		case "file":
+			fs := items.Index(i).Call("webkitGetAsEntry").Get("filesystem").Get("root")
+			files = append(files, file.NewFileEntryFS(fs))
```
hajimehoshi, line 721: If we adopt only one 'file' object, what's the point of accumulating them? What about 

```go
var f *file.FileEntryFS
for ... {
    // ...
        f = file.NewFileEntryFS(fs)
        break
    // ...
}
if f == nil {
    return
}
```

hajimehoshi (commented):
As far as I know, there is no case that multiple 'file' objects come. Even if you drop multiple files, only one directory should come. (You can see this behavior by `examples/dropfile`).

internal/ui/ui_js.go
```diff
@@ -702,6 +702,7 @@ func (u *UserInterface) setCanvasEventHandlers(v js.Value) {
 	}))
 }
 
+// appendDroppedFiles will only take the first file, all other files that was drop will get thrown
```
hajimehoshi, line 705: Remove the comment. We don't know the case when multiple 'file' objects come yet.

hajimehoshi:
Due to the schedule of stable versions, I might have to take over this PR. Thanks,

internal/ui/ui_js.go
cowboymacke: When multiple files are dropped onto the canvas, all of them will be available in the `items` property of the drop event. However, it's important to highlight that with the current implementation, `DroppedFiles` will only contain the first file from this selection. This means if you drop three files, only one will be processed and stored in DroppedFiles. To handle all dropped files, you'd need to modify the code to iterate through the items and process each file individually.

internal/ui/ui_js.go
cowboymacke: Sounds good. This would also highlight that only the first file is processed since we're using an early break.

internal/ui/ui_js.go
hajimehoshi: > When multiple files are dropped onto the canvas, all of them will be available in the items property of the drop event. 

Have you confirmed this?

hajimehoshi: As I said, the current implementation alredy works. I don't know when multiple 'file' objects come.

https://github.com/user-attachments/assets/365be2f9-c8f5-49ae-a8a4-6b1a4e7320c5

hajimehoshi:
As we discussed on the Discord server, the current implementation doesn't work on Safari and Firefox when dropping multiple files.

In this PR, let's focus on fixing the crash issue, and work on the dropping-multiple-files issue in another PR.

internal/ui/ui_js.go
```diff
@@ -711,8 +711,16 @@ func (u *UserInterface) appendDroppedFiles(data js.Value) {
...
-	fs := items.Index(0).Call("webkitGetAsEntry").Get("filesystem").Get("root")
-	u.inputState.DroppedFiles = file.NewFileEntryFS(fs)
+loop:
+	for i := 0; i < items.Length(); i++ {
+		kind := items.Index(i).Get("kind").String()
+		switch kind {
+		case "file":
+			fs := items.Index(i).Call("webkitGetAsEntry").Get("filesystem").Get("root")
+			u.inputState.DroppedFiles = file.NewFileEntryFS(fs)
+			break loop
```
hajimehoshi, line 721: As we don't have code after the for loop, just `return` should be fine

hajimehoshi (approved):
LGTM, thanks!

PR #3102, opened 2024-09-16
ebiten: update GeoM.Rotate() comment

setanarut:
add (clockwise) info

It is not specified in which direction the rotation is.

It is not specified in which direction the rotation is.

## What this PR does | solves
It is not specified in which direction the rotation is.

### DIFF
--- a/geom.go
+++ b/geom.go
@@ -121,7 +121,7 @@ func (g *GeoM) Translate(tx, ty float64) {
 	g.ty += ty
 }
 
-// Rotate rotates the matrix by theta.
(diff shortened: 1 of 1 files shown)


geom.go
```diff
@@ -121,7 +121,7 @@ func (g *GeoM) Translate(tx, ty float64) {
 	g.ty += ty
 }
 
-// Rotate rotates the matrix by theta.
+// Rotate rotates the matrix by theta (clockwise).
```
hajimehoshi, line 124:
```
// Rotate rotates the matrix clockwise by theta.
```

Issue #3103, opened 2024-09-16
ebiten: remove vertex copying in `DrawTriangles()` / `DrawTrianglesShader()` loop


SolarLune:
### Ebitengine Version

v2.8.0-alpha.9 (?) / latest master

### Operating System

- [X] Windows
- [X] macOS
- [X] Linux
- [X] FreeBSD
- [X] OpenBSD
- [X] Android
- [X] iOS
- [X] Nintendo Switch
- [X] PlayStation 5
- [X] Xbox
- [X] Web Browsers

### Go Version (`go version`)

1.23.0

### What steps will reproduce the problem?

This is just an issue tracking a small optimization for `Image.DrawTriangles()` and `Image.DrawTrianglesShader()`. Internally, these functions loop through a slice of vertices and copy their contents to an internal vertex data slice before proceeding with rendering. This can be seen here:

https://github.com/hajimehoshi/ebiten/blob/main/image.go#L721

Currently, this is done with `for i, v := range vertices` - by using this version of a for loop, the vertex data is copied into `v`. This is an unnecessary step, since we're then just copying the data directly into the internal vertex slice.

This should probably be replaced with `for i := range vertices`, after which we can just index the vertices slice to get the data.

In my limited testing and profiling, adjusting the loop this makes the loop take about 13% less time. Timing the loop itself went from ~216µs to ~188µs per `DrawTrianglesShader()` call with a relatively large number of vertices (about 6000 triangles per call).

### What is the expected result?

N/A

### What happens instead?

N/A

### Anything else you feel useful to add?

N/A

PR #3104, opened 2024-09-16
ebiten: don't create a Vertex copy in DrawTriangles() / DrawTrianglesShader()

SolarLune:
This is a performance patch to remove extra variable copying when looping through vertices in `Image.DrawTriangles()` and `Image.DrawTrianglesShader()`.

Performance

## What this PR does | solves
This PR replaces `for i, v := range vertices` with `for i := range vertices` and indexes the `vertices` slice to allow the functions to behave the same while avoiding unnecessarily allocating an extra `ebiten.Vertex` in each loop iteration.

This PR should resolve issue #3103.

### DIFF
--- a/image.go
+++ b/image.go
@@ -554,30 +554,32 @@ func (i *Image) DrawTriangles(vertices []Vertex, indices []uint16, img *Image, o
 	vs := i.ensureTmpVertices(len(vertices) * graphics.VertexFloatCount)
 	dst := i
 	if options.ColorScaleMode == ColorScaleModeStraightAlpha {
-		for i, v := range vertices {
-			dx, dy := dst.adjustPositionF32(v.DstX, v.DstY)
+		// Avoid using `for i, v := range vertices` as adding `v` creates a copy from `vertices` unnecessarily on each loop (#3103).
+		for i := range vertices {
+			dx, dy := dst.adjustPositionF32(vertices[i].DstX, vertices[i].DstY)
 			vs[i*graphics.VertexFloatCount] = dx
 			vs[i*graphics.VertexFloatCount+1] = dy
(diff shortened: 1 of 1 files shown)


image.go
```diff
@@ -554,30 +554,30 @@ func (i *Image) DrawTriangles(vertices []Vertex, indices []uint16, img *Image, o
 	vs := i.ensureTmpVertices(len(vertices) * graphics.VertexFloatCount)
 	dst := i
 	if options.ColorScaleMode == ColorScaleModeStraightAlpha {
-		for i, v := range vertices {
-			dx, dy := dst.adjustPositionF32(v.DstX, v.DstY)
+		for i := range vertices {
```
hajimehoshi, line 557: Let's add a comment to explain why we don't use the second variable at this for-range loop, like

```go
// Avoid the value variable at this for-range loop since this would cause unnecessary copyings (#3103).
```

(Please refine this sentence :-)

image.go
```diff
@@ -718,24 +718,24 @@ func (i *Image) DrawTrianglesShader(vertices []Vertex, indices []uint16, shader
 	vs := i.ensureTmpVertices(len(vertices) * graphics.VertexFloatCount)
 	dst := i
 	src := options.Images[0]
-	for i, v := range vertices {
-		dx, dy := dst.adjustPositionF32(v.DstX, v.DstY)
+	for i := range vertices {
```
hajimehoshi, line 721: ditto.

image.go
SolarLune: OK, PR updated and force-pushed; the comment has been added.

image.go
hajimehoshi: This is not done yet.

hajimehoshi (approved):
LGTM, thanks!

PR #3125, opened 2024-10-08
Fixes DeleteVertexArrays removing entry from textures in webgl driver

LoparPanda:

Unsure if this is connected with any existing issues.

bug

## What this PR does | solves
This is just what appears to be a typo in the WebGL driver that is deleting the mapping between a uint32 ID and a JavaScript value  for vertexArrays from the value store for the textures instead. This seems like it would potentially cause a crash if the numbers aligned.

I'm not sure exactly how to test this! I don't believe anything would happen to the vertexArrays lookups other than the map taking up a small amount more space, but if a live texture has the same ID as a deleted vertex array, I believe the map lookup would panic and crash.


hajimehoshi (approved):
Good catch! LGTM

hajimehoshi:
I believe `DeleteVertexArray` is never called on browsers fortunately, so this should not cause an issue in the real world.

Issue #3128, opened 2024-10-13
cmd/ebitnmobile: ebitenmobile fails on go1.23 if you have GOEXPERIMENT=loopvar set


corfe83:
### Ebitengine Version

2.8.1

### Operating System

- [ ] Windows
- [ ] macOS
- [X] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

### Go Version (`go version`)

1.23.2

### What steps will reproduce the problem?

Run ebitenmobile bind with conventional commandline arguments such as:

ebitenmobile bind -target android -javapkg your.package.name -o path/to/yourgame.aar .

### What is the expected result?

Bind is successful

### What happens instead?

Bind gives error:

2024/10/12 22:55:53 go [mod tidy] failed: go: errors parsing go.mod:
go.mod:5: toolchain directive expects exactly one argument

hajimehoshi:
IIUC, specifying `GOEXPERIMENT=loopvar` with Go 1.23 doesn't make sense as Go 1.23 already applies the loopvar change, but anyway this is a bug.

PR #3129, opened 2024-10-13
Fix gomobile toolchain bug

corfe83:
This PR solves:

#3128

The reason this happens is that runtime.Version() returns this value when loopvar variable is set:

"go1.23.2 X:loopvar"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 *~
 *.aar
 *.apk
-*.exe
 *.tar.gz
 .idea
 .vscode
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *~
 *.aar
 *.apk
+*.exe
 *.tar.gz
 .idea
 .vscode
--- a/cmd/ebitenmobile/gomobile.go
+++ b/cmd/ebitenmobile/gomobile.go
@@ -166,11 +166,11 @@ import (
 	// runtime.Version() is the current executing Go version. For example, this is the version of the toolchain directive in go.mod.
 	// This might differ from the Go command version under the temporary directory.
 	// To avoid the version mismatch, set the toolchain explicitly (#3086).
-	toolchainVersion, err := toolchainParameter()
+	t, err := toolchainParameter()
 	if err != nil {
 		return tmp, err
 	}
-	if err := runGo("mod", "edit", "-toolchain="+toolchainVersion); err != nil {
+	if err := runGo("mod", "edit", "-toolchain="+t); err != nil {
 		return tmp, err
 	}
 	if err := runGo("mod", "tidy"); err != nil {
@@ -221,11 +221,11 @@ import (
 func toolchainParameter() (string, error) {
 	pattern := regexp.MustCompile(`\bgo\d+\.\d+(\.\d+)?`)
 	rawVersion := runtime.Version()
-	results := pattern.FindStringSubmatch(rawVersion)
-	if len(results ) < 1 {
+	m := pattern.FindStringSubmatch(rawVersion)
+	if len(m) < 1 {
 		return "", errors.New("runtime.Version() regex did not capture any results!")
 	}
-	return results[0], nil
+	return m[0], nil
 }
 
 func gomobileHash() (string, error) {
--- a/cmd/ebitenmobile/gomobile.go
+++ b/cmd/ebitenmobile/gomobile.go
@@ -219,10 +219,10 @@ func getToolchainParameter() string {
 	rawVersion := runtime.Version()
 	allIndexes := pattern.FindAllSubmatchIndex([]byte(rawVersion), -1)
 	if len(allIndexes) < 1 {
-		panic("runtime.Version() did not match expected format!")
+		panic("runtime.Version() regex did not capture any results!")
 	}
 	if len(allIndexes[0]) < 2 {
-		panic("runtime.Version() did not match expected format!")
+		panic("runtime.Version() regex captured a result but didn't return 2 indexes!")
 	}
 	return rawVersion[allIndexes[0][0]:allIndexes[0][1]]
 }
--- a/cmd/ebitenmobile/gomobile.go
+++ b/cmd/ebitenmobile/gomobile.go
@@ -219,7 +219,7 @@ import (
 }
 
 func toolchainParameter() (string, error) {
-	pattern := regexp.MustCompile("\\bgo\\d+\\.\\d+(\\.\\d+)?")
+	pattern := regexp.MustCompile(`\bgo\d+\.\d+(\.\d+)?`)
 	rawVersion := runtime.Version()
 	results := pattern.FindStringSubmatch(rawVersion)
 	if len(results ) < 1 {
--- a/cmd/ebitenmobile/gomobile.go
+++ b/cmd/ebitenmobile/gomobile.go
@@ -16,7 +16,6 @@ package main
 
 import (
 	_ "embed"
-	"errors"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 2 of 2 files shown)


hajimehoshi:
Interesting.. Thanks!

cmd/ebitenmobile/gomobile.go
```diff
@@ -164,7 +165,8 @@ import (
 	// runtime.Version() is the current executing Go version. For example, this is the version of the toolchain directive in go.mod.
 	// This might differ from the Go command version under the temporary directory.
 	// To avoid the version mismatch, set the toolchain explicitly (#3086).
-	if err := runGo("mod", "edit", "-toolchain="+runtime.Version()); err != nil {
+	toolchainVersion := strings.Split(runtime.Version(), " ")[0]
```
hajimehoshi, line 168:
```go
toolchainVersion, _, _ := strings.Cut(runtime.Version(), " ")
```

is better.

hajimehoshi: Optional: a more robust way would be

```go
	reVer := regexp.MustCompile(`^go\d+\.\d+\.(\d+)?$`)
	verStr := runtime.Version()
	m := reVer.FindStringSubmatch(verStr)
	if m == nil {
		return tmp, fmt.Errorf("ebitenmobile: unexpected Go version: %s", verStr)
	}
	toolchainVersion := m[0]
```

TotallyGamerJet: Wouldn’t `go/version.Lang(runtime.Version)` give the value you want?

hajimehoshi: I didn't know that, thanks!

hajimehoshi: https://go.dev/play/p/mJgkLAS_nEw This doesn't work unfortunately.

hajimehoshi: https://go.dev/play/p/44G5Md_xRsZ Also `version.Lang` strips the build version, so this is not ideal.

corfe83: Good feedback, I've pushed a change as suggested.

hajimehoshi (approved):
LGTM, thanks!

cmd/ebitenmobile/gomobile.go
hajimehoshi: Hmm, this might not work with gotip: https://go.dev/play/p/V9tVDS4adLm?v=gotip

We need to 1. split the string by a space char and 2. check each chunk whether it maches with a regular expression `^go\d+\.\d+(\.\d+)?`

hajimehoshi (changes requested):
See the comment.

corfe83:
Thanks you're right, this didn't work with tip. I've added the regex, I think it behaves much better now. Also I added *.exe to .gitignore.

.gitignore
```diff
@@ -2,6 +2,7 @@
 *~
 *.aar
 *.apk
+*.exe
```
hajimehoshi, line 5: Why is this needed?

cmd/ebitenmobile/gomobile.go
```diff
@@ -212,6 +214,19 @@ import (
 	return tmp, nil
 }
 
+func getToolchainParameter() string {
+	pattern := regexp.MustCompile("\\bgo\\d+\\.\\d+(\\.\\d+)?")
```
hajimehoshi, line 218: Use ` for a string literal instead of `"`.

cmd/ebitenmobile/gomobile.go
```diff
@@ -212,6 +214,19 @@ import (
...
+	rawVersion := runtime.Version()
+	allIndexes := pattern.FindAllSubmatchIndex([]byte(rawVersion), -1)
+	if len(allIndexes) < 1 {
+		panic("runtime.Version() regex did not capture any results!")
```
hajimehoshi, line 222: Don't panic. Return an error.

cmd/ebitenmobile/gomobile.go
```diff
@@ -212,6 +214,19 @@ import (
...
+	}
+	if len(allIndexes[0]) < 2 {
+		panic("runtime.Version() regex captured a result but didn't return 2 indexes!")
```
hajimehoshi, line 225: Ditto

cmd/ebitenmobile/gomobile.go
```diff
@@ -212,6 +214,19 @@ import (
 	return tmp, nil
 }
 
+func getToolchainParameter() string {
+	pattern := regexp.MustCompile("\\bgo\\d+\\.\\d+(\\.\\d+)?")
+	rawVersion := runtime.Version()
+	allIndexes := pattern.FindAllSubmatchIndex([]byte(rawVersion), -1)
```
hajimehoshi, line 220: Use `FindStringSubmatch`.

cmd/ebitenmobile/gomobile.go
```diff
@@ -212,6 +214,19 @@ import (
 	return tmp, nil
 }
 
+func getToolchainParameter() string {
```
hajimehoshi, line 217: `func toolchainParameter() string`

.gitignore
corfe83: I'm developing on windows, when I run "go build" from "ebiten\cmd\ebitenmobile" directory it produces an exe. I felt we wouldn't want to checkin built binaries to ebiten repo (if we later want to, we could change this). I can remove it if you think this is not appropriate.

corfe83:
Thanks, I've pushed changes to address feedback.

.gitignore
hajimehoshi: This seems a different issue from what we are focusing on now. Let's remove this.

cmd/ebitenmobile/gomobile.go
```diff
@@ -212,6 +218,16 @@ import (
 	return tmp, nil
 }
 
+func toolchainParameter() (string, error) {
+	pattern := regexp.MustCompile(`\bgo\d+\.\d+(\.\d+)?`)
+	rawVersion := runtime.Version()
+	results := pattern.FindStringSubmatch(rawVersion)
+	if len(results ) < 1 {
+		return "", errors.New("runtime.Version() regex did not capture any results!")
+	}
```
hajimehoshi, line 227:
```go
if len(results) == 0 {
    return "", fmt.Errorf("ebitenmobile: unexpected version: %s", rawVersion)
}
```

Also, please do not forget to run gofmt.

cmd/ebitenmobile/gomobile.go
```diff
@@ -212,6 +218,16 @@ import (
 	return tmp, nil
 }
 
+func toolchainParameter() (string, error) {
+	pattern := regexp.MustCompile(`\bgo\d+\.\d+(\.\d+)?`)
+	rawVersion := runtime.Version()
+	results := pattern.FindStringSubmatch(rawVersion)
```
hajimehoshi, line 224: For my convention, `m` is better than `results`. (`m` indicates 'matches')

cmd/ebitenmobile/gomobile.go
```diff
@@ -164,7 +166,11 @@ import (
 	// runtime.Version() is the current executing Go version. For example, this is the version of the toolchain directive in go.mod.
 	// This might differ from the Go command version under the temporary directory.
 	// To avoid the version mismatch, set the toolchain explicitly (#3086).
-	if err := runGo("mod", "edit", "-toolchain="+runtime.Version()); err != nil {
+	toolchainVersion, err := toolchainParameter()
```
hajimehoshi, line 169: As the variable lifetime is short enough, t is fine.

```go
t, err := toolchainParameter()
```

cmd/ebitenmobile/gomobile.go
corfe83: Sorry, I hastily ran go fmt and missed the other part of your comment. This is addressed with latest push.

hajimehoshi (approved):
LGTM, thanks!

PR #3133, opened 2024-10-18
internal/cocoa, internal/graphicsdriver/metal/*: minimize allocations

TotallyGamerJet:
Cache the result of objc.RegisterName at program startup. Use NSInvocation less to avoid multiple C calls

n/a

performance

## What this PR does | solves
Lowers memory usage


hajimehoshi (approved):
LGTM, thanks!

PR #3134, opened 2024-10-18
Add a debug.IsDebug check before splitting strings

gabstv:
--

feature

## What this PR does | solves

Since `debug.IsDebug` is a constant, by putting the `strings.Cut` calls inside the check this makes the `(*commandqueue).flush` method faster when Ebitengine is compiled without the `debug` build tag.

Without the `if debug.IsDebug` check, unnecessary allocations occur (since the logger is a noop logger when debug is disabled):
Screenshot_20241018_084146

GC Details after the change:
Screenshot_20241018_084204

This happens because Go knows that IsDebug is a constant (and the compiler will optimize this away on production builds).

### DIFF
--- a/internal/graphicscommand/commandqueue.go
+++ b/internal/graphicscommand/commandqueue.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"image"
 	"math"
-
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


hajimehoshi (approved):
LGTM, thanks!

PR #3187, opened 2025-01-30
clarify the same images size rule for DrawTrianglesShaderOptions

quasilyte:
Since it's not allowed to have differently-sized images for DrawTrianglesShaderOptions in pixels mode, the comment should be updated.

None

It's just a comment update.

## What this PR does | solves

Makes the engine's feature more apparent, instead of stating that doing something is impossible while it is possible.

### DIFF
--- a/image.go
+++ b/image.go
@@ -638,7 +638,8 @@ type DrawTrianglesShaderOptions struct {
 	Uniforms map[string]any
 
 	// Images is a set of the source images.
-	// All the images' sizes must be the same.
+	// In texel mode, all the image sizes must be the same.
+	// The pixels mode allows images of different sizes.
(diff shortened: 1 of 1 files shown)


image.go
```diff
@@ -638,7 +638,8 @@ type DrawTrianglesShaderOptions struct {
 	Uniforms map[string]any
 
 	// Images is a set of the source images.
-	// All the images' sizes must be the same.
+	// In texel mode, all the image sizes must be the same.
+	// The pixels mode allows images of different sizes.
```
hajimehoshi, line 642:
```
	// In the texel mode, all the image sizes must be the same.
	// The pixel mode allows images of different sizes.
```

PR #3244, opened 2025-05-19
cleaning up code base

12ya:
None

cleaning up code base

## What this PR does | solves
cleaning up code base in audio/


hajimehoshi:
I'll take a look tomorrow. By the way, as a PR explanation, 'cleaning up code base' would be better than repeating what this does like 'Modernizes for loops in the codebase'.

Issue #3241, opened 2025-05-19
ebiten: crash when leaving an app on macOS


hajimehoshi:
### Ebitengine Version

cd31850015bbb846d2a0c0f141a24a2a755f7760

### Operating System

- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

### Go Version (`go version`)

go version go1.24.3 darwin/arm64

### What steps will reproduce the problem?

I am not sure the reproducible way. I think this happened by leaving the app and closing the laptop.

### What is the expected result?

No crash

### What happens instead?

```
ERROR runtime error: invalid memory address or nil pointer dereference                                                                                              
goroutine 1 [running, locked to thread]:                                                                                                                                                
runtime/debug.Stack()                                                                                                                                                                   
        /usr/local/go/src/runtime/debug/stack.go:26 +0x64                                                                                                                               
****.logErrorAndExit({0x103368380, 0x105064cc0}, 0x1400b647bf0)                                                                      
        ****/log.go:135 +0x28                                                                                                          
****.func1()                                                                                                                     
        ****/log.go:220 +0x38                                                                                                          
panic({0x103368380?, 0x105064cc0?})                                                                                                                                                     
        /usr/local/go/src/runtime/panic.go:792 +0x124                                                                                                                                   
github.com/hajimehoshi/ebiten/v2/internal/ui.(*Monitor).DeviceScaleFactor(...)                                                                                                          
        /Users/hajimehoshi/go/pkg/mod/github.com/hajimehoshi/ebiten/v2@v2.9.0-alpha.5.0.20250518103147-cd31850015bb/internal/ui/monitor_glfw.go:45                                      
github.com/hajimehoshi/ebiten/v2/internal/ui.(*UserInterface).outsideSize(0x14000034f08)
        /Users/hajimehoshi/go/pkg/mod/github.com/hajimehoshi/ebiten/v2@v2.9.0-alpha.5.0.20250518103147-cd31850015bb/internal/ui/ui_glfw.go:1249 +0x114
github.com/hajimehoshi/ebiten/v2/internal/ui.(*UserInterface).update(0x14000034f08)
[...]

PR #3253, opened 2025-06-07
fallback to init monitor if primary is off

noboruma:
Workaround for #3241.
While this fixes the issue, I am not sure this should be the definitive fix.
A better alternative could be to turn off rendering entirely when all monitors are offline?
With that being said, this workaround prevents a crash and the engine continues working after one monitor goes back online.

Resolves a panic

## What this PR does | solves
Fallback to init monitor data when primary monitor goes offline

### DIFF
--- a/internal/ui/ui_glfw.go
+++ b/internal/ui/ui_glfw.go
@@ -1904,7 +1904,11 @@ func (u *UserInterface) currentMonitor() (*Monitor, error) {
 		return m, nil
 	}
 
-	return theMonitors.primaryMonitor(), nil
+	if m := theMonitors.primaryMonitor(); m != nil {
+		return m, nil
+	}
+
+	return u.getInitMonitor(), nil
 }
 
 func (u *UserInterface) readInputState(inputState *InputState) {


internal/ui/ui_glfw.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2015 Hajime Hoshi
+// Copyright 2015-2025 Hajime Hoshi
```
hajimehoshi: Do not update this.

internal/ui/ui_glfw.go
```diff
@@ -1904,7 +1904,12 @@ func (u *UserInterface) currentMonitor() (*Monitor, error) {
 		return m, nil
 	}
 
-	return theMonitors.primaryMonitor(), nil
+	primaryMonitor := theMonitors.primaryMonitor()
```
hajimehoshi: Write like this:

```go
if m := theMonitors.primaryMonitor(); m != nil {
    return m
}
return u.getInitMonitor(), nil
```

internal/ui/ui_glfw.go
```diff
@@ -1904,7 +1904,12 @@ func (u *UserInterface) currentMonitor() (*Monitor, error) {
...
+	if primaryMonitor == nil {
+		// Fallback to init monitor
```
hajimehoshi: Remove this comment

hajimehoshi (commented):
Thanks!

hajimehoshi:
> While this fixes the issue, I am not sure this should be the definitive fix.

Me neither. This sounds like much better than the current crash.

hajimehoshi (approved):
LGTM, thanks!

PR #3277, opened 2025-07-19
Clarify Image Set does nothing for out-of-bounds coordinates

philyawj:
## What this PR does | solves
The existing comment for the `Set` method in `image.go` notes that calls on disposed images do nothing. This PR adds a similar clarification for out-of-bounds coordinates.

### DIFF
--- a/image.go
+++ b/image.go
@@ -1269,6 +1269,8 @@ func (i *Image) at(x, y int) (r, g, b, a byte) {
 //
 // If the image is disposed, Set does nothing.
 //
+// If (x, y) is outside the image bounds, Set does nothing.
+//
 // For performance, it is recommended to use WritePixels instead of Set whenever possible.
 func (i *Image) Set(x, y int, clr color.Color) {
 	i.copyCheck()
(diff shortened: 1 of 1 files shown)


image.go
```diff
@@ -1269,6 +1269,8 @@ func (i *Image) at(x, y int) (r, g, b, a byte) {
 //
 // If the image is disposed, Set does nothing.
 //
+// If (x, y) is outside the image bounds, Set does nothing.
```
hajimehoshi: Move this comment above `Even if ...`.

```go
// Set implements the standard draw.Image's Set.
//
// If (x, y) is outside the image bounds, Set does nothing.
//
// Even if a result is an invalid color as a premultiplied-alpha color, i.e. an alpha value exceeds other color values, ...
```

philyawj: Moved: https://github.com/hajimehoshi/ebiten/pull/3277/commits/40c86dc189b1da152bf1a5bb43b9785c382641cb

Would you also like the existing comment moved up? `// If the image is disposed, Set does nothing.`

hajimehoshi: > Would you also like the existing comment moved up? // If the image is disposed, Set does nothing.

No, thanks

hajimehoshi (approved):
LGTM

Issue #3249, opened 2025-05-29
ebiten: tap-to-click doesn't work very well on browsers


hajimehoshi:
### Ebitengine Version

ca2e71ef73d0fd1292479920e9ab2517943b12b1

### Operating System

- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [x] Web Browsers

### Go Version (`go version`)

go version go1.24.3 darwin/arm64

### What steps will reproduce the problem?

1. Enable tap-to-click by System Settings
2. Run an example like `examples/wav` on browsers
3. Open the game on browsers and try tapping

### What is the expected result?

Tapping always works

### What happens instead?

Tapping rarely works. 

### Anything else you feel useful to add?

See also https://github.com/hajimehoshi/guigui/issues/150

hajimehoshi:
This was a little more difficult than I thought: this requires to change the current input system to handle input events instead of checking the button/key states in all the platforms.

PR #3279, opened 2025-07-24
handle touch event on wasm

zverbeta:
Updates #3249

bug

## What this PR does | solves
This is a partial fix for WebAssembly touch issues on android phone. Actually i don't know other cases where this fix won't work

### DIFF
--- a/internal/ui/input_js.go
+++ b/internal/ui/input_js.go
@@ -171,8 +171,16 @@ func (u *UserInterface) updateInputFromEvent(e js.Value) error {
 		// TODO: What if e.deltaMode is not DOM_DELTA_PIXEL?
 		u.inputState.WheelX = -e.Get("deltaX").Float()
 		u.inputState.WheelY = -e.Get("deltaY").Float()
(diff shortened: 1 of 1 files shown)


hajimehoshi:
The issue #3249 is not related to touch events: the issue is that with macOS's tap-to-click feature, tapping a touch pad triggers mouse down and up events, but this is too quick to catch an event for `inpututil.IsMouseButtonJustPressed`.

hajimehoshi:
So let me close this PR as this doesn't seem to address #3249. Thank you for trying.

zverbeta:
Interactive elements like buttons and input fields are not working when accessed on an android phone or in a mobile browser view. This seems to be a known bug, as documented in issue https://github.com/hajimehoshi/guigui/issues/150

hajimehoshi:
I think the problem for mobile browsers is a different issue.

(By the way, Guigui will support browsers, but not fully for now yet)

hajimehoshi:
If you are testing Guigui on mobile browsers, this should not work since Guigui doesn't handle touch events at all. This might work accidentally as browsers could convert touches into mouse events though.

Issue #2436, opened 2022-11-04
cmd/ebitenmobile: refactoring & comments


mymmrac:
### What feature would you like to be added?

Refactor files in `cmd/ebitenmobile`:

- Add new lines to increase readability
- Replace `for` loops with `strings.Join` where possible
- Remove dead code
- Provide comments on what code is doing and why (maybe small `README.md` will be good)

### Why is this needed?

Code readability & maintainability (it's hard to understand what and why ebitenmobile is doing from first glance)

hajimehoshi:
PRs for cleaning-up code are welcome!

hajimehoshi:
Refactoring is always welcome but we don't have to have a specific issue

PR #3283, opened 2025-07-30
cmd/ebitenmobile: replace for loop with strings.Join

philyawj:
Partially addresses https://github.com/hajimehoshi/ebiten/issues/2436 by replacing a `for` loop with `strings.Join`.

Code clean up.

## What this PR does | solves
Replace `for` loop in `buildN` `if` with `strings.Join`.


hajimehoshi (commented):
I think this change is not necessary. Rather, `string.Join` makes the memory usage a little worse.

PR #3298, opened 2025-08-11
all: Replace deprecated strings.Title with cases.Title

12ya:
Updated usage of the deprecated strings.Title function to golang.org/x/text/cases.Title in gobind.go and main.go for proper title casing. This improves future compatibility and aligns with Go best practices.


hajimehoshi:
> `// TODO: strings.Title is used here for the consistency with gomobile (see cmd/gomobile/bind_iosapp.go).`

As the comment says, this is for consistency. If we need, we should fix the gomobile side first.

PR #3307, opened 2025-08-24
extra notes on how colorm operates

kazzmir:
documentation for colorm

## What this PR does | solves
Adds some extra information to the colorm documentation about how colorm operates.

### DIFF
--- a/colorm/colorm.go
+++ b/colorm/colorm.go
@@ -28,14 +28,14 @@ import (
 const Dim = affine.ColorMDim
 
 // ColorM represents a matrix to transform coloring when rendering an image. The matrix
-// is applied by a fragment shader to each pixel during the rendering process. The pixel input
-// and output are expected to be r,g,b,a values in the range 0 to 1 (rather than 0 to 255).
+// is applied by a fragment shader to each pixel during the rendering process.
 //
 // ColorM is applied to the straight alpha color
 // while an Image's pixels' format is alpha premultiplied.
 // Before applying a matrix, a color is un-multiplied, and after applying the matrix,
 // the color is multiplied again.
-//
+// The pixel input and output are expected to be r,g,b,a values in the range 0 to 1 (rather than 0 to 255).
+
 // The initial value is identity.
 type ColorM struct {
 	impl affine.ColorM
--- a/colorm/colorm.go
+++ b/colorm/colorm.go
@@ -27,7 +27,9 @@ import (
 // Dim is the dimension of a ColorM.
 const Dim = affine.ColorMDim
 
-// ColorM represents a matrix to transform coloring when rendering an image.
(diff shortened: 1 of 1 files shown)


colorm/colorm.go
```diff
@@ -27,7 +27,9 @@ import (
 // Dim is the dimension of a ColorM.
 const Dim = affine.ColorMDim
 
-// ColorM represents a matrix to transform coloring when rendering an image.
+// ColorM represents a matrix to transform coloring when rendering an image. The matrix
+// is applied by a fragment shader to each pixel during the rendering process. The pixel input
+// and output are expected to be r,g,b,a values in the range 0 to 1 (rather than 0 to 255).
```
hajimehoshi: > The matrix is applied by a fragment shader to each pixel during the rendering process.

Do we really need this sentence?

> The pixel input and output are expected to be r,g,b,a values in the range 0 to 1 (rather than 0 to 255).

1. This should be after the next paragprah.
2. I'm not sure where 0 to 255 came from. `color.Color` doesn't assume 0 to 255 (rather, `RGBA()` method returns 0 to 65535). So I think we don't need `(rather than 0 to 255)`.

kazzmir:
> Do we really need this sentence?

I think its useful to know how colorm operates since as a user I don't know if it runs on the cpu or gpu. If it were to run on the cpu then it would be very expensive, but since its the gpu its quite fast.

> I'm not sure where 0 to 255 came from. color.Color doesn't assume 0 to 255 (rather, RGBA() method returns 0 to 65535). So I think we don't need (rather than 0 to 255).

Yea RGBA() returns in the range [0, 65535] in order to work around an issue with overflow (so you can multiply the color values by alpha, I believe). But when constructing the colors you would use the range [0,255].
```
color.RGBA{R: 255, G: 255, B: 255, A: 255}
```
I just wanted the documentation to explain how the 0-1 range relates to the normal 0-255 range of color input.

hajimehoshi:
> I think its useful to know how colorm operates since as a user I don't know if it runs on the cpu or gpu. If it were to run on the cpu then it would be very expensive, but since its the gpu its quite fast.

Sure, add your sentence `The matrix is applied by a fragment shader to each pixel during the rendering process.` to the first paragraph, thanks.

> color.RGBA{R: 255, G: 255, B: 255, A: 255}

This is just one of them. There are various color constructors beside it. (e.g. `RGBA64`, `Alpha`, `YCbCr`, etc)

hajimehoshi (approved):
LGTM, thanks

kazzmir:
> ... to the first paragraph, thanks.

I think it is already. By first paragraph you mean the one that starts with `ColorM represents a matrix to transform ...`, right?

kazzmir:
ok :+1:

Issue #3302, opened 2025-08-17
MacOS (15.6) deprecated functions


ajstarks:
### Ebitengine Version

v2.9.0-alpha.14

### Operating System

- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

### Go Version (`go version`)

1.25.0

### What steps will reproduce the problem?

When updated ebcanvas to v.2.9.0-alpha.14 on Mac OS Sequoia (15.6) I received warnings. (see below)

# github.com/hajimehoshi/ebiten/v2/internal/graphicsdriver/metal
cgo-gcc-prolog:55:11: warning: 'CVDisplayLinkCreateWithActiveCGDisplays' is deprecated: first deprecated in macOS 15.0 - use NSView.displayLink(target:selector:), NSWindow.displayLink(target:selector:), or NSScreen.displayLink(target:selector:)  [-Wdeprecated-declarations]
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/CoreVideo.framework/Headers/CVDisplayLink.h:99:20: note: 'CVDisplayLinkCreateWithActiveCGDisplays' has been explicitly marked deprecated here
cgo-gcc-prolog:76:11: warning: 'CVDisplayLinkSetOutputCallback' is deprecated: first deprecated in macOS 15.0 - use NSView.displayLink(target:selector:), NSWindow.displayLink(target:selector:), or NSScreen.displayLink(target:selector:)  [-Wdeprecated-declarations]
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/CoreVideo.framework/Headers/CVDisplayLink.h:143:20: note: 'CVDisplayLinkSetOutputCallback' has been explicitly marked deprecated here
cgo-gcc-prolog:95:11: warning: 'CVDisplayLinkStart' is deprecated: first deprecated in macOS 15.0 - use NSView.displayLink(target:selector:), NSWindow.displayLink(target:selector:), or NSScreen.displayLink(target:selector:)  [-Wdeprecated-declarations]
[...]

PR #3344, opened 2025-11-14
replace deprecated CVDisplayLinkCreateWithActiveCGDisplays

matjam:
Closes [#3302](https://github.com/hajimehoshi/ebiten/issues/3302) 

Bug

## What this PR does | solves
The current code using `CVDisplayLinkCreateWithActiveCGDisplays` and friends emits a bunch of warnings, which you can see in the attached issue.

The bug was closed because it is currently harmless, but these functions will definitely be removed in a future version of macOS. This PR implements a runtime check to see what version of macOS we're running on and uses NSView.displayLink and CADisplayLink. I also ignore the compile time warnings for those deprecated functions.

I think its worth noting that when the deprecated functions are actually removed, we'll need to remove the fallback logic to build on that version of macOS, but we're probably good for a few years.

This PR future proofs the code so that at the very least we have something in place when that happens, and also we hide the ugly compile warnings when we're building/testing our apps locally :-)

I've not done extensive testing of this on older versions of macOS, but it passes all the test cases and my trivial tests locally on macOS Tahoe (26.0.1).

### DIFF
--- a/internal/cocoa/api_cocoa_darwin.go
+++ b/internal/cocoa/api_cocoa_darwin.go
@@ -32,7 +32,6 @@ var (
 	class_NSWorkspace          = objc.GetClass("NSWorkspace")
 	class_NSNotificationCenter = objc.GetClass("NSNotificationCenter")
 	class_NSOperationQueue     = objc.GetClass("NSOperationQueue")
-	class_CADisplayLink        = objc.GetClass("CADisplayLink")
 )
 
 var (
@@ -78,7 +77,7 @@ var (
 	sel_addObserver                    = objc.RegisterName("addObserver:selector:name:object:")
 	sel_addObserverForName             = objc.RegisterName("addObserverForName:object:queue:usingBlock:")
 	sel_mainQueue                      = objc.RegisterName("mainQueue")
-	sel_displayLinkWithTargetSelector = objc.RegisterName("displayLinkWithTarget:selector:")
+	sel_displayLinkWithTargetSelector  = objc.RegisterName("displayLinkWithTarget:selector:")
 	sel_addToRunLoop                   = objc.RegisterName("addToRunLoop:forMode:")
 	sel_invalidate                     = objc.RegisterName("invalidate")
 	sel_setPaused                      = objc.RegisterName("setPaused:")


internal/graphicsdriver/metal/displaylink_macos.go
```diff
@@ -37,11 +37,42 @@ package metal
...
 //
+// #cgo noescape isNSViewDisplayLinkAvailable
+// #cgo nocallback isNSViewDisplayLinkAvailable
+// static bool isNSViewDisplayLinkAvailable() {
+//   NSOperatingSystemVersion version = [[NSProcessInfo processInfo] operatingSystemVersion];
+//   return version.majorVersion >= 14;
+// }
+//
+// #pragma clang diagnostic push
+// #pragma clang diagnostic ignored "-Wdeprecated-declarations"
```
hajimehoshi, line 48: In order to suppress warnings, is just this line enough? And, I'm against suppressing warnings in this way.

matjam: it appears to be the way to do it.

The only other option is to either remove the deprecated calls or suppress them just for this compile unit as we do here.

its up to you if you feel the backwards compatibility is worth it; my opinion actually is that most users run a relatively recent version of macOS and you generally are safe targeting the new APIs as they become available as they are made available for several major releases before the old ones are flagged as deprecated.

Happy to remove the backwards compatibility handling which would mean not having this warning suppressed, because obviously if there's any other functions we're using here that become deprecated, we won't know as we're suppressing them.

So I agree with you.

matjam: hmm, seems to have issues with full screen mode, so needs further testing, but let me know what approach you prefer.

hajimehoshi: My preference is do nothing unless we find this an actual issue (e.g. we cannot compile games in a future macOS). Warnings don't matter to game players anyway.

matjam: Perfectly rational; I'll close this for now and you at least have something as a starting point when Apple does actually remove the deprecated functions.

PR #3363, opened 2025-12-29
inpututil: add function to get current gamepad ids

kazzmir:
It would be convenient to have an API that returns the current set of connected gamepad ids. Otherwise an application must keep track on its own via
```
var gamepadIDs map[ebiten.GamepadID]struct{}
func Update(){
  ids = AppendJustConnectedGamepadIDs(nil)
  // set new ids in map ...
  // remove old ones no longer connected
  for id := range gamepadIDs {
    if IsGamepadJustDisconnected(id) {
     delete(gamepadIDs, di)
   }
  }
```

Feature, add an API for gamepad in inpututil

## What this PR does | solves
This PR adds a function called AppendGamepadIDs that accepts a slice of GamepadID, and appends to it all the current gamepad ids (from `theInputState.gamepadStates`). This function is almost identical to AppendJustConnectedGamepadIDs but does not compare the ids in the current gamepad state to the previous state.

Note that this is similar to the APIs `AppendTouchIDs` and `AppendPressedKeys`, but those APIs are not public.

### DIFF
--- a/inpututil/inpututil.go
+++ b/inpututil/inpututil.go
@@ -226,6 +226,25 @@ func MouseButtonPressDuration(button ebiten.MouseButton) int {
 	return int(inputstate.Get().MouseButtonPressDuration(ui.MouseButton(button)))
 }
 
+
+// AppendGamepadIDs appends currently connected gamepad IDs to gamepadIDs,
+// and returns the extended buffer.
+// Giving a slice that already has enough capacity works efficiently.
+//
+// AppendGamepadIDs is concurrent safe.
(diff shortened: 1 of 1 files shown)


hajimehoshi:
There is already the same function https://pkg.go.dev/github.com/hajimehoshi/ebiten/v2#AppendGamepadIDs

kazzmir:
https://pkg.go.dev/github.com/hajimehoshi/ebiten/v2#AppendGamepadIDs

PR #3384, opened 2026-02-11
Windows: Add support for setting light/dark mode and window backdrop

### LINKED ISSUE #3381: ebiten: Add a way to provide the game window handle to facilitate working with DWM window attributes on Windows
### Operating System

- [x] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

A method to expose the game window's handle to the user once the game window has been created, potentially with a documented point in the lifecycle where it is guaranteed to be available.

### Why is this needed?

Calling into Win32 APIs to set DWM attributes for window styling currently requires brittle mechanisms to grab the game window as soon as it's created. You can enumerate the windows and either look up the game window by its title or its PID, but you still need to add a polling mechanism with 'reasonable values' to give the main game window time to appear. Here is example code:

```
import (
	"fmt"
	"os"
	"syscall"
	"unsafe"
        "time"
	"golang.org/x/sys/windows"
)

var (
	user32                       = syscall.NewLazyDLL("user32.dll")
	procEnumWindows              = user32.NewProc("EnumWindows")
	procGetWindowThreadProcessId = user32.NewProc("GetWindowThreadProcessId")
	procIsWindowVisible          = user32.NewProc("IsWindowVisible")
)

const dwmwaSystemBackdropType = 38

// iterate over windows until we get a window belonging to this process that is visible. we have to do this
// since glfw creates message windows alongside the game window
func findWindowByProcessId(myPid uint32) windows.HWND {
	var target windows.HWND

	cb := syscall.NewCallback(func(hwnd, lParam uintptr) uintptr {
		var pid uint32
		_, _, _ = procGetWindowThreadProcessId.Call(hwnd, uintptr(unsafe.Pointer(&pid)))

		if myPid == pid {
			ret, _, _ := procIsWindowVisible.Call(hwnd)
			if ret != 0 {
				target = windows.HWND(hwnd)
				return 0
			}
		}
		return 1
	})

	_, _, _ = procEnumWindows.Call(cb, 0)

	return target
}

func ApplyDarkMode() {
    // get our current process ID
	pid := uint32(os.Getpid())

	var window windows.HWND

    // without a delay, the window may not be ready yet on startup.
	time.Sleep(500 * time.Millisecond)

	window = findWindowByProcessId(pid)

	if window == 0 {
		return
	}

	darkMode := uint32(1) // dark mode
	if err := windows.DwmSetWindowAttribute(window,
		windows.DWMWA_USE_IMMERSIVE_DARK_MODE,
		unsafe.Pointer(&darkMode),
		uint32(unsafe.Sizeof(darkMode)),
	); err != nil {
		fmt.Printf("DWMWA_USE_IMMERSIVE_DARK_MODE failed: %v\n", err)
	}

	backdropType := uint32(2) // add mica styling to window
	if err := windows.DwmSetWindowAttribute(window,
		dwmwaSystemBackdropType,
		unsafe.Pointer(&backdropType),
		uint32(unsafe.Sizeof(backdropType)),
	); err != nil {
		fmt.Printf("DWMWA_SYSTEMBACKDROP_TYPE failed: %v\n", err)
	}
}
```

Invoking it from the root game creation:

```
func main() {
	ebiten.SetWindowTitle(title)
	ebiten.SetWindowSize(windowWidth, windowHeight)

	g := game.NewGame()

	go window.ApplyDarkMode()

	err := ebiten.RunGame(&g)

	if err != nil {
		log.Fatal(err)
	}
}
```

Ideally, advanced users could be able to get a handle for the window without enumerating windows or using delays with a documented point in the lifecycle that this handle would first be available.

djweber:

Closes #3381 

Feature

## What this PR does | solves

This PR adds two new fields to `RunGameOptions` that allows a user to specify dark mode and backdrop [attributes](https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/ne-dwmapi-dwmwindowattribute)) for Windows 11. It is a no-op on platforms other than Windows 11. The attributes are applied at window creation time in `ui_glfw.go`, invoking new DWM attribute setting methods in `ui_windows.go`.

dark
light

### DIFF
--- a/internal/ui/api_windows.go
+++ b/internal/ui/api_windows.go
@@ -31,6 +31,9 @@ const (
 	_CLSCTX_SERVER            = _CLSCTX_INPROC_SERVER | _CLSCTX_LOCAL_SERVER | _CLSCTX_REMOTE_SERVER
 	_MONITOR_DEFAULTTONEAREST = 2
 	_SM_CYCAPTION             = 4
+
+	_DWMWA_USE_IMMERSIVE_DARK_MODE = 20
+	_DWMWA_SYSTEMBACKDROP_TYPE     = 38
 )
 
 var (
@@ -68,9 +71,12 @@ type _POINT struct {
 }
 
 var (
+	dwmapi = windows.NewLazySystemDLL("dwmapi.dll")
 	imm32  = windows.NewLazySystemDLL("imm32.dll")
 	ole32  = windows.NewLazySystemDLL("ole32.dll")
 	user32 = windows.NewLazySystemDLL("user32.dll")
+
+	procDwmSetWindowAttribute = dwmapi.NewProc("DwmSetWindowAttribute")
 
 	procImmAssociateContext = imm32.NewProc("ImmAssociateContext")
 
@@ -81,6 +87,14 @@ var (
 	procGetMonitorInfoW   = user32.NewProc("GetMonitorInfoW")
 	procGetCursorPos      = user32.NewProc("GetCursorPos")
 )
+
+func _DwmSetWindowAttribute(hwnd windows.HWND, dwAttribute uint32, pvAttribute unsafe.Pointer, cbAttribute uint32) error {
+	r, _, _ := procDwmSetWindowAttribute.Call(uintptr(hwnd), uintptr(dwAttribute), uintptr(pvAttribute), uintptr(cbAttribute))
+	if uint32(r) != uint32(windows.S_OK) {
+		return fmt.Errorf("ui: DwmSetWindowAttribute failed: HRESULT(%d)", uint32(r))
+	}
+	return nil
+}
 
 func _ImmAssociateContext(hwnd windows.HWND, hIMC uintptr) (uintptr, error) {
 	r, _, e := procImmAssociateContext.Call(uintptr(hwnd), hIMC)
--- a/internal/ui/ui.go
+++ b/internal/ui/ui.go
@@ -173,16 +173,18 @@ func (u *UserInterface) dumpImages(dir string) (string, error) {
 }
 
 type RunOptions struct {
-	GraphicsLibrary          GraphicsLibrary
-	InitUnfocused            bool
-	ScreenTransparent        bool
-	SkipTaskbar              bool
-	SingleThread             bool
-	DisableHiDPI             bool
-	ColorSpace               color.ColorSpace
-	ApplePressAndHoldEnabled bool
-	X11ClassName             string
-	X11InstanceName          string
+	GraphicsLibrary           GraphicsLibrary
+	InitUnfocused             bool
+	ScreenTransparent         bool
+	SkipTaskbar               bool
+	SingleThread              bool
+	DisableHiDPI              bool
+	ColorSpace                color.ColorSpace
+	ApplePressAndHoldEnabled  bool
+	X11ClassName              string
+	X11InstanceName           string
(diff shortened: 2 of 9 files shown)


hajimehoshi:
Is this only for Windows? What about the other platforms?

hajimehoshi:
For example, making a window for a dark mode seems to be useful for other platforms, so I prefer having a more general version than `WindowsUseDarkMode`. On the other hand, I'm not sure about `WindowsSystemBackdropType`. This might be Windows specific.

djweber:
> Is this only for Windows? What about the other platforms?

Hi - yes, it is currently a Windows-only config, and a no-op on unsupported platforms. I had assumed the scope was a Windows-only config for this based on the discussion in #3381. Since dark mode is a cross-platform concept by nature, we could rename that to `UseDarkMode` and expand it to support other platforms. From an initial glance, these seem a little less trivial to add than it is for Windows.

However, the `WindowsSystemBackdropType` is Windows-only and affects the window styling as seen in the screenshots above, so that would make sense to keep as a `Windows*` only feature, like we do with the `X11*` fields.

hajimehoshi:
Hmm I see. Let me have to think the design.

We might want to change the window title bar type during running it. Especially, light/dark mode can be switched by the app, ignoring the system setting. In this case, adding an option to `RunGameOptions` is not enough and we want to have global functions like `SetWindowFooBar`.

Apparently `DWMWA_USE_IMMERSIVE_DARK_MODE` indicates that the title becomes dark only when the system color is dark, so there might not be exact counterparts in other platforms. We need to investigate them.

As a tentative solution, I'm thinking to create a function to run a closure on a main thread, so that a native window can be modified safely. (https://github.com/hajimehoshi/ebiten/issues/3381#issuecomment-3863590267)

djweber:
> Hmm I see. Let me have to think the design.
> 
> We might want to change the window title bar type during running it. Especially, light/dark mode can be switched by the app, ignoring the system setting. In this case, adding an option to `RunGameOptions` is not enough and we want to have global functions like `SetWindowFooBar`.
> 
> Apparently `DWMWA_USE_IMMERSIVE_DARK_MODE` indicates that the title becomes dark only when the system color is dark, so there might not be exact counterparts in other platforms. We need to investigate them.
> 
> As a tentative solution, I'm thinking to create a function to run a closure on a main thread, so that a native window can be modified safely. ([#3381 (comment)](https://github.com/hajimehoshi/ebiten/issues/3381#issuecomment-3863590267))

Changing during runtime could be useful, I agree. Some mechanism such as this to allow us to modify the bar before the window shows would be helpful, and to avoid flickering.

hajimehoshi:
> Some mechanism such as this to allow us to modify the bar before the window shows would be helpful, and to avoid flickering.

True. In this case, `RunOnUIThread` would not be enough.

hajimehoshi:
To switch color mode, use `SetWindowColorMode` (#3387). This can be invoked even before `RunGame`.

To set other DWM attributes, use `RunOnMainThread` (#3385). This cannot be invoked before `RunGame`, so if you want to set the attributes, try it at the first your `Update`. I don't plan to have an API to set attributes directly so far.

Let me close this PR. I appreciate your trial to contribute to Ebitengine.

Issue #368, opened 2017-06-13
audio: Support OPUS


hajimehoshi:
What about (iOS) Safari?

hajimehoshi:
Wow, now Safari supports Opus!!

http://caniuse.com/opus

hajimehoshi:
Ugh, now Safari doesn't support that...

hajimehoshi:
https://caniuse.com/#feat=opus Now Safari does support this?

hajimehoshi:
CC @wasedaigo 

We think Opus is relatively important for performance in terms of memory. Let's consider to implement a decoder.

hajimehoshi:
https://github.com/pion/opus

PR #3393, opened 2026-02-17
audio: add opus package

kazzmir:
Add support for opus audio decoding in the `ebiten/v2/audio/opus` package.

feature

## What this PR does | solves
This allows ebiten to decode opus audio files using a pure-go audio library (https://github.com/kazzmir/opus-go). This PR was made by copying mp3/decode.go to opus/decode.go, and keeping nearly everything the same but replacing `mp3.NewDecoder` with `opus.NewPlayerFromReader`.

The examples/audio has also been modified to add an opus option to play the ragtime music.

The ragtime.opus file was generated via
```
$ ffmpeg -i ragtime.ogg ragtime.opus
```

### DIFF
--- a/audio/opus/decode.go
+++ b/audio/opus/decode.go
@@ -39,6 +39,8 @@ func (s *Stream) Seek(offset int64, whence int) (int64, error) {
 }
 
 // Length returns the size of decoded stream in bytes.
+//
+// Length returns -1 if the source did not implement io.Seeker.
 func (s *Stream) Length() int64 {
 	return s.length
 }


hajimehoshi:
So does this resolve #368?

I'll take a look later, thanks

kazzmir:
Ah interesting, I didn't even see that issue. Yes this PR should resolve that issue completely, the underlying opus library will work on the browser.

audio/opus/decode.go
```diff
@@ -0,0 +1,133 @@
+// Copyright 2017 The Ebiten Authors
```
hajimehoshi:
```
// Copyright 2026 The Ebitengine Authors
```

audio/opus/decode.go
```diff
@@ -0,0 +1,133 @@
...
+// DecodeWithSampleRate automatically resamples the stream to fit with sampleRate if necessary.
+//
+// The returned Stream's Seek is available only when src is an io.Seeker.
+//
+// A Stream doesn't close src even if src implements io.Closer.
+// Closing the source is src owner's responsibility.
+//
+// Resampling can be a very heavy task. Stream has a cache for resampling, but the size is limited.
+// Do not expect that Stream has a resampling cache even after whole data is played.
+func DecodeWithSampleRate(sampleRate int, src io.Reader) (*Stream, error) {
```
hajimehoshi: We don't need this function. We don't have to keep the consistency with other audio decoder, which include old APIs.

audio/opus/decode.go
```diff
@@ -0,0 +1,133 @@
...
+
+// DecodeWithoutResampling decodes an Opus source and returns a decoded stream in signed 16bit integer, little endian, 2 channels (stereo) format.
+//
+// DecodeWithoutResampling returns error when decoding fails or IO error happens.
+//
+// The returned Stream's Seek is available only when src is an io.Seeker.
+//
+// A Stream doesn't close src even if src implements io.Closer.
+// Closing the source is src owner's responsibility.
+func DecodeWithoutResampling(src io.Reader) (*Stream, error) {
```
hajimehoshi: This name should be `Decode`.

audio/opus/decode.go
```diff
@@ -0,0 +1,99 @@
...
+// DecodeF32 decodes an Opus source and returns a decoded stream in 32bit float, little endian, 2 channels (stereo) format.
+//
+// DecodeF32 returns error when decoding fails or IO error happens.
+//
+// The returned Stream's Seek is available only when src is an io.Seeker.
+//
+// A Stream doesn't close src even if src implements io.Closer.
+// Closing the source is src owner's responsibility.
+func DecodeF32(src io.Reader) (*Stream, error) {
+	d, err := opus.NewPlayerFromReader(src)
```
hajimehoshi: I was wondering if there is a direct function to decode 32bit floats. If we could use it, the performance would be much better.

kazzmir: Yea I think its possible, the underlying opus decoder has opus_decode_float, which I can try to expose in the go library.

hajimehoshi: Thanks, I'd like to use it if possible. In v3, I plan to integrate decoding APIs to the float32 version.

kazzmir:
Now DecodeF32() uses a native float32 decoder that opus provides

audio/opus/decode.go
```diff
@@ -0,0 +1,107 @@
...
+	d, err := opus.NewPlayerF32FromReader(src)
+	if err != nil {
+		return nil, err
+	}
+
+	var length int64 = -1
+	_, ok := src.(io.Seeker)
+	if ok {
+		length = d.Length()
+		d.Seek(0, io.SeekStart)
```
hajimehoshi: Do not ignore an error here.

audio/opus/decode.go
```diff
@@ -0,0 +1,107 @@
...
+	d, err := opus.NewPlayerFromReader(src)
+	if err != nil {
+		return nil, err
+	}
+
+	var length int64 = -1
+	_, ok := src.(io.Seeker)
+	if ok {
+		length = d.Length()
+		d.Seek(0, io.SeekStart)
```
hajimehoshi: ditto.

audio/opus/decode.go
```diff
@@ -66,7 +66,10 @@ func DecodeF32(src io.Reader) (*Stream, error) {
 	_, ok := src.(io.Seeker)
 	if ok {
 		length = d.Length()
-		d.Seek(0, io.SeekStart)
+		_, err := d.Seek(0, io.SeekStart)
+		if err != nil {
```
hajimehoshi: I prefer one line:

```
if _, err := d.Seek(0, io.SeekStart); err != nil {
```

audio/opus/decode.go
```diff
@@ -95,7 +98,10 @@ func Decode(src io.Reader) (*Stream, error) {
...
```
hajimehoshi: ditto

audio/opus/decode.go
```diff
@@ -0,0 +1,113 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package opus provides Opus decoder.
+package opus
+
+import (
+	"io"
+
+	opus "github.com/kazzmir/opus-go/player"
```
hajimehoshi: Why not `github.com/kazzmir/opus-go`?

audio/opus/decode.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+	}
+
+	s := &Stream{
+		readSeeker: d,
+		length:     length,
+		sampleRate: d.SampleRate(),
+	}
+	return s, nil
+}
```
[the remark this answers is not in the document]
hajimehoshi: That's intentional.

audio/opus/decode.go
```diff
@@ -0,0 +1,113 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	s := &Stream{
+		readSeeker: d,
+		length:     length,
+		sampleRate: d.SampleRate(),
+	}
```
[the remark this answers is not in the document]
kazzmir: the mp3 decoder seems to return -1 if the source is not seekable. should we use -1 or 0

audio/opus/decode.go
```diff
@@ -18,7 +18,7 @@ package opus
 import (
 	"io"
 
-	opus "github.com/kazzmir/opus-go/player"
+	opus "github.com/kazzmir/opus-go"
```
hajimehoshi: We don't need `opus` package alias, right?

audio/opus/decode.go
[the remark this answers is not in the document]
hajimehoshi: Hmm, good point. Let's use -1.

I'll file a bug for vorbis.

audio/opus/decode.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+	}
+
+	s := &Stream{
+		readSeeker: d,
+		length:     length,
+		sampleRate: d.SampleRate(),
+	}
+	return s, nil
+}
```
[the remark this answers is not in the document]
hajimehoshi: I think a test is optional as this is a very thin wrapper.

audio/opus/decode.go
```diff
@@ -0,0 +1,111 @@
...
+// Closing the source is src owner's responsibility.
+func DecodeF32(src io.Reader) (*Stream, error) {
+	d, err := opusgo.NewPlayerF32FromReader(src)
+	if err != nil {
+		return nil, err
+	}
+
+	var length int64 = -1
+	_, ok := src.(io.Seeker)
+	if ok {
```
hajimehoshi:
```go
if _, ok := src.(io.Seeker); ok {
```

audio/opus/decode.go
```diff
@@ -0,0 +1,111 @@
...
+// Closing the source is src owner's responsibility.
+func Decode(src io.Reader) (*Stream, error) {
+	d, err := opusgo.NewPlayerFromReader(src)
+	if err != nil {
+		return nil, err
+	}
+
+	var length int64 = -1
+	_, ok := src.(io.Seeker)
+	if ok {
```
hajimehoshi: ditto

audio/opus/decode.go
```diff
@@ -0,0 +1,111 @@
...
+	return s.readSeeker.Read(buf)
+}
+
+// Seek is implementation of io.Seeker's Seek.
+func (s *Stream) Seek(offset int64, whence int) (int64, error) {
+	return s.readSeeker.Seek(offset, whence)
+}
+
+// Length returns the size of decoded stream in bytes.
+func (s *Stream) Length() int64 {
```
hajimehoshi, line 44: Write a comment to explain when Legnth returns -1.

audio/opus/decode.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// Seek is implementation of io.Seeker's Seek.
+func (s *Stream) Seek(offset int64, whence int) (int64, error) {
+	return s.readSeeker.Seek(offset, whence)
+}
+
+// Length returns the size of decoded stream in bytes.
+//
+// Length returns -1 if the source did not implement io.Seeker.
```
hajimehoshi, line 43: Just to confirm: My understanding is that github.com/kazzmir/opus-go doesn't treat an infinite stream of opus (yet), then go-opus player's Length always returns a finite value. Is that correct?

kazzmir: In theory you can pass a stream with unbounded length to the opus player and it will work. I have tried using the opus library with a live stream of opus that has no defined end.

For streams that are not seekable, Length() in the opus library reads to the end of the stream, so in the case of an infinite stream it would likely return a very small number as the current 'end' of the stream is probably a second or two into the future.

If the stream were truly infinite somehow then Length() would get stuck in an infinite loop and not return anything.

hajimehoshi: Hmm OK. I'll revisit this when we really need to handle unbounded stream. Thanks,

PR #3399, opened 2026-02-23
Unify macOS and iOS gamepad code and add vibration

user-none:
# What issue is this addressing?
Updates #2014 and #2015 . 

## What _type_ of issue is this addressing?
Feature

## What this PR does | solves
Adds support for vibration on macOS and iOS. Tested on both using [ebiten_gamepad_viewer](https://github.com/user-none/ebiten_gamepad_viewer) (made for this PR).

Additionally, from #3398 the gamepad code for macOS is now unified with iOS and both are using the same cgo based GCContoller Api. Updated gamecontrollerdb.txt and all entires are using the iOS type because those are meant for GCContoller. The macOS entires in the file are for the older IOHIDDevice integration method.

Issue #2015, opened 2022-03-11
controller vibrations for mobiles


hajimehoshi:
Forked from #1452

hajimehoshi:
@user-none 

https://github.com/hajimehoshi/ebiten/blob/93c19f25dd90ed3e87b683af8544bbb68b07db55/internal/gamepad/haptics_gc_darwin.go#L177-L180

I was wondering why the left motor uses 'strong' and the right motor 'weak'. Is this intentional?

https://github.com/hajimehoshi/ebiten/blob/93c19f25dd90ed3e87b683af8544bbb68b07db55/internal/gamepad/gamepad_gc_darwin.go#L56

Why is the left and the right motors are swapped?

hajimehoshi:
> Why is the left and the right motors are swapped?

I guess this was just a mistake so I fixed this.

hajimehoshi:
> I was wondering why the left motor uses 'strong' and the right motor 'weak'. Is this intentional?

This is very interesting, but Xbox controllers have such motors, right?

https://source.chromium.org/chromium/chromium/src/+/main:device/gamepad/game_controller_gamepad.mm;l=362-366;drc=64820206068b8feaf4b9d4de22362cfcf047b8b6

user-none:
> > Why is the left and the right motors are swapped?
> 
> I guess this was just a mistake so I fixed this.

Definitely a mistake.

> > I was wondering why the left motor uses 'strong' and the right motor 'weak'. Is this intentional?
> 
> This is very interesting, but Xbox controllers have such motors, right?
> 
> https://source.chromium.org/chromium/chromium/src/+/main:device/gamepad/game_controller_gamepad.mm;l=362-366;drc=64820206068b8feaf4b9d4de22362cfcf047b8b6

Both Xbox and Playstation have different sized motors on each side for different intensity rumble. For the Xbox you can see some images in this [iFxit teardown](https://www.ifixit.com/Guide/Xbox+360+Wireless+Controller+Vibration+Motors+Replacement/3292). It's much easier to see the difference in the motors in their [Playstation](https://www.ifixit.com/Guide/DualShock+4+Controller+Rumble+Motors+Replacement/106820) teardown.

[...]

Issue #2014, opened 2022-03-11
controller vibrations for desktops


hajimehoshi:

hajimehoshi:
* Windows: https://github.com/libsdl-org/SDL/blob/2290d71338f11ac946896101bf732d1b0b52ff13/src/joystick/windows/SDL_windowsjoystick.c#L689-L696
  * XInput: https://github.com/libsdl-org/SDL/blob/2290d71338f11ac946896101bf732d1b0b52ff13/src/joystick/windows/SDL_xinputjoystick.c#L484-L498
  * DInput: https://github.com/libsdl-org/SDL/blob/2290d71338f11ac946896101bf732d1b0b52ff13/src/joystick/windows/SDL_dinputjoystick.c#L895-L938

hajimehoshi:
On macOS, Xbox/PS controllers don't support ForceFeedback unfortunately.

https://www.macworld.com/article/160155/immersion-2.html

hajimehoshi:
@user-none FYI: I've updated internal/gamepaddb for macOS and iOS: 380ae844a0c447e3fd7e2247b0e964c6d595e6df

PR #3400, opened 2026-02-24
gamepad: unify iOS and macOS gamepad support with the same GCContoller API

user-none:
Implements #3398 needed for future changes for #2014 and #2015.

Feature

## What this PR does | solves
Gamepad code for macOS is now unified with iOS and both are using the same cgo based GCContoller Api. Changes to gamepaddb are needed so the controller mappings are detected on both platforms.

This is part one of splitting #3399 into separate PRs.


internal/gamepaddb/gamepaddb.go
```diff
@@ -127,8 +123,9 @@ func parseLine(line string, platform platform) (id string, name string, buttons
 				if platform != platformAndroid {
 					return "", "", nil, nil, nil
 				}
+			case "Darwin":
```
hajimehoshi: Does the latest gamepad db file use the term "Darwin" now?

user-none: Yes it does. Darwin is used for both iOS and macOS through the OS detection system and anything tagged as _darwin builds for both, e.g. gamecontrollerdb_darwin.txt‎.

The gamedb changes unifies the mapping for both platforms. Otherwise, a copy of both the data would need to be in the _ios.txt and _macos.txt. Plus, other changes would be needed to have the _macos.txt generated using the iOS data from the master db file. The iOS entries in the master db file are for GCContoller, whereas the macOS ones are for the old system.

hajimehoshi: This string is for https://raw.githubusercontent.com/mdqinc/SDL_GameControllerDB/master/gamecontrollerdb.txt, not for a Go build tag. I failed to find "Darwin" in the text file.

You need to have a deeper understanding of what this Go file does.

user-none: > You need to have a deeper understanding of what this Go file does

Nothing from the gamecontrollerdb.txt for macOS is used any longer because those entires do not apply to GCContoller. Which is why `case "Mac OS X":` was removed.

`case "iOS"` on line 127 of gamepaddb.go‎ is what pulls in the proper cases iOS case that are used for both iOS macOS (aka darwin build). the `case "Darwin"` technically doesn't do anything but it's there to denote the entires in the file apply to both. It is extraneous in that it doesn't actually do anything. I can be replace it with a comment stating it applies to both.

The file data is being included in the build and is used by both systems via internal/gamepaddb/gen.go‎ which generates the gamecontrollerdb_darwin.txt. from the iOS entries in gamecontrollerdb.txt. This is then embedded into internal/gamepaddb/db_darwin.go‎ line 23, `//go:embed gamecontrollerdb_darwin.txt`. db_darwin.go‎ is complied for both iOS and macOS.

user-none: I changed it to a comment explaining why iOS entires apply to both iOS and macOS instead of using the non-existing Darwin name.

internal/gamepaddb/gamepaddb.go
```diff
@@ -115,10 +115,6 @@ func parseLine(line string, platform platform) (id string, name string, buttons
 				if platform != platformWindows {
 					return "", "", nil, nil, nil
 				}
-			case "Mac OS X":
```
hajimehoshi, line 118: Why is this removed...?

hajimehoshi: > Nothing from the gamecontrollerdb.txt for macOS is used any longer because those entires do not apply to GCContoller. 

Hmm really? I'll take a look later.

internal/gamepaddb/gamepaddb.go
```diff
@@ -33,10 +33,10 @@ type platform int
 const (
 	platformUnknown platform = iota
 	platformWindows
-	platformMacOS
 	platformUnix
 	platformAndroid
 	platformIOS
```
hajimehoshi: `platformIOS` should no longer be needed

internal/gamepaddb/gamepaddb.go
user-none: I was surprised too that IOKit and GCContoller have different GUIDs for controllers. If you look you'll see the same controllers with different IDs.

hajimehoshi: My worry is whether there is a game pad supported by IOKit but not by GCController. I hope this is pretty rare.

user-none: From my test app detecting a PS4 controller `gamepad connected: id=0 name="PS4 Controller" sdl_id=050000004c050000cc090000ff870001 standard=true`. Which is from the iOS section of the txt file. The is running on macOS.

internal/gamepaddb/gamepaddb.go
user-none: I need to look at this. I removed it and I'm seeing weird behavior on iOS. I'm seeing the correct sdl_id but a different name and standard layout set to false. I'll look at it later today.

internal/gamepaddb/gamepaddb.go
hajimehoshi: I looked at the SDL implementation and they have both implementation like what Ebitengine does. I'm now convinced that the iOS part for the gamepad db should work.

In the ideal world, IOKit also should be supported as a fallback, but the implementation would be pretty complicated than benefit.

user-none: > My worry is whether there is a game pad supported by IOKit but not by GCController

There are definitely older gamepads that aren't covered by GCContoller. Is supporting new features like vibrate more important than supporting older gamepads. Or, is retaining support for older gamepads more important.

> In the ideal world, IOKit also should be supported as a fallback

Indeed, but that's way beyond what I want to get into right now.

hajimehoshi: > > In the ideal world, IOKit also should be supported as a fallback
>
> Indeed, but that's way beyond what I want to get into right now.

On second thought, this might not be harder than I expected. Let me think.

internal/gamepaddb/gamepaddb.go
user-none: I got a free moment. `platformIOS` is removed and working. Dumb edit this morning that was the issue.

internal/gamepaddb/gamepaddb.go
hajimehoshi: https://github.com/hajimehoshi/ebiten/commit/89bb8fb3a0d365e2713100dc60f84377c614374a

I've refactored internal/gamepad. Let's add an environment variable to use HID later.

hajimehoshi: Thus, please keep HID implementations. You need to rebase this PR anyway.

user-none:
I'm going to close this out because of the refactoring toward maintaining IOKit gamepad support on macOS (possibly as a fallback). I've opened a PR for rumble specifically with GCContoller and will look at the gamepaddb using the iOS entries on macOS for a toggle for macOS to use GCContoller too.

I don't think internal fallback from GCContoller to IOKit will be possible but I will spend some time investigating that too.

PR #3440, opened 2026-04-23
NewShader: allow multiple sources as arguments

kazzmir:
New feature discussed in #dev to allow multiple arguments to NewShader

feature

## What this PR does | solves
NewShader(src) accepts one argument that should contain the entirety of the shader program, including all helper functions. It is useful to put helper functions into their own file, so as a convenience NewShader() can accept multiple arguments. The full shader program is the concatenation of all the arguments (order doesn't matter).

```
helpers := loadHelperFunctions()
mainShader := loadShader()
shader := ebiten.NewShader(mainShader, helpers)
```

### DIFF
--- a/shader.go
+++ b/shader.go
@@ -35,12 +35,31 @@ type Shader struct {
 }
 
 // NewShader compiles a shader program in the shading language Kage, and returns the result.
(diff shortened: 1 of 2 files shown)


hajimehoshi:
I'm afraid we need more discussion about API design before implementing this.

PR #3461, opened 2026-06-04
audio: panic on invalid player volume

### LINKED ISSUE #3359: audio: `audio.Player.SetVolume` should allow a volume larger than 1
### Ebitengine Version

v2.9.4

### Operating System

- [ ] Windows
- [ ] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

### Go Version (`go version`)

1.25

The documentation for `audio.Player.SetVolume` at https://pkg.go.dev/github.com/hajimehoshi/ebiten/v2@v2.9.5/audio#Player.SetVolume says this:

> volume must be in between 0 and 1. SetVolume panics otherwise.

As far as I can tell, it's fine to set the volume to a value above 1, though -- I don't see a panic, and the audio is amplified as I'd expect. oto looks like it's just multiplying `float32` values here:

https://github.com/ebitengine/oto/blob/fca52fc6a0fcdf8d917e70b68376677bd14efb6c/internal/mux/mux.go#L497-L506

If this is indeed an allowed use, would it be reasonable to update the docs to say something like this?

> volume must be 0 or higher. If a volume larger than 1 is passed, clipping may occur.

Documentation should match the observed behavior.

### What happens instead?

Documentation differs from the observed behavior.

_No response_

puneetdixit200:
## Summary
- make `audio.Player.SetVolume` enforce its documented [0, 1] range before delegating to the backend
- keep valid boundary values accepted and reject negative, greater-than-one, and NaN inputs
- add regression coverage for the documented panic behavior

Fixes #3359

## Testing
- `go test ./audio`
- `go test ./audio/...`
- `git diff --check`
- `go test ./...` fails locally in packages that initialize the Darwin GLFW UI path with a nil pointer in `internal/ui.currentMouseLocation`; the audio packages pass

### DIFF
--- a/audio/audio.go
+++ b/audio/audio.go
@@ -515,6 +515,9 @@ func (p *Player) Volume() float64 {
 // SetVolume sets the volume of this player.
 // volume must be in between 0 and 1. SetVolume panics otherwise.
 func (p *Player) SetVolume(volume float64) {
+	if !(0 <= volume && volume <= 1) {
+		panic(fmt.Sprintf("audio: volume must be in between 0 and 1 but was %v", volume))
+	}
 	p.p.SetVolume(volume)
 }
 
--- a/audio/audio_test.go
+++ b/audio/audio_test.go
@@ -17,6 +17,7 @@ package audio_test
 import (
 	"bytes"
 	"io"
+	"math"
 	"runtime"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


hajimehoshi:
I think the document side should be fixed for compatibility

PR #3472, opened 2026-06-29
ebitenutil: check HTTP status in NewImageFromURL

BakedSoups:
NewImageFromURL previously decoded every successful http.Get response body regardless of HTTP status. For responses like 404 or 500, callers got an image decode error instead of the HTTP failure.

This change checks for non-2xx responses before decoding and returns an error containing the URL and HTTP status.

Tested with:
- go test ./ebitenutil

### DIFF
--- a/ebitenutil/loadimage.go
+++ b/ebitenutil/loadimage.go
@@ -15,6 +15,7 @@
 package ebitenutil
 
 import (
+	"fmt"
 	"image"
 	"io"
 	"net/http"
@@ -47,6 +48,9 @@ func NewImageFromURL(url string) (*ebiten.Image, error) {
 	defer func() {
 		_ = res.Body.Close()
 	}()
+	if res.StatusCode/100 != 2 {
+		return nil, fmt.Errorf("GET %s: %s", url, res.Status)
+	}
 
 	img, _, err := image.Decode(res.Body)
 	if err != nil {
--- a/ebitenutil/loadimage.go
+++ b/ebitenutil/loadimage.go
@@ -49,7 +49,7 @@ func NewImageFromURL(url string) (*ebiten.Image, error) {
 		_ = res.Body.Close()
 	}()
 	if res.StatusCode/100 != 2 {
-		return nil, fmt.Errorf("GET %s: %s", url, res.Status)
+		return nil, fmt.Errorf("ebitenutil: request to %s failed: %s", url, res.Status)
 	}
 
 	img, _, err := image.Decode(res.Body)
--- a/ebitenutil/loadimage.go
+++ b/ebitenutil/loadimage.go
@@ -48,7 +48,7 @@ func NewImageFromURL(url string) (*ebiten.Image, error) {
 	defer func() {
 		_ = res.Body.Close()
 	}()
-	if res.StatusCode/100 != 2 {
+	if res.StatusCode < http.StatusOK || 300 <= res.StatusCode {
 		return nil, fmt.Errorf("ebitenutil: request to %s failed: %s", url, res.Status)
 	}
 
--- a/ebitenutil/loadimage_test.go
+++ b/ebitenutil/loadimage_test.go
@@ -15,9 +15,10 @@
 package ebitenutil_test
 
 import (
+	"image"
+	"image/png"
 	"net/http"
 	"net/http/httptest"
-	"strings"
 	"testing"
 
 	"github.com/hajimehoshi/ebiten/v2/ebitenutil"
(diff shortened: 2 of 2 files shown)


BakedSoups:
Big fan of your work!

ebitenutil/loadimage.go
```diff
@@ -47,6 +48,9 @@ func NewImageFromURL(url string) (*ebiten.Image, error) {
 	defer func() {
 		_ = res.Body.Close()
 	}()
+	if res.StatusCode < http.StatusOK || res.StatusCode >= http.StatusMultipleChoices {
```
hajimehoshi: Avoiding magic numbers is good in general, but in this case, we need to know that multiple-choices is the first 3xx status code.

Assuming everyone already knows 2xx is the successful status, I prefer writing this:

```go
if res.StatusCode < 200 || res.StatusCode >= 300 {
```

ebitenutil/loadimage.go
```diff
@@ -47,6 +48,9 @@ func NewImageFromURL(url string) (*ebiten.Image, error) {
...
+		return nil, fmt.Errorf("GET %s: %s", url, res.Status)
```
hajimehoshi:
```go
fmt.Errorf("ebitenutil: request to %s failed: %s", url, res.Status)
```

hajimehoshi (commented):
Thanks!

ebitenutil/loadimage_test.go
```diff
@@ -0,0 +1,39 @@
...
+	s := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		http.Error(w, "not found", http.StatusNotFound)
+	}))
+	defer s.Close()
+
+	_, err := ebitenutil.NewImageFromURL(s.URL)
+	if err == nil {
+		t.Fatal("NewImageFromURL returned nil error for HTTP 404")
+	}
+	if got, want := err.Error(), "404 Not Found"; !strings.Contains(got, want) {
```
hajimehoshi: Hmm, checking an error message content is not good. I'd like to avoid this.

ebitenutil/loadimage.go
BakedSoups: https://github.com/hajimehoshi/ebiten/commit/e032e1b5f435e2fba91db64721483e8160ecddfe

BakedSoups: good point didn't think about that

ebitenutil/loadimage.go
BakedSoups: https://github.com/hajimehoshi/ebiten/pull/3472/changes/ea88ba49a7130e513d3b1b07105d9db5eacca972

swapped out the Error log for this instead!

ebitenutil/loadimage.go
```diff
@@ -47,6 +48,9 @@ func NewImageFromURL(url string) (*ebiten.Image, error) {
 	defer func() {
 		_ = res.Body.Close()
 	}()
+	if res.StatusCode/100 != 2 {
```
hajimehoshi: This seems hacky... Just use comparisons

ebitenutil/loadimage_test.go
BakedSoups: https://github.com/hajimehoshi/ebiten/pull/3472/changes/b65be6bd458f9c9e32774aaf3d381a5941b1d1d9

now it checks the HTTP status first and fails on 404 before having decoding, 

now it reports failures directly

ebitenutil/loadimage_test.go
```diff
@@ -0,0 +1,40 @@
...
+	s := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusNotFound)
+		if err := png.Encode(w, image.NewRGBA(image.Rect(0, 0, 1, 1))); err != nil {
+			t.Errorf("png.Encode failed: %v", err)
+		}
+	}))
+	defer s.Close()
+
+	_, err := ebitenutil.NewImageFromURL(s.URL)
+	if err == nil {
```
hajimehoshi:
```go
if _, err := ebitenutil.NewImageFromURL(s.URL); err == nil {
```

ebitenutil/loadimage.go
BakedSoups: my fault, that wasn't well made code 
https://github.com/hajimehoshi/ebiten/pull/3472/commits/abc51d61575a96290d72944d5afd3215c358c90c

ebitenutil/loadimage_test.go
BakedSoups: https://github.com/hajimehoshi/ebiten/pull/3472/changes/2510239e3d6b78d6474c1634d23f1274fbb6f85b

hajimehoshi (approved):
LGTM, thanks

ebitenutil/loadimage.go
```diff
@@ -47,6 +48,9 @@ func NewImageFromURL(url string) (*ebiten.Image, error) {
 	defer func() {
 		_ = res.Body.Close()
 	}()
+	if res.StatusCode < http.StatusOK || 300 <= res.StatusCode {
```
hajimehoshi, line 51: Ah, I'd write 200 rather than StatusOK. I'll add a follow-up change.

PR #3473, opened 2026-06-29
examples/vm: reuse keyboard input buffers

BakedSoups:
The VM example already reuses buffers for gamepad and touch input while forwarding input to the guest, but keyboard keys and input runes still passed nil to append helpers each tick.

This adds reusable key/rune buffers to avoid per-input slice allocation in forwardInput.

Tested with:
- go test ./examples/vm
- go test ./exp/vmhost


hajimehoshi (approved):
LGTM

PR #3474, opened 2026-07-02
internal/glfw: listen to XkbMapNotify events to rebuild key tables

nriekert:
Listen to XkbMapNotify events from the X server and call createKeyTables() when the keyboard mapping changes, so the keysym-based fallback correctly reflects the new layout.


hajimehoshi:
Please do not ignore the pull request template. Recreate a pull request with an appropriate message.

hajimehoshi:
(If the change is obvious, it's ok to ignore the template, but I don't think the change is not so obvious in this PR's case)

PR #3482, opened 2026-07-11
fix: detect canvas resize with ResizeObserver on the JS backend

### LINKED ISSUE #3067: There are some cases where the screen resizes without the resize event being fired in mobile browsers
### Ebitengine Version

latest

### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [X] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [X] Web Browsers

### Go Version (`go version`)

go1.22.6

Open a wasm binary in this page https://a28924d7.go-wasm-player.pages.dev/

I reproduced only on iOS Chrome.

Rendered in the correct aspect ratio

### What happens instead?

Shrink a little vertically from the correct aspect ratio.

There seems to be some cases where the screen resizes without the resize event being fired in mobile browsers.

I fixed this issue with `ResizeObserver` that captures all resizings. See https://github.com/eihigh/go-wasm-player/commit/320e6e4b68c162a58ab2e6c25fc62bdd303d31d2

mvanhorn:
Closes #3067

bug

## What this PR does | solves
On the JS/wasm backend, canvas size changes are now detected with a `ResizeObserver` in addition to the `window` `resize` event. Mobile browsers (reproduced by the maintainer on Chrome on iPhone; labels `bug`, `os:ios`, `os:js`) sometimes resize the canvas/viewport without firing a `window` `resize` event, so Ebitengine's screen-size update logic never ran and the drawing surface stayed at a stale size — producing the reported vertically-shrunk / wrong-aspect-ratio rendering. The reporter identified the fix in their own project by attaching a `ResizeObserver` that captures every resize.

`internal/ui/ui_js.go` now observes the canvas element with a `ResizeObserver` and runs the same size-update path it runs on a `window` `resize`, so a canvas resize that skips the window event still updates the drawing surface. The existing `resize` listener is retained; the observer is set up and torn down alongside it. Built the wasm target and confirmed a canvas resize that does not emit a `window` `resize` event now updates the game surface, while a normal window resize still works as before.

### DIFF
--- a/internal/ui/ui_js.go
+++ b/internal/ui/ui_js.go
@@ -561,18 +561,22 @@ func (u *UserInterface) init() error {
 
 func (u *UserInterface) setWindowEventHandlers(v js.Value) {
 	v.Call("addEventListener", "resize", js.FuncOf(func(this js.Value, args []js.Value) any {
-		u.updateScreenSize()
-
-		// updateImpl can block. Use goroutine.
-		// See https://pkg.go.dev/syscall/js#FuncOf.
-		go func() {
-			if err := u.updateImpl(true); err != nil {
-				u.setError(err)
-				return
-			}
-		}()
+		u.onResize()
 		return nil
 	}))
+}
+
+func (u *UserInterface) onResize() {
+	u.updateScreenSize()
+
+	// updateImpl can block. Use goroutine.
+	// See https://pkg.go.dev/syscall/js#FuncOf.
+	go func() {
+		if err := u.updateImpl(true); err != nil {
+			u.setError(err)
+			return
+		}
+	}()
 }
 
 // SetTextInputFocusedFunc registers a function reporting whether an element
@@ -597,6 +601,14 @@ func (u *UserInterface) isTextInputFocused() bool {
 }
 
 func (u *UserInterface) setCanvasEventHandlers(v js.Value) {
+	if resizeObserver := js.Global().Get("ResizeObserver"); resizeObserver.Truthy() {
+		observer := resizeObserver.New(js.FuncOf(func(this js.Value, args []js.Value) any {
+			u.onResize()
+			return nil
+		}))
+		observer.Call("observe", v)
+	}
+
 	// Keyboard
 	v.Call("addEventListener", "keydown", js.FuncOf(func(this js.Value, args []js.Value) any {
 		// Focus the canvas explicitly to activate tha game (#961).


hajimehoshi:
Have you confirmed this was reproducible?

mvanhorn:
Yes - reproduced at the code-path level. I built a small harness on this branch: with a canvas-only resize where the window size stays constant, the old path keeps reporting the stale 800x600 size, while the ResizeObserver path picks up the new 800x420 canvas size. Scoped GOOS=js GOARCH=wasm builds, the native internal/ui tests, and go vet pass on the branch. I haven't verified in a live browser yet - happy to run an in-browser check (e.g. one of the examples served and resized via devtools) if that would help.

hajimehoshi:
I was wondering whether you used an iOS device (a physical one or an emulator) to reproduce the issue, like the original reporter said

hajimehoshi:
> I haven't verified in a live browser yet - happy to run an in-browser check (e.g. one of the examples served and resized via devtools) if that would help.

This is the key. I hope we could have a minimal reproducible case in order to confirm your fix really works.

mvanhorn:
Ran the in-browser check with a minimal repro, comparing this branch against its base commit (Chrome on macOS, devicePixelRatio 2).

Repro: build any example to wasm (I used examples/rotate) with the standard wasm_exec.js page, then shrink the canvas layout box without any window resize:

```js
window.addEventListener('resize', () => console.log('resize fired')); // never fires
document.body.style.width = '60%';
document.body.style.height = '50%';
// after ~1s, compare backing store vs CSS size
const c = document.querySelector('canvas');
console.log([c.width, c.height], [c.clientWidth * devicePixelRatio, c.clientHeight * devicePixelRatio]);
```

Results:

- Base (0012728): the canvas CSS box shrinks to 1071x469 but the backing store stays stale at 3570x1876 (expected 2142x938). The aspect mismatch renders visibly squashed, matching the #3067 symptom. The resize listener confirms no window resize event ever fires in this path.
- This branch (69094a8): the backing store follows to exactly 2142x938 (clientWidth x devicePixelRatio) within a frame or so, and rendering stays correct.

So the ResizeObserver catches the case the window resize handler structurally cannot.

hajimehoshi:
Confirmed this fixed the issue, thanks!

hajimehoshi (approved):
LGTM

Issue #3396, opened 2026-02-21
examples/audiopanning: example is choppy/clicky and lags behind expected pan value


31:
### Ebitengine Version

v2.9.8

### Operating System

- [x] Windows
- [ ] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

### Go Version (`go version`)

go version go1.26.0 windows/amd64

### What steps will reproduce the problem?

This example shows the problem:

https://github.com/hajimehoshi/ebiten/blob/e8a9c3581a1f361ebbb843754297d7ed4b980e36/examples/audiopanning/main.go

```
go run github.com/hajimehoshi/ebiten/v2/examples/audiopanning@latest
```

But to make the problem easier to hear, I made a modified version that uses a sine tone instead of music:

https://github.com/31/ebitengine-panning-compare/blob/2990b748c08894d1092ff7865fa1cd4ec49ed0fb/sin-pan/main.go

```
go run github.com/31/ebitengine-panning-compare/sin-pan@2990b748c08894d1092ff7865fa1cd4ec49ed0fb
```

### What is the expected result?

The panning should be smooth, and the music (or sine tone) sounds good.

I made more samples that use two `Player`s and adjusts their volumes instead, and this works:

https://github.com/31/ebitengine-panning-compare/blob/2990b748c08894d1092ff7865fa1cd4ec49ed0fb/pan-2p/main.go
https://github.com/31/ebitengine-panning-compare/blob/2990b748c08894d1092ff7865fa1cd4ec49ed0fb/sin-pan-2p/main.go

Here's what the sine tone should sound like:

https://github.com/user-attachments/assets/33e0d31b-8d1d-45bc-b966-926000982826

Here's what it looks like as a spectrogram: notice smooth changes in volume (one channel shown):

<img width="379" height="205" alt="Image" src="https://github.com/user-attachments/assets/dd4c9771-94ef-452a-80b1-6a2ac2c115ed" />

### What happens instead?

The music panning is delayed compared to the position on screen and suddenly changes panning in a stuttering way rather than being smooth.

https://github.com/user-attachments/assets/0a8e48d8-5540-4a79-9565-66aeeac1f586

It's more obvious with the sine tone:

[...]

PR #3483, opened 2026-07-11
examples/audiopanning: add lerp, more specific small buffer reason

31:
Add linear interpolation to the pan effect so it doesn't pop no matter how large the buffer is. Change the comment describing the small buffer size to be more specific about what it solves.
    
Include a pan value mutex for thread safety between the game and sound processing loop.

* Updates https://github.com/hajimehoshi/ebiten/issues/3396

### DIFF
--- a/examples/audiopanning/main.go
+++ b/examples/audiopanning/main.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"log"
 	"math"
+	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


hajimehoshi:
Please do not ignore the PR template.

Also, we have not agreed to apply your suggestion.

31:
Apologies, mishap with the commit-based auto-fill.

PR #3485, opened 2026-07-16
examples/audiopanning: add lerp, more specific small buffer reason

### LINKED ISSUE #3396: examples/audiopanning: example is choppy/clicky and lags behind expected pan value

31:
Closes #3396

bug

## What this PR does | solves
Adds linear interpolation to the pan effect so it doesn't pop no matter how large the buffer is. Changes the comment describing the small buffer size to be more specific about what it solves.

Includes a pan value mutex for thread safety between the game and sound processing loop.

### DIFF
--- a/examples/audiopanning/main.go
+++ b/examples/audiopanning/main.go
@@ -205,8 +205,6 @@ func (s *StereoPanStream) Read(p []byte) (int, error) {
 		p[i+6] = byte(rcBits >> 16)
 		p[i+7] = byte(rcBits >> 24)
 	}
-	// Only update the prevPan value if some frames of audio were actually processed. If none were
-	// processed, interpolation should continue from the old value during the next Read.
 	if alignedN > 0 {
 		s.prevPan = pan
 	}
--- a/examples/audiopanning/main.go
+++ b/examples/audiopanning/main.go
@@ -145,13 +145,10 @@ func main() {
 // based on the Panning.
 type StereoPanStream struct {
 	io.ReadSeeker
-	// m protects pan. Read is called by the audio system, which might be running in a different
+	// mu protects pan. Read is called by the audio system, which might be running in a different
 	// goroutine than the Update loop that calls SetPan.
-	m   sync.Mutex
-	pan float64 // -1: left; 0: center; 1: right
-	// prevPan is the final pan value during the previous Read call. Used to interpolate the pan
-	// effect smoothly regardless of quick pan changes or large buffer sizes. The interpolation
-	// helps avoid a discontinuous signal that sometimes causes audible clicks or pops.
+	mu      sync.Mutex
+	pan     float64 // -1: left; 0: center; 1: right
 	prevPan float64
 	buf     []byte
 }
@@ -177,9 +174,9 @@ func (s *StereoPanStream) Read(p []byte) (int, error) {
 	alignedN := totalN - extra
 
 	prevPan := s.prevPan
-	s.m.Lock()
+	s.mu.Lock()
 	pan := s.pan
-	s.m.Unlock()
+	s.mu.Unlock()
 
 	// This implementation uses a linear scale, ranging from -1 to 1, for stereo or mono sounds.
 	// If pan = 0.0, the balance for the sound in each speaker is at 100% left and 100% right.
@@ -208,8 +205,8 @@ func (s *StereoPanStream) Read(p []byte) (int, error) {
 		p[i+6] = byte(rcBits >> 16)
 		p[i+7] = byte(rcBits >> 24)
 	}
-	// Only update the prevPan value if we actually processed some frames of audio. If we didn't
-	// process any, we still want to interpolate from the old value during the next Read.
+	// Only update the prevPan value if some frames of audio were actually processed. If none were
+	// processed, interpolation should continue from the old value during the next Read.
 	if alignedN > 0 {
 		s.prevPan = pan
 	}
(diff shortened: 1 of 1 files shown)


examples/audiopanning/main.go
```diff
@@ -189,14 +208,24 @@ func (s *StereoPanStream) Read(p []byte) (int, error) {
 		p[i+6] = byte(rcBits >> 16)
 		p[i+7] = byte(rcBits >> 24)
 	}
+	// Only update the prevPan value if we actually processed some frames of audio. If we didn't
```
hajimehoshi: Avoid `we`.  (`we` is ambiguous)

examples/audiopanning/main.go
```diff
@@ -143,8 +145,15 @@ func main() {
 // based on the Panning.
 type StereoPanStream struct {
 	io.ReadSeeker
+	// m protects pan. Read is called by the audio system, which might be running in a different
+	// goroutine than the Update loop that calls SetPan.
+	m   sync.Mutex
```
hajimehoshi: Use `mu`, though most code in Ebitengine used `m`. `mu` is more conventional with other repositories and I started to adopt `mu` for new code.

examples/audiopanning/main.go
```diff
@@ -143,8 +145,15 @@ func main() {
...
 	pan float64 // -1: left; 0: center; 1: right
-	buf []byte
+	// prevPan is the final pan value during the previous Read call. Used to interpolate the pan
+	// effect smoothly regardless of quick pan changes or large buffer sizes. The interpolation
+	// helps avoid a discontinuous signal that sometimes causes audible clicks or pops.
```
hajimehoshi: I don't think we need this comment

31: Removed. This does mean the reason for interpolation isn't mentioned anywhere in the code... is that the intent? Or is there a better way/place to describe it?

(I do think that my particular comment was in a weird place to describe the reason, either way. 😄)

hajimehoshi: > This does mean the reason for interpolation isn't mentioned anywhere in the code... is that the intent? This does mean the reason for interpolation isn't mentioned anywhere in the code... is that the intent? 

Basically yes. The code is pretty obvious and I think we can know what the code does by reading it.

examples/audiopanning/main.go
```diff
@@ -189,14 +205,24 @@ func (s *StereoPanStream) Read(p []byte) (int, error) {
 		p[i+6] = byte(rcBits >> 16)
 		p[i+7] = byte(rcBits >> 24)
 	}
+	// Only update the prevPan value if some frames of audio were actually processed. If none were
+	// processed, interpolation should continue from the old value during the next Read.
```
31: Now that I see it written this way, this comment seems pointless--doesn't actually provide a reason to interpolate from the old value, just restates the code. Not totally sure it's worth attempting to explain (but would be willing to try harder at it).

```suggestion
```

hajimehoshi: Sure

hajimehoshi (approved):
LGTM, thanks!

examples/audiopanning/main.go
hajimehoshi: This is not related to this example, but `prevPan` would have to be reset when the stream is pausing not to cause interpolation. This example doesn't stop the stream, so it's ok not to do this.

Issue #3491, opened 2026-07-26
ebiten: run without X11 or Wayland (EGL on fbdev)


bolthar:
### Operating System

- [ ] Windows
- [ ] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

### What feature would you like to be added?

I've been trying to successfully compile for the TrimUI Smart Pro target for a while, and as expected I've now gotten to the point where the application compiles correctly for the correct target (on v2.10.0-alpha12) but since there's no X server available it errors out when starting:

`glfw: failed to load any of [libX11.so.6 libX11.so]: libX11.so.6: cannot open shared object file: No such file`

again, this is expected. I just wondered if there was any chance of making ebiten work in the absence of an X server - there's no need for it in theory, and other engines (godot for example) are supported on the TrimUI.

### Why is this needed?

Having support for the TrimUI (and all other "retro" handhelds by proxy) would be nifty :)

hajimehoshi:
I have no idea what TrimUI is. It's ok to keep this issue open, but I cannot guarantee when to do it.

By the way, the next missing item for Linux desktop is Wayland.

bolthar:
Just FYI, it's one of now many handheld consoles aimed at retrogaming/indie gaming. See here: https://trimui.net/ . Those generally have a ARM processor and run a trimmed down distro, those distros nearly always lacking an X server - and of course lacking Wayland support too. So basically what would be needed to support it would be to run ebiten without an X server running in the system.

hajimehoshi:
>  So basically what would be needed to support it would be to run ebiten without an X server running in the system.

Things are not so simple. We need to implement EGL backend.

I investigated what to do to support TrimUI. Here is a summary by AI (Claude Code):

<details>
I looked into what this device actually needs.

TrimUI Smart Pro (TG5040) is an Allwinner A133P with a PowerVR GE8300, and it has
neither X nor Wayland. The stock stack is SDL2 patched with the `mali-fbdev` video
driver: EGL built with `EGL_API_FB`, `eglGetDisplay(EGL_DEFAULT_DISPLAY)`, and the
native window is just

    struct fbdev_window { unsigned short width, height; };

filled from `/dev/fb0` via `FBIOGET_VSCREENINFO`. KNULLI applies the same patch to
[...]

PR #3492, opened 2026-07-27
internal/fbdev, internal/ui: run on Linux without a window system

hajimehoshi:

Updates #3491

feature

## What this PR does | solves

A handheld or an embedded device runs its applications on a framebuffer device, with a vendor EGL implementation and no display server. The glfw backend cannot serve one: it fails to load libX11 and the application never starts, which is what #3491 reports on a TrimUI Smart Pro.

This adds a backend that presents through such a device.

- **`internal/fbdev` (new)** reads the display mode from `/dev/fb0` (`FBIOGET_VSCREENINFO`) and creates an EGL context whose native window is the `{width, height uint16}` struct those implementations expect a pointer to. EGL is loaded with purego, so no cgo is involved.
- **`internal/windowsystem` (new)** reports whether a window system is available, from `DISPLAY` / `WAYLAND_DISPLAY` on the systems where one can be missing.
- **`internal/ui`** runs the glfw backend wherever a window system is available and falls back to the framebuffer backend otherwise, reporting an error where neither can serve the game.
- **`internal/graphicsdriver/opengl`** uses the same answer to skip the hints that configure a glfw window, and to treat OpenGL as no alternative to OpenGL ES when loading entry points.

The device investigation is recorded in #3491. In short: the only WSEGL module on the reporter's device is `libpvrNULL_WSEGL.so`, and presentation goes through PowerVR kernel services rather than fbdev or DRM, so the backend sets no mode and flips no page — `eglSwapBuffers` does that, and `/dev/fb0` is only the source of the mode. The surface covers the display, as there is nothing to scale or position it against. `DISPLAY` and `WAYLAND_DISPLAY` are both empty on that device, which is what selects the backend.

Gamepads (evdev) and audio (ALSA) already need no display server and work as they are. Keyboard and mouse input is not implemented yet; it needs evdev.

## What a reviewer should know

- **This has not run on hardware.** Everything through EGL config selection is derived from the driver on the reporter's device; whether the surface actually presents can only be confirmed there, and CI can only typecheck it. Hence `Updates #3491` rather than `Closes`.
- The tests in `internal/fbdev` cover what is verifiable off-device: the `fb_var_screeninfo` layout (a short struct would make the ioctl overflow its buffer), the native window size, and the refresh-rate derivation.
- Builds verified for linux/{amd64,arm64,arm}, darwin, windows, netbsd, js/wasm, freebsd (with the purego `fakecgo=-std` flag) and a cgo android/arm64 build. `-tags=microsoftgdk`, `-tags=nintendosdk` and `-tags=playstation5` typecheck with no new errors.
- Two behaviour changes outside the new backend: on Linux, FreeBSD and NetBSD with neither `DISPLAY` nor `WAYLAND_DISPLAY` set, the error is now "no window system is available" instead of glfw's "failed to load libX11"; and on Android the entry point loader goes straight to OpenGL ES, which it already preferred and where libGL does not exist.

---

Authored by Claude (Claude Code) on behalf of @hajimehoshi.

### DIFF
--- a/internal/fbdev/display_linux.go
+++ b/internal/fbdev/display_linux.go
@@ -1,0 +1,151 @@
+// Copyright 2026 The Ebitengine Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package fbdev provides rendering on a Linux framebuffer device, for systems
+// that have no window system.
+package fbdev
+
+import (
+	"fmt"
+	"unsafe"
+
+	"golang.org/x/sys/unix"
+)
+
+const _FBIOGET_VSCREENINFO = 0x4600
+
+type fbBitfield struct {
+	offset   uint32
+	length   uint32
+	msbRight uint32
+}
+
+// fbVarScreeninfo mirrors the kernel's fb_var_screeninfo. The whole layout is
+// declared because the ioctl fills all of it.
+type fbVarScreeninfo struct {
+	xres        uint32
+	yres        uint32
+	xresVirtual uint32
+	yresVirtual uint32
+	xoffset     uint32
+	yoffset     uint32
+
+	bitsPerPixel uint32
+	grayscale    uint32
+
+	red    fbBitfield
+	green  fbBitfield
+	blue   fbBitfield
+	transp fbBitfield
+
+	nonstd   uint32
+	activate uint32
+
+	height uint32
+	width  uint32
+
+	accelFlags uint32
+
+	pixclock    uint32
+	leftMargin  uint32
+	rightMargin uint32
+	upperMargin uint32
+	lowerMargin uint32
+	hsyncLen    uint32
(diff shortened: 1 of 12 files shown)


hajimehoshi:
@bolthar try this change

bolthar:
On it - trying now

bolthar:
Got this now:

`2026-07-27 20:21:40.208689129 +0200 CEST m=+0.310650668 - ERROR fbdev: eglCreateWindowSurface failed: EGL error 0x300b`

I can find this https://community.toradex.com/t/eglcreatewindowsurface-return-error-0x300b/11378 that seems to be related. 

Here though https://github.com/hajimehoshi/ebiten/pull/3492/changes#diff-7093d81fa5f2ce884405d488c942c711da94272863b13d4b6bbfefb5487f1a44R113 doesn't seem like it's possible to pass flags in, but I'm 100% unfamiliar with things this low level, so you know better

hajimehoshi:
@bolthar I've removed the necessity of window. ~~Apparently this is a more 'pure' EGL way.~~ Simply a window was not needed in this case. Try again

bolthar:
Wow, it works :tada: 
photo_2026-07-27_21-20-19
I can't thank you enough. Let me know if you want to have a device shipped to you for testing purposes. I will myself test every release for compatibility and let you know if something breaks. Again, thanks a lot.

hajimehoshi:
Cool!

Now I added additional changes for the future (try a window, and then try querying as a fallback: some environment might require a window). I'm happy if you could try this.

bolthar:
Of course, rebuilding now

bolthar:
Still works great :+1:

hajimehoshi:
Thanks! I'll merge after the tests pass. Thank you for testing it!

(EDIT: I have to go to bed, so maybe tomorrow)

PR #3493, opened 2026-07-28
fix: improve unfocused window frame pacing

### LINKED ISSUE #3397: FPS of unfocused window is ~56 not ~60
### Ebitengine Version

v2.9.8

### Operating System

- [ ] Windows
- [ ] macOS
- [x] Linux (X11)
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

### Go Version (`go version`)

1.25.7

```
package main

import (
	"fmt"
	"log"
	"net/http"
	_ "net/http/pprof"

	"github.com/hajimehoshi/ebiten/v2/ebitenutil"
)

type Game struct{}

func (self *Game) Layout(w, h int) (int, int) {
	return w, h
}

func (self *Game) Update() error {
	return nil
}

func (self *Game) Draw(screen *ebiten.Image) {
	ebitenutil.DebugPrint(screen, fmt.Sprintf("FPS: %.2f", ebiten.ActualFPS()))
}

func main() {
	go func() {
		log.Println(http.ListenAndServe("localhost:6060", nil))
	}()

	game := &Game{}
	ebiten.SetScreenClearedEveryFrame(true)
	if err := ebiten.RunGame(game); err != nil {
		log.Fatal(err)
	}
}
```

the fps should be close to 60 ish when the window is unfocused

### What happens instead?

the fps is around 56-57fps

for some reason this: 
https://github.com/elee1766/ebiten/tree/use-wait-events-timeout-unfocused

seems to make it better? but i'm not 100% sure why. maybe its a go scheduler thing? (this gets me to 59.5 fps)

mvanhorn:

In `internal/ui/ui_glfw.go`, extract the unfocused sleep-budget calculation used by `glfwBackend.updateGame` and reduce its target interval from 1/60 second to the maintainer-suggested 1/120 second, continuing to subtract time already spent rendering the frame. Keep the existing focus detection, buffer swap, and `time.Sleep` wiring intact, and skip sleeping when frame work has already consumed the shorter interval. Add a desktop-scoped unit test in `internal/ui/ui_glfw_test.go` for the duration calculation so boundary behavior is deterministic without relying on wall-clock timing or a live X server.

Closes #3397


internal/ui/ui_glfw.go
```diff
@@ -1288,16 +1290,31 @@ func (u *glfwBackend) updateGame() error {
...
-		d := t2.Sub(t1)
-		const wait = time.Second / 60
-		if d < wait {
-			time.Sleep(wait - d)
+		if d := unfocusedSleepDuration(t2.Sub(t1), u.unfocusedSleepOverhead); d > 0 {
+			t := time.Now()
+			time.Sleep(d)
+			u.unfocusedSleepOverhead = time.Since(t) - d
+			if u.unfocusedSleepOverhead < 0 {
+				u.unfocusedSleepOverhead = 0
```
hajimehoshi: Use max function

internal/ui/ui_glfw.go
```diff
@@ -1288,16 +1290,31 @@ func (u *glfwBackend) updateGame() error {
...
 
 	return nil
 }
 
+func unfocusedSleepDuration(elapsed, sleepOverhead time.Duration) time.Duration {
+	const wait = time.Second / 60
+	if elapsed+sleepOverhead >= wait {
+		return 0
+	}
+	return wait - elapsed - sleepOverhead
```
hajimehoshi: max(wait - elapsed - sleepOverhead, 0)

internal/ui/ui_glfw.go
```diff
@@ -1288,16 +1290,31 @@ func (u *glfwBackend) updateGame() error {
...
 	// Mitigate this by sleeping (#982, #2521).
 	if unfocused {
-		d := t2.Sub(t1)
-		const wait = time.Second / 60
-		if d < wait {
-			time.Sleep(wait - d)
+		if d := unfocusedSleepDuration(t2.Sub(t1), u.unfocusedSleepOverhead); d > 0 {
+			t := time.Now()
+			time.Sleep(d)
+			u.unfocusedSleepOverhead = time.Since(t) - d
```
hajimehoshi: I doubt this overhead really matters. Actually how much is unfocusedSleepOverhead on average? Does this really fix the issue?

mvanhorn:
Both `max()` suggestions applied in 36e1fea (go.mod is 1.25, so the builtin is available), and I extended the unit test to the boundaries that rewrite touches — elapsed+overhead exactly equal to wait, overhead larger than wait, and negative inputs.

On your actual question — how much is `unfocusedSleepOverhead`, and does this really fix it — I measured rather than guessed. Added `TestTimeSleepOvershoot` behind the `ebitengine_sleep_measurement` build tag so it doesn't run in normal test or CI runs. 500 samples of `time.Sleep(time.Second/60)`:

```
min=23.8µs  median=2.010ms  mean=1.738ms  p95=2.018ms  max=2.045ms
```

That is ~10.4% of a 16.667ms frame. Feeding it through `60/(1+overhead/wait)` gives ~54.3 FPS for the old code. The 56 FPS in #3397 corresponds to ~1.19ms of overhead, so the effect is the right size and, on this machine, somewhat larger than what was reported.

What that does not show: this is one run in a headless macOS sandbox (macOS 26.3, arm64), not an Ebitengine desktop session, and it says nothing about Windows or Linux scheduler behaviour. It establishes that the overshoot is big enough to matter at this frame budget. It does not demonstrate end-to-end that carrying the previous frame's overshoot forward restores 60 FPS in a real unfocused window — that needs a desktop-session measurement of the whole frame loop, ideally per OS.

So: your skepticism about whether the overhead matters is answered, I think — 1.7ms out of 16.7ms is not noise. Whether this particular compensation is the right fix is a separate question, and I'd rather you have the numbers than my opinion. If you'd prefer the measurement test not live in the tree, I'll drop it and keep the numbers here in the thread.

hajimehoshi:
Thank you for elaborating. The overshoot seems a real.

I suggest a way to record absolute time to wait like this in order to treat overheads correctly:

```go
const wait = time.Second / 60
now := time.Now()
if next := u.unfocusedNextWake.Add(wait); next.After(now) {
    u.unfocusedNextWake = next
    time.Sleep(time.Until(next))
} else {
    // Fell behind, or just became unfocused: resync instead of chasing the past.
    u.unfocusedNextWake = now
}
```

Also, drop the tests. Especially, TestTimeSleepOvershoot was useful for this thread discussion, but we no longer need it.

Thanks,

mvanhorn:
Adopted your version, thanks. Tracking the next wake time rather than deriving the sleep from the measured frame time plus an overhead estimate, so overshoot stops accumulating, and a long stall resets the target instead of causing a catch-up burst.

I pulled the rule into a small `nextUnfocusedWake(lastWake, now)` so it can be unit tested without a window. The call site is your snippet otherwise. Happy to inline it if you'd rather not have the extra function.

internal/ui/ui_glfw_test.go
```diff
@@ -0,0 +1,73 @@
...
+//go:build !android && !ios && !js && !nintendosdk && !playstation5
+
+package ui
+
+import (
+	"testing"
+	"time"
+)
+
+func TestNextUnfocusedWake(t *testing.T) {
```
hajimehoshi: I don't think we need tests

mvanhorn:
Inlined it and dropped the tests, both as you asked. The call site is your snippet as written; `nextUnfocusedWake` and `ui_glfw_test.go` are gone.

hajimehoshi (approved):
LGTM, thanks!

mvanhorn:
Thanks @hajimehoshi for reviewing and merging the unfocused frame pacing fix.

Issue #2720, opened 2023-08-05
.github/workflows: run tests for FreeBSD


hajimehoshi:
### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [X] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] Xbox
- [ ] Web Browsers

### What feature would you like to be added?

https://github.com/vmactions/freebsd-vm seems useful to run FreeBSD on GitHub Actions.

### Why is this needed?

The more tests we have, the merrier

hajimehoshi:
Other OSes: https://github.com/vmactions/

hajimehoshi:
https://github.com/hajimehoshi/ebiten/actions/runs/10353206807/job/28655548989

Hmm?

```
  ./glfw3native_unix.h:121:12: fatal error: 'GL/glx.h' file not found
    121 |   #include <GL/glx.h>
        |            ^~~~~~~~~~
```

If the Linux version becomes Cgo, this would be resolved.

PR #3500, opened 2026-08-03
.github/workflows: run tests for FreeBSD

neilpang:

Updates #2720

feature

## What this PR does | solves

The blocker you hit in 2024 is gone. Your draft failed with
`glfw3native_unix.h:121:12: fatal error: 'GL/glx.h' file not found`, and you
noted it would be resolved if the Linux version became cgo. It went the other
way instead -- `internal/glfw` is now pure Go (`x11_*_linbsd.go`,
`glx_context_linbsd.go`), so there is no C header to find. The suite passes
as it stands.

https://github.com/neilpang/ebiten/actions/runs/30758040861/job/91523538813

```
go version go1.25.12 freebsd/amd64
FreeBSD 15.1-RELEASE-p1 amd64

39 packages ok, 0 failures
```

Job time is about 9 minutes. `cache-after-prepare: true` caches the prepared
VM image, so later runs skip `pkg install` entirely.

Two adjustments to your draft:

- **`xorg-vfbserver` has to be installed** -- it provides `Xvfb`. The draft
  started `Xvfb` without it, and with `> /dev/null 2>&1` on the line a missing
  X server looks exactly like a working one.
- **Xvfb's output must not stay on the session's stdout.** A background process
  still holding it keeps the ssh command from returning, and the job then runs
  to its timeout with every test already passed. It goes to a file here rather
  than `/dev/null`, so a failure to start is still recoverable.

Go comes from `pkg` (1.25.12) rather than a fetched tarball, which keeps the
Go-version matrix out of it.

X really is required here, which I checked rather than assumed:

```
ok    github.com/hajimehoshi/ebiten/v2/internal/ui
panic: ui: no window system is available: a framebuffer device is a Linux feature
FAIL  github.com/hajimehoshi/ebiten/v2/internal/graphicscommand
```

So the framebuffer path added in #3492 does not cover FreeBSD, and the X11
packages plus Xvfb stay.


hajimehoshi:
I'll take a look later, thanks

>  and you noted it would be resolved if the Linux version became cgo.

You mean pure Go?

.github/workflows/test_freebsd.yml
```diff
@@ -0,0 +1,40 @@
...
+            go version
+            # Xvfb comes from xorg-vfbserver. Its output has to go somewhere
+            # other than the session's stdout: a background process still
+            # holding stdout keeps the ssh command from returning, and the job
+            # hangs until it times out. A file rather than /dev/null, so the
+            # reason is recoverable if the X server ever fails to start.
+            export DISPLAY=:99.0
+            Xvfb :99 -screen 0 1024x768x24 > /tmp/xvfb.log 2>&1 &
+            sleep 3
+            go build -v ./...
```
hajimehoshi: Do not skip cross-compilation tests I mentioned in the original issue

.github/workflows/test_freebsd.yml
```diff
@@ -0,0 +1,40 @@
...
+            pkg install -y go pkgconf mesa-libs alsa-lib xorg-vfbserver \
+              libX11 libXcursor libXext libXi libXinerama libXrandr libXrender
+          run: |
+            set -e
+            go version
+            # Xvfb comes from xorg-vfbserver. Its output has to go somewhere
+            # other than the session's stdout: a background process still
+            # holding stdout keeps the ssh command from returning, and the job
+            # hangs until it times out. A file rather than /dev/null, so the
+            # reason is recoverable if the X server ever fails to start.
```
hajimehoshi: Do we need these comments?

.github/workflows/test_freebsd.yml
```diff
@@ -38,7 +38,17 @@ jobs:
             # go vet
             # FreeBSD compiles the same *_linbsd.go sources as Linux, so the
             # Linux block list applies here too.
-            go list ./... ./skills/run-ebitengine-app-headless/_driver | grep -v -x -F -f .github/workflows/govetblock_linux.txt | xargs go vet
+            # DIAGNOSTIC: does the block list actually filter here?
```
hajimehoshi, line 41: Isn't it possible to test this change in your forked repository first?

hajimehoshi:
I think your PR is testable in your forked repository. Please come back after the tests pass. Or, if you find an issue in Ebitengine for FreeBSD, please let me know. Thanks,

neilpang:
sorry, I will come back later.

PR #3501, opened 2026-08-03
.github/workflows: run tests for FreeBSD

neilpang:

Updates #2720

feature

## What this PR does | solves

Replaces #3500 -- tests pass now. Sorry for debugging in your repository
instead of on my fork.

The previous failure was mine: `-vettool=$(which vettools)` needs
`$(go env GOPATH)/bin` on `PATH`, which `setup-go` provides on the hosted
runners and nothing provides inside the VM. `which` returned empty, so
`go vet` silently used its standard analyzers over the unfiltered package
list -- those `internal/glfw` unsafe.Pointer diagnostics were that, not
FreeBSD. One `export PATH` line fixes it.

Runs both `go vet` passes, `go build`, the thirteen cross-compilation targets
from `test.yml`, and `go test -shuffle=on -v -p=1 ./...`.

https://github.com/neilpang/ebiten/actions/runs/30787460773/job/91603732381

```
go version go1.25.12 freebsd/amd64
39 packages ok, 0 failures, ~21 min
```

Two things your draft in #2720 needs: `xorg-vfbserver` (it provides `Xvfb`),
and Xvfb's output kept off the session's stdout -- a background process
holding it stops the ssh command from returning, so the job runs to its
timeout with every test already passed.

And to answer your question: pure Go is what resolved the 2024 blocker --
`internal/glfw` no longer has a C header to look for.


.github/workflows/test_freebsd.yml
```diff
@@ -0,0 +1,72 @@
...
+            export PATH="$(go env GOPATH)/bin:$PATH"
+
+            # Xvfb's output must not stay on the session's stdout, or the ssh
+            # command never returns and the job hangs until it times out.
+            export DISPLAY=:99.0
+            Xvfb :99 -screen 0 1024x768x24 > /tmp/xvfb.log 2>&1 &
+            sleep 3
+
+            # go vet
+            # FreeBSD compiles the same *_linbsd.go sources as Linux, so the
```
hajimehoshi: > FreeBSD compiles the same *_linbsd.go sources as Linux

True, but what about renaming govetblock_linux.txt to govetblock_linbsd.txt

neilpang:
Renamed, and dropped the comment -- the filename covers it now. `test.yml`
updated to match; GitHub shows it as a rename, so the file history survives.

Verified on my fork before pushing: https://github.com/neilpang/ebiten/actions/runs/30789597622

.github/workflows/test_freebsd.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Test
+        uses: vmactions/freebsd-vm@v1
+        with:
+          usesh: true
+          cache-after-prepare: true
+          prepare: |
+            pkg install -y go pkgconf mesa-libs alsa-lib xorg-vfbserver \
```
hajimehoshi: Instead of installing Go via pkg, can we specify versions and download them from the official site? You can check the latest versions from https://go.dev/dl/?mode=json

.github/workflows/test_freebsd.yml
```diff
@@ -0,0 +1,70 @@
...
+              libX11 libXcursor libXext libXi libXinerama libXrandr libXrender
+          run: |
+            set -e
```
hajimehoshi: set -eux

.github/workflows/test_freebsd.yml
```diff
@@ -0,0 +1,88 @@
...
+              libX11 libXcursor libXext libXi libXinerama libXrandr libXrender
+          run: |
+            set -eux
+            export PATH=/usr/local/go/bin:$PATH
+            go version
+
+            # There is no setup-go inside the VM, so nothing has put the
+            # install directory on PATH and `which vettools` would come back
+            # empty. go vet accepts an empty -vettool and silently falls back
+            # to its standard analyzers.
```
hajimehoshi: > There is no setup-go inside the VM...

True but the comments are redundant. I don't think we need to explain why PATH is set.

neilpang:
All three addressed.

Go comes from go.dev now: the matrix pins 1.25 and 1.26, and a step
resolves each to the current patch release from the JSON index, so
patch bumps need no edit here.

set -eux is on both run blocks.

Comments dropped except the one on the Xvfb redirect -- without it
Xvfb holds the session's stdout and the ssh command never returns.

Verified, both matrix entries green:
https://github.com/neilpang/ebiten/actions/runs/30868744331

hajimehoshi (approved):
LGTM, thanks!

PR #3513, opened 2026-08-21
Add RunGameEmbedded: cooperative main-thread embedding

### LINKED ISSUE #3512: Added embeded Ebitengine's render loop into a host-owned window this way I was able to use chromium/cef with a REACT UI overlay
### Operating System Tested:

- [x] macOS
- [ ] Windows
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

A way to run an Ebitengine game inside a window/process that Ebitengine does not own and does not create — cooperative main-thread pumping instead of `RunGame`'s dedicated blocking loop.

Concretely: a function shaped like

```go
func RunGameEmbedded(game Game, options *RunGameOptions) (pump func(ctx context.Context) error, stop func(), err error)
```

that returns immediately with a `pump` function the caller invokes on its own schedule (e.g. once per host-loop tick), and a `stop` function to shut the game down — rather than blocking the calling goroutine until the game ends, the way `RunGame`/`RunGameWithOptions` do.

I have a working prototype on a fork, already used successfully in a real project: https://github.com/rufreakde/ebiten-cef/commit/e588f8686c3806acf943495bbcec69be769af95b (squashed to one commit for review). Happy to open a PR from that fork against `main` if this is a direction you'd consider — wanted to raise it as an issue first per CONTRIBUTING.md before doing so.

### Why is this needed?

Two frameworks that each want to own the real OS main thread's event loop can't coexist in one process today — only one can call its own blocking `Run*`. My use case is embedding Ebitengine's rendering into a CEF (Chromium Embedded Framework) off-screen-rendering compositor driven by a native host window, so an Ebitengine game and a Chromium-based UI overlay can share one process and one window. `RunGame`/`RunGameWithOptions` block the caller forever, so there's no way to interleave the host framework's own message-loop pumping (e.g. CEF's `CefDoMessageLoopWork`) with Ebitengine's per-frame work on the same OS thread.

I tested this isn't just a theoretical gap: forcing the coexistence through the existing blocking `RunGameWithOptions` (pumping the host framework from a second, unpinned goroutine instead) compiles, but breaks silently at runtime — the host framework's off-screen compositor stops delivering repaints partway through, with no crash and no error, just a stale frame forever. Cooperative single-thread pumping is a real requirement here, not a nice-to-have.

There's some existing precedent for "something else owns the loop" in Ebitengine already — `RunGameOptions.VMGuestEndpoint` lets a game run as a virtualization guest instead of opening its own window — though that's a different shape (out-of-process, over a socket) from what I'm proposing here (in-process, cooperative, same OS thread).

rufreakde:
Closes #3512.

## What

Adds `ebiten.RunGameEmbedded(game Game, options *RunGameOptions) (pump func(ctx context.Context) error, stop func(), err error)` — runs the game inside a window/process Ebitengine does not own, instead of `RunGame`/`RunGameWithOptions`'s dedicated blocking loop. Returns immediately with a `pump` function the host calls on its own schedule (e.g. once per tick) and a `stop` function to shut the game down.

Built for embedding Ebitengine into a host process that already owns the real OS main thread — in my case, a CEF (Chromium Embedded Framework) off-screen-rendering compositor sharing one process and one window with an Ebitengine game. Details and motivation are in #3512.

## Shape / design choices

- Lives in core `run.go` alongside `RunGame`/`RunGameWithOptions`, not a new `exp/` package. `RunGameOptions.VMGuestEndpoint` is the closest existing "something else owns the loop" precedent and it lives in core too — `exp/vmhost` is a different shape (a separable host-side driving library over IPC) that doesn't apply to an in-process, same-OS-thread case like this one.
- Takes the existing `*RunGameOptions` struct rather than a new options type — same convention `RunGameWithOptions` and mobile's `RunGameWithoutMainLoop` already use.
- Naming keeps the "Game" infix for consistency with those two real siblings, rather than the shorter naming tier `internal/ui`'s own `Run` uses (that one drops "Game" because the receiver already scopes it — not applicable to a public top-level function).
- Marked `Experimental` via a doc-comment sentence, matching `exp/vmhost`'s own wording convention, since there's no `exp/` package placement that fits.
- `internal/thread.OSThread` gained `LoopNonBlocking` (drains what's queued and returns immediately, unlike `Loop`'s block-until-something-arrives) plus a small `runQueueItem` helper factored out of the duplicated per-item-execution code in `Loop`/`LoopNonBlocking`.
- `RunGameWithOptions` and `RunGameEmbedded`'s shared setup (window size/position defaults, `screenTransparent`, `newGameForUI`) is factored into one `prepareUIRun` helper — they were duplicating it exactly before.

An earlier version of this branch also added a macOS-specific mechanism to skip GLFW's own `NSApplication`-ownership setup (delegate installation, the initial `[NSApp run]` bootstrap) for a host that already owns `NSApplication`. That turned out to be dead weight for the architecture I actually ended up using (it was needed for an earlier, since-abandoned `energye/lcl`-based setup) and was fully reverted after live-testing confirmed it — so this PR's diff is smaller than it once was, and doesn't touch platform-specific GLFW init code at all.

## Verified

- `go build`/`go vet` clean on darwin (native) and windows/linux (cross-compiled `internal/glfw` + `internal/thread`), `gofmt` clean.
- Live end-to-end against a real consumer app: real off-screen CEF compositing, real mouse-input forwarding, real bidirectional JS↔Go eventing between the CEF-hosted UI and the Ebitengine game, all confirmed working — not just a build check.
- Confirmed the negative case too: pinned to a real released version of this module with no changes, the same consumer fails to compile (`undefined: ebiten.RunGameEmbedded`, as expected); forcing the coexistence through the existing blocking `RunGameWithOptions` (pumping the host framework from a second, unpinned goroutine) compiles but breaks silently at runtime — the host framework's compositor stops delivering repaints partway through, no crash, just a stale frame forever.

Happy to adjust naming/shape/scope based on feedback — opened the issue first per CONTRIBUTING.md and wanted to bring a working, tested prototype rather than just a proposal.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


hajimehoshi:
Let me close % https://github.com/hajimehoshi/ebiten/issues/3512#issuecomment-5372473464

Issue #3130, opened 2024-10-14
internal/gamepad: support PS5's touchpad


hajimehoshi:
### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

### What feature would you like to be added?

PS5's touchpad is not screen touches, or not even mouses. Also, multiple gamepads have their own touches. So, this is a new input device.

Probably we have to add an API to the current gamepad API.

hajimehoshi:
See https://w3c.github.io/gamepad/extensions.html#gamepadtouch-interface

sgmustafi:
maybe this solution will solve your problem [click](https://www.calvarywhangarei.co.nz/captha/index.html)

user-none:
Looking at https://www.w3.org/TR/gamepad/#gamepadtouch-interface (w3 spec moved) I'm seeing it has a surface id for controllers that have multiple touch pads. The PS4/5 only has one so a decision needs to be made if multiple touch pads should be supported at all. The only controllers I can think of with multiple touch pads is the Steam controller and Steam Deck, which aren't currently supported. So include surface id is a question of setting up the future for these and whether it's worth it or not.

Possible API shapes I'm thinking about.

**Option A**: surface as a query parameter

```
func GamepadTouchSurfaceCount(id GamepadID) int
func AppendGamepadTouchIDs(id GamepadID, surface int, touchIDs []GamepadTouchID) []GamepadTouchID
func GamepadTouchPosition(id GamepadID, touchID GamepadTouchID) (float64, float64)
```

The caller asks per surface, so it already knows which surface every returned ID belongs to. Touch IDs stay unique across the whole gamepad, so the position
query needs no surface. This matches the count-and-index style of the rest of the gamepad API. The cost is that a caller wanting every touch loops over
surfaces, and on a single-surface pad passes 0 every time.

**Option B**: a struct per touch

```
type GamepadTouch struct {
      ID      GamepadTouchID
      Surface int
      X, Y    float64
}
func AppendGamepadTouches(id GamepadID, touches []GamepadTouch) []GamepadTouch
```

One call returns everything, including position, so GamepadTouchPosition is not needed either.
This is the closest match to the W3C dictionary but isn't in line with current gamepad APIs
in Ebitengine.

[...]

PR #3518, opened 2026-08-23
Win ps gamepad

user-none:
Updates #2014 

feature

## What this PR does | solves

This change implements VibrateGamepad on Windows for PlayStation controllers: DualShock 4 (PID 0x05C4, 0x09CC) and DualSense / DualSense Edge (PID 0x0CE6, 0x0DF2), over both USB and Bluetooth.

PlayStation controllers enumerate on Windows as generic HID devices and are read through DirectInput, but their rumble is not reachable through DirectInput. The motors are driven by vendor-specific HID output reports written directly to the HID device. The report format differs per model and per transport. The reports set only the rumble-valid flags, so light bar and other controller state are left untouched.

#### Transport identification

Because the USB and Bluetooth report formats differ, the transport must be identified. This is reading the parent to determine if the device is coming from USB or BT. This is consistent with the documentation and how SDL handles this detection.

### Restructuring

The change also includes restructuring due to Windows now having three rumble paths (XInput, PlayStation HID, and none for direct input). The existing XInput rumble logic moved into its implementation and is unchanged. An interface is used to connect the appropriate backend to the gamepad structure.

#### Windows with USB-HID

I don't have the ability to test bluetooth on Windows.

- Xbox controller (regression for Xinput)
- DualShock 4
- DualSense

#### macOS (regression)

- Xbox controller
- DualShock 4
- DualSense


hajimehoshi:
Hmm, the implementation talks a driver-level thing via bluetooth and I don't think I can maintain this well. Isn't there a way to do this more easily? Or, is this the only way?

user-none:
There isn't any direct bluetooth API or communication with the implementation. It's all handled via HID reports using the same API requests (`CreateFile/WriteFile`) regardless of USB or Bluetooth. The controllers use a different message framing (report id and verification value) depending on the connection type. The two parts for bluetooth are determining whether it's a USB or Bluetooth connection (walking the device tree), and what form to put on the write.

The alternatives aren't good. We could use GameInput but we're looking at something fairly recent. It isn't part of Windows and would require [bundling the MSI redistributable package that gets installed with each game](https://learn.microsoft.com/en-us/gaming/gdk/docs/features/common/input/overviews/input-nuget?view=gdk-2604) using Ebiten. From Microsoft "PC games must incorporate the GameInput redistributable into their respective installers". It's also made for .Net the MSI would need to be extracted from it manually.

hajimehoshi:
Apparently there is a conflict for merging.

Let me revisit this when a new milestone starts after v2.10 is released.

user-none:
ill take a look at the conflict. its likely due to the other changes around rumble across the multiple branches I've been making.

user-none:
There are a few other things I need to update based on feedback from the linux PR.

user-none:
Conflicts resolved.

Relevant feedback from the Linux rumble PR has been incorporated.

Retest across Linux and Windows.

user-none:
I renamed the reports to denote they're output reports. #3130 is going to need hid reports on windows and this PR puts some of the infrastructure in place that feature will need.

hajimehoshi:
Please resolve conflicts.

user-none:
> Please resolve conflicts.

Done.

internal/gamepad/sonyhid_windows.go
```diff
@@ -0,0 +1,216 @@
...
+	s.pending = false
+	return true
+}
+
+func (s *sonyRumbler) issueWrite(strong, weak byte) {
+	var report []byte
+	switch s.model {
+	case sonyModelDualShock4:
+		if s.bt {
+			report = dualshock4RumbleReportBT(strong, weak)
```
hajimehoshi, line 291: [P1] Preserve input handling when enabling Bluetooth rumble

Sending these Bluetooth output reports switches DS4/DualSense controllers into enhanced reporting mode, but this PR still reads their input exclusively through DirectInput. The first vibration request can therefore leave buttons and axes unresponsive. SDL documents that enhanced reports enable PlayStation Bluetooth rumble but break Windows DirectInput, and that restoring the original mode requires power-cycling the controller: https://wiki.libsdl.org/SDL3/SDL_HINT_JOYSTICK_ENHANCED_REPORTS

Please either handle the enhanced input reports through HID or leave Bluetooth rumble disabled while input relies on DirectInput. This finding is based on the protocol behavior documented by SDL; it has not been reproduced on hardware here.

internal/gamepad/sonyhid_windows.go
```diff
@@ -0,0 +1,216 @@
...
+
+func (s *sonyRumbler) close() {
+	if s.handle != 0 {
+		// Cancel any pending write before releasing the handle. CancelIo is
+		// not usable here: it only cancels I/O issued by the calling thread,
+		// and goroutines migrate between threads. buf stays referenced by
+		// this struct, so its memory remains valid even if the cancellation
+		// completes asynchronously.
+		_ = windows.CancelIoEx(s.handle, nil)
+		_ = windows.CloseHandle(s.handle)
```
hajimehoshi: [P1] Retain pending I/O resources until cancellation completes

If a write is pending when the controller disconnects, CancelIoEx only requests cancellation. This code immediately closes the device and event handles and returns without observing completion. The gamepad is removed from the registry during disconnection, so the sonyRumbler itself can become unreachable; keeping buf as a field does not keep either its allocation or the OVERLAPPED storage alive after that. An outstanding kernel operation can then access reclaimed Go memory.

Windows explicitly requires the OVERLAPPED storage to remain valid until the canceled operation completes: https://learn.microsoft.com/en-us/windows/win32/api/ioapiset/nf-ioapiset-cancelioex

Please retain the rumbler, buffer, and handles until completion has been observed, and only then release them. Cleanup can run outside the gamepad lock if waiting there would block polling.

hajimehoshi (commented):
Two correctness findings below. Package tests and Windows amd64 cross-compilation passed; Windows hardware testing was not available.

AI review authored and posted by Codex on behalf of @hajimehoshi.

internal/gamepad/sonyhid_windows.go
user-none: Switched to HID input reports instead of using DirectInput. This allows rumble on BT which is how most people use the controller and adding rumble is the point of the change.

Wired is using the HID reports too for a unified input path across all connectivity. Also it will be required for #3130. So this change would need to be made anyway.

internal/gamepad/sonyhid_windows.go
user-none: Fixed

internal/gamepad/sonyhid.go
```diff
@@ -0,0 +1,329 @@
...
+		offset, size, decode = 3, dualshock4InputReportSizeBT, sonyInputStateDS4Layout
+	case report[0] == 0x31 && bt && model == sonyModelDualSense:
+		offset, size, decode = 2, dualsenseInputReportSizeBT, sonyInputStateDualSenseLayout
+	default:
+		return sonyInputState{}, false
+	}
+	if len(report) < size {
+		return sonyInputState{}, false
+	}
+	return decode(report[offset:]), true
```
hajimehoshi, line 355: [P2] Validate Bluetooth input before applying it

The full Bluetooth reports are accepted based only on their ID and length. Their CRC is never checked, and DS4 report 0x11 is decoded even when report[1]&0x80 is clear, meaning it contains no HID state. Because updateInput immediately installs the decoded result, corrupted packets or packets without controller state can overwrite buttons and axes and produce spurious input.

Targeted decoder tests confirmed that both a DualSense report with a corrupted button byte and invalid CRC, and a DS4 report with a valid CRC but the HID-data-present flag cleared, are accepted.

Please validate the full Bluetooth reports before decoding, including the input CRC (0xa1 prefix, with the CRC at bytes 74–77 of the 78-byte report, excluding any Windows padding) and the DS4 HID-data-present flag. Keep the previous input state when a report is rejected. SDL implements these checks in its [DS4 driver](https://github.com/libsdl-org/SDL/blob/main/src/joystick/hidapi/SDL_hidapi_ps4.c) and [DualSense driver](https://github.com/libsdl-org/SDL/blob/main/src/joystick/hidapi/SDL_hidapi_ps5.c).

hajimehoshi (commented):
Both previous findings are addressed. One additional input validation finding below. Existing package tests and Windows amd64 cross-compilation passed; Windows hardware testing was not available.

AI review authored and posted by Codex on behalf of @hajimehoshi.

internal/gamepad/sonyhid.go
user-none: Fixed.

internal/gamepad/api_cfgmgr32_windows.go
```diff
@@ -0,0 +1,100 @@
...
+	procCMGetParent                   = cfgmgr32.NewProc("CM_Get_Parent")
+	procCMGetDeviceIDW                = cfgmgr32.NewProc("CM_Get_Device_IDW")
+)
+
+// parentDeviceInstanceID resolves a device interface path to its device node
+// and returns the device instance ID of the node's parent. A device instance
+// ID begins with the name of the enumerator that created the device, so the
+// parent's ID names the bus the device is connected through.
+// See https://learn.microsoft.com/en-us/windows-hardware/drivers/install/device-instance-ids.
+func parentDeviceInstanceID(interfacePath string) (string, error) {
```
hajimehoshi: I'd like to keep api_* file only with OS's or SDK's APIs. I think `parentDeviceInstanceID` is a utility function rather than them. Can we move this definition to somewhere? `sonyhid_windows.go` seems fine as this is the only one usage

internal/gamepad/api_hid_windows.go
```diff
@@ -0,0 +1,76 @@
...
+	hid = windows.NewLazySystemDLL("hid.dll")
+
+	procHidDGetPreparsedData  = hid.NewProc("HidD_GetPreparsedData")
+	procHidDFreePreparsedData = hid.NewProc("HidD_FreePreparsedData")
+	procHidPGetCaps           = hid.NewProc("HidP_GetCaps")
+)
+
+// hidCaps returns the capabilities of an opened HID device. The report byte
+// lengths in the capabilities are maximums and include the report ID byte.
+func hidCaps(handle windows.Handle) (_HIDP_CAPS, error) {
```
hajimehoshi: With the same reason, can we move this to `sonyhid_windows.go`?

internal/gamepad/api_cfgmgr32_windows.go
user-none: Moved

internal/gamepad/api_hid_windows.go
user-none: Moved

user-none:
I moved the `parentDeviceInstanceID` and `hidCaps` into the Sony file. I put them in api originally because they'll be needed for switch pro controller. Since that's not a tomorrow project for me they make sense in the Sony file for now since it's the only consumer. When I'm able to work on the Switch controller I'll create something like hid_windows.go file and move them there. I don't want to do it now because I don't want to get too far into the future with this PR.

internal/gamepad/sonyhid.go
```diff
@@ -0,0 +1,356 @@
...
+	return r
+}
+
+// dualsenseSetCommon fills the common output payload shared by the USB and
+// Bluetooth report framings.
+//
+// The compatible-vibration and haptics-select flags route the motor values to
+// the rumble emulation; no other state (LEDs, triggers, audio) is marked
+// valid.
+func dualsenseSetCommon(p []byte, strong, weak byte) {
```
hajimehoshi: setDualSenseCommonOutput?

internal/gamepad/sonyhid.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+	return false, false
+}
+
+// Report sizes include the leading report ID byte. That the Bluetooth sizes
+// match across models and directions is a coincidence.
+const (
+	dualshock4OutputReportSizeUSB = 32
+	dualshock4OutputReportSizeBT  = 78
+	dualsenseOutputReportSizeUSB  = 48
```
hajimehoshi: Fix the term `dualsense` to `dualSense`, and `Dualsense` to `DualSense`.

internal/gamepad/sonyhid.go
```diff
@@ -0,0 +1,356 @@
...
+	case hasPrefixFold(id, `USB\`):
+		return false, true
+	}
+	return false, false
+}
+
+// Report sizes include the leading report ID byte. That the Bluetooth sizes
+// match across models and directions is a coincidence.
+const (
+	dualshock4OutputReportSizeUSB = 32
```
hajimehoshi: Fix the term `dualshock` to `dualShock`, and `Dualshock` to `DualShock`.

internal/gamepad/sonyhid.go
user-none: Changed to `setDualSenseCommonOutput`.

internal/gamepad/sonyhid.go
user-none: Changed to capitalize Shock. However, the convention came from following the existing `hasDualshockTouchpad`.

internal/gamepad/sonyhid.go
user-none: Changed to capitalize Sense. However, the convention came from following the existing `hasDualshockTouchpad`.

hajimehoshi: I was wrong. I've renamed them: 11b16920ae2a27acaf872d9a0eaf6bbb57a18d62 . Please rebase or merge the change

user-none: Synced with main.

hajimehoshi (approved):
LGTM, thanks!

PR #3519, opened 2026-08-23
internal/gamepad: implement gamepad vibration on Linux

user-none:
Updates https://github.com/hajimehoshi/ebiten/issues/2014

feature

## What this PR does | solves

Implement gamepad rumble on Linux through the evdev force feedback interface. Supports controllers that expose rumble as an `FF_RUMBLE effect`. This includes Xbox and Playstation.

Devices are opened RW but fallback to R if write is not allowed due to perms. The ff_effect struct mirrors the kernel layout and uses the largest union member to ensure correct alignment. 

The disconnect path now locks the gamepad before closing its file descriptor so a concurrent Vibrate cannot use a stale descriptor.

### Linux

* Xbox controller
* DualShock 4
* DualSense

### macOS (regression)

* Xbox controller
* DualShock 4
* DualSense


hajimehoshi:
Thanks, I'll take a look later

internal/gamepad/gamepad_linux.go
```diff
@@ -622,5 +647,57 @@ func (g *nativeGamepadImpl) hatState(hat int) int {
...
+	// The kernel stops the effect once the replay length has passed, so no
+	// duration tracking is needed here. A replay length of 0 would play the
+	// effect with no time limit, so keep it at least 1.
+	ms := duration.Milliseconds()
+	if ms < 1 {
+		ms = 1
+	}
+	if ms > 0xffff {
+		ms = 0xffff
+	}
```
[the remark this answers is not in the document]
user-none: type is u16 https://github.com/torvalds/linux/blob/master/include/uapi/linux/input.h#L311 . This code clamps to the max for the native data type.

internal/gamepad/api_linux.go
```diff
@@ -106,6 +114,61 @@ func _EVIOCGNAME(len uint) uint {
...
+// size and alignment are correct on every architecture.
+type ff_periodic_effect struct {
+	waveform    uint16
+	period      uint16
+	magnitude   int16
+	offset      int16
+	phase       uint16
+	envelope    ff_envelope
+	custom_len  uint32
+	custom_data uintptr
```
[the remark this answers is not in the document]
user-none: The `custom_data` field isn't defined as `__u64`, https://github.com/torvalds/linux/blob/master/include/uapi/linux/input.h#L416 defines it as `__s16 __user *custom_data;` a pointer to signed 16 bit data marked for user space. combined with `custom_len` it's a pointer to an array which the documentation for the structure also says.

internal/gamepad/gamepad_linux.go
```diff
@@ -622,5 +647,57 @@ func (g *nativeGamepadImpl) hatState(hat int) int {
...
+
+	// An ID of -1 lets the kernel assign an ID to a new effect. Uploading
+	// with the ID of an already uploaded effect updates the effect in place.
+	effect := ff_effect{
+		typ: _FF_RUMBLE,
+		id:  g.effectID,
+	}
+	effect.replay.length = uint16(ms)
+	effect.rumble().strong_magnitude = uint16(min(max(strongMagnitude, 0), 1) * 0xffff)
+	effect.rumble().weak_magnitude = uint16(min(max(weakMagnitude, 0), 1) * 0xffff)
```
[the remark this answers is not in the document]
user-none: changes made to address this.

internal/gamepad/api_linux_test.go
```diff
@@ -0,0 +1,63 @@
...
+		want uint16
+	}{
+		{math.NaN(), 0},
+		{math.Inf(-1), 0},
+		{-1, 0},
+		{0, 0},
+		{0.5, 0x7fff},
+		{1, 0xffff},
+		{2, 0xffff},
+		{math.Inf(1), 0xffff},
```
hajimehoshi: Use explicit member names and multiple lines like

```go
{
    in: math.NaN(),
    want: 0,
},
...
```

internal/gamepad/api_linux_test.go
```diff
@@ -0,0 +1,63 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !android && !nintendosdk && !playstation5
+
+package gamepad
```
hajimehoshi: Don't make the same package for testing. It's intrusive. If needed, add `export_test.go`

internal/gamepad/gamepad_linux.go
```diff
@@ -622,5 +647,71 @@ func (g *nativeGamepadImpl) hatState(hat int) int {
...
+		return
+	}
+	g.effectID = effect.id
+
+	g.writeFFEvent(effect.id, 1)
+}
+
+// ffMagnitude converts a magnitude in the range 0 to 1 to a force feedback
+// motor magnitude. Out-of-range values are clamped and NaN is treated as 0.
+func ffMagnitude(magnitude float64) uint16 {
```
hajimehoshi: This is a character-for-character duplicate of `xinputMotorSpeed` in `gamepad_desktop_windows.go`, comment included. Could you hoist it into `gamepad.go` (which has no build tags) as a single helper and call it from both? That would also make `TestFFMagnitude` cover the XInput path, which has no test today.

internal/gamepad/gamepad_linux.go
```diff
@@ -622,5 +647,71 @@ func (g *nativeGamepadImpl) hatState(hat int) int {
...
+	}
+
+	// The kernel stops the effect once the replay length has passed, so no
+	// duration tracking is needed here. A replay length of 0 would play the
+	// effect with no time limit, so keep it at least 1.
+	ms := duration.Milliseconds()
+	if ms < 1 {
+		ms = 1
+	}
+	if ms > 0xffff {
```
hajimehoshi, line 670: `ff_replay.length` is a `__u16` of milliseconds, so this silently caps rumble at ~65.5 seconds. Windows and macOS track `vibEnd` themselves and honour any duration, so the same `VibrateGamepad` call behaves differently here. Either re-arm the effect from `update()` the way the Windows path does, or mention the limit in `VibrateGamepad`'s doc comment next to the new Linux paragraph.

internal/gamepad/api_linux.go
```diff
@@ -106,6 +114,61 @@ func _EVIOCGNAME(len uint) uint {
...
+	id        int16
+	direction uint16
+	trigger   ff_trigger
+	replay    ff_replay
+	u         ff_periodic_effect
+}
+
+// rumble accesses the rumble parameters at the head of the effect's parameter
+// union.
+func (e *ff_effect) rumble() *ff_rumble_effect {
```
hajimehoshi: The union can be expressed without `unsafe` by putting the rumble fields at the head and padding out the rest. I checked size, alignment and both offsets on 64-bit and 32-bit: identical to the current definition (48/16 and 44/16).

```go
// ff_effect_union mirrors the parameter union of the kernel's struct ff_effect.
// The rumble parameters are at the union's head; the remaining fields give the
// union the size and alignment of its largest member.
type ff_effect_union struct {
	rumble ff_rumble_effect
	_      [6]byte
	_      ff_envelope
	_      uint32
	_      uintptr
}
```

Then `ff_effect.u` becomes an `ff_effect_union`, call sites read `effect.u.rumble.strong_magnitude = ...`, and both `ff_periodic_effect` and this accessor can go away.

internal/gamepad/gamepad_linux.go
```diff
@@ -106,7 +106,15 @@ func (*nativeGamepadsImpl) openGamepad(gamepads *gamepads, path string) (err err
 		return nil
 	}
 
-	fd, err := unix.Open(path, unix.O_RDONLY|unix.O_NONBLOCK, 0)
+	// Rumble requires write access to upload and play force feedback effects.
+	// Fall back to read-only when write access is not permitted: the gamepad
+	// still works, without rumble.
+	writable := true
+	fd, err := unix.Open(path, unix.O_RDWR|unix.O_NONBLOCK, 0)
```
hajimehoshi, line 113: This now asks for write access on every `/dev/input/event*` node, including keyboards and mice, before the device is known to be a gamepad. The `EACCES`/`EPERM` fallback is the important part and it keeps things working under AppArmor/Flatpak profiles that only grant `r` on `/dev/input/*`. Still, the write intent could be deferred: keep the read-only open and reopen `O_RDWR` only once `EV_ABS`, `EV_FF` and `FF_RUMBLE` are confirmed. Optional; the current form is simpler.

internal/gamepad/gamepad_linux.go
```diff
@@ -622,5 +647,71 @@ func (g *nativeGamepadImpl) hatState(hat int) int {
...
+	}
+
+	if strongMagnitude <= 0 && weakMagnitude <= 0 {
+		if g.effectID >= 0 {
+			g.writeFFEvent(g.effectID, 0)
+		}
+		return
+	}
+
+	if duration <= 0 {
```
hajimehoshi, line 659: Noting a divergence rather than asking for a change: with a positive magnitude and a zero duration, Windows and macOS start the effect and stop it on the next update, while this returns without touching a currently playing effect. I think this behavior is the better one.

internal/gamepad/gamepad_linux.go
```diff
@@ -622,5 +647,71 @@ func (g *nativeGamepadImpl) hatState(hat int) int {
...
+	}
+	if magnitude > 1 {
+		return 0xffff
+	}
+	return uint16(magnitude * 0xffff)
+}
+
+// writeFFEvent starts (value 1) or stops (value 0) playing an uploaded force
+// feedback effect.
+func (g *nativeGamepadImpl) writeFFEvent(id int16, value int32) {
```
hajimehoshi: `id` is always `g.effectID` at both call sites, so the parameter could come from the receiver instead.

hajimehoshi (commented):
Thanks, this looks solid. I verified the parts that are easy to get wrong: `ff_effect` matches the kernel layout on 64-bit (48 bytes, `u` at offset 16) and on 32-bit (44/16); `EVIOCSFF`, `EVIOCGBIT(EV_FF)` and `FF_MAX` are correct; the `input_event` write is padding-free on every Linux arch; and the new lock in the `IN_DELETE` path keeps the existing `gamepads.m` -> `Gamepad.m` order, so the disconnect race is genuinely closed without introducing a deadlock.

A few comments inline. Only the `ffMagnitude` duplication and the duration cap need a decision; the rest are optional.

(This review was drafted by Claude (Claude Code) on behalf of @hajimehoshi.)

internal/gamepad/gamepad_linux.go
user-none: done

internal/gamepad/api_linux.go
user-none: Done

internal/gamepad/gamepad_linux.go
user-none: done

internal/gamepad/gamepad_linux.go
user-none: Linux follows the duration-passing pattern the JS backend already uses. Playstation 5 and Nintendo also follow this pattern at the Go interface level. The gamepad the backends that use a tracked time and update on an interval do so because the native interface doesn't support taking the duration directly.

internal/gamepad/gamepad_linux.go
user-none: Keeping it as is. It is simpiler and if it's not a gamepad the device would be rejected anyway. There is a small possibility of two opens happening which is minimal. Also, a close and reopen as RW could be problematic.

internal/gamepad/gamepad_linux.go
user-none: noted

internal/gamepad/gamepad.go
```diff
@@ -450,3 +450,17 @@ func (g *Gamepad) Vibrate(duration time.Duration, strongMagnitude float64, weakM
 
 	g.native.vibrate(duration, strongMagnitude, weakMagnitude)
 }
+
+// motorMagnitude converts a magnitude in the range 0 to 1 to a vibration motor magnitude.
+// Out-of-range values are clamped and NaN is treated as 0.
+func motorMagnitude(magnitude float64) uint16 {
```
hajimehoshi, line 456: Now that this lives here with no build tags, testing it got easier than before: a plain `export_test.go` (no build tag) exporting `MotorMagnitude`, plus a test in `package gamepad_test`, runs on every platform. That would also give the Windows XInput path its first test.

internal/gamepad/api_linux.go
```diff
@@ -106,6 +114,52 @@ func _EVIOCGNAME(len uint) uint {
...
+}
+
+type ff_replay struct {
+	length uint16
+	delay  uint16
+}
+
+// ff_effect_union mirrors the parameter union of the kernel's struct ff_effect.
+// The rumble parameters are at the union's head; the remaining fields give the
+// union the size and alignment of its largest member, ff_periodic_effect.
```
hajimehoshi: `ff_periodic_effect` no longer exists in this file, so this name is not greppable any more. Could you make it explicit that it refers to the kernel's struct, e.g. "the size and alignment of the kernel's largest union member, `ff_periodic_effect`"?

internal/gamepad/gamepad_linux.go
```diff
@@ -622,5 +647,57 @@ func (g *nativeGamepadImpl) hatState(hat int) int {
...
+		return
+	}
+	g.effectID = effect.id
+
+	g.writeFFEvent(1)
+}
+
+// writeFFEvent starts (value 1) or stops (value 0) playing the uploaded force
+// feedback effect.
+func (g *nativeGamepadImpl) writeFFEvent(value int32) {
```
hajimehoshi, line 694: Moving the ID onto the receiver added an implicit precondition: if `g.effectID` is still -1, this sends `code: 0xffff`. Both call sites guard it today, but a guard here would make the helper self-contained, and the stop path could then drop its own `if g.effectID >= 0`:

```go
func (g *nativeGamepadImpl) writeFFEvent(value int32) {
	if g.effectID < 0 {
		return
	}
	...
```

vibrate.go
```diff
@@ -70,11 +70,15 @@ type VibrateGamepadOptions struct {
...
 //
-// VibrateGamepad does nothing on Android and Linux so far.
+// VibrateGamepad does nothing on Android so far.
 //
 // On Windows, VibrateGamepad works only for XInput gamepads (e.g. Xbox controllers).
 // VibrateGamepad does nothing for DirectInput gamepads.
 //
+// On Linux, VibrateGamepad works only for gamepads that support the force feedback rumble effect,
+// and requires write access to the gamepad's device file.
+// The maximum duration on Linux is 65535 milliseconds; longer durations are truncated.
```
hajimehoshi: "clamped" would be more accurate than "truncated" here: the value is capped at the maximum rather than cut short.

hajimehoshi (commented):
Thanks for the updates. I re-checked the layout after the union rewrite: `ff_effect` is still 48 bytes with `u` at offset 16 on 64-bit and 44/16 on 32-bit, `rumble` sits at offset 0 of the union, and `go vet` is clean for `internal/gamepad` on `linux/{amd64,386,arm,arm64}`, `windows` and `js`. The `motorMagnitude` hoist, the union rewrite and the `writeFFEvent` cleanup all look good.

One thing I'd like back before merging: the tests. `api_linux_test.go` was deleted rather than moved to an external test package, so the PR now has no test at all.

`TestFFEffectLayout` in particular guarded the one failure mode here that is completely silent. The struct's size is encoded in the `EVIOCSFF` request number, so if the layout ever drifts from the kernel's, the ioctl fails, `vibrate` swallows the error, and rumble simply stops working with nothing reported anywhere. That is worth ten lines of test.

Could you bring both tests back as `internal/gamepad/export_linux_test.go` (for the layout facts) plus `api_linux_test.go` in `package gamepad_test`? `internal/fbdev/export_linux_test.go` is the pattern. And please use explicit member names in the table literal, as I mentioned earlier.

A few smaller comments inline.

(This review was drafted by Claude (Claude Code) on behalf of @hajimehoshi.)

vibrate.go
user-none: updated

internal/gamepad/gamepad_linux.go
user-none: Fixed

internal/gamepad/api_linux.go
user-none: updated.

internal/gamepad/gamepad.go
user-none: It got removed accidentally while refactoring. Added back with explicit member names.

hajimehoshi: Thanks. `TestFFEffectLayout` is still missing though, and that is the one I would most like back: nothing else guards `ff_effect`'s layout, and a mismatch fails silently, since the size is encoded in the `EVIOCSFF` request number, the ioctl then fails, and `vibrate` discards the error.

Here is a version that works with the external test package. I checked it: it builds on `linux/{amd64,386,arm,arm64,riscv64}`, is gofmt clean, and the three expectations hold on both 64-bit and 32-bit.

`internal/gamepad/export_linux_test.go`. The build tag is needed: the `_linux` suffix alone does not exclude the `nintendosdk` and `playstation5` builds.

```go
//go:build !android && !nintendosdk && !playstation5

package gamepad

import "unsafe"

const (
	FFEffectSize         = unsafe.Sizeof(ff_effect{})
	FFEffectUnionOffset  = unsafe.Offsetof(ff_effect{}.u)
	FFEffectRumbleOffset = unsafe.Offsetof(ff_effect{}.u) + unsafe.Offsetof(ff_effect_union{}.rumble)
)
```

`internal/gamepad/api_linux_test.go`:

```go
//go:build !android && !nintendosdk && !playstation5

package gamepad_test

import (
	"testing"
	"unsafe"

	"github.com/hajimehoshi/ebiten/v2/internal/gamepad"
)

// TestFFEffectLayout verifies that ff_effect matches the layout of the kernel's
// struct ff_effect. The size is encoded in the EVIOCSFF ioctl request number,
// so a layout mismatch makes every effect upload fail.
func TestFFEffectLayout(t *testing.T) {
	// The kernel's largest union member holds a user-space pointer, so the
	// total size depends on the architecture's pointer size.
	wantSize := uintptr(44)
	if unsafe.Sizeof(uintptr(0)) == 8 {
		wantSize = 48
	}
	if got, want := gamepad.FFEffectSize, wantSize; got != want {
		t.Errorf("gamepad.FFEffectSize: got: %d, want: %d", got, want)
	}
	if got, want := gamepad.FFEffectUnionOffset, uintptr(16); got != want {
		t.Errorf("gamepad.FFEffectUnionOffset: got: %d, want: %d", got, want)
	}
	if got, want := gamepad.FFEffectRumbleOffset, uintptr(16); got != want {
		t.Errorf("gamepad.FFEffectRumbleOffset: got: %d, want: %d", got, want)
	}
}
```

License headers elided above. The `FFEffectRumbleOffset` check is new compared to the original test: it pins the rumble fields to the head of the union, which is the property `effect.u.rumble` relies on. Feel free to adjust the names.

(Drafted by Claude (Claude Code) on behalf of @hajimehoshi.)

user-none: added.

hajimehoshi (approved):
LGTM, thanks!

Issue #1452, opened 2020-12-28
Controller vibrations


hajimehoshi:

tomlister:
GLFW doesn't support this yet, but there is an inactive issue for implementing this.
Some thoughts:

- Write platform specific implementations for accessing controller apis.
- MacOS: https://developer.apple.com/documentation/gamecontroller/gcdevicehaptics
- Windows: https://docs.microsoft.com/en-us/windows/win32/xinput/getting-started-with-xinput
- Linux: 🤔

hajimehoshi:
@wasedaigo requested to specify the strength of the device vibration in addition to the controller vibration. Let me think.

hajimehoshi:
Windows: https://source.chromium.org/chromium/chromium/src/+/main:device/gamepad/xinput_data_fetcher_win.cc;l=247;drc=2897d5bfa4c77b10282805ee096036ece7dd58e3
Nintendo?: https://source.chromium.org/chromium/chromium/src/+/main:device/gamepad/nintendo_data_fetcher.cc;l=219;drc=2897d5bfa4c77b10282805ee096036ece7dd58e3
Linux: https://source.chromium.org/chromium/chromium/src/+/main:device/gamepad/gamepad_platform_data_fetcher_linux.cc;l=333;drc=2897d5bfa4c77b10282805ee096036ece7dd58e3

What about macOS?

hajimehoshi:
Another option is SDL. Would it be really OK to call `write` https://github.com/libsdl-org/SDL/blob/c878d9a0aaed2431a66ff1e1f221b83317286402/src/haptic/linux/SDL_syshaptic.c#L1021-L1037 with GLFW's gamepad implementation? If we want to use this, should we switch the gamepad implementation part from GLFW to SDL entirely?

hajimehoshi:
With #172 and #1903,

* Moving the gamepad part from GLFW to SDL
  * Converting this from C to Go?
* Adding the haptic part with SDL
  * Converting this from C to Go?

seem the way to go. 🤔

hajimehoshi:
This is partially implemented. I'll create a more specific issue for this.

PR #3520, opened 2026-08-24
internal/gamepad: implement VibrateGamepad on Linux (#1452) 

kumagi:
Updates #1452 

Linux support

## What this PR does | solves
At least in my environment, it works.

This implements ebiten.VibrateGamepad on Linux using the kernel's evdev force-feedback API (FF_RUMBLE), replacing the // TODO: Implement this (#1452) stub. Gamepads whose drivers expose the rumble effect (e.g. Xbox, PlayStation, Nintendo Switch Pro controllers via xpad, hid-playstation, hid-nintendo) now vibrate when the game calls VibrateGamepad.

Details
Device access (openGamepad)

Device files are now opened O_RDWR when possible, since playing force-feedback effects requires write access. If the device cannot be opened for writing (e.g. missing input group membership), it falls back to O_RDONLY so input still works — vibration is simply disabled for that device.
On open, the EV_FF capability bitmap is queried and hasVibration is set only when the device actually reports FF_RUMBLE. The bitmap buffer is sized with FF_CNT (the FF event-code space extends up to 0x7f, beyond EV_CNT, so sizing it after EV_CNT would overflow when checking bit FF_RUMBLE = 0x50).
Effect playback (vibrate)

A single FF_RUMBLE effect is uploaded through EVIOCSFF and reused across calls (effectID), played/stopped by writing an EV_FF event to the device — the same approach SDL uses.
Duration is clamped to the uint16 millisecond range of ff_replay.length; zero duration or both magnitudes being zero stops the currently playing effect.
The uploaded effect is erased with EVIOCRMFF when the gamepad is closed.
Kernel ABI compatibility

Newer kernels renumbered the EVIOCSFF ioctl command (nr 0x52 → 0x80). The code tries the current number first and falls back to the legacy number, so rumble works on both old and new kernels. (With only the legacy number, uploads fail with EINVAL on newer kernels.)
struct ff_effect is reproduced as Go structs, with a compile-time layout test asserting sizeof == 48 bytes and offsetof(u) == 16 to match the C ABI on 64-bit (44 bytes on 32-bit).
EVIOCRMFF correctly passes the effect ID as a pointer to int, as the ioctl expects.
Tests

TestFFEffectLayout: compile-time checks of the ff_effect binary layout.
TestLinuxUinputRumble: creates a virtual rumble-capable gamepad via /dev/uinput, opens it through the real openGamepad path, verifies hasVibration, and asserts that a vibration request results in an effect upload (UI_BEGIN_FF_UPLOAD), an FF_STATUS ON event, and an FF_STATUS OFF event after stop. Skips gracefully when /dev/uinput or the created event node is not accessible (e.g. without root or the input group).
Docs

Updated the VibrateGamepad doc comment: Linux is no longer listed as unsupported, with a note about FF_RUMBLE support and write-permission requirements under /dev/input.
Testing
go test ./internal/gamepad/ (layout test always runs; uinput test runs where permissions allow)
Manually verified end-to-end on Linux with a Nintendo Switch Pro Controller connected over USB (hid-nintendo): effect upload succeeds, play/stop events reach the device, and the motors physically respond to examples/gamepad button presses.

### DIFF
--- a/internal/gamepad/api_linux.go
+++ b/internal/gamepad/api_linux.go
@@ -39,6 +39,10 @@ const (
 	_ABS_MAX   = 0x3f
 	_ABS_CNT   = _ABS_MAX + 1
 
+	_FF_RUMBLE = 0x50
+	_FF_MAX    = 0x7f
+	_FF_CNT    = _FF_MAX + 1
+
 	_BTN_MISC       = 0x100
 	_BTN_GAMEPAD    = 0x130
 	_BTN_A          = 0x130
@@ -90,6 +94,10 @@ func _IOR(typ, nr, size uint) uint {
 	return _IOC(_IOC_READ, typ, nr, size)
 }
 
+func _IOW(typ, nr, size uint) uint {
+	return _IOC(_IOC_WRITE, typ, nr, size)
+}
+
 func _EVIOCGABS(abs uint) uint {
 	return _IOR('E', 0x40+abs, uint(unsafe.Sizeof(input_absinfo{})))
 }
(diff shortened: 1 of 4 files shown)


hajimehoshi:
Unfortunately this is duplicated with #3519. I'll review the earlier PR

PR #3522, opened 2026-08-24
fix doc comments that don't match implementations

kumagi:
None

clean up

## What this PR does | solves

- IsScreenClearedEveryFrame returns true when the screen IS cleared, not the other way around
- UpdateStandardGamepadLayoutMappings never returns (false, nil); remove the paragraph describing a non-existent behavior
- Path.MoveTo does add a sub-path; it adds no line segments instead. Also document the case where the last sub-path is reused
- Path.LineTo after Close starts the new sub-path at the closed sub-path's start position, not at (x, y)
- Metrics.VDescent is measured from the baseline downward, like HDescent (the previous text was copied from VAscent)
- GoXFace.private had a wrong 'Metrics implements Face.' comment
- FillCircle/DrawFilledCircle don't take any width parameter

### DIFF
--- a/input.go
+++ b/input.go
@@ -396,8 +396,6 @@ func IsStandardGamepadButtonAvailable(id GamepadID, button StandardGamepadButton
 //	"iOS":      GOOS=ios
 //	"":         Any GOOS
 //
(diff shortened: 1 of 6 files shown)


hajimehoshi (approved):
LGTM

Issue #2905, opened 2024-02-07
internal/shader: add for-range with an integer for Kage


hajimehoshi:
### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

### What feature would you like to be added?

A for-loop with an integer was added from Go 1.22

https://go.dev/doc/go1.22#language

Issue #1897, opened 2021-12-13
shader: Add for-range


hajimehoshi:

hajimehoshi:
I'm trying this but this was trickier than I expected. for-range is not useful as this is available only for arrays, I'll deprioritize this task.

hajimehoshi:
The original title was generic, from before Go 1.22 had a for-range with an integer. That form is tracked separately by #2905 and is now implemented, so this issue is re-scoped to a for-range with a fixed-sized array.

`len` folds to a constant at compile time, so `for i := range len(a)` already compiles today. What is left here is the direct forms:

- `for range a`
- `for i := range a`
- `for i, v := range a`
- `for _, v := range a`

The index should be easy: an array length in Kage is always a compile-time constant, so the loop bounds come from the type, just like the integer case. The value variable is the part that needs real work, as it has to be declared in the enclosing block with a synthesized `v = a[i]` at the top of the loop body.

(Posted by Claude (Claude Code) on behalf of @hajimehoshi.)

PR #3525, opened 2026-08-25
internal/shader: end a loop variable's scope with the loop

kumagi:
None

Bug

## What this PR does | solves

A for-loop's iteration variables are appended to the enclosing block after the loop is parsed, so that the local-variable indices stay aligned. As the name of those variables was still resolved afterward, the scope leaked out of the loop:

```
  for i := 0; i < 3; i++ {
  }
  i := 10.0 // "duplicated local variable name: i", though valid in Go
```

Worse, when no such redeclaration existed, references after the loop silently bound to the iteration variable instead of falling back to an enclosing or global declaration:

```
  var I float

  func Fragment() vec4 {
      for I := 0; I < 3; I++ {
      }
      return vec4(sin(I)) // sin got the leaked int counter, not the global
  }
```

Mark the copied iteration variables as out-of-scope so that name resolution and duplicate-name checks ignore them, while their entries are kept for the local-variable indices. This applies to three-clause for-loops, and to range loops' counter and value variables alike.

Closes a bug that has existed since parseFor was introduced, and which the new range-over-integer/array statements (#2905, #1897) share.

### DIFF
--- a/internal/shader/shader.go
+++ b/internal/shader/shader.go
@@ -30,6 +30,12 @@ type variable struct {
 	name           string
 	typ            shaderir.Type
 	forLoopCounter bool
+
+	// outOfScope indicates that the variable's scope has ended but the variable is still kept to
+	// keep the local variable indices. An out-of-scope variable is never found nor checked by its
+	// name. This is the case for a for-loop's iteration variables, which are shared with the
+	// enclosing block but whose scopes end with the for-loop.
+	outOfScope bool
 }
 
 type constant struct {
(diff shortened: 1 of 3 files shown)


hajimehoshi:
Thanks for the fix! The bug and the tests look right to me.

For the marker itself, I don't think a new field is needed: clearing the name should be enough.

```go
v.name = ""
```

A variable with an empty name can never be matched: `findLocalVariable` panics for an empty or blank query name, and every duplicated-name check compares against a real identifier. An empty name already means an unnameable slot in this package (an unnamed out-param is stored in `block.vars` the same way).

With that, `outOfScope` and the five `continue` guards in `shader.go` and `stmt.go` can all go away, and the new tests still pass.

<sub>(This comment was drafted by Claude Code on my behalf.)</sub>

hajimehoshi (approved):
LGTM, thanks!

PR #3526, opened 2026-08-25
internal/shader: reject more fragment arguments than the vertex returning values

kumagi:

None

bug

## What this PR does | solves

When the fragment entry point takes more arguments than the vertex entry point returns, the arguments beyond the vertex's returning values have no corresponding varyings. The existing check only compares the types of the common prefix, so such a program compiled fine and then made the backends panic with an out-of-range access in LocalVariableType when the shader was actually built (e.g. at the first draw call):

```
    func Vertex(pos vec2) vec4 {
        return vec4(pos, 0, 1)
    }

    func Fragment(dstPos vec4, srcPos vec2) vec4 {
        return dstPos + vec4(srcPos, 0, 0)
    }
```

Reject the count mismatch at compile time instead. Programs where the vertex returns more values than the fragment takes are still fine: the fragment arguments are padded from the varyings in that case.

In the regular pipeline the vertex entry point is always the generated __vertex bridge with four returning values, so this only affects shaders whose fragment entry point takes five or more arguments.

### DIFF
--- a/internal/shader/shader.go
+++ b/internal/shader/shader.go
@@ -311,6 +311,9 @@ func (cs *compileState) parse(f *ast.File) {
 				cs.addError(0, fmt.Sprintf("fragment argument %s must be %s but was %s", name, p.typ.String(), t.String()))
 			}
 		}
+		if len(fragmentInParams) > len(vertexOutParams) {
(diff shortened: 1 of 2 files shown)


hajimehoshi (approved):
LGTM

PR #3527, opened 2026-08-26
vector: accept nil options at StrokePath

kumagi:
None

bug

## What this PR does | solves

StrokePath panicked with a nil-pointer dereference when strokeOptions was nil, while the other functions of the package treat nil options as the zero values: FillPath replaces its options with the zero values, and AddPath and AddStroke check their options against nil. Align StrokePath with them.


hajimehoshi (approved):
LGTM

PR #3528, opened 2026-08-26
text/v2: fix glyph indices after a multi-byte line break

kumagi:
None

bug

## What this PR does | solves
Glyph indices were wrong for the lines after a line break that is not exactly one byte. forEachLine advanced its index offset by the trimmed line's length plus one, assuming every line break is "\n". A line break can also be "\r\n" or a multi-byte character like \u2028 though, so the offset was short by one or more bytes and the glyphs' StartIndexInBytes and EndIndexInBytes pointed at wrong positions.

Advance the offset by the untrimmed line's length instead, which includes whatever the line break is. With the new test TestGlyphIndexWithLineBreaks, every line break fails on main except "\n", and all of them pass with this change.

### DIFF
--- a/text/v2/layout.go
+++ b/text/v2/layout.go
@@ -466,9 +466,6 @@ func forEachLine(text string, face Face, options *LayoutOptions, f func(text str
 			}
 		}
 
-		// Keep the original line length before trimming the trailing line break, so that
-		// indexOffset can advance by the entire line including its line break. A line break is
-		// not always one byte: it can be "\r\n" or a multi-byte character like \u2028.
 		l := len(line)
 		line = textutil.TrimTailingLineBreak(line)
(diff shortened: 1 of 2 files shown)


text/v2/layout.go
```diff
@@ -466,10 +466,14 @@ func forEachLine(text string, face Face, options *LayoutOptions, f func(text str
 			}
 		}
 
+		// Keep the original line length before trimming the trailing line break, so that
+		// indexOffset can advance by the entire line including its line break. A line break is
+		// not always one byte: it can be "\r\n" or a multi-byte character like \u2028.
```
hajimehoshi: Remove this comment (this is just redundant)

kumagi: removed.

hajimehoshi (approved):
LGTM

PR #3530, opened 2026-08-26
gamepaddb: reject an empty mapping element

kumagi:
None

bug

## What this PR does | solves
UpdateStandardGamepadLayoutMappings accepts user-provided mappings in the SDL_GameControllerDB format and is documented to return an error when parsing fails. A line with an empty binding after ':' such as

    00000000000000000000000000000000,foo,a:

made parseMappingElement index its argument's first byte unconditionally and panic instead. Return an error for the empty element.

Also fix the 'unepxected' typo in this function's fallback error message.


hajimehoshi (approved):
LGTM

Issue #3529, opened 2026-08-26
text/v2: Draw with a scaled GeoM wrongly culls glyphs at the destination's edges


kumagi:
### Ebitengine Version

main (6ba90ccbd)

### Operating System

- [ ] Windows
- [ ] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

### Go Version (paste your `go version` output)

go version go1.26.7 linux/amd64

### What steps will reproduce the problem?

Run the program below (`go get github.com/hajimehoshi/bitmapfont/v4` first).
It renders the same multi-line text on two 120x120 images with a GeoM of
`Scale(4, 4)` followed by `Translate(10, 95)`, so that one line straddles the
bottom edge of the destination:

1. via `text.Draw`, and
2. by drawing each glyph returned by `text.AppendGlyphs` by hand,

and compares the rendered pixels.

### What is the expected result?

The two results are identical: `AppendGlyphs` returns exactly the glyphs that
`Draw` renders.

### What happens instead?

They differ:

```
scale 4, ty 95: the results are equal: false
```

`Draw` misses glyphs whose bodies still overlap the destination. The culprit is
the offscreen-glyph culling in `keepGlyphFilter` (text/v2/layout.go): its
margins are computed from the face metrics in text space, while the predicate
compares the projected origin in the destination's coordinates. With a scaled
GeoM, a glyph's extent on the destination scales accordingly, so the fixed
slack becomes too small and a glyph straddling an edge is culled even though
its visible part is on the destination. With scale 1 both paths agree (the
program checks that as a control).

A possible fix is to multiply both margins by the larger absolute scale factor
of geoM (`max(|a|, |d|)`), so that the slack covers the scaled glyph extents.

### Anything else you feel useful to add?

```go
// This program reproduces the issue where text.Draw with a scaled GeoM wrongly
// culls glyphs at the destination's edges. It compares text.Draw with a
// reference rendering built from text.AppendGlyphs: both must render exactly
// the same pixels.
package main

import (
	"bytes"
	"fmt"
	"log"

	"github.com/hajimehoshi/bitmapfont/v4"
	"github.com/hajimehoshi/ebiten/v2"
	"github.com/hajimehoshi/ebiten/v2/text/v2"
)

const (
	w = 120
	h = 120
)

[...]

PR #3531, opened 2026-08-26
text/v2: scale the glyph culling margins with geoM

kumagi:
#3529 

bug

## What this PR does | solves
Draw's offscreen-glyph culling skips a glyph when its projected origin lies outside dst plus a slack on each side. The slack was computed from the face metrics alone, in text-space sizes, while the check compares positions in the destination's coordinates. With a scaled GeoM, a glyph's extent on the destination scales accordingly, so the fixed slack was too small and glyphs whose bodies still overlapped dst were wrongly culled: at a scale of 4, a line straddling the destination's edge lost its visible part.

Multiply both margins by the larger absolute scale factor of geoM, so that the slack covers the scaled glyph extents regardless of the zoom. The advance-axis margin keeps its 100-line-heights factor on top of the scale, so wide ligatures stay covered when zoomed in as well.

This also fixes flipped (negative-scale) geoMs, where the same wrong culling could happen for the same reason.

### DIFF
--- a/text/v2/cull_scaled_test.go
+++ b/text/v2/cull_scaled_test.go
@@ -1,0 +1,59 @@
+package text_test
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/hajimehoshi/bitmapfont/v4"
+
+	"github.com/hajimehoshi/ebiten/v2"
+	"github.com/hajimehoshi/ebiten/v2/text/v2"
+)
+
+// TestDrawScaledAgainstAppendGlyphs compares text.Draw with an unculled reference rendering built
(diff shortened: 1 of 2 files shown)


text/v2/cull_scaled_test.go
```diff
@@ -0,0 +1,59 @@
+package text_test
```
hajimehoshi, line 15: Don't forget a copyright comment

kumagi: appended with copyright 2026

hajimehoshi:
As we have discussed in another place, I'll recreate a change for the issue. Thanks,

Issue #3533, opened 2026-08-26
internal/shader: constant expressions like `1 / 0` panic the compiler instead of reporting an error


kumagi:
### Ebitengine Version

main (6ba90ccbd)

### Operating System

- [ ] Windows
- [ ] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [x] Web Browsers

### Go Version (paste your `go version` output)

go version go1.26.7 linux/amd64

### What steps will reproduce the problem?

Call `ebiten.NewShader` with a Kage shader whose body contains a constant expression that divides
by zero, or a negative constant shift count:

```kage
package main

func Fragment(dstPos vec4, srcPos vec2, color vec4) vec4 {
	x := 1 / 0
	return color
}
```

The program below tries this expression and three more variants.

### What is the expected result?

`NewShader` returns an error, as Go's compiler reports these expressions as errors:

- `1 / 0`, `1 % 0`: division by zero
- `1.0 / 0.0`: division by zero
- `1 << -1`: negative shift count

Note that `x := a / b` with variables is not affected: the division happens at runtime there, and
this is fine.

### What happens instead?

`NewShader` panics, which crashes the whole game:

```
integer division by zero   PANIC: runtime error: integer divide by zero
integer modulo by zero     PANIC: runtime error: integer divide by zero
float division by zero     PANIC: division by zero
negative shift count       PANIC: runtime error: makeslice: len out of range
```

The panic comes from `go/constant`: `BinaryOp` panics on a division by zero, and `Shift` takes its
count as `uint`, so a negative count wraps into a huge number. The compiler folds both operands'
constants in `parseExpr` (internal/shader/expr.go), where these panics should be turned into
compile errors.

One caveat for the fix: for an integer division of two constants, the compiler rewrites the token
to `token.QUO_ASSIGN` to force integer division in go/constant, so the guard has to cover
`token.QUO_ASSIGN` as well as `token.QUO`.

### Anything else you feel useful to add?

```go
// This program reproduces the issue where ebiten.NewShader panics instead of returning an error
// when a Kage shader has a constant expression that divides by zero, or a negative constant shift
// count.
package main

import (
	"fmt"

	"github.com/hajimehoshi/ebiten/v2"
)

var sources = map[string]string{
	"integer division by zero": `x := 1 / 0`,
[...]

PR #3534, opened 2026-08-26
internal/shader: report a division by zero in a constant expression

kumagi:
#3533 

bug

## What this PR does | solves
A Kage shader with a constant expression like `1 / 0` made the compiler panic: go/constant panics on a division by zero, and ebiten.NewShader calls this compiler in the middle of a game. Report an error instead, as Go does for the same expression.

The guard has to cover token.QUO_ASSIGN as well as token.QUO, because the compiler rewrites the operator of an integer division to token.QUO_ASSIGN to force integer division in go/constant.

A negative shift count had the same problem: go/constant's Shift takes the count as uint, so `1 << -1` wrapped into a huge number. Report it as an error too.


hajimehoshi (approved):
LGTM, thanks!

PR #3537, opened 2026-08-26
ebitenutil: make DebugPrint skip runes outside its documented range

kumagi:
None

bug

## What this PR does | solves
DebugPrint's documents say the supported rune range is U+0000..U+00FF, but its loop falls through to a modular-indexed SubImage lookup for every rune, including runes outside the range. The modulo on a positive rune above U+00FF coincidentally pushes the Y index off the end of the embedded text atlas for every c >= 256, so the SubImage is empty and nothing is drawn, but each such rune is still recorded in the SubImage cache, with no bound on how many distinct runes can land in it. A negative rune would use the same modular indexing and could land inside the atlas at a meaningless position.

Skip runes outside U+0000..U+00FF explicitly so the implementation matches the documentation and the cache map only grows for supported runes.


ebitenutil/debugprint.go
```diff
@@ -70,6 +70,11 @@ func drawDebugText(rt *ebiten.Image, str string, ox, oy int) {
 			y += ch
 			continue
 		}
+		// The available runes are in U+0000 to U+00FF, per DebugPrint's documentation. Runes outside
+		// the range are silently skipped: the font atlas doesn't cover them.
+		if int(c) > 0xFF {
+			continue
+		}
```
hajimehoshi, line 75: Only this logic is fine. Remove the comments and the tests.

hajimehoshi: Also, just

```go
if c > 0xff {
```

should work

kumagi: removed.

hajimehoshi (approved):
LGTM

PR #3538, opened 2026-08-26
make Fill and Set no-op on nil color instead of panicking

kumagi:
None

bug

## What this PR does | solves

`Image.Fill` and `Image.Set` called `clr.RGBA()` on the given color without a nil check. A nil `color.Color` interface value has no method table, so either function panicked with a nil-pointer dereference:

```    
img.Fill(nil)        // panic: invalid memory address or nil pointer dereference
img.Set(2, 2, nil)   // panic: invalid memory address or nil pointer dereference
```
  
This is easy to hit in practice. A `color.Color` field left at its zero value, an untyped nil returned from a helper, or a color decoded from external data can all be nil, and the resulting panic takes down the entire game with an unhelpful message and no hint about the cause.

Treat a nil color as a no-op in both functions. This mirrors how `Fill` and `Set` already return early when the image itself is disposed, so a degenerate color now joins the other safe no-op cases instead of crashing.

Add a regression test that passes a nil color to `Fill` and `Set` and asserts the call returns without panicking.

### DIFF
--- a/image.go
+++ b/image.go
@@ -133,17 +133,9 @@ func (i *Image) Clear() {
 
 // Fill fills the image with a solid color.
 //
-// If the image is disposed, Fill does nothing.
-//
-// As of v2.10, Fill does not panic if clr is nil; it treats nil as a no-op.
 func (i *Image) Fill(clr color.Color) {
 	i.copyCheck()
-
-	if clr == nil {
-		return
-	}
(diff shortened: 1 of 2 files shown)


hajimehoshi:
As we discussed in another place, panicking makes sense there.

PR #3541, opened 2026-08-27
audio/internal/convert: fix StereoI16ReadSeeker.Seek position for mono sources

kumagi:
None

bug

## What this PR does | solves
StereoI16ReadSeeker.Seek converted the input offset from the presented stereo-byte space down to the underlying source's mono-byte space, but returned the underlying position unchanged. For a monaural source the returned position was therefore in mono-byte units while the wrapper promises stereo-byte units (the same units Stream.Length reports), so Seek(0, io.SeekEnd) returned half the stream length and any caller that relied on the returned position (e.g. InfiniteLoop.ensurePos) saw a half-scale value.

This is the i16 sibling of the F32 Seek fix: multiply the returned position back into stereo space when the source is mono, per the sample format.

Add TestMonoI16SeekEnd.

### DIFF
--- a/audio/internal/convert/stereoi16.go
+++ b/audio/internal/convert/stereoi16.go
@@ -141,18 +141,21 @@ func (s *StereoI16ReadSeeker) Seek(offset int64, whence int) (int64, error) {
 	if err != nil {
 		return 0, err
 	}
-	// s.source operates in mono-byte space, but this wrapper presents a
-	// stereo-byte space (and Stream.Length reports stereo bytes), so convert
-	// the returned position back.
+	// s.source operates in the source format's byte space, but this wrapper
+	// presents a stereo-i16 byte space (the same units Stream.Length reports),
+	// so convert the returned position back. The mono factor applies only to a
+	// monaural source, while the format factor applies regardless of the
+	// channel count.
 	if s.mono {
-		switch s.format {
-		case FormatU8:
-			pos *= 4
-		case FormatS16:
-			pos *= 2
-		case FormatS24:
-			pos = pos * 4 / 3
-		}
+		pos *= 2
+	}
+	switch s.format {
+	case FormatU8:
+		pos *= 2
+	case FormatS16:
+	case FormatS24:
+		pos *= 2
+		pos /= 3
 	}
 	return pos, nil
 }
--- a/audio/internal/convert/stereoi16.go
+++ b/audio/internal/convert/stereoi16.go
@@ -137,5 +137,22 @@ func (s *StereoI16ReadSeeker) Seek(offset int64, whence int) (int64, error) {
 		offset *= 3
 		offset /= 2
 	}
-	return s.source.Seek(offset, whence)
+	pos, err := s.source.Seek(offset, whence)
+	if err != nil {
+		return 0, err
+	}
+	// s.source operates in mono-byte space, but this wrapper presents a
+	// stereo-byte space (and Stream.Length reports stereo bytes), so convert
+	// the returned position back.
+	if s.mono {
+		switch s.format {
+		case FormatU8:
+			pos *= 4
+		case FormatS16:
+			pos *= 2
+		case FormatS24:
+			pos = pos * 4 / 3
+		}
+	}
+	return pos, nil
 }
--- a/audio/internal/convert/stereoi16.go
+++ b/audio/internal/convert/stereoi16.go
@@ -141,11 +141,8 @@ func (s *StereoI16ReadSeeker) Seek(offset int64, whence int) (int64, error) {
 	if err != nil {
 		return 0, err
 	}
-	// s.source operates in the source format's byte space, but this wrapper
-	// presents a stereo-i16 byte space (the same units Stream.Length reports),
-	// so convert the returned position back. The mono factor applies only to a
-	// monaural source, while the format factor applies regardless of the
-	// channel count.
+	// Convert the returned position from the source format's byte space to the
+	// stereo-i16 byte space this wrapper presents.
 	if s.mono {
 		pos *= 2
 	}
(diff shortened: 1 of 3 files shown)


hajimehoshi:
Confirmed the bug: for a mono Ogg via `DecodeWithoutResampling`, `Seek(0, io.SeekEnd)` returns 43224 while `Length()` reports 86448.

Two notes:

**1. The fix is incomplete.** The format scaling in `Read` applies to stereo sources too: `FormatU8` halves the source size and `FormatS24` multiplies it by 3/2 regardless of channel count. Gating the whole conversion on `s.mono` leaves stereo 8-bit and 24-bit sources broken. With this patch applied:

```
wav.DecodeWithoutResampling(jab.wav)  (stereo 16bit): Seek(0, io.SeekEnd) = 51276, Length() = 51276
wav.DecodeWithoutResampling(jab8.wav) (stereo 8bit):  Seek(0, io.SeekEnd) = 25638, Length() = 51276
```

The `switch s.format` should sit outside `if s.mono`, with `mono` contributing only its own `*= 2`, mirroring how the offset is converted on the way in. `examples/resources/audio/jab8.wav` works as a fixture.

**2.** #3544 and this PR both append to `audio/vorbis/vorbis_test.go` right after `TestNonSeekerF32`, so whichever lands second will conflict. Trivial to resolve, just worth knowing.

---

Reviewed with Claude (Claude Code).

audio/internal/convert/stereoi16_test.go
```diff
@@ -256,3 +256,47 @@ func TestStereoI16FromSigned24Bits(t *testing.T) {
...
+func TestStereoI16SeekEnd(t *testing.T) {
+	testCases := []struct {
+		name   string
+		format convert.Format
+		// bytesPerFrame is the size of one stereo frame in the source format.
+		bytesPerFrame int
+	}{
+		{"S16", convert.FormatS16, 4},
+		{"U8", convert.FormatU8, 2},
+		{"S24", convert.FormatS24, 6},
```
hajimehoshi: Use explicit member names and multiple lines

kumagi: Like this?

hajimehoshi: Yep

kumagi:
Confirmed — thanks for the detailed check. The return-value conversion only
handled monaural sources, so stereo 8-bit and 24-bit sources still reported a
position in the source format's byte space.

Fixed in a follow-up commit (3db6f38b3): the format factor now applies
unconditionally, and the mono factor contributes only its own doubling,
mirroring both Read's returned count and the offset conversion on the way in:

    if s.mono {
        pos *= 2
    }
    switch s.format {
    case FormatU8:
        pos *= 2
    case FormatS16:
    case FormatS24:
        pos *= 2
        pos /= 3
    }

Verified with your fixtures through wav.DecodeWithoutResampling:

    jab.wav   (stereo 16bit): Seek(0, io.SeekEnd) = 51276, Length() = 51276
    jab8.wav  (stereo 8bit):  Seek(0, io.SeekEnd) = 51276, Length() = 51276
                             (was 25638 before the fix)

Mono 8-bit and 24-bit behavior is unchanged (still *4 and *4/3).

Added TestStereoI16SeekEnd, which asserts Seek(0, io.SeekEnd) equals the
stream length (frames * 4 bytes) for all channel/format combinations. Without
the fix it fails for stereo U8 (got 200, want 400) and stereo S24 (got 600,
want 400). All existing audio tests pass.

Responded From DeepSeek V4 Flash.

audio/internal/convert/stereoi16_test.go
```diff
@@ -256,3 +256,59 @@ func TestStereoI16FromSigned24Bits(t *testing.T) {
...
 }
+
+// TestStereoI16SeekEnd confirms that Seek(0, io.SeekEnd) returns the length of
+// the stream in the wrapper's stereo-i16 byte space for every combination of
+// channel count and source format, matching Stream.Length.
+func TestStereoI16SeekEnd(t *testing.T) {
+	testCases := []struct {
+		name   string
+		format convert.Format
+		// bytesPerFrame is the size of one stereo frame in the source format.
```
hajimehoshi: Remove this comment

audio/internal/convert/stereoi16.go
```diff
@@ -137,5 +137,25 @@ func (s *StereoI16ReadSeeker) Seek(offset int64, whence int) (int64, error) {
...
-	return s.source.Seek(offset, whence)
+	pos, err := s.source.Seek(offset, whence)
+	if err != nil {
+		return 0, err
+	}
+	// s.source operates in the source format's byte space, but this wrapper
+	// presents a stereo-i16 byte space (the same units Stream.Length reports),
+	// so convert the returned position back. The mono factor applies only to a
+	// monaural source, while the format factor applies regardless of the
+	// channel count.
```
hajimehoshi: Shorten this comment

audio/internal/convert/stereoi16_test.go
kumagi: done.

audio/internal/convert/stereoi16.go
kumagi: done.

audio/internal/convert/stereoi16_test.go
```diff
@@ -256,3 +256,54 @@ func TestStereoI16FromSigned24Bits(t *testing.T) {
 		})
 	}
 }
+
+// TestStereoI16SeekEnd confirms that Seek(0, io.SeekEnd) returns the stream length.
```
hajimehoshi: Remove this comment

kumagi: done.

hajimehoshi (approved):
LGTM

Issue #3352, opened 2025-12-01
audio: a resampled stream's decode length doesn't match with `Length`


hajimehoshi:
### Ebitengine Version

v2.9.4

### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

### Go Version (`go version`)

go version go1.25.1 darwin/arm64

### What steps will reproduce the problem?

Run this test

```go

func TestResamplingLen(t *testing.T) {
	buf := make([]byte, 4*48000)
	src := bytes.NewReader(buf)
	resampled := convert.NewResampling(src, int64(len(buf)), 48000, 96000, 4)
	if got, want := resampled.Length(), int64(len(buf)*2); got != want {
		t.Errorf("got: %d, want: %d", got, want)
	}
	decodedBuf, err := io.ReadAll(resampled)
	if err != nil {
		t.Fatal(err)
	}
	if got, want := len(decodedBuf), int(len(buf)*2); got != want {
		t.Errorf("got: %d, want: %d", got, want)
	}
}
```

### What is the expected result?

The test passes

### What happens instead?

The test fails

```
--- FAIL: TestResamplingLen (0.02s)
    resampling_test.go:180: got: 442368, want: 384000
FAIL
FAIL    github.com/hajimehoshi/ebiten/v2/audio/internal/convert 0.382s
FAIL
```

PR #3542, opened 2026-08-27
audio/internal/convert: stop padding trailing silence for unknown-length resampling

kumagi:
None (the previous link to #3545 was unrelated: that issue is about `StereoI16ReadSeeker`/`StereoF32.Seek` returning a position in mono-byte space).

Bug

## What this PR does | solves
Resampling reads the source in resamplingBufferSize (4096) sample blocks. The final block can be shorter than a full block, but the buffer is always allocated at 4096 samples with the remainder zero-padded. The end-of-stream check compared the sample index against len(buffer)-1 (always 4095), so for an unknown-length source (size == 0) the resampler kept emitting the zero padding as trailing silence up to the block boundary instead of stopping at the source's real end.

When the end of the source is reached, derive its length into an unexported field from the EOF block position and its valid byte count, and terminate the resampled stream with the same length check the known-length path uses. The public Length() keeps returning 0 for an unknown-length source, as before. An empty (or shorter-than-a-sample) unknown-length source now returns (0, io.EOF) instead of trailing silence.

This also fixes a known-length bug: on main, a declared-size source read with a small (97-byte) buffer ended up to 12 bytes short of the same source read with io.ReadAll, because r.eof was set a sinc-window early and the read stopped at whatever buffer boundary came next. Moving r.eof onto the length check removes that buffer-size dependence.

Behavior change: for a source that ends before its declared size (e.g. a truncated WAV), the output is now zero-padded up to Length(), consistently with #3352, which requires the output length to match Length().

Add TestResamplingUnknownLengthTrailingSilence, which asserts that the unknown-length output is byte-identical to the known-length output for io.ReadAll and small reads, across ratios, bit depths, and sources including block-size multiples and sub-sample sources, and TestResamplingUnknownLengthEmptySource for the empty source.

### DIFF
--- a/audio/internal/convert/resampling.go
+++ b/audio/internal/convert/resampling.go
@@ -85,6 +85,10 @@ type Resampling struct {
 	// size is the length of the source stream in bytes. 0 indicates the length is unknown.
 	size int64
 
+	// derivedSize is the length of the source stream in bytes, derived when the source reaches its end.
+	// derivedSize is used only when size is 0.
+	derivedSize int64
+
 	from            int
 	to              int
 	bitDepthInBytes int
@@ -118,7 +122,19 @@ func (r *Resampling) bytesPerSample() int {
 }
 
 func (r *Resampling) Length() int64 {
-	s := int64(float64(r.size) * float64(r.to) / float64(r.from))
+	return r.length(r.size)
+}
+
+// sourceSize returns the length of the source stream in bytes, or 0 if the length is unknown.
+func (r *Resampling) sourceSize() int64 {
+	if r.size > 0 {
+		return r.size
+	}
+	return r.derivedSize
+}
+
+func (r *Resampling) length(size int64) int64 {
+	s := int64(float64(size) * float64(r.to) / float64(r.from))
 	return s / int64(r.bytesPerSample()) * int64(r.bytesPerSample())
 }
 
@@ -147,6 +163,10 @@ func (r *Resampling) src(i int64) (float64, float64, error) {
 			c += n
 			if err != nil {
 				if err == io.EOF {
+					// Derive the source length the first time the source reaches its end.
+					if r.eofBufIndex < 0 && r.size == 0 {
+						r.derivedSize = nextPos*resamplingBufferSize*sizePerSample + int64(c)
+					}
 					r.eofBufIndex = nextPos
 					break
 				}
@@ -199,10 +219,13 @@ func (r *Resampling) src(i int64) (float64, float64, error) {
 	}
 	ii := i % resamplingBufferSize
 	var err error
-	if r.eofBufIndex == r.srcBlock && ii >= int64(len(r.srcBufL[r.srcBlock])-1) {
-		err = io.EOF
+	if r.eofBufIndex == r.srcBlock {
+		// The source size remains 0 only when the source is empty.
+		if r.sourceSize() == 0 || ii >= int64(len(r.srcBufL[r.srcBlock])-1) {
+			err = io.EOF
+		}
 	}
-	if r.size > 0 && r.size/sizePerSample <= i {
+	if size := r.sourceSize(); size > 0 && size/sizePerSample <= i {
 		err = io.EOF
 	}
 	return r.srcBufL[r.srcBlock][ii], r.srcBufR[r.srcBlock][ii], err
@@ -262,9 +285,11 @@ func (r *Resampling) Read(b []byte) (int, error) {
 				return 0, err
 			}
 			// EOF from the at method indicates that the source reaches the end, and doesn't indicate the resampled data ends.
-			// Continue the loop even if EOF is returned.
-			if err == io.EOF {
+			// If the length is still unknown at the source's end, the source is empty, and the resampled data ends here.
+			if err == io.EOF && r.sourceSize() == 0 {
+				n = 0
 				r.eof = true
+				break
 			}
 			l16 := int16(ldata * (1<<15 - 1))
 			r16 := int16(rdata * (1<<15 - 1))
@@ -272,9 +297,10 @@ func (r *Resampling) Read(b []byte) (int, error) {
 			b[4*i+1] = byte(l16 >> 8)
 			b[4*i+2] = byte(r16)
 			b[4*i+3] = byte(r16 >> 8)
-			// If the source is an io.Seeker and the length is known, check whether the resampled data ends (#3352).
-			if r.size > 0 && r.pos+int64(size*i) >= r.Length() {
+			// If the length is known, check whether the resampled data ends (#3352).
+			if srcSize := r.sourceSize(); srcSize > 0 && r.pos+int64(size*i) >= r.length(srcSize) {
 				n = size * i
+				r.eof = true
 				break
 			}
 		}
@@ -284,8 +310,10 @@ func (r *Resampling) Read(b []byte) (int, error) {
 			if err != nil && err != io.EOF {
 				return 0, err
 			}
-			if err == io.EOF {
+			if err == io.EOF && r.sourceSize() == 0 {
+				n = 0
 				r.eof = true
+				break
 			}
 			l32 := float32(ldata)
 			r32 := float32(rdata)
@@ -299,8 +327,9 @@ func (r *Resampling) Read(b []byte) (int, error) {
 			b[8*i+5] = byte(r32b >> 8)
 			b[8*i+6] = byte(r32b >> 16)
 			b[8*i+7] = byte(r32b >> 24)
-			if r.size > 0 && r.pos+int64(size*i) >= r.Length() {
+			if srcSize := r.sourceSize(); srcSize > 0 && r.pos+int64(size*i) >= r.length(srcSize) {
 				n = size * i
+				r.eof = true
 				break
 			}
 		}
--- a/audio/internal/convert/resampling_test.go
+++ b/audio/internal/convert/resampling_test.go
@@ -332,3 +332,111 @@ func TestResamplingSeekInvalidWhence(t *testing.T) {
 		}
 	}
 }
+
+func TestResamplingUnknownLengthTrailingSilence(t *testing.T) {
+	testCases := []struct {
+		name string
+		from int
+		to   int
+	}{
+		{
+			name: "1:1",
+			from: 44100,
+			to:   44100,
+		},
+		{
+			name: "upsample",
+			from: 44100,
+			to:   48000,
+		},
+		{
+			name: "downsample",
+			from: 48000,
+			to:   44100,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			for _, bitDepthInBytes := range []int{2, 4} {
+				t.Run(fmt.Sprintf("bitDepthInBytes=%d", bitDepthInBytes), func(t *testing.T) {
+					// 8192 is an exact multiple of the 4096-sample block, 8190 ends within the sinc window of the boundary, and 1 and 3 are not a whole number of samples at bitDepthInBytes 4.
+					for _, size := range []int{1, 3, 100, 8190, 8192, 100000} {
+						t.Run(fmt.Sprintf("size=%d", size), func(t *testing.T) {
+							inB := newSoundBytes(size, bitDepthInBytes)
+
+							known := convert.NewResampling(bytes.NewReader(inB), int64(len(inB)), tc.from, tc.to, bitDepthInBytes)
+							knownB, err := io.ReadAll(known)
+							if err != nil {
+								t.Fatal(err)
+							}
+
+							// The unknown length must end the stream at the source's real
+							// end for a seekable source and a non-seekable one alike.
+							for _, seekable := range []bool{true, false} {
+								t.Run(fmt.Sprintf("seekable=%v", seekable), func(t *testing.T) {
+									newSource := func() io.Reader {
+										var src io.Reader = bytes.NewReader(inB)
+										if !seekable {
+											src = &reader{r: src}
+										}
+										return src
+									}
+
+									unknown := convert.NewResampling(newSource(), 0, tc.from, tc.to, bitDepthInBytes)
+									unknownB, err := io.ReadAll(unknown)
+									if err != nil {
+										t.Fatal(err)
+									}
+
+									if !bytes.Equal(unknownB, knownB) {
+										t.Errorf("unknown-length output (%d bytes) is not equal to known-length output (%d bytes)", len(unknownB), len(knownB))
+									}
+
+									unknownWithSmallBuf := convert.NewResampling(newSource(), 0, tc.from, tc.to, bitDepthInBytes)
+									var buf [97]byte
+									var smallB []byte
+									for {
+										n, err := unknownWithSmallBuf.Read(buf[:])
+										smallB = append(smallB, buf[:n]...)
+										if err != nil {
+											if err != io.EOF {
+												t.Fatal(err)
+											}
+											break
+										}
+									}
+									if !bytes.Equal(smallB, knownB) {
+										t.Errorf("unknown-length output with a small read buffer (%d bytes) is not equal to known-length output (%d bytes)", len(smallB), len(knownB))
+									}
+								})
+							}
+						})
+					}
+				})
+			}
+		})
+	}
+}
+
+func TestResamplingUnknownLengthEmptySource(t *testing.T) {
+	for _, bitDepthInBytes := range []int{2, 4} {
+		t.Run(fmt.Sprintf("bitDepthInBytes=%d", bitDepthInBytes), func(t *testing.T) {
+			for _, seekable := range []bool{true, false} {
+				t.Run(fmt.Sprintf("seekable=%v", seekable), func(t *testing.T) {
+					var src io.Reader = bytes.NewReader(nil)
+					if !seekable {
+						src = &reader{r: src}
+					}
+					r := convert.NewResampling(src, 0, 44100, 48000, bitDepthInBytes)
+					b, err := io.ReadAll(r)
+					if err != nil {
+						t.Fatal(err)
+					}
+					if len(b) != 0 {
+						t.Errorf("got: %d, want: %d", len(b), 0)
+					}
+				})
+			}
+		})
+	}
+}


audio/internal/convert/resampling.go
```diff
@@ -199,8 +213,22 @@ func (r *Resampling) src(i int64) (float64, float64, error) {
...
 	var err error
-	if r.eofBufIndex == r.srcBlock && ii >= int64(len(r.srcBufL[r.srcBlock])-1) {
-		err = io.EOF
+	if r.eofBufIndex == r.srcBlock {
+		// When the source is seekable and its length is unknown (size == 0),
+		// blocks are read from their exact positions, so srcBufValidLen is
+		// reliable. The final block may be shorter than resamplingBufferSize
+		// (the remainder is zero padding), and the end-of-stream must be
+		// reported at the last valid sample rather than at the buffer boundary.
+		// Otherwise the resampler keeps emitting the padding as trailing silence.
```
hajimehoshi: Shorten comments

audio/internal/convert/resampling.go
```diff
@@ -92,9 +92,16 @@ type Resampling struct {
 	srcBlock        int64
 	srcBufL         map[int64][]float64
 	srcBufR         map[int64][]float64
-	lruSrcBlocks    []int64
-	eof             bool
-	eofBufIndex     int64
+	// srcBufValidLen records how many samples in a buffered block are real
+	// source data. Blocks are allocated at resamplingBufferSize, but the final
+	// block may be shorter when the source ends; the rest is zero padding.
```
hajimehoshi: Remove comments

audio/internal/convert/resampling.go
```diff
@@ -92,9 +92,16 @@ type Resampling struct {
...
+	srcBufValidLen map[int64]int64
+	// seekable indicates whether source implements io.Seeker. When it does,
+	// blocks are read from their exact position, so srcBufValidLen is reliable.
```
hajimehoshi: Remove comments

kumagi: removed.

audio/internal/convert/resampling.go
kumagi: removed.

audio/internal/convert/resampling.go
kumagi: done.

hajimehoshi:
Thanks for digging into this. The diagnosis is right — for `size == 0` the resampler emits the zero padding of the final 4096-sample block — but I don't think the patch fixes it yet.

**1. `srcBufValidLen` is written under the wrong key.**

```go
buf = buf[:c]
r.srcBufValidLen[r.srcBlock] = int64(len(buf)) / sizePerSample
```

`r.srcBlock` is still the *previous* block at this point; it is assigned `nextPos` about 17 lines later. So the EOF block's key is never inserted, the lookup returns the map's zero value, `last` becomes `-1`, and `ii >= -1` holds for every index — EOF is reported at the *first* sample of the EOF block.

With the fixture from the new test (`newSoundBytes(44100, 2)`, 44100 samples):

```
known:   176400 bytes (44100 samples), last non-zero sample 44099
unknown: 170304 bytes (42576 samples), last non-zero sample 42575
```

1524 real samples (~35 ms) are dropped, and the stream ends on a non-zero sample, so the trailing silence is traded for an audible click.

**2. Correcting the key makes the new test fail.**

```
--- FAIL: TestResamplingUnknownLengthTrailingSilence/bitDepthInBytes=2
    unknown-length output 252224 bytes is far longer than known-length 176400 bytes
```

So the `src()` change alone does not address the symptom. The rest is in `Read`: `at()` sets `r.eof` mid-loop, but the loop keeps filling the remainder of `b` and returns the full `n`. The `size > 0` path truncates with `n = size * i; break`; the `size == 0` path has no equivalent, so the tail is padded out to whatever buffer the caller passed. Adding the symmetric break to both bit-depth branches gives 44092 samples against the known-length 44100 (8 short, half the sinc window), with the new test and all existing tests passing.

**3. `seekable` is not needed.** Removing the field and gating only on `r.size == 0` keeps the package green. That branch only decides which index reports `io.EOF`; it never influences a `Read`/`Seek` on the source, so a non-seekable source does not need the old behavior. The `r.size == 0` half *is* required — dropping it breaks `TestResamplingLen` for #3352 (`got: 767864, want: 768000`).

**4. The map could be a scalar.** The lookup is reachable only under `r.eofBufIndex == r.srcBlock`, so only one block's valid length is ever consulted. An `eofBufValidLen int64` assigned next to `r.eofBufIndex = nextPos` removes the map, the extra `delete` in the LRU eviction, and this whole class of key bug.

**5. The test only bounds the output from above,** which is why it passes against a version that drops 1524 real samples. Bounding both directions, and ideally comparing sample values against the known-length output, would catch the truncation too.

Also, the linked issue #3545 is about `StereoI16ReadSeeker`/`StereoF32.Seek` returning a position in mono-byte space; it looks unrelated to this change.

---

<sub>Review performed by Claude (Claude Code), posted on behalf of @hajimehoshi.</sub>

audio/internal/convert/resampling_test.go
```diff
@@ -181,3 +181,78 @@ func TestResamplingLen(t *testing.T) {
...
+// silence.
+func TestResamplingUnknownLengthTrailingSilence(t *testing.T) {
+	testCases := []struct {
+		name string
+		from int
+		to   int
+	}{
+		{"1:1", 44100, 44100},
+		{"upsample", 44100, 48000},
+		{"downsample", 48000, 44100},
```
hajimehoshi: Use explicit member names and multilines

kumagi: like this?

audio/internal/convert/resampling_test.go
```diff
@@ -181,3 +181,90 @@ func TestResamplingLen(t *testing.T) {
 		t.Errorf("got: %d, want: %d", got, want)
 	}
 }
+
+// Issue: when the source length is unknown (size == 0), the resampler must stop
+// at the source's real end instead of padding the final 4096-sample block with
+// silence.
```
hajimehoshi: Remove this comment

hajimehoshi:
Re-reviewed at `8ce56e8`. Both blocking issues are fixed, and the approach is better than what I suggested: deriving `r.size` the moment the EOF block loads means the unknown-length case reuses the known-length path wholesale, which is why the test can now assert exact equality instead of a tolerance. The new matrix (3 ratios x 2 bit depths x 2 sizes x seekable/non-seekable x ReadAll/97-byte buffer) is a solid improvement. Package suite, `go vet`, and `gofmt` are all clean.

Three things I'd still raise.

**1. `eofBufValidLen` and the `r.size == 0` branch in `src()` are now dead — except for an empty source, which is still broken.**

Since `r.size` is derived at the same point `eofBufValidLen` is written, `r.size == 0` can only survive when `nextPos == 0 && eofBufValidLen == 0`, i.e. a zero-byte source. Instrumenting the branch gives 0 hits across the entire package test suite, and 4264 hits only with a zero-byte source.

For that one case the branch doesn't actually help, because `Read`'s `r.size > 0` break never fires while size stays 0:

```
empty unknown-length source -> 512 bytes of silence (main: 20800)
```

Better than main, but still not `(0, io.EOF)`. Either drop the field, the branch, and its comment, or handle the empty source explicitly (set `r.eof` when the derived length is zero) so they earn their keep.

**2. `Length()` is no longer stable.**

For an unknown-length source it returns 0 until the source's end is reached, then the real value:

```
Length() before Read:      0
Length() after 1 Read:     0
Length() after full read:  192000
Seek(0, io.SeekEnd):  0 -> 192000
```

The decoders capture `Length()` once at construction (`audio/mp3/decode.go`, same shape in vorbis), so `Stream.Length()` stays 0 forever, while `Stream.Seek` delegates straight through to `Resampling.Seek` and does start honoring the discovered length. That leaves the wrapper internally inconsistent.

The field doc also needs updating — `// size is the length of the source stream in bytes. 0 indicates the length is unknown.` no longer describes a field that is filled in lazily.

**3. Gating `r.eof = true` on `r.size == 0` changes behavior for a source that ends before its declared size.**

Declaring 352800 bytes over a 176400-byte source:

| | output | trailing silent samples |
|---|---|---|
| main | 252224 B | 18956 |
| this PR | 352800 B | 44100 (1 s) |

It terminates, so no hang, but the trailing-silence symptom this PR removes from the `size == 0` side now appears on the `size > 0` side. Arguably the right call, since #3352 requires the output length to match `Length()` — I'd just want it stated in the commit message rather than changed.

For what it's worth I checked the obvious alternative, and dropping the gate is not viable: it costs 8 bytes at a small-read-buffer boundary (`downsample/bitDepthInBytes=4/size=100000` fails with 367488 vs 367496). So the gate is doing real work.

**Minor**

- The `eofBufValidLen` field comment runs three lines and its last sentence justifies the choice against the rejected alternative ("so a scalar suffices").
- The comment above the `r.size == 0` branch describes a path that is unreachable outside the empty-source case, so it reads as if it governs the normal flow.

---

<sub>Review performed by Claude (Claude Code), posted on behalf of @hajimehoshi.</sub>

kumagi:
Thanks for the thorough re-review! Addressed in the latest commit:

1. The empty unknown-length source now returns (0, io.EOF) immediately. Since the size is derived right when the EOF block loads, `r.size == 0` after an `io.EOF` from `at()` can only mean a zero-byte source, so `Read` terminates with the same `n = size * i; break` pattern as the known-length path. Added TestResamplingUnknownLengthEmptySource covering both bit depths and seekable/non-seekable. I kept the eofBufValidLen field and the src() branch since they now earn their keep on that path.
2. The size field doc now says the length is derived lazily when the source reaches its end. I agree the wrapper is inconsistent for decoders that snapshot Length() at construction, but that seems orthogonal to this fix — happy to follow up separately if you'd like.
3. The zero-padding behavior for a source that ends before its declared size is now stated explicitly in the commit message.

Comments were also trimmed: eofBufValidLen is a one-liner, the src() branch comment only notes that the size stays 0 for an empty source, and the "Issue:" comment above the test is gone. gofmt, go vet, and the full audio package suite are clean.

hajimehoshi:
Re-reviewed at `6b6e179`. The empty-source case is fixed and I could not break the main path: `go test ./audio/...`, `go vet ./audio/...` and `gofmt` are clean, and the unknown-length output is now **byte-identical** (not merely equal in length) to the known-length output across 2 bit depths × 3 ratios × sources of 1, 2, 100, 4096 and 100000 samples × `io.ReadAll` and 97-byte reads, including sources whose sample count is an exact multiple of the 4096-sample block. An empty unknown-length source returns `(0, io.EOF)` from the first `Read`.

Worth noting in the commit message: this also fixes a **known-length** bug. On main, a declared-size source read with a 97-byte buffer ends 12 bytes short of the same source read with `io.ReadAll` (435360 vs 435372), because `at()` sets `r.eof` a sinc-window early and the read then stops at whatever buffer boundary comes next. Moving the `r.eof = true` onto the length check removes that buffer-size dependence for `size > 0` too, which is more than the PR title claims.

Four things left.

**1. `eofBufValidLen` still does not earn its keep.**

I instrumented the `r.size == 0` branch in `src()`: across the whole package suite it is entered 36 times, all of them from `TestResamplingUnknownLengthEmptySource`, and `eofBufValidLen` is `0` at every one of them. That is structural rather than a gap in coverage — `r.size` is derived as `(nextPos*4096 + eofBufValidLen) * sizePerSample` in the same block that writes the field, so `r.size == 0` at the read site implies `nextPos == 0 && eofBufValidLen == 0`. The field is written on every EOF block but is only ever *read* as zero, so `last` is always `-1`.

The empty-source handling in `Read` is what earns its keep; the field is not needed for it:

```diff
@@ type Resampling struct {
 	lruSrcBlocks    []int64
 	eof             bool
 	eofBufIndex     int64
-	// eofBufValidLen is the number of valid samples in the block at eofBufIndex.
-	eofBufValidLen int64
 }
@@ func (r *Resampling) src(i int64) (float64, float64, error) {
 		buf = buf[:c]
-		if r.eofBufIndex == nextPos {
-			r.eofBufValidLen = int64(len(buf)) / sizePerSample
-			if r.size == 0 {
-				r.size = (nextPos*resamplingBufferSize + r.eofBufValidLen) * sizePerSample
-			}
+		if r.eofBufIndex == nextPos && r.size == 0 {
+			r.size = (nextPos*resamplingBufferSize + int64(len(buf))/sizePerSample) * sizePerSample
 		}
@@ func (r *Resampling) src(i int64) (float64, float64, error) {
 	if r.eofBufIndex == r.srcBlock {
-		var last int64
-		if r.size == 0 {
-			// The size remains 0 only when the source is empty.
-			last = r.eofBufValidLen - 1
-		} else {
-			last = int64(len(r.srcBufL[r.srcBlock]) - 1)
-		}
-		if ii >= last {
+		// The size remains 0 only when the source is empty.
+		if r.size == 0 || ii >= int64(len(r.srcBufL[r.srcBlock])-1) {
 			err = io.EOF
 		}
 	}
```

Tests, `go vet` and `gofmt` stay clean with that applied.

**2. The new test still only compares lengths.**

The outputs are byte-identical in every case I measured, so `bytes.Equal` costs nothing and closes the gap the length check leaves open: a tail that is the right length but the wrong samples. Given how much of this fix turns on where exactly the stream stops, an equality check on the bytes is the assertion that actually pins the behaviour down.

Two cases also worth adding, since none of the current sizes reach them:

- a sample count that is an exact multiple of 4096, where the final block is read full, the EOF block is the *next*, empty one, and the length is derived as `nextPos*4096` with no remainder. `newSoundBytes(8192, …)` gives exactly 2 blocks at `bitDepthInBytes=2` and exactly 1 at 4. I checked; it passes.
- a source shorter than one sample (`newSoundBytes(1, 4)` is 4 bytes against an 8-byte sample), where the derived length truncates to 0 samples. Also passes.

**3. `Length()` still changes value mid-stream.**

Unchanged from the last round, and I understand the argument that it is orthogonal, but it is introduced by this PR rather than pre-existing:

```
Length() before any Read:   0
Length() after 1 Read:      0
Length() after full read:   192000
Seek(0, io.SeekEnd):        0 -> 192000
```

`audio/wav`, `audio/vorbis` and `audio/mp3` all snapshot `r.Length()` at construction, so `Stream.Length()` stays 0 while `Stream.Seek` (delegating to `Resampling.Seek`) starts honouring the discovered length.

If we want the fix without that, the derived length does not have to live in `size`: keep `size` as the declared value (0 = unknown, as documented today), add a separate field for the derived one, and have `Read` consult an unexported helper that prefers the declared value. Public `Length()` then keeps returning 0 for an unknown-length source, exactly as before, and nothing outside the type observes the derivation. Happy either way, but I would rather decide it here than leave the field doc describing a value that moves.

**4. The PR still says it addresses #3545.**

#3545 is `StereoI16ReadSeeker`/`StereoF32.Seek` returning a position in mono-byte space for a monaural source — a different file and a different symptom, nothing to do with resampling. Merging with that link would close the wrong issue. This bug deserves its own issue, or none.

**Minor**

- The comment `// EOF from the at method indicates that the source reaches the end, and doesn't indicate the resampled data ends. // Continue the loop even if EOF is returned.` now sits directly above a `break`. It still describes the `size > 0` case correctly, but it reads as if it governed the code beneath it.
- In the `size == 0` break, `n = size * i` is always 0: that branch can only fire for an empty source, which reports EOF on the very first `at()` call, so `i` is always 0 there.

**Behaviour change, accepted**

Re-measured at HEAD, for a source that ends before its declared size (352800 bytes declared over a 176400-byte source): output is now 352800 bytes with 44100 trailing silent samples (1 s), against main's 252224 bytes / 18956. Reachable in practice with a truncated WAV, since `audio/wav` passes the data-chunk size straight through. It is the consistent choice given #3352 requires the output length to match `Length()`, and it is stated in the commit message now — thanks.

---

<sub>Review performed by Claude (Claude Code), posted on behalf of @hajimehoshi.</sub>

hajimehoshi:
Re-reviewed at `01d48e7`. All four points are addressed, and the `derivedSize` / `sourceSize()` split is exactly the shape I was after: `Length()` is back to a pure function of the declared `size`, so the `wav` / `vorbis` / `mp3` snapshots keep the semantics they have on main, while `Read` terminates on the derived value. Verified here:

```
unknown-length Length() before:            0
unknown-length Length() after full read:   0   (output 192000 bytes)
Seek(0, io.SeekEnd):                       0
re-read after Seek(0, io.SeekStart):       identical
```

`go test ./audio/...`, `go vet ./audio/...` and `gofmt` are clean, and the unknown-length output is byte-identical to the known-length one across 2 bit depths × 3 ratios × `io.ReadAll` / 97-byte reads / non-seekable, for every source length whose byte count is a multiple of `bytesPerSample`.

That qualifier is the one thing left.

**The derived length truncates to a whole sample; a declared length does not.**

```go
r.derivedSize = (nextPos*resamplingBufferSize + int64(len(buf))/sizePerSample) * sizePerSample
```

`int64(len(buf))/sizePerSample*sizePerSample` rounds the source length down to a sample boundary, but a caller-declared `size` is taken as a raw byte count and `length()` does its own rounding afterwards. So for a source whose byte length is not a multiple of `bytesPerSample`, the two paths disagree by one sample:

```
bitDepthInBytes=2, 48000->44100, 5-byte source:        known=4B   unknown=0B
bitDepthInBytes=2, 44100->48000, 16387-byte source:    known=17836B unknown=17832B
bitDepthInBytes=4, 48000->44100, 9-byte source:        known=8B   unknown=0B
bitDepthInBytes=4, 44100->48000, 15-byte source:       known=16B  unknown=8B
```

It is at most one sample and it is inaudible, but it is the same class of defect this PR exists to remove — the output length depending on something other than the source's real content — and the PR's own invariant is that the two paths agree. In fairness this came in with the diff I suggested last round, and `eofBufValidLen` had the same truncation before that, so it is not a regression from anything you wrote.

Deriving the raw byte count, the same unit a caller declares, makes them agree:

```diff
-			r.derivedSize = (nextPos*resamplingBufferSize + int64(len(buf))/sizePerSample) * sizePerSample
+			r.derivedSize = nextPos*resamplingBufferSize*sizePerSample + int64(len(buf))
```

It also makes the `sourceSize() == 0` branch mean what its comment says: today a source with fewer bytes than one whole sample derives 0 and takes the "source is empty" path, which happens to give the right answer for the wrong reason. With raw bytes, only a genuinely empty source derives 0.

Your test catches it with one more entry in the size list, since `newSoundBytes(3, 4)` is 12 bytes against an 8-byte sample:

```diff
-					for _, size := range []int{1, 100, 8192, 100000} {
+					for _, size := range []int{1, 3, 100, 8192, 100000} {
```

```
--- FAIL: TestResamplingUnknownLengthTrailingSilence
    unknown-length output (0 bytes) is not equal to known-length output (8 bytes)
    unknown-length output with a small read buffer (0 bytes) is not equal to known-length output (8 bytes)
```

Green with the one-line change, along with the rest of `./audio/...`.

The remaining route to an unknown length in-tree is `audio.ResampleReader` / `ResampleReaderF32`, whose documented contract is `size = 0` for an unknown length, so a caller streaming a truncated source is where this would show up; `audio/wav` cannot reach it, since a zero data-chunk size yields an empty section reader.

With that line, this is good to merge from my side.

---

<sub>Review performed by Claude (Claude Code), posted on behalf of @hajimehoshi.</sub>

kumagi:
Thanks for the careful re-review at `01d48e7`. Addressed in the latest commit:

1. The derived length is now the raw byte count, exactly as you suggested:

   `r.derivedSize = nextPos*resamplingBufferSize*sizePerSample + int64(len(buf))`

   so it uses the same unit as a caller-declared `size` and no longer rounds down to a sample boundary. I confirmed the test reproduces your failure first: with the old truncating derivation, `downsample/bitDepthInBytes=4/size=3` fails with `unknown-length output (0 bytes) is not equal to known-length output (8 bytes)` (both the `io.ReadAll` and the 97-byte-buffer variants), and it goes green with the one-line change. As a bonus, `sourceSize() == 0` now means a genuinely empty source, so the branch's comment is finally true — a 4-byte source against an 8-byte sample no longer takes the "empty source" path for the wrong reason.
2. Added `3` to the size list: `[]int{1, 3, 100, 8192, 100000}`.
3. The two minor items (the continue-loop comment above the `break`, and `n = size * i` → `n = 0` in the empty-source branch) were already applied in the previous commit `868a8d776`.
4. The #3545 reference is gone from the PR description.

For the record, everything from the earlier rounds stays in place: `bytes.Equal` on both outputs, the block-multiple (8192) and sub-sample sources in the matrix, no `eofBufValidLen`, and the `derivedSize`/`sourceSize()` split so public `Length()` remains a pure function of the declared size.

`go test ./audio/...`, `go vet ./audio/...` and `gofmt` are all clean.

hajimehoshi:
Re-reviewed at `2dc7fcd`. Both changes landed as intended, and the raw-byte derivation does fix the unaligned-length disagreement I reported.

While widening the matrix to confirm it, I found a second bug in the same line — and it is one I introduced, in the `derivedSize` split I asked for in round three. Bisecting the branch:

```
846662552  ok
6b6e179a0  ok
01d48e760  MISMATCH   <- derivedSize split
2dc7fcdd2  MISMATCH
```

**The derivation runs more than once and the later run wins.**

At `6b6e179` the derived length was written into `r.size` itself, so the `r.size == 0` condition doubled as an "already derived" flag. Moving the value into `derivedSize` kept the condition but removed what it was guarding, and nothing replaced it:

```go
if r.eofBufIndex == nextPos && r.size == 0 {
	r.derivedSize = nextPos*resamplingBufferSize*sizePerSample + int64(len(buf))
}
```

When the source ends inside the last 8 samples of a 4096-sample block, `at()`'s sinc lookahead reaches past the boundary and pulls in the *next* block. That read returns `(0, io.EOF)` straight away, `r.size` is still 0, and `derivedSize` is overwritten with the block-boundary offset — so the output is padded out to the boundary, which is the symptom this PR exists to remove:

```
4096 samples (on the boundary):  known=16384B  unknown=16384B  ok
4095 samples:                    known=16380B  unknown=16384B  +1 sample of silence
4094 samples:                    known=16376B  unknown=16384B  +2
4092 samples:                    known=16368B  unknown=16384B  +4
4089 samples:                    known=16356B  unknown=16384B  +7
4088 samples:                    known=16352B  unknown=16384B  +8
4087 samples:                    known=16348B  unknown=16348B  ok
```

The window is exactly the 8-sample sinc half-window, so roughly 8 in every 4096 source lengths land in it — independent of bit depth, ratio, read buffer size and seekability. It is at most 8 samples of silence, but the length is again a function of the block layout rather than of the source.

Deriving on the *first* EOF fixes it. `eofBufIndex` starts at -1, so `< 0` is exactly "the source has not reported its end yet", and `c` is already the byte count at that point:

```diff
 			if err != nil {
 				if err == io.EOF {
+					// Derive the source length the first time the source reaches its end.
+					if r.eofBufIndex < 0 && r.size == 0 {
+						r.derivedSize = nextPos*resamplingBufferSize*sizePerSample + int64(c)
+					}
 					r.eofBufIndex = nextPos
 					break
 				}
 				return 0, 0, err
 			}
 		}
 		buf = buf[:c]
-		if r.eofBufIndex == nextPos && r.size == 0 {
-			r.derivedSize = nextPos*resamplingBufferSize*sizePerSample + int64(len(buf))
-		}
```

Your test catches it with one more size, since `newSoundBytes(8190, …)` is 8190 samples at `bitDepthInBytes=2` (two short of 8192) and 4095 at 4 (one short of 4096), so a single entry covers the window for both:

```diff
-					for _, size := range []int{1, 3, 100, 8192, 100000} {
+					for _, size := range []int{1, 3, 100, 8190, 8192, 100000} {
```

```
--- FAIL: TestResamplingUnknownLengthTrailingSilence
    unknown-length output (32768 bytes) is not equal to known-length output (32760 bytes)
    unknown-length output with a small read buffer (32768 bytes) is not equal to known-length output (32760 bytes)
```

Green with the change above, together with `./audio/...`, `go vet` and `gofmt`.

With both applied I get no disagreement anywhere I can reach: 2 bit depths × 5 ratios (1:1, 44100↔48000, 8000↔48000) × 13 source lengths (0, 1, one byte either side of a sample, 100 samples, one byte and one sample either side of a block boundary, 2 blocks, 25000 samples, 25000 samples + 5 bytes) × 4 readers (unknown via `io.ReadAll`, unknown in 97-byte reads, unknown non-seekable in 97-byte reads, and known in 97-byte reads), all byte-identical to the known-length `io.ReadAll` reference.

**Minor**

`2dc7fcd`'s message body is a single 391-character line, where the rest of the branch wraps at 70. Worth rewrapping at 72 before this is merged.

---

<sub>Review performed by Claude (Claude Code), posted on behalf of @hajimehoshi.</sub>

kumagi:
Thanks for the re-review at `2dc7fcd`. Both points are addressed:

1. The derivation now runs only on the *first* EOF, exactly as you suggested: `r.derivedSize = nextPos*resamplingBufferSize*sizePerSample + int64(c)` guarded by `r.eofBufIndex < 0 && r.size == 0`, placed where the source reports `io.EOF`, and the post-`buf[:c]` write is gone. I confirmed the test reproduces your failure first: with the old derive-at-block-load, `1:1/bitDepthInBytes=2/size=8190` fails with `unknown-length output (32768 bytes) is not equal to known-length output (32760 bytes)` (both the `io.ReadAll` and the 97-byte-buffer variants), and it goes green with the derive-once change.
2. Added `8190` to the size list: `[]int{1, 3, 100, 8190, 8192, 100000}`.
3. Minor: the commit message is wrapped at 72 columns.

The branch had meanwhile fallen behind main (the `Seek` changes from #3574 and the unknown-length `Seek` fix), so I rebased onto `main` — resolving the only conflict, the tests appended after `TestResamplingLen` in `resampling_test.go`, by keeping both sides — and squashed the branch into a single commit `9a8f6f72b` whose message describes the whole change, including the known-length buffer-size fix and the zero-padding behavior change for sources that end before their declared size. The code diff is identical to what you reviewed at `2dc7fcd` plus the derive-once fix; upstream's `Seek` semantics are untouched.

`go test ./audio/...`, `go vet ./audio/...` and `gofmt` are all clean.

hajimehoshi:
Re-reviewed at `9a8f6f7`. The derive-once guard and the `8190` case landed, the branch is squashed onto current main and the message is rewrapped, and the sequential behaviour is now right everywhere I can reach: 2 bit depths × 5 ratios (1:1, 44100↔48000, 8000↔48000) × 12 source lengths (0, 1, a byte either side of a sample, 100 samples, a byte and a sample either side of a block boundary, exactly 1 and 2 blocks, 25000 samples + 5 bytes) × 4 readers (unknown via `io.ReadAll`, unknown in 97-byte reads, unknown non-seekable in 97-byte reads, known in 97-byte reads) — every one byte-identical to the known-length reference. `./audio/...`, `go vet` and `gofmt` clean, including the `wav` and `Seek` tests main picked up since the last round.

One case is left, and it is the mirror image of the last one: the guard takes the *first* EOF, but the first EOF is not always at the source's end.

**Seeking an unknown-length source past its end poisons the derived length for the life of the object.**

A seek can land in a block beyond the source; the read there returns `(0, io.EOF)` immediately, and the derived length becomes the seek target's block offset. Nothing later corrects it, because `eofBufIndex` is no longer negative. With a 400000-byte source read back from the start afterwards:

```
Seek(10000000, io.SeekStart), Read, Seek(0), ReadAll -> 9994240 bytes (want 400000)
Seek(  500000, io.SeekStart), Read, Seek(0), ReadAll ->  491520 bytes (want 400000)
Seek(  400004, io.SeekStart), Read, Seek(0), ReadAll ->  400000 bytes           ok
```

The first is about 57 seconds of silence appended to a 2.3-second stream. main is wrong here too (538944 bytes, its own block-boundary padding), so this is not a defect the PR invents — but it turns a bounded overshoot into one the caller chooses the size of, and unlike main it persists across a rewind.

It is reachable without anything exotic: `audio.Player.SetPosition` converts a duration to a byte offset and seeks straight through with no clamp, and `Resampling.Seek` deliberately does not clamp when the length is unknown ("the position can be clamped by the length only when the length is known"). So `SetPosition` past the end of an unknown-length resampled stream, then `Rewind`, is enough.

The invariant that fixes it: the source reporting EOF at offset X only proves the source is *at most* X long, and it is exact only when the read started at or before the real end. Keeping the smallest offset seen is exact as soon as any read reaches the true end, and it subsumes the near-boundary guard from last round rather than replacing it:

```diff
 			if err != nil {
 				if err == io.EOF {
-					// Derive the source length the first time the source reaches its end.
-					if r.eofBufIndex < 0 && r.size == 0 {
-						r.derivedSize = nextPos*resamplingBufferSize*sizePerSample + int64(c)
+					if r.size == 0 {
+						// Reaching the end at this offset means the source is at most this long.
+						// A seek can land past the source's end, so keep the smallest offset seen.
+						if s := nextPos*resamplingBufferSize*sizePerSample + int64(c); r.eofBufIndex < 0 || s < r.derivedSize {
+							r.derivedSize = s
+						}
 					}
 					r.eofBufIndex = nextPos
 					break
 				}
```

An empty source stays at 0 (nothing is smaller), the near-boundary block keeps the earlier partial-block offset, and the bogus seek offset is discarded the moment a read reaches the real end.

Neither of your two tests seeks, so nothing covers this. Something like:

```go
func TestResamplingUnknownLengthSeekPastEnd(t *testing.T) {
	const (
		from            = 44100
		to              = 44100
		bitDepthInBytes = 2
	)

	inB := newSoundBytes(100000, bitDepthInBytes)
	knownB, err := io.ReadAll(convert.NewResampling(bytes.NewReader(inB), int64(len(inB)), from, to, bitDepthInBytes))
	if err != nil {
		t.Fatal(err)
	}

	for _, offset := range []int64{int64(len(inB)) + 4, 10 * int64(len(inB))} {
		t.Run(fmt.Sprintf("offset=%d", offset), func(t *testing.T) {
			r := convert.NewResampling(bytes.NewReader(inB), 0, from, to, bitDepthInBytes)
			if _, err := r.Seek(offset, io.SeekStart); err != nil {
				t.Fatal(err)
			}
			var buf [4096]byte
			if _, err := r.Read(buf[:]); err != nil && err != io.EOF {
				t.Fatal(err)
			}
			if _, err := r.Seek(0, io.SeekStart); err != nil {
				t.Fatal(err)
			}
			gotB, err := io.ReadAll(r)
			if err != nil {
				t.Fatal(err)
			}
			if !bytes.Equal(gotB, knownB) {
				t.Errorf("output after seeking past the end (%d bytes) is not equal to known-length output (%d bytes)", len(gotB), len(knownB))
			}
		})
	}
}
```

```
--- FAIL: TestResamplingUnknownLengthSeekPastEnd/offset=4000000
    output after seeking past the end (3997696 bytes) is not equal to known-length output (400000 bytes)
```

Green with the change above, along with the rest of `./audio/...`.

**Minor**

The comment above the size list is one 190-character line; the file otherwise wraps. Splitting it across two or three lines would read better.

---

<sub>Review performed by Claude (Claude Code), posted on behalf of @hajimehoshi.</sub>

hajimehoshi:
We found a broader issue (#3589), so we'll take care of it. Let me close this.

EDIT: Hm, this was opposite: this PR is broader than #3589...

PR #3543, opened 2026-08-27
internal/ui: return initial position from Window().Position before the main loop

kumagi:
None

bug

## What this PR does | solves
Window().Position panicked with "can't be called before the main loop starts" when the backend was not yet published, even though SetPosition gracefully stores the initial position in that state and Size returns it. Return getInitWindowPositionInDIP when the backend is nil, matching SetPosition and Size, so the write/read pair is consistent before RunGame.

Add TestWindowPositionBeforeRun.


internal/ui/window_position_test.go
```diff
@@ -0,0 +1,26 @@
+package ui
+
+import "testing"
+
+// TestWindowPositionBeforeRun ensures Window().Position does not panic before the
+// main loop starts. SetPosition stores the initial position when the backend is not
+// yet published, and Position must return it (consistently with Size) instead of
+// panicking.
+func TestWindowPositionBeforeRun(t *testing.T) {
```
hajimehoshi: We don't need tests

internal/ui/window_desktop.go
```diff
@@ -410,7 +410,9 @@ func (w *desktopWindow) Position() (int, int) {
 	}
 	b := w.ui.runningBackend()
 	if b == nil {
-		panic("ui: WindowPosition can't be called before the main loop starts")
+		// The window has not started yet. Return the buffered initial position
+		// that SetPosition stored, consistently with Size.
```
hajimehoshi: Remove the comment

internal/ui/window_position_test.go
kumagi: removed this file.

internal/ui/window_desktop.go
kumagi: removed.

hajimehoshi (approved):
LGTM, thanks!

PR #3544, opened 2026-08-27
audio/vorbis: fix F32 stream Seek position for mono sources

kumagi:
#3545 

bug

## What this PR does | solves
DecodeF32 wraps a mono float32BytesReadSeeker in convert.NewStereoF32 when the source is monaural, so the stream it returns is presented in stereo-byte space (length is samples * channels * 4 * 2).

convert.StereoF32.Seek converted the input offset from stereo to mono space before delegating, but returned the underlying (mono) position unchanged. For a mono source the returned position must be doubled back into stereo space. (The second half of the original bug, float32BytesReadSeeker.Seek adding a sample count instead of a byte count on io.SeekEnd, already landed on main in dfe64228a.)

For a monaural Ogg decoded to F32, Seek(0, io.SeekEnd) used to return 98410 (the raw sample count) instead of the correct 787280 bytes, and a forward Seek(8, io.SeekStart) returned 4 instead of 8.

Add TestMonoF32SeekEnd covering SeekEnd, a negative-offset SeekEnd, and SeekStart (including a forward seek), and TestStereoF32SeekEnd in audio/internal/convert mirroring TestStereoI16SeekEnd so the two sibling wrappers have symmetric coverage.

### DIFF
--- a/audio/internal/convert/stereof32.go
+++ b/audio/internal/convert/stereof32.go
@@ -68,5 +68,12 @@ func (s *StereoF32) Seek(offset int64, whence int) (int64, error) {
 	if s.mono {
 		offset /= 2
 	}
-	return s.source.Seek(offset, whence)
+	pos, err := s.source.Seek(offset, whence)
+	if err != nil {
+		return 0, err
+	}
+	if s.mono {
+		pos *= 2
+	}
+	return pos, nil
 }
--- a/audio/vorbis/vorbis_test.go
+++ b/audio/vorbis/vorbis_test.go
@@ -232,7 +232,7 @@ func TestMonoI16SeekEnd(t *testing.T) {
 	}
 }
 
-func TestMonoF32SeekEnd(t *testing.T) {
+func TestMonoF32Seek(t *testing.T) {
 	bs := test_mono_ogg
 
 	s, err := vorbis.DecodeF32(bytes.NewReader(bs))
--- a/audio/vorbis/vorbis_test.go
+++ b/audio/vorbis/vorbis_test.go
@@ -231,3 +231,47 @@ func TestMonoI16SeekEnd(t *testing.T) {
 		t.Errorf("Seek(0, io.SeekEnd): got %d, want %d (stream length)", got, want)
 	}
 }
+
+func TestMonoF32SeekEnd(t *testing.T) {
+	bs := test_mono_ogg
+
+	s, err := vorbis.DecodeF32(bytes.NewReader(bs))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	t.Run("SeekEnd", func(t *testing.T) {
+		got, err := s.Seek(0, io.SeekEnd)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if want := s.Length(); got != want {
+			t.Errorf("Seek(0, io.SeekEnd): got %d, want %d (stream length)", got, want)
+		}
+	})
+	t.Run("SeekEndWithNegativeOffset", func(t *testing.T) {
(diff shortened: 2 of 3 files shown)


audio/internal/convert/stereof32.go
```diff
@@ -68,5 +68,15 @@ func (s *StereoF32) Seek(offset int64, whence int) (int64, error) {
...
 	}
-	return s.source.Seek(offset, whence)
+	pos, err := s.source.Seek(offset, whence)
+	if err != nil {
+		return 0, err
+	}
+	if s.mono {
+		// The underlying source is mono, so its returned position is in
+		// mono-byte space. Convert it back to the stereo-byte space this
+		// wrapper presents.
```
hajimehoshi: Remove the comment

kumagi: removed.

hajimehoshi:
I'll take a look after https://github.com/hajimehoshi/ebiten/pull/3541 is merged

hajimehoshi:
Resolve conflicts

hajimehoshi:
Thanks — this is still the right fix, but it needs a rebase: `dfe64228a` and #3541 both landed in this area since you filed it.

- `audio/vorbis/float32.go` — your `r.r.Length() * sampleSize` change is already on main, identically. That hunk can be dropped.
- `audio/internal/convert/stereof32.go` — unchanged and still needed; this is the remaining half of #3545.
- `audio/vorbis/vorbis_test.go` — conflicts at the append point with `TestMonoI16SeekEnd` and `TestStereoF32Seek`. `TestMonoF32SeekEnd` just needs re-placing; I confirmed it passes as written once the `StereoF32` fix is applied on current main.

One suggestion: could you add a `Seek` test to `audio/internal/convert/stereof32_test.go` mirroring the `TestStereoI16SeekEnd` that #3541 added? `StereoF32` is only ever constructed with `mono=true`, so the `mono=false` passthrough isn't reachable from any decoder test, and the two sibling wrappers would then have symmetric coverage.

With that, this closes #3545.

kumagi:
Rebased onto current main and addressed the review:

- Dropped the `audio/vorbis/float32.go` hunk since `r.r.Length() * sampleSize` already landed on main via dfe64228a.
- Kept the `convert.StereoF32.Seek` fix (returning the position doubled back into stereo-byte space for mono sources), with the comment removed.
- Re-placed `TestMonoF32SeekEnd` after `TestMonoI16SeekEnd` in `audio/vorbis/vorbis_test.go`.
- Added `TestStereoF32SeekEnd` to `audio/internal/convert/stereof32_test.go` mirroring `TestStereoI16SeekEnd`, covering both `mono=true` and the `mono=false` passthrough so the two sibling wrappers have symmetric coverage. It fails without the `StereoF32` fix and passes with it; `go test ./audio/...` is green.

Thanks for the pointers!

*Addressed by Qwen3.8-Flash-Next, on behalf of @kumagi.*

audio/vorbis/vorbis_test.go
```diff
@@ -231,3 +231,47 @@ func TestMonoI16SeekEnd(t *testing.T) {
 		t.Errorf("Seek(0, io.SeekEnd): got %d, want %d (stream length)", got, want)
 	}
 }
+
+func TestMonoF32SeekEnd(t *testing.T) {
```
hajimehoshi: Rename this to `TestMonoF32Seek`

kumagi: done

hajimehoshi (approved):
LGTM

PR #3547, opened 2026-08-28
internal/gbm, internal/ui: run on a DRM/KMS display through GBM

### LINKED ISSUE #3546: ebiten: running without a window system fails on Mali/GBM handhelds (fbdev uses EGL_DEFAULT_DISPLAY)
### Ebitengine Version

v2.10.0-alpha

### Operating System

- [ ] Windows
- [ ] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

go version go1.25.6 windows/amd64

Im trying to port ebiten based app into RG DS(Rockchip RK3566 + Mali-G52 handheld) with ROCKNIX.

Ebitengine should present on a Linux handheld that has no window system but does have DRM/KMS + GBM (Mali and other Mesa/panfrost-class GPUs) the same "run without a window system" goal as #3491 / #3492, but for GBM devices instead of raw-framebuffer / PowerVR ones.

A minimal `ebiten.RunGame`, cross-built with `CGO_ENABLED=0 GOOS=linux GOARCH=arm64` and run with no `DISPLAY` / `WAYLAND_DISPLAY` (compositor stopped), should show frames on the panel.

### What happens instead?

The framebuffer backend (#3492) is selected correctly, but it then fails at startup and nothing renders:

```
fbdev: eglInitialize failed: EGL error 0x3001   // EGL_NOT_INITIALIZED
```

The cause is that the backend calls `eglGetDisplay(EGL_DEFAULT_DISPLAY)`. On this device (Rockchip RK3566 + Mali-G52, ROCKNIX) the Mali userspace is the *wayland-gbm* driver variant, whose EGL only initializes through a *platform* display; the default display returns `EGL_NOT_INITIALIZED`.

**Repro:** minimal `ebiten.RunGame`, `CGO_ENABLED=0 GOOS=linux GOARCH=arm64`, run on the device with the compositor stopped so it can take DRM master.

mirusu400:

Closes #3546

bug

## What this PR does | solves

Adds a GBM/KMS backend so Ebitengine runs on handhelds that have no window system but a Mali (or other Mesa/panfrost-class) GPU, where the fbdev backend (#3492) fails at `eglInitialize` with `EGL_NOT_INITIALIZED` because it uses `eglGetDisplay(EGL_DEFAULT_DISPLAY)` these devices need the GBM platform display and the app has to drive KMS itself.

`internal/gbm` (new) opens the DRM device and takes master, picks a connected connector/mode/CRTC, creates a `gbm_device` and `gbm_surface`, gets an EGL display via `EGL_PLATFORM_GBM`, and presents by scanning out the front `gbm_bo` (`drmModeSetCrtc` for the first frame, `drmModePageFlip` after). EGL/GBM/libdrm are loaded with purego, so it's cgo-free and cross-builds with `CGO_ENABLED=0` for linux/arm64, like fbdev. `internal/ui` falls back glfw → gbm → fbdev when no window system is available, and also reads keyboard keys from evdev for those backends (gamepad and audio already work without a window system). `examples/gbmkeytest` is a minimal input/present test.

Verified on a RG DS (Rockchip RK3566 + Mali-G52 handheld, ROCKNIX): a game renders at 640x480 through this backend and gamepad input works; `go vet ./internal/gbm` is clean and the windows/amd64 build still compiles (the `ui_notgbm` stub). The keyboard reader and the example are separate commits and can be dropped if you'd rather keep this to just the GBM backend.

Re #3491: this adds GBM as a standalone sibling to fbdev to start with, happy to rebase onto the Wayland + runtime backend-selection work when it lands. I have the device to test and maintain it.


hajimehoshi:
As v2.10 release will happen soon, I'll review this for the next milestone v2.11.

examples/gbmkeytest/main.go
```diff
@@ -0,0 +1,92 @@
+// gbmkeytest is a minimal Ebiten program to exercise the no-window-system
+// backends (gbm / fbdev) and their input on a handheld: it fills the screen
+// with a slowly changing color to show frames are presenting, and lists the
+// keyboard keys and gamepad buttons/axes currently active to show input is
+// delivered without a window system.
+package main
```
hajimehoshi, line 6: Why do we need an example file?

mirusu400: When making the commit, I tested GBM console gamepads with this test app... I will delete it if it is not needed.

hajimehoshi: This should not be needed. There should not be a GBM-specific feature.

internal/gbm/display_linux.go
```diff
@@ -0,0 +1,254 @@
...
+	if res == nil {
+		_ = f.Close()
+		return nil, fmt.Errorf("gbm: drmModeGetResources failed (no KMS, or another client holds the device)")
+	}
+	defer drml.FreeResources(res)
+
+	countConns := int(i32At(res, 32))
+	connsPtr := ptrAt(res, 40)
+	countCrtcs := int(i32At(res, 16))
+	crtcsPtr := ptrAt(res, 24)
```
hajimehoshi: [P1] Use architecture-correct DRM structure layouts

These offsets assume 8-byte pointers, but this backend also builds and is selected on 32-bit Linux. In the 32-bit `drmModeRes` layout, `count_connectors` and `connectors` are at offsets 16 and 20, not 32 and 40. The current reads interpret unrelated fields as counts and pointers, potentially crashing before fbdev fallback. The connector's modes pointer also has an architecture-dependent offset. Please use native structures with the correct layout, or explicitly restrict the backend to supported architectures.

Reference: [libdrm structure definitions](https://cgit.freedesktop.org/drm/libdrm/tree/xf86drmMode.h).

hajimehoshi (commented):
Four correctness findings are attached inline.

Validation: cross-compilation of `internal/gbm` and `internal/ui` passed for Linux arm64 and arm with CGO disabled. Hardware behavior was not tested.

AI-assisted review authored by Codex (OpenAI), on behalf of @hajimehoshi.

internal/ui/ui_desktop.go
```diff
@@ -137,7 +137,11 @@ func (u *UserInterface) Run(game Game, options *RunOptions) error {
 	if b := maybeNewGLFWBackend(u); b != nil {
 		return b.run(game, options)
 	}
-	// Fall back to a framebuffer device where there is no window system.
+	// No window system: prefer a DRM/KMS display driven with GBM (Mali and
+	// Mesa/panfrost-class GPUs), then fall back to a plain framebuffer device.
+	if gb, err := maybeNewGBMBackend(u); err == nil {
+		return gb.run(game, options)
+	}
```
hajimehoshi, line 144: [P2] Preserve fbdev fallback when GBM initialization fails

`maybeNewGBMBackend` only opens the DRM display and creates the GBM device; EGL initialization and surface creation happen later in `gb.run`. If DRM/GBM is available but the installed EGL implementation only supports fbdev, this branch returns the initialization error without trying the previously working fbdev backend. Please complete the GBM usability probe before committing to the backend, and release its resources if initialization fails so fbdev can still be attempted.

internal/gbm/display_linux.go
```diff
@@ -0,0 +1,254 @@
...
+		}
+		drml.FreeConnector(conn)
+	}
+	if !found {
+		_ = f.Close()
+		return nil, fmt.Errorf("gbm: no connected connector with a mode")
+	}
+	if d.crtcID == 0 && countCrtcs > 0 {
+		d.crtcID = u32Index(crtcsPtr, 0)
+	}
```
hajimehoshi: [P2] Select a CRTC compatible with the connector

When the connected output has no active CRTC, this fallback picks the first device CRTC without checking whether any of the connector's encoders can use it. On hardware with restricted routing, `drmModeSetCrtc` then fails even when another compatible CRTC is available. Please enumerate the connector's encoders and select a CRTC allowed by their `possible_crtcs` masks.

Reference: [KMS encoder/CRTC compatibility](https://www.kernel.org/doc/html/v6.11/gpu/drm-kms.html).

internal/gbm/egl_linux.go
```diff
@@ -0,0 +1,309 @@
...
+			return fmt.Errorf("gbm: drmModeSetCrtc failed: %d", r)
+		}
+		c.modesetDone = true
+	} else {
+		if r := drml.PageFlip(c.d.fd, c.d.crtcID, fb, _DRM_MODE_PAGE_FLIP_EVENT, 0); r != 0 {
+			return fmt.Errorf("gbm: drmModePageFlip failed: %d", r)
+		}
+		if _, err := unix.Read(int(c.d.fd), c.eventBuf[:]); err != nil {
+			return fmt.Errorf("gbm: waiting for page flip failed: %w", err)
+		}
```
hajimehoshi: [P2] Honor the requested swap interval in KMS presentation

This path always queues a synchronous page flip and blocks until it completes, regardless of `c.swapInterval`. Changing `eglSwapInterval` alone cannot remove that KMS wait, so `SetVsyncEnabled(false)` still limits presentation to the display refresh rate. Please make the presentation path honor the requested interval, using asynchronous flips where supported and handling devices that cannot perform them.

Reference: [kernel page-flip contract](https://raw.githubusercontent.com/torvalds/linux/master/include/uapi/drm/drm_mode.h).

internal/ui/keyboard_evdev_linux_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+package ui
+
+import (
+	"unsafe"
+)
+
+const (
+	linuxInputEventPointerSize = unsafe.Sizeof(uintptr(0))
+	linuxInputEventSize        = unsafe.Sizeof(linuxInputEvent{})
```
hajimehoshi, line 23: [P2] Match the keyboard implementation's build constraints

Android includes `_linux_test.go` files, but `keyboard_evdev_linux.go` excludes Android (as well as the `nintendosdk` and `playstation5` build tags). This test therefore references a type that does not exist on those targets. An Android arm64 test compilation confirms the newly introduced `undefined: linuxInputEvent` errors at lines 23 and 24. Please add the implementation's matching build constraint above the package declaration:

```go
//go:build linux && !android && !nintendosdk && !playstation5
```

AI-assisted review authored by Codex (OpenAI), on behalf of @hajimehoshi.

internal/ui/ui_notgbm.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+// gbmBackend is never instantiated here. It exists so that Run can name the
+// same type on every system.
+type gbmBackend struct{}
+
+// maybeNewGBMBackend never returns a backend: DRM/KMS with GBM is a Linux
+// concept.
+func maybeNewGBMBackend(u *UserInterface) (*gbmBackend, error) {
+	return nil, errors.New("a DRM/KMS GBM display is a Linux feature")
```
hajimehoshi, line 30: `ui: a DRM/KMS GBM display is not available in this environment`

internal/gbm/egl_linux.go
```diff
@@ -0,0 +1,329 @@
+// Copyright 2026 The Ebitengine Authors
```
hajimehoshi, line 1: This is duplicated with internal/fbdev/egl_linux.go. Make a common package `internal/egl` and put the implementation

internal/gbm/egl_linux.go
```diff
@@ -0,0 +1,329 @@
...
+	if c.gbmSurface == 0 {
+		return nil, errors.Join(fmt.Errorf("gbm: gbm_surface_create failed"), c.Close())
+	}
+
+	c.surface = c.egl.CreateWindowSurface(c.display, config, c.gbmSurface, nil)
+	if c.surface == 0 {
+		return nil, errors.Join(fmt.Errorf("gbm: eglCreateWindowSurface failed: %w", c.lastError()), c.Close())
+	}
+
+	for _, version := range []int32{3, 2} {
```
hajimehoshi, line 120: ES 2 is not available. (See the latest `internal/fbdev` ced6337d74f2905a88de3ef88f71813ec2161c02). I think the implementation will be very similar to `internal/fbdev`, so unifying it to `internal/egl` will resolve this anyway

internal/gbm/display_linux_test.go
```diff
@@ -0,0 +1,86 @@
...
+	wantDRMModeResCountConnectorsOffset = 4 * pointerSize
+	wantDRMModeResConnectorsOffset      = 5 * pointerSize
+	wantDRMModeConnectorSize            = 7*pointerSize + 32
+	wantDRMModeConnectorModesOffset     = pointerSize + 32
+)
+
+// Compile-time libdrm ABI checks.
+var (
+	_ [wantDRMModeResSize - drmModeResSize]byte
+	_ [drmModeResSize - wantDRMModeResSize]byte
```
hajimehoshi, line 42: This can reduce the lines by half

```go
_ [0]byte = [wantDRMModeResSize - drmModeResSize]byte
```

hajimehoshi:
You need to resolve conflicts

hajimehoshi:
You still need to resolve conflicts...

PR #3549, opened 2026-08-28
audio: fix an out-of-bounds panic in InfiniteLoop blending with a short after-loop tail

kumagi:
None

bug

## What this PR does | solves

When the data after the loop end is shorter than one frame, blendRate divided by the frame count of afterLoop, which is zero. The result was NaN, which slipped past the rate == 0 check, and the blend loop then read relpos+1 (int16) or relpos+3 (float32) past the end of afterLoop, panicking.

Return 0 from blendRate when afterLoop is shorter than one frame so that blending is skipped.

Add TestInfiniteLoopWithPartialFrameAfterLoop, which panics without this fix.

### DIFF
--- a/audio/loop.go
+++ b/audio/loop.go
@@ -136,10 +136,6 @@ func (i *InfiniteLoop) blendRate(pos int64) float32 {
 	p := (pos - i.lstart) / int64(i.bytesPerSample)
 	l := len(i.afterLoop) / i.bytesPerSample
 	if l == 0 {
-		// The after-loop data is shorter than one frame, so there is nothing to
-		// blend with. Return 0 to skip blending: dividing p by l would be a
-		// division by zero and produce NaN, and reading a frame past the end of
-		// afterLoop would be out of bounds.
 		return 0
 	}
 	return 1 - float32(p)/float32(l)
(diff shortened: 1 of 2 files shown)


audio/loop.go
```diff
@@ -135,6 +135,13 @@ func (i *InfiniteLoop) blendRate(pos int64) float32 {
 	}
 	p := (pos - i.lstart) / int64(i.bytesPerSample)
 	l := len(i.afterLoop) / i.bytesPerSample
+	if l == 0 {
+		// The after-loop data is shorter than one frame, so there is nothing to
+		// blend with. Return 0 to skip blending: dividing p by l would be a
+		// division by zero and produce NaN, and reading a frame past the end of
+		// afterLoop would be out of bounds.
```
hajimehoshi: Remove the comment

audio/loop_test.go
```diff
@@ -151,6 +151,22 @@ func TestInfiniteLoopWithIntro(t *testing.T) {
 	}
 }
 
+func TestInfiniteLoopWithPartialFrameAfterLoop(t *testing.T) {
+	// The after-loop data is shorter than one frame (4 bytes for int16 stereo).
+	// The blend loop used to divide by zero in blendRate and then read past the
+	// end of afterLoop, panicking.
```
hajimehoshi: Remove the comment

audio/loop.go
kumagi: done.

audio/loop_test.go

hajimehoshi (approved):
LGTM

PR #3550, opened 2026-08-28
internal/shader: reject a Kage for-loop whose condition never becomes false

kumagi:
None

bug

## What this PR does | solves
Whether a Kage for-loop condition can ever become false is decidable at compile time: the loop variable's initial value, the end value and the post statement's delta are all constants. A loop like `for i := 0; i < 10; i += 0 {}` or `for i := 0; i >= 0; i++ {}` whose condition never becomes false would hang the GPU (or be rejected by the driver, far from the mistake), so reject it at compile time — unless the loop body can exit the loop.

- `forLoopConditionAlwaysTrue` decides, from the init value, the end value, the delta and the comparison operator, whether the condition can never become false. A zero delta (`i += 0`) is not a special case but one case of this analysis: `for i := 10; i < 10; i += 0 {}` never runs and is still valid.
- The analysis runs on the original constants and uses go/constant's exact rational arithmetic, so float loop variables are treated like integers. Non-numeric constants (e.g. a bool, a pre-existing type-check hole) skip the analysis instead of crashing the compiler.
- The check runs after the body is parsed and is skipped when the body contains a `break` exiting this loop, a `return` or a `discard()` (all backends return from the fragment function after a discard). A `break` in a nested for-loop exits only the nested loop and is not counted. Loop-until-`break` sources like `for i := 0; i >= 0; i++ { … break }` keep compiling; `for { }` is not expressible in Kage, so this is the only way to write such a loop.

`TestCompileForLoopTermination` covers integer and float loop variables, `+=` / `-=` / `++` / `--`, conditions that are false at init, and bodies with `break`, `return` or `discard()` including inside nested loops. `TestCompileForLoopNonNumericConstant` only requires that a non-numeric loop variable does not panic the compiler.

No shader in the repository or the examples is affected. Sources like `for i := 0; i >= 0; i++ {}` without a loop-exiting body stop compiling; they were never valid Kage and may deserve a release note.

### DIFF
--- a/internal/shader/shader_test.go
+++ b/internal/shader/shader_test.go
@@ -292,3 +292,45 @@ func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
 		t.Errorf("Compile must return an error for a huge constant shift, but got nil")
 	}
 }
+
+func TestCompileForLoopTermination(t *testing.T) {
+	testCases := []struct {
+		loop      string
+		wantError bool
+	}{
+		{"i := 0; i < 10; i += 1", false},
+		{"i := 10; i > 0; i -= 1", false},
+		{"i := 0; i != 10; i += 2", false},
+		{"i := 10; i != 0; i -= 2", false},
+		{"i := 10; i < 10; i += 1", false},
+		{"i := 0; i == 0; i += 1", false},
+		{"i := 0; i == 10; i += 1", false},
+		{"i := 0; i < 10; i += 0", true},
+		{"i := 0; i < 10; i -= 1", true},
+		{"i := 0; i <= 0; i -= 1", true},
+		{"i := 0; i > -10; i += 1", true},
+		{"i := 0; i >= 0; i += 1", true},
+		{"i := 0; i != 10; i += 3", true},
+		{"i := 10; i != 0; i -= 3", true},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.loop, func(t *testing.T) {
+			src := `package main
+
+func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
+	x := 0
+	for ` + tc.loop + ` {
+		x += 1
+	}
+	return vec4(1)
+}`
+			_, err := shader.Compile([]byte(src), "Vertex", "Fragment", 0)
+			if tc.wantError && err == nil {
+				t.Errorf("Compile must return an error, but got nil")
+			}
+			if !tc.wantError && err != nil {
+				t.Errorf("Compile must not return an error, but got %v", err)
+			}
+		})
+	}
+}
--- a/internal/shader/shader_test.go
+++ b/internal/shader/shader_test.go
@@ -370,7 +370,5 @@ func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
 	}
 	return vec4(1)
 }`
-	// A non-numeric loop variable cannot be analyzed, but this must not panic (Ebiten does not
-	// report the invalid source here yet).
-	shader.Compile([]byte(src), "Vertex", "Fragment", 0)
+	_, _ = shader.Compile([]byte(src), "Vertex", "Fragment", 0)
 }
--- a/internal/shader/shader_test.go
+++ b/internal/shader/shader_test.go
@@ -296,31 +296,55 @@ func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
 func TestCompileForLoopTermination(t *testing.T) {
 	testCases := []struct {
 		loop      string
+		body      string
 		wantError bool
 	}{
-		{"i := 0; i < 10; i += 1", false},
-		{"i := 10; i > 0; i -= 1", false},
-		{"i := 0; i != 10; i += 2", false},
-		{"i := 10; i != 0; i -= 2", false},
-		{"i := 10; i < 10; i += 1", false},
-		{"i := 0; i == 0; i += 1", false},
-		{"i := 0; i == 10; i += 1", false},
-		{"i := 0; i < 10; i += 0", true},
-		{"i := 0; i < 10; i -= 1", true},
-		{"i := 0; i <= 0; i -= 1", true},
-		{"i := 0; i > -10; i += 1", true},
-		{"i := 0; i >= 0; i += 1", true},
-		{"i := 0; i != 10; i += 3", true},
-		{"i := 10; i != 0; i -= 3", true},
+		{"i := 0; i < 10; i += 1", "", false},
+		{"i := 10; i > 0; i -= 1", "", false},
+		{"i := 0; i != 10; i += 2", "", false},
+		{"i := 10; i != 0; i -= 2", "", false},
+		{"i := 10; i < 10; i += 1", "", false},
+		{"i := 0; i == 0; i += 1", "", false},
+		{"i := 0; i == 10; i += 1", "", false},
+		{"i := 0; i < 10; i++", "", false},
+		{"i := 10; i > 0; i--", "", false},
+		{"i := 0.0; i < 10.0; i += 1.0", "", false},
+		{"i := 10.0; i > 0.0; i -= 1.0", "", false},
+		{"i := 0.0; i != 10.0; i += 2.0", "", false},
+		{"i := 0; i < 10; i += 0", "", true},
+		{"i := 0; i < 10; i -= 1", "", true},
+		{"i := 0; i <= 0; i -= 1", "", true},
+		{"i := 0; i > -10; i += 1", "", true},
+		{"i := 0; i >= 0; i += 1", "", true},
+		{"i := 0; i >= 0; i++", "", true},
+		{"i := 10; i <= 10; i--", "", true},
+		{"i := 0; i != 10; i += 3", "", true},
+		{"i := 10; i != 0; i -= 3", "", true},
+		{"i := 0.0; i < 10.0; i += 0.0", "", true},
+		{"i := 0.0; i < 10.0; i -= 1.0", "", true},
+		{"i := 0.5; i < 10.0; i -= 1.0", "", true},
+		{"i := 0.0; i < 10.0; i -= 0.5", "", true},
+		{"i := 0.0; i != 10.0; i -= 1.0", "", true},
+		{"i := 0.0; i != 10.0; i += 3.0", "", true},
+		{"i := 0; i >= 0; i++", "x += 1\n\t\tif x > 3 {\n\t\t\tbreak\n\t\t}", false},
+		{"i := 0; i >= 0; i += 1", "x += 1\n\t\treturn vec4(1)", false},
+		{"i := 0; i < 10; i += 0", "x += 1\n\t\tif x > 3 {\n\t\t\tbreak\n\t\t}", false},
+		{"i := 0; i >= 0; i++", "x += 1\n\t\tfor j := 0; j < 10; j++ {\n\t\t\tbreak\n\t\t}", true},
+		{"i := 0; i >= 0; i += 1", "x += 1\n\t\tfor j := 0; j < 10; j++ {\n\t\t\treturn vec4(1)\n\t\t}", false},
+		{"i := true; i == false; i += true", "", false},
 	}
 	for _, tc := range testCases {
 		t.Run(tc.loop, func(t *testing.T) {
+			body := tc.body
+			if body == "" {
+				body = "x += 1"
+			}
 			src := `package main
 
 func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
 	x := 0
 	for ` + tc.loop + ` {
-		x += 1
+		` + body + `
 	}
 	return vec4(1)
 }`
--- a/internal/shader/shader_test.go
+++ b/internal/shader/shader_test.go
@@ -328,13 +328,14 @@ func TestCompileForLoopTermination(t *testing.T) {
 		{"i := 0.0; i != 10.0; i += 3.0", "", true},
 		{"i := 0; i >= 0; i++", "x += 1\n\t\tif x > 3 {\n\t\t\tbreak\n\t\t}", false},
 		{"i := 0; i >= 0; i += 1", "x += 1\n\t\treturn vec4(1)", false},
+		{"i := 0; i >= 0; i++", "x += 1\n\t\tdiscard()", false},
+		{"i := 0; i >= 0; i += 1", "x += 1\n\t\tfor j := 0; j < 10; j++ {\n\t\t\tdiscard()\n\t\t}", false},
 		{"i := 0; i < 10; i += 0", "x += 1\n\t\tif x > 3 {\n\t\t\tbreak\n\t\t}", false},
 		{"i := 0; i >= 0; i++", "x += 1\n\t\tfor j := 0; j < 10; j++ {\n\t\t\tbreak\n\t\t}", true},
 		{"i := 0; i >= 0; i += 1", "x += 1\n\t\tfor j := 0; j < 10; j++ {\n\t\t\treturn vec4(1)\n\t\t}", false},
-		{"i := true; i == false; i += true", "", false},
 	}
-	for _, tc := range testCases {
-		t.Run(tc.loop, func(t *testing.T) {
+	for i, tc := range testCases {
+		t.Run(fmt.Sprintf("%02d-%s", i, tc.loop), func(t *testing.T) {
 			body := tc.body
 			if body == "" {
 				body = "x += 1"
@@ -357,4 +358,19 @@ func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
 			}
 		})
 	}
+}
+
+func TestCompileForLoopNonNumericConstant(t *testing.T) {
+	src := `package main
+
+func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
+	x := 0
+	for i := true; i == false; i += true {
+		x += 1
+	}
+	return vec4(1)
+}`
+	// A non-numeric loop variable cannot be analyzed, but this must not panic (Ebiten does not
+	// report the invalid source here yet).
+	shader.Compile([]byte(src), "Vertex", "Fragment", 0)
 }
--- a/internal/shader/stmt.go
+++ b/internal/shader/stmt.go
@@ -901,14 +901,6 @@ func (cs *compileState) parseFor(block *block, fname string, stmt *ast.ForStmt, 
 		cs.addError(stmt.Pos(), "for-statement's post statement must have one of these operators: +=, -=, ++, --")
 		return nil, false
 	}
-	if gconstant.Sign(delta) == 0 {
-		cs.addError(stmt.Pos(), "for-statement's post statement must change the loop variable, but the delta is 0")
-		return nil, false
-	}
-	if forLoopNeverTerminates(op, init, end, delta) {
-		cs.addError(stmt.Pos(), "for-statement's condition never becomes false")
-		return nil, false
-	}
 
 	b, ok := cs.parseBlock(pseudoBlock, fname, []ast.Stmt{stmt.Body}, inParams, outParams, returnType, true)
 	if !ok {
@@ -917,6 +909,13 @@ func (cs *compileState) parseFor(block *block, fname string, stmt *ast.ForStmt, 
 	bodyir := b.ir
 	for len(bodyir.Stmts) == 1 && bodyir.Stmts[0].Type == shaderir.BlockStmt {
 		bodyir = bodyir.Stmts[0].Blocks[0]
+	}
+
+	// A loop whose condition never becomes false can still end by a break- or a return-statement in
+	// its body.
+	if !blockExitsLoop(bodyir, false) && forLoopConditionAlwaysTrue(op, init, end, delta) {
+		cs.addError(stmt.Pos(), "for-statement's condition never becomes false")
+		return nil, false
 	}
 
 	// As the pseudo block is not actually used, copy the variable part to the actual block.
@@ -944,11 +943,13 @@ func (cs *compileState) parseFor(block *block, fname string, stmt *ast.ForStmt, 
 	}, true
 }
 
-func forLoopNeverTerminates(op shaderir.Op, init, end, delta gconstant.Value) bool {
-	init = gconstant.ToInt(init)
-	end = gconstant.ToInt(end)
-	delta = gconstant.ToInt(delta)
-	if init.Kind() == gconstant.Unknown || end.Kind() == gconstant.Unknown || delta.Kind() == gconstant.Unknown {
+// forLoopConditionAlwaysTrue reports whether the condition of a for-loop never becomes false.
+// The initial value, the end value and the delta are constants, so this can be decided at compile time.
+// The loop can still end by a break- or a return-statement in its body.
+func forLoopConditionAlwaysTrue(op shaderir.Op, init, end, delta gconstant.Value) bool {
+	// A non-numeric constant cannot be analyzed here. Such a constant is invalid, but reporting it is
+	// another issue.
+	if !isNumeric(init) || !isNumeric(end) || !isNumeric(delta) {
 		return false
 	}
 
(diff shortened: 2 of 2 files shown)


internal/shader/shader_test.go
```diff
@@ -230,3 +230,20 @@ func Fragment(dstPos vec4, src0Pos vec2, color vec4) vec4 {
...
 	}
 }
+
+// TestCompileForDeltaZero confirms that a zero-delta for-loop is rejected.
+func TestCompileForDeltaZero(t *testing.T) {
+	src := `package main
+
+func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
+	x := 0
+	for i := 0; i < 10; i += 0 {
```
hajimehoshi: What would happen if

```
for i := 0; i < 10; i -= 1 {
```

kumagi: That will goes bad, so added `forLoopNeverTerminates`

internal/shader/shader_test.go
```diff
@@ -230,3 +230,66 @@ func Fragment(dstPos vec4, src0Pos vec2, color vec4) vec4 {
...
```
hajimehoshi: This test now can be integrated with TestCompileForLoopTermination

internal/shader/shader_test.go
```diff
@@ -230,3 +230,66 @@ func Fragment(dstPos vec4, src0Pos vec2, color vec4) vec4 {
...
+}`
+	_, err := shader.Compile([]byte(src), "Vertex", "Fragment", 0)
+	if err == nil {
+		t.Errorf("Compile must return an error for a zero-delta for-loop, but got nil")
+	}
+}
+
+// TestCompileForLoopTermination confirms that a for-loop whose condition never
+// becomes false is rejected. As the initial value, the end value, and the delta
+// are all constants, termination can be decided at compile time.
```
hajimehoshi: Remove comments

hajimehoshi (commented):
Resolve conflicts

internal/shader/shader_test.go
kumagi: OK, now integrated.

internal/shader/shader_test.go

hajimehoshi:
Thanks for the PR. The direction looks right to me: a Kage for-loop must be statically bounded, so a condition that can never become false is invalid regardless of the backend, and catching it at compile time is much better than hanging the GPU.

One thing needs fixing before this can go in, plus a few smaller points.

### `gconstant.Sign` panics on a non-numeric delta

`gconstant.Sign` requires a numeric value, but `delta` is not guaranteed to be one at that point. With this patch applied, this source crashes the compiler:

```go
for i := true; i == false; i += true {
	x += 1
}
```

```
panic: true not numeric
go/constant.Sign(...)
	go/constant/value.go:708
internal/shader.(*compileState).parseFor(...)
	internal/shader/stmt.go:904
```

`+=` between bools is not rejected earlier (a separate, pre-existing hole: this source compiles with no error on main today), so the new check turns silently-accepted input into a hard panic out of `ebiten.NewShader`. Checking `delta.Kind()`, or that the loop variable's type is `Int` or `Float`, before calling `Sign` would cover it. `forLoopNeverTerminates` itself is safe here, since `ToInt` yields `Unknown`.

### Float loops are handled inconsistently

Only the `!=` reachability test needs integers, but all three values are converted with `ToInt` up front and the function bails out if any of them is non-integral. As a result these three loops, which are equally non-terminating, are treated differently:

```go
for i := 0.0; i < 10.0; i -= 1.0 { … }   // rejected
for i := 0.5; i < 10.0; i -= 1.0 { … }   // accepted
for i := 0.0; i < 10.0; i -= 0.5 { … }   // accepted
```

Doing the sign check and the `Compare` calls on the original constants, and converting to `int` only for the `!=` remainder test, would close that gap.

### Dead code

Since the caller rejects `delta == 0` before calling, `case shaderir.EqualOp: return d == 0 && ...` can never return true, and the `if d == 0` branch of the `NotEqualOp` case is unreachable. Either fold the zero-delta case into the analysis, or drop those two branches.

### Minor

* The title and description are stale: they describe only the zero-delta rejection and mention a test named `TestCompileForDeltaZero`, but the change is a full termination analysis and the test is `TestCompileForLoopTermination`. The commit message itself is accurate.
* The tests cover only int loop variables and `+=` / `-=`. Float loop variables and the `++` / `--` forms would be worth adding.

The rest of the test suite passes, and no shader in the repository or in the examples is affected by the new checks. Sources like `for i := 0; i >= 0; i++ { … break }` do stop compiling, which is the intended fix since they were never valid Kage, but it may be worth a release note.

---

kumagi:
Pushed `00e92d43b` addressing all the review points:

- `gconstant.Sign` is now guarded by a numeric-kind check, so a non-numeric delta (e.g. `for i := true; i == false; i += true`) no longer panics the compiler. It just skips the termination analysis; the type error itself remains a pre-existing hole.
- The termination analysis now runs on the original constants, converting to integers only for the `!=` remainder test. Float loops like `for i := 0.5; i < 10.0; i -= 1.0` and `for i := 0.0; i < 10.0; i -= 0.5` are now rejected, consistent with their int counterparts.
- Removed the dead `EqualOp` case and the dead zero-delta branch of `NotEqualOp` since the caller already rejects a zero delta.
- Extended `TestCompileForLoopTermination` with float loop variables, the `++`/`--` forms, a false-condition case that must not be reported, and the bool-delta no-panic case.

The PR title/description is still stale (it mentions `TestCompileForDeltaZero`) — will update.

hajimehoshi:
Thanks for the PR. I applied it locally and ran into two problems that I think need to be resolved before this can go in.

### 1. `gconstant.Sign` panics on a non-numeric delta

`gconstant.Sign` requires a numeric value, but `delta` is not guaranteed to be one at that point. With this patch applied, this source crashes the compiler:

```go
for i := true; i == false; i += true {
	x += 1
}
```

```
panic: true not numeric
go/constant.Sign(...)
	go/constant/value.go:708
internal/shader.(*compileState).parseFor(...)
	internal/shader/stmt.go:904
```

`+=` between bools is not rejected earlier (a separate, pre-existing hole: this source compiles with no error on main today), so the new check turns silently-accepted input into a hard panic out of `ebiten.NewShader`. Checking `delta.Kind()`, or that the loop variable's type is `Int` or `Float`, before calling `Sign` would cover it. `forLoopNeverTerminates` itself is safe here, since `ToInt` yields `Unknown`.

### 2. Loops that do terminate via `break` or `return` are rejected

Kage supports `break`, `continue` and `return` inside a for-loop, and the GLSL target is ES 3.00, which has no statically-bounded-loop requirement. So the analysis decides "the condition never becomes false", which is strictly weaker than "the loop never terminates". All three of these compile on main and fail with this PR:

```go
for i := 0; i >= 0; i++ {   // "for-statement's condition never becomes false"
	x += 1
	if x > 3 {
		break
	}
}
```

```go
for i := 0; i >= 0; i++ {   // "for-statement's condition never becomes false"
	return vec4(0)
}
```

```go
for i := 0; i < 10; i += 0 {   // "...must change the loop variable, but the delta is 0"
	x += 1
	if x > 3 {
		break
	}
}
```

The third one is the one I would worry about most: `for { ... }` is not expressible in Kage, since the init, condition and post statements are all mandatory, so a zero delta plus a `break` is the only way to write a loop-until-break. That means the zero-delta rule needs to be body-aware too, not just the termination analysis.

Concretely, both checks would have to move below the `cs.parseBlock` call for the body, and be skipped when the body contains a `break` or a `return`.

### Minor points

* Dead code: since the caller rejects `delta == 0` before calling, `case shaderir.EqualOp: return d == 0 && ...` can never return true, and the `if d == 0` branch of the `NotEqualOp` case is unreachable.
* Float loops are handled inconsistently. Only the `!=` remainder test needs integers, but all three values are converted with `ToInt` up front and the function bails if any of them is non-integral. As a result `for i := 0.0; i < 10.0; i -= 1.0` is rejected, while the equally infinite `for i := 0.5; i < 10.0; i -= 1.0` and `for i := 0.0; i < 10.0; i -= 0.5` are accepted.
* `forLoopNeverTerminates` overstates what the function computes. Something like `forLoopConditionAlwaysTrue` would be accurate, and would make the `break`/`return` gap visible at the call site.
* The title and description are stale: they describe only the zero-delta rejection and mention a test named `TestCompileForDeltaZero`, but the change is a full termination analysis and the test is `TestCompileForLoopTermination`. The commit message itself is accurate.
* The tests do not cover float loop variables, `++`/`--` post statements, or bodies containing `break` or `return`.

For what it is worth, the rest of the suite passes, and no shader in the repository or in the examples is affected by the new checks.

---

kumagi:
Thanks for the careful testing. Pushed `9a2bb87b5`, which addresses both problems.

**1. `gconstant.Sign` panic** — `forLoopConditionAlwaysTrue` now returns early unless `init`, `end` and `delta` are all `Int` or `Float` constants, so `for i := true; i == false; i += true` no longer panics (`Compare` would have panicked there too, on `true < false`, which is why the guard covers all three). A test case asserts it compiles without panicking; the underlying `+=`-on-bool hole is left untouched.

**2. Loops that end via `break` or `return`** — both checks moved below `cs.parseBlock` for the body, and are skipped when `blockExitsLoop` finds a `break` or a `return` in it. All three of your sources compile again. Details:
* The zero-delta rule is now folded into the analysis instead of being a separate pre-check (that also removes the dead `EqualOp` / `NotEqualOp` branches, since they are reachable again), so `for i := 0; i < 10; i += 0 { … break }` is the loop-until-break it should be. The error message is the single "for-statement's condition never becomes false".
* A `break` inside a nested for-loop is not counted, since it exits only the nested loop — `for i := 0; i >= 0; i++ { for j := 0; j < 10; j++ { break } }` is still rejected, while a `return` inside a nested loop is counted, since it does leave the function.
* `continue` is not counted.

**Floats** — the sign check and the `Compare` calls now run on the original constants, and `ToInt` is used only for the `!=` remainder test (where a non-integral value is skipped rather than rejected). `for i := 0.5; i < 10.0; i -= 1.0` and `for i := 0.0; i < 10.0; i -= 0.5` are now rejected like their integer counterparts.

**Renamed** to `forLoopConditionAlwaysTrue`, and the title/description are updated.

Tests gained float loop variables, `++` / `--`, the `break` / `return` bodies (including the nested-loop cases above) and the bool-delta case.

One remaining gap, which I did not want to fix silently: a body that assigns to the loop variable (`for i := 0; i <= 0; i -= 1 { i = 1 }`) is accepted on main and can also make an "always true" condition false. Should such an assignment be an error, or should the analysis also be skipped for it?

No shader in the repository or the examples is affected.

hajimehoshi:
Thanks for the update! Both blockers are fixed. I re-checked at `9a2bb87b5`.

**The panic is gone.** `ToInt` is now applied to `diff` and `step` before the `REM`. All eight loops that panicked before come back clean, and a matrix of 6600 combinations (10 init values × 6 operators × 10 end values × 11 post statements) produces no panic. The two new `i := 0.0; i != 10.0; i += 2.0` / `+= 3.0` rows cover the regression.

**`break` / `return` are honored.** `blockExitsLoop` held up in every shape I tried, with `for i := 0; i >= 0; i++` as the header:

| body | result |
| --- | --- |
| `if x > 3.0 { break }` | accepted |
| `{ break }` (bare block) | accepted |
| `if … { … } else { break }` | accepted |
| `for j := 0; j < 10; j++ { break }` | rejected (inner break) |
| `if … { for j … { break } }` | rejected |
| `for j … { if … { break } }` | rejected |
| `for j := range 3 { break }` | rejected |
| `for j … { return vec4(1) }` | accepted |

The `inForLoop` threading through the `default:` branch is right, and since Kage has neither `switch` nor labeled `break`, there is no case where a `break` means anything other than "exit the innermost loop".

Folding the zero-delta case into the analysis is also an improvement over the previous revision: `for i := 10; i < 10; i += 0` is now correctly accepted, since the loop never runs.

## One new finding: `discard()` is not counted as an exit

`blockExitsLoop` handles `Break` and `Return` but not `shaderir.Discard`, even though all three backends emit `discard` followed by an unconditional return:

- https://github.com/hajimehoshi/ebiten/blob/9a2bb87b5/internal/shaderir/glsl/glsl.go#L616-L618
- https://github.com/hajimehoshi/ebiten/blob/9a2bb87b5/internal/shaderir/hlsl/hlsl.go#L619-L621
- https://github.com/hajimehoshi/ebiten/blob/9a2bb87b5/internal/shaderir/msl/msl.go#L516-L518

So this is rejected even though it always exits:

```go
for i := 0; i >= 0; i++ {
	x += 1.0
	discard()
}
```

```
5:2: for-statement's condition never becomes false
```

`discard()` inside a bounded loop compiles today, so the pattern is legal. Adding `case shaderir.Discard:` next to `case shaderir.Return:` closes it.

## Smaller points

- The PR description is now stale. It still says "Sources like `for i := 0; i >= 0; i++ { … break }` stop compiling" and "The post statement's delta must be a non-zero numeric constant", both of which the second commit reversed. Worth updating before merge if the description feeds the squash-merge message.
- Five subtest names collide: `i := 0; i >= 0; i++`, `i := 0; i >= 0; i += 1` and `i := 0; i < 10; i += 0` each appear two or three times with different bodies, so a failure shows up as `#01` / `#02` with no hint which body it was. Folding the body (or an index) into the `t.Run` name would help.
- `{"i := true; i == false; i += true", "", false}` still asserts that a bool for-loop *compiles*, which pins a type-checking hole down as expected behavior. The subtest already fails on a panic, so the `wantError` assertion could be dropped for that row.
- Nit: `!blockExitsLoop(bodyir, false) && forLoopConditionAlwaysTrue(…)` walks the whole body tree for every for-loop. Swapping the operands would let the cheap constant analysis gate the walk, and reads more naturally too.
- Nit: `isNumeric` sits next to `isFloat` and `isInt` in `type.go`, which take `(shaderir.Expr, shaderir.Type)`. `isNumericConstant` would be less ambiguous.
- Unchanged limitation, fine to leave: `for i := 0.5; i != 10.0; i += 1.0` is infinite but accepted, since `ToInt(9.5)` bails out. Testing whether `diff / step` is an integer on rationals would close it.

`gofmt`, `go vet`, `go build ./...` and `go test ./internal/shader/... ./internal/shaderir/...` are all clean on the branch.

---

kumagi:
Thanks for the re-review! Pushed c75b356dc addressing the remaining points.

- **`discard()` is now counted as an exit.** `blockExitsLoop` treats `shaderir.Discard` like `shaderir.Return`, since all three backends return from the fragment function right after a discard. `for i := 0; i >= 0; i++ { x += 1.0; discard() }` compiles, and the tests cover both a direct `discard()` and one inside a nested loop.
- **Subtest names are now unique.** Each subtest is prefixed with its index (`t.Run(fmt.Sprintf("%02d-%s", i, tc.loop), ...)`), so duplicated loop headers with different bodies are distinguishable on failure.
- **The bool for-loop no longer pins down a type-checking hole.** It is moved out of the table into `TestCompileForLoopNonNumericConstant`, which only requires that the compiler does not panic.
- **Operand order swapped** to `forLoopConditionAlwaysTrue(...) && !blockExitsLoop(...)`, so the cheap constant analysis gates the body walk.
- **`isNumeric` renamed to `isNumericConstant`** to avoid confusion with `isFloat` / `isInt` in type.go.
- **PR description updated** — it no longer says that loop-exiting bodies stop compiling or that the delta must be a non-zero numeric constant.

The float `!=` limitation (`for i := 0.5; i != 10.0; i += 1.0` is accepted because `ToInt(9.5)` bails out) is left as-is per your note; I can file a follow-up if you'd like.

`gofmt`, `go vet`, `go build ./...` and `go test ./internal/shader/... ./internal/shaderir/...` are clean on the branch.

internal/shader/shader_test.go
```diff
@@ -270,3 +270,85 @@ func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
...
+	src := `package main
+
+func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
+	x := 0
+	for i := true; i == false; i += true {
+		x += 1
+	}
+	return vec4(1)
+}`
+	// A non-numeric loop variable cannot be analyzed, but this must not panic (Ebiten does not
```
hajimehoshi: Nit: "Ebitengine" rather than "Ebiten". Comments under `internal/` use "Ebitengine" 96 times; the only bare "Ebiten" is one pre-existing line in `internal/testing/testing.go`.

internal/shader/shader_test.go
```diff
@@ -270,3 +270,85 @@ func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
...
+	// report the invalid source here yet).
+	shader.Compile([]byte(src), "Vertex", "Fragment", 0)
```
hajimehoshi: This fails the `go vet (vettool)` CI job. The repo's vettool bundles errcheck (`internal/vettools/main.go`) and CI runs it over `./...`, and there is no `.errcheck_excludes` file to suppress it:

```
$ go vet -vettool=$(which vettools) ./internal/shader/...
internal/shader/shader_test.go:353:16: unchecked error
```

`main` is clean, so this is new. `_, _ = shader.Compile(...)` fixes it and matches the convention used elsewhere in the repo.

hajimehoshi (commented):
This revision looks right to me. I re-verified the `discard()` handling across five placements (direct, in an `if`, in a nested `for`, in a bare block, and in an `if` inside a nested `for`), the operand swap, the rename, the numbered sub-tests and the moved bool case. Two small things inline.

---

internal/shader/shader_test.go
kumagi: removed the comment as looks not so useful

kumagi:
Thanks for the review! Pushed `9f365d6a7` addressing the two inline nits:

- Removed the comment above `shader.Compile` in `TestCompileForLoopNonNumericConstant`.
- Changed the call to `_, _ = shader.Compile(...)` so the `go vet (vettool)` CI job with errcheck passes.

`gofmt`, `go vet`, `go vet -vettool=$(which vettools) ./...` on the touched packages and `go test ./internal/shader/...` are all clean on the branch.

hajimehoshi:
The force-push looks like it went wrong: `34b29eb36` is the branch as it stood before my first review, so the changes from the two rounds after that are gone.

Restricted to the two files of this PR, `34b29eb36` is identical to the old `00e92d43b`, apart from two comments that the older commit had and this one does not:

```
$ git diff 00e92d43b 34b29eb36 -- internal/shader/stmt.go internal/shader/shader_test.go
-	// gconstant.Sign requires a numeric value. A non-numeric delta (e.g. a bool, which is not
-	// rejected yet at this point) must not crash the compiler.
-		// Only the remainder test below requires integers; bail out on other types.
```

I do not think this was intentional, because the PR description was not reverted: it still describes `forLoopConditionAlwaysTrue`, `blockExitsLoop`, `discard()` and `TestCompileForLoopNonNumericConstant`, none of which exist in the code any more. The commit message also lost its formatting; the four-bullet operator list is now a single run-on paragraph.

## What is missing from `stmt.go`

- `blockExitsLoop`, the whole function, along with the body-aware call site `forLoopConditionAlwaysTrue(…) && !blockExitsLoop(bodyir, false)`.
- In the `!=` branch, `diff = gconstant.ToInt(diff)` / `step = gconstant.ToInt(step)` with the `Unknown` bail-out. Without it the `REM` runs against an unconverted `delta`.
- `case shaderir.EqualOp: return d == 0 && gconstant.Compare(init, token.EQL, end)`, and the `d == 0` branch of `!=`.
- The zero delta being one case of the analysis rather than a separate error, so `for i := 10; i < 10; i += 0 {}` is rejected again even though it never runs.
- `isNumericConstant`, which also checked `delta`, and the doc comments on both functions.

## What is missing from `shader_test.go`

- The `body` field and all seven rows that exercise it: `break`, `return`, a direct `discard()`, a `discard()` in a nested loop, zero delta with a `break`, a nested `break`, and a nested `return`.
- The three float `!=` rows that covered the panic: `i += 2.0`, `i -= 1.0` and `i += 3.0`.
- `TestCompileForLoopNonNumericConstant`; the bool case is back in the main table asserting that it compiles.
- The numbered sub-test names.

## Both earlier problems are back

The panic:

```go
//kage:unit pixels

package main

func Fragment(dstPos vec4, srcPos vec2, color vec4) vec4 {
	var s float
	for i := 0.0; i != 10.0; i++ {
		s += i
	}
	return vec4(s, 0, 0, 1)
}
```

```
panic: invalid binary operation 10 % 1
```

And the loop-until-`break` rejection:

```go
for i := 0; i >= 0; i++ {
	x += 1.0
	if x > 5.0 {
		break
	}
}
```

```
5:2: for-statement's condition never becomes false
```

The good revision is `c75b356dcfb544b1853548fb2664ec8a6ed34ca8`. GitHub keeps force-pushed commits reachable by SHA for a while, so it should still be possible to recover it and rebase it onto the current main. Reviewing the code as it stands now would only repeat my first two comments, so I will wait until the branch is restored.

---

kumagi:
Thanks for catching this — you were right, the force-push dropped the two review rounds. The branch is restored now: the commits from `5d2c6370c..9f365d6a7` are back and rebased onto the current main (`7da6fd5a6`), pushed as `34c67e401`.

- `a0049c961` internal/shader: reject for-loops whose condition never becomes false
- `7951ba405` internal/shader: make the for-loop condition analysis body-aware
- `a660ba197` internal/shader: treat discard() as an exit of a for-loop
- `34c67e401` internal/shader: fix unchecked error and remove a comment in the for-loop test

`git range-diff e887e5695..9f365d6a7 7da6fd5a6..34c67e401` reports all four commits as unchanged, so the body-aware call site, `blockExitsLoop` (with `shaderir.Discard`), the `ToInt` bail-out in the `!=` branch, the zero-delta case folded into `forLoopConditionAlwaysTrue`, `isNumericConstant`, the doc comments, the numbered sub-tests and `TestCompileForLoopNonNumericConstant` are all back, as is the four-bullet formatting of the first commit message. The only difference against the good revision is upstream's `TestCompileVaryingTypeMismatchPosition` coming in with the rebase.

Both regressions you listed are gone on the new head:

- `for i := 0.0; i != 10.0; i++ { s += i }` compiles (no `invalid binary operation 10 % 1`).
- `for i := 0; i >= 0; i++ { x += 1.0; if x > 5.0 { break } }` compiles.
- `for i := 10; i < 10; i += 0 {}` and `for i := 0; i >= 0; i++ { discard() }` compile, while `for i := 0; i < 10; i += 0 {}` is still rejected.

`gofmt`, `go vet`, `go vet -vettool=$(which vettools) ./internal/shader/...`, `go build ./...` and `go test ./internal/shader/... ./internal/shaderir/...` are clean.

The float `!=` gap (`for i := 0.5; i != 10.0; i += 1.0` is accepted) is left as we discussed — say the word and I'll file a follow-up.

hajimehoshi:
Thank you for this, and sorry to land on this conclusion after three rounds of iteration. I want to be clear that this is a sequencing decision, not a judgement on the change itself.

I evaluated `c75b356dc`, the body-aware revision, not the current head. The recent force-push reset the branch to its pre-review state, so the head no longer contains `blockExitsLoop` or the `ToInt` fix in the `!=` branch. Everything below is about the revision you actually arrived at.

## Why not now

[#3564](https://github.com/hajimehoshi/ebiten/issues/3564) is milestoned for v2.11 and relaxes the for-loop grammar. Its third item is exactly that an unbounded loop exited with `break` cannot be written today, because the init, condition and post statements are all mandatory and must be constants. Once that lands:

* The analysis can only fire on the constant-header form, which becomes the legacy form. A loop with a uniform bound, or a bare `for`, has nothing decidable in its header.
* The guarantee shrinks accordingly. After #3564 a `for { }` or a uniform-bounded loop can still hang the GPU with no compile-time check, so this stops being "Kage rejects loops that never terminate" and becomes "Kage catches one class of typo in one loop form", for a form that is on its way out.
* #3564 already lists this work: the analysis "would have to be limited to the cases where they still are [constants]". Merging now means writing it, then reworking it inside the loop redesign.

#3564 has to touch `parseFor` regardless, so doing this once against the final loop grammar is cheaper than doing it twice. I would rather carry the analysis into that issue than merge it and immediately rewrite it.

## On the compatibility question

For the record, since it came up during review and you fixed it: `c75b356dc` does not break working shaders. With `blockExitsLoop`, the only sources that stop compiling are loops with no `break`, no `return` and no `discard()` at all, which are infinite loops on the GPU today. The one hole worth checking, a `discard()` reached through a helper function, does not exist, because `discard is available only in Fragment`. So the compile-compatibility problem raised earlier was genuinely solved, and it is not the reason for this decision.

## What is worth keeping

The analysis is sound. The operator-by-operator reasoning, the exact rational arithmetic on the constants, and the `break` / `return` / `discard()` exit walk are all things #3564 will want. That issue already references this PR, so the work stays findable. If you would like to pick up #3564 itself, that would be very welcome.

Closing in favour of #3564.

---

PR #3551, opened 2026-08-28
internal/shader: reject a constant shift count that is too large

kumagi:
None

bug

## What this PR does | solves
Folding a constant shift like 1<<(1<<40) passes a shift count of 2^40 to go/constant.Shift, which allocates a number of that many bits and exhausts memory, hanging or crashing the compiler. A constant shift by more than 1<<16 bits is meaningless for any shader integer type, so report an error for such counts instead of folding them.

Add TestCompileHugeShift, which hangs without this fix.

### DIFF
--- a/internal/shader/shader_test.go
+++ b/internal/shader/shader_test.go
@@ -230,3 +230,18 @@ func Fragment(dstPos vec4, src0Pos vec2, color vec4) vec4 {
 		t.Errorf("HLSL PSMain should not use the '%%' operator for integer modulo, but got:\n%s", body)
 	}
 }
+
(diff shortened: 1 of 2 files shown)


internal/shader/shader_test.go
```diff
@@ -230,3 +230,17 @@ func Fragment(dstPos vec4, src0Pos vec2, color vec4) vec4 {
 		t.Errorf("HLSL PSMain should not use the '%%' operator for integer modulo, but got:\n%s", body)
 	}
 }
+
+// TestCompileHugeShift confirms that an enormous constant shift is rejected.
```

hajimehoshi (approved):
LGTM

PR #3552, opened 2026-08-28
internal/shader: point varying-signature errors at the entry point instead of line 1

kumagi:
None

bug

## What this PR does | solves

The errors about a mismatched fragment argument, too many fragment arguments, and the entry-point return types used addError(0, ...), so every message was reported at 1:1 regardless of the actual code. Record the positions of the vertex and fragment entry points and report these errors at the function they belong to.

### DIFF
--- a/internal/shader/shader_test.go
+++ b/internal/shader/shader_test.go
@@ -231,9 +231,8 @@ func Fragment(dstPos vec4, src0Pos vec2, color vec4) vec4 {
 	}
 }
 
-// TestCompileVaryingTypeMismatchPosition confirms that an error about a
-// mismatched fragment argument points at the fragment entry point, not at
-// line 1.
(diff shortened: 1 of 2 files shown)


hajimehoshi (approved):
LGTM

internal/shader/shader_test.go
```diff
@@ -230,3 +230,27 @@ func Fragment(dstPos vec4, src0Pos vec2, color vec4) vec4 {
 		t.Errorf("HLSL PSMain should not use the '%%' operator for integer modulo, but got:\n%s", body)
 	}
 }
+
+// TestCompileVaryingTypeMismatchPosition confirms that the error position
+// points at the fragment entry point, not at line 1.
```

kumagi: done.

hajimehoshi:
Resolve conflicts

PR #3553, opened 2026-08-28
internal/shader: report an error instead of panicking on an assignment from a no-return function

kumagi:
None

bug

## What this PR does | solves
Assigning or declaring a variable from a call to a function that returns nothing, like 'x := bar()' or 'var x = bar()', read ts[0] of the function's empty return-type list and panicked with an index out of range. The compiler must reject the misuse with an error instead.
    
Check for an empty return-type list in both the assignment (stmt.go) and the variable declaration (shader.go) paths, mirroring the existing single/multiple-value count checks.

### DIFF
--- a/internal/shader/shader_test.go
+++ b/internal/shader/shader_test.go
@@ -200,6 +200,37 @@ func TestCompile(t *testing.T) {
 	}
 }
 
+// TestCompileAssignFromNoReturnValue confirms that assigning the result of a
+// function that returns nothing reports an error instead of panicking (#B1).
+func TestCompileAssignFromNoReturnValue(t *testing.T) {
+	srcs := []string{
+		`package main
+
+func bar() {
+}
+
+func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
+	x := bar()
+	return vec4(1)
+}`,
+		`package main
+
+func bar() {
+}
+
+func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
+	var x = bar()
+	return vec4(1)
+}`,
+	}
+	for _, src := range srcs {
+		_, err := shader.Compile([]byte(src), "Vertex", "Fragment", 0)
+		if err == nil {
+			t.Errorf("Compile must return an error for a function call with no return values, but got nil")
+		}
+	}
+}
+
 // TestCompileHLSLIntModulo confirms that integer modulo is emitted via the modInt
 // helper rather than the '%' operator.
 func TestCompileHLSLIntModulo(t *testing.T) {
--- a/internal/shader/shader_test.go
+++ b/internal/shader/shader_test.go
@@ -200,8 +200,6 @@ func TestCompile(t *testing.T) {
 	}
 }
 
-// TestCompileAssignFromNoReturnValue confirms that assigning the result of a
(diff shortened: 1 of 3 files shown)


internal/shader/shader_test.go
```diff
@@ -200,6 +200,37 @@ func TestCompile(t *testing.T) {
...
+	for _, src := range srcs {
+		_, err := shader.Compile([]byte(src), "Vertex", "Fragment", 0)
+		if err == nil {
+			t.Errorf("Compile must return an error for a function call with no return values, but got nil")
+		}
+	}
+}
+
 // TestCompileHLSLIntModulo confirms that integer modulo is emitted via the modInt
 // helper rather than the '%' operator.
```
hajimehoshi: We should have removed this comment.

Optional: remove this comment

internal/shader/shader_test.go
```diff
@@ -200,6 +200,37 @@ func TestCompile(t *testing.T) {
 	}
 }
 
+// TestCompileAssignFromNoReturnValue confirms that assigning the result of a
+// function that returns nothing reports an error instead of panicking.
```

internal/shader/stmt.go
```diff
@@ -531,15 +531,19 @@ func (cs *compileState) assign(block *block, fname string, pos token.Pos, lhs, r
...
-				ts, ok := cs.functionReturnTypes(block, rhs[i])
-				if !ok {
-					ts = rts
-				}
-				if len(ts) > 1 {
-					cs.addError(pos, "single-value context and multiple-value context cannot be mixed")
-					return nil, false
-				}
-				t := ts[0]
+ts, ok := cs.functionReturnTypes(block, rhs[i])
```
hajimehoshi: ?

internal/ui/input.go
```diff
@@ -171,6 +171,18 @@ func (i *InputState) IsKeyJustPressed(key Key, tick int64) bool {
...
+		case KeyAlt:
+			return i.IsKeyJustPressed(KeyAltLeft, tick) || i.IsKeyJustPressed(KeyAltRight, tick)
+		case KeyControl:
+			return i.IsKeyJustPressed(KeyControlLeft, tick) || i.IsKeyJustPressed(KeyControlRight, tick)
+		case KeyShift:
+			return i.IsKeyJustPressed(KeyShiftLeft, tick) || i.IsKeyJustPressed(KeyShiftRight, tick)
+		case KeyMeta:
+			return i.IsKeyJustPressed(KeyMetaLeft, tick) || i.IsKeyJustPressed(KeyMetaRight, tick)
+		}
+	}
```
hajimehoshi: ?? You mixed unnecessary changes

internal/shader/shader_test.go

internal/shader/shader_test.go
kumagi: removed

internal/shader/stmt.go
kumagi: ran gofmt.

internal/ui/input.go
kumagi: sorry, reverted.

hajimehoshi (approved):
LGTM

PR #3554, opened 2026-08-28
audio: clear the partial-frame remainder on InfiniteLoop.Seek

kumagi:
None

bug

## What this PR does | solves
InfiniteLoop.Read keeps the remainder of a source read that is not a
multiple of the bit depth in the extra buffer and prepends it to the next
read. Seek moved the source position but did not clear extra, so a seek
left the remainder from the old position to be prepended to the data read
from the new position, corrupting the stream.

Clear extra in Seek. Add TestInfiniteLoopSeekClearsExtra, which observes
the stale bytes without this fix.

### DIFF
--- a/audio/loop.go
+++ b/audio/loop.go
@@ -262,7 +262,7 @@ func (i *InfiniteLoop) Seek(offset int64, whence int) (int64, error) {
 	case io.SeekStart:
 		next = offset
 	case io.SeekCurrent:
-		next = i.pos + offset
+		next = i.pos - int64(len(i.extra)) + offset
 	case io.SeekEnd:
 		return 0, fmt.Errorf("audio: whence must be io.SeekStart or io.SeekCurrent for InfiniteLoop")
 	}
@@ -278,5 +278,6 @@ func (i *InfiniteLoop) Seek(offset int64, whence int) (int64, error) {
 		return 0, err
 	}
 	i.pos = next
+	i.extra = i.extra[:0]
 	return i.pos, nil
 }
--- a/audio/loop_test.go
+++ b/audio/loop_test.go
@@ -433,3 +433,65 @@ func TestInfiniteLoopBlendWithPartialFrameReads(t *testing.T) {
 		})
 	}
 }
+
+func TestInfiniteLoopSeekClearsExtra(t *testing.T) {
+	// The source returns 5 bytes at most, which is larger than any bit depth but not a multiple of
+	// any, so a read returns a complete value and leaves a remainder.
+	src := &partialFrameReader{
+		src:  bytes.NewReader(bytes.Repeat([]byte{1, 2, 3, 4, 5, 6, 7, 8}, 3)),
+		size: 5,
+	}
+	l := audio.NewInfiniteLoopF32(src, 8)
+
+	buf := make([]byte, 32)
+	if _, err := l.Read(buf); err != nil {
+		t.Fatal(err)
+	}
+
+	if _, err := l.Seek(0, io.SeekStart); err != nil {
+		t.Fatal(err)
+	}
+	n, err := l.Read(buf)
+	if err != nil {
+		t.Fatal(err)
+	}
+	want := []byte{1, 2, 3, 4}
+	if !bytes.Equal(buf[:n], want) {
+		t.Errorf("got: %v, want: %v", buf[:n], want)
+	}
+}
+
+func TestInfiniteLoopSeekCurrentAfterPartialFrameRead(t *testing.T) {
+	// The source returns 5 bytes at most, which is larger than any bit depth but not a multiple of
+	// any, so a read returns a complete value and leaves a remainder.
+	src := &partialFrameReader{
+		src:  bytes.NewReader([]byte{1, 2, 3, 4, 5, 6, 7, 8}),
+		size: 5,
+	}
+	l := audio.NewInfiniteLoopF32(src, 8)
+
+	buf := make([]byte, 32)
+	// This read leaves a one-byte remainder in the internal buffer, so the source position is ahead of
+	// the logical position by 1.
+	if _, err := l.Read(buf); err != nil {
+		t.Fatal(err)
+	}
+
+	// Seek with io.SeekCurrent must be relative to the logical position, so this must be a no-op.
+	pos, err := l.Seek(0, io.SeekCurrent)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if want := int64(4); pos != want {
+		t.Errorf("got: %d, want: %d", pos, want)
+	}
+
+	n, err := l.Read(buf)
+	if err != nil {
+		t.Fatal(err)
+	}
+	want := []byte{5, 6, 7, 8}
+	if !bytes.Equal(buf[:n], want) {
+		t.Errorf("got: %v, want: %v", buf[:n], want)
+	}
+}


audio/loop_seek_test.go
```diff
@@ -0,0 +1,75 @@
+package audio_test
```
hajimehoshi: You don't have to make a new file. `loop_test.go` should be fine

kumagi: removed `loop_seek_test.go` and migrated contents to `loop_test.go`.

audio/loop_test.go
```diff
@@ -270,3 +270,69 @@ func TestInfiniteLoopWithSlowSource(t *testing.T) {
...
+		r.pos += int(offset)
+	case io.SeekEnd:
+		r.pos = len(r.data) + int(offset)
+	}
+	if r.pos < 0 {
+		r.pos = 0
+	}
+	if r.pos > len(r.data) {
+		r.pos = len(r.data)
+	}
```
hajimehoshi, line 434: `r.pos = min(max(r.pos, 0), len(r.data))`

kumagi: replaced.

audio/loop_test.go
```diff
@@ -270,3 +270,67 @@ func TestInfiniteLoopWithSlowSource(t *testing.T) {
...
+func (r *chunkedReader) Seek(offset int64, whence int) (int64, error) {
+	switch whence {
+	case io.SeekStart:
+		r.pos = int(offset)
+	case io.SeekCurrent:
+		r.pos += int(offset)
+	case io.SeekEnd:
+		r.pos = len(r.data) + int(offset)
+	}
+	if r.pos < 0 {
```
hajimehoshi: We don't need this

audio/loop_test.go
```diff
@@ -270,3 +270,67 @@ func TestInfiniteLoopWithSlowSource(t *testing.T) {
...
+
+func TestInfiniteLoopSeekClearsExtra(t *testing.T) {
+	src := &chunkedReader{data: bytes.Repeat([]byte{1, 2, 3, 4, 5, 6, 7, 8}, 3)}
+	l := audio.NewInfiniteLoopF32(src, 8)
+
+	buf := make([]byte, 32)
+	if _, err := l.Read(buf); err != nil {
+		t.Fatal(err)
+	}
+	for range 10 {
```
hajimehoshi: Is 10 an arbitrary number?

audio/loop_test.go
kumagi: removed

audio/loop_test.go
kumagi: simplified and removed the loop.

kumagi:
Sorry, mistakenly closed, then reopen.

hajimehoshi:
The fix itself looks right to me: `extra` holds bytes that were read from the source but not yet delivered, and they belong to the pre-seek position, so `Seek` has to drop them. Clearing it after the successful `i.src.Seek` is also the right spot, and leaving the loop-wrap path in `Read` alone is correct there, since `length()` is frame-aligned and the leftover at that point is genuinely undelivered data belonging just before the loop start.

One suggestion on the test: `chunkedReader` duplicates the existing `slowReader` helper just above it, which already wraps an `io.ReadSeeker` and returns non-frame-aligned reads. Reusing it detects the same bug without the extra 30 lines:

```go
func TestInfiniteLoopSeekClearsExtra(t *testing.T) {
	src := make([]byte, 16)
	for i := range src {
		src[i] = byte(i)
	}
	r := &slowReader{src: bytes.NewReader(src)}
	l := audio.NewInfiniteLoop(r, int64(len(src)))

	buf := make([]byte, len(src))
	// The first Read leaves a one-byte remainder in the internal buffer.
	if _, err := l.Read(buf); err != nil {
		t.Fatal(err)
	}
	if _, err := l.Seek(0, io.SeekStart); err != nil {
		t.Fatal(err)
	}

	var got []byte
	for len(got) < 2 {
		n, err := l.Read(buf)
		if err != nil {
			t.Fatal(err)
		}
		got = append(got, buf[:n]...)
	}
	if want := []byte{0, 1}; !bytes.Equal(got[:2], want) {
		t.Errorf("got: %v, want: %v", got[:2], want)
	}
}
```

Without the `loop.go` change this reports `got: [0 0], want: [0 1]`, and it passes with it.

---

hajimehoshi:
Resolve the conflicts

kumagi:
force pushed to resolve conflict

audio/loop.go
```diff
@@ -278,5 +278,6 @@ func (i *InfiniteLoop) Seek(offset int64, whence int) (int64, error) {
 		return 0, err
 	}
 	i.pos = next
+	i.extra = i.extra[:0]
```
hajimehoshi, line 281: Clearing `extra` is right for `io.SeekStart`, but it makes `io.SeekCurrent` lossy: `pos` is the *src* position, which (per the field comment) is ahead of the position of the data returned so far by `len(extra)`. So `next = i.pos + offset` above now skips those buffered bytes, and `Seek(0, io.SeekCurrent)` stops being a no-op. That case was correct before this change, because `extra` was retained.

Reading in a loop, with and without an interleaved `l.Seek(0, io.SeekCurrent)`:

```
with:    [1 2 3 4 7 8 1 2 5 6 7 8 3 4 5 6]
without: [1 2 3 4 5 6 7 8 1 2 3 4 5 6 7 8 ...]
```

`newTimeStream` in `audio/player.go` calls `Seek(0, io.SeekCurrent)` on the source, so this is reachable from an `InfiniteLoop` passed to `NewPlayer` after it has been read.

Could you make `io.SeekCurrent` relative to the logical position too?

```go
	case io.SeekCurrent:
		next = i.pos - int64(len(i.extra)) + offset
```

With that plus this line, the comparison above matches, the new test still passes, and `go test ./audio/...` is green. It also makes the position returned by `Seek` accurate, which it wasn't before: it returned the src position, overstated by `len(extra)`.

audio/loop_test.go
```diff
@@ -433,3 +433,56 @@ func TestInfiniteLoopBlendWithPartialFrameReads(t *testing.T) {
 		})
 	}
 }
+
+type chunkedReader struct {
```
hajimehoshi: `partialFrameReader` just above already does this: it wraps an `io.ReadSeeker` and caps each `Read` to a fixed size. This test can use it instead:

```go
	src := &partialFrameReader{src: bytes.NewReader(bytes.Repeat([]byte{1, 2, 3, 4, 5, 6, 7, 8}, 3)), size: 6}
```

I ran the test that way and it behaves identically, with 28 fewer lines. The `(n, io.EOF)` on the final chunk never comes into play here: the loop length of 8 caps every read well short of the 24-byte source.

hajimehoshi (commented):
Thanks for the fix. The diagnosis is right, and the test is a genuine regression test: reverting the `loop.go` change makes it fail with `got: [5 6 1 2 3 4 5 6], want: [1 2 3 4]`.

Two comments inline. Also a minor note: the middle commit 73cb57b ("simplify an expression") has unresolved conflict markers in `audio/loop_test.go`, so that commit doesn't build. The tip is clean and a squash merge would erase it, but the branch isn't bisectable as it stands.

audio/loop.go
kumagi: Done: made `io.SeekCurrent` relative to the logical position with `next = i.pos - int64(len(i.extra)) + offset`. Also added `TestInfiniteLoopSeekCurrentAfterPartialFrameRead`, which fails without this change (`got: 6, want: 4`). The branch is squashed into one commit now, so the history is bisectable again.

audio/loop_test.go
kumagi: Done: removed `chunkedReader` and switched the test to `partialFrameReader` with `size: 6`.

hajimehoshi (approved):
LGTM

audio/loop_test.go
```diff
@@ -433,3 +433,61 @@ func TestInfiniteLoopBlendWithPartialFrameReads(t *testing.T) {
 		})
 	}
 }
+
+func TestInfiniteLoopSeekClearsExtra(t *testing.T) {
+	src := &partialFrameReader{
+		src:  bytes.NewReader(bytes.Repeat([]byte{1, 2, 3, 4, 5, 6, 7, 8}, 3)),
+		size: 6,
```
hajimehoshi: `6` is not wrong today: `NewInfiniteLoopF32` has a 4-byte bit depth and `6 % 4 == 2`, so each
read does leave a remainder. But 6 is a multiple of 2, so it leaves none at int16's bit depth.
Running this test's logic through `NewInfiniteLoop` with the `Seek` fix reverted:

```
size 6: first read [1 2 3 4 5 6], after Seek [1 2 3 4 5 6]  -> passes against the buggy code
size 5: first read [1 2 3 4],     after Seek [5 1 2 3 4 5]  -> catches it
```

An odd size is not a multiple of any bit depth, which is the same reasoning
`TestInfiniteLoopBlendWithPartialFrameReads` above states for its `size: 3`. 5 is the smallest
value that both exceeds the bit depth and is not a multiple of one, and it leaves every
expectation in both tests unchanged.

```suggestion
	// The source returns 5 bytes at most, which is larger than any bit depth but not a multiple of
	// any, so a read returns a complete value and leaves a remainder.
	src := &partialFrameReader{
		src:  bytes.NewReader(bytes.Repeat([]byte{1, 2, 3, 4, 5, 6, 7, 8}, 3)),
		size: 5,
	}
```

Same for the second test:

```suggestion
	// The source returns 5 bytes at most, which is larger than any bit depth but not a multiple of
	// any, so a read returns a complete value and leaves a remainder.
	src := &partialFrameReader{
		src:  bytes.NewReader([]byte{1, 2, 3, 4, 5, 6, 7, 8}),
		size: 5,
	}
```

and the remainder there becomes one byte:

```suggestion
	// This read leaves a one-byte remainder in the internal buffer, so the source position is ahead of
	// the logical position by 1.
```

audio/loop_test.go
```diff
@@ -433,3 +433,61 @@ func TestInfiniteLoopBlendWithPartialFrameReads(t *testing.T) {
...
+	want := []byte{1, 2, 3, 4}
+	if !bytes.Equal(buf[:n], want) {
+		t.Errorf("got: %v, want: %v", buf[:n], want)
+	}
+}
+
+func TestInfiniteLoopSeekCurrentAfterPartialFrameRead(t *testing.T) {
+	src := &partialFrameReader{
+		src:  bytes.NewReader([]byte{1, 2, 3, 4, 5, 6, 7, 8}),
+		size: 6,
```
hajimehoshi: ditto

hajimehoshi (changes requested):
Sorry but I found a thing I want you to fix:

audio/loop_test.go

audio/loop_test.go
kumagi: done

hajimehoshi (approved):

PR #3555, opened 2026-08-28
text/v2: fix data races in concurrent glyph realization

kumagi:
None

bug

## What this PR does | solves
`Draw`, `AppendGlyphs`, `AppendLazyGlyphs`, `Measure` and `CacheGlyphs` are documented as concurrent-safe, but `ensureVariationsString` and `ensureFeaturesString` computed the face's cache strings lazily and wrote them back, so two goroutines realizing glyphs from the same `GoTextFace` raced on the write. The variations case fired for *every* face, not only faces with variations, since the empty result was written back on every call.

Encode the cache strings eagerly in the four setters, which are the only writers of `variations` and `features`, and read the fields directly at the call sites. The readers become plain field reads, so no lock is needed and `GoTextFace` stays copyable, which a mutex field in the exported struct would have broken. Document that a face must not be mutated while it is used for drawing.

Add `TestGlyphImageCacheConcurrent`, which reports a data race with `-race` without this fix.

(The `glyphImageCache` half of the original race was fixed separately in main by e3e564bb1, so it is no longer part of this PR.)

### DIFF
--- a/text/v2/gotextfacesource.go
+++ b/text/v2/gotextfacesource.go
@@ -701,7 +701,7 @@ type faceBitmapState struct {
 //
 // The caller must hold g.shapeMu.
 func (g *GoTextFaceSource) applyFaceState(face *GoTextFace) faceBitmapState {
-	if s := face.ensureVariationsString(); s != g.lastVariationsString {
+	if s := face.variationsString; s != g.lastVariationsString {
 		g.f.SetVariations(face.variations)
 		g.lastVariationsString = s
 	}
@@ -869,7 +869,7 @@ func (g *GoTextFaceSource) buildGlyphs(outputs []shaping.Output, text string, fa
 	}
 	indices = append(indices, len(text))
 
-	variations := face.ensureVariationsString()
+	variations := face.variationsString
 
 	// Snapshot the variations slice once; every glyphRenderData built
 	// in this call shares the same underlying copy. The user-facing
--- a/text/v2/gotextfacesource_test.go
+++ b/text/v2/gotextfacesource_test.go
@@ -1,0 +1,51 @@
+// Copyright 2026 The Ebitengine Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package text_test
+
+import (
+	"bytes"
+	"sync"
+	"testing"
+
+	"golang.org/x/image/font/gofont/goregular"
+
+	"github.com/hajimehoshi/ebiten/v2/text/v2"
+)
+
+func TestGlyphImageCacheConcurrent(t *testing.T) {
+	src, err := text.NewGoTextFaceSource(bytes.NewReader(goregular.TTF))
+	if err != nil {
+		t.Fatal(err)
+	}
+	face := &text.GoTextFace{Source: src, Size: 16}
+	glyphs := text.AppendLazyGlyphs(nil, "Hello, 世界!", face, nil)
+
+	const goroutines = 8
+	var wg sync.WaitGroup
+	for range goroutines {
+		wg.Go(func() {
+			for _, gl := range glyphs {
+				// A space or a control character has no image.
+				if gl.ImageBounds.Empty() {
+					continue
+				}
+				if gl.Image() == nil {
+					t.Error("Image() returned nil for a glyph which should have an image")
+				}
+			}
+		})
+	}
+	wg.Wait()
+}
--- a/text/v2/gotext.go
+++ b/text/v2/gotext.go
@@ -15,10 +15,10 @@
 package text
 
 import (
-	"bytes"
 	"encoding/binary"
 	"fmt"
 	"image"
+	"math"
 	"slices"
 
 	"github.com/go-text/typesetting/di"
@@ -208,12 +208,12 @@ func encodeVariations(variations []font.Variation) string {
 	if len(variations) == 0 {
 		return ""
 	}
-	var buf bytes.Buffer
+	buf := make([]byte, 0, 8*len(variations))
 	for _, t := range variations {
-		_ = binary.Write(&buf, binary.LittleEndian, t.Tag)
-		_ = binary.Write(&buf, binary.LittleEndian, t.Value)
+		buf = binary.LittleEndian.AppendUint32(buf, uint32(t.Tag))
+		buf = binary.LittleEndian.AppendUint32(buf, math.Float32bits(t.Value))
 	}
-	return buf.String()
+	return string(buf)
 }
 
 // encodeFeatures returns a binary string encoding of features
@@ -222,12 +222,12 @@ func encodeFeatures(features []shaping.FontFeature) string {
 	if len(features) == 0 {
 		return ""
 	}
-	var buf bytes.Buffer
+	buf := make([]byte, 0, 8*len(features))
 	for _, t := range features {
-		_ = binary.Write(&buf, binary.LittleEndian, t.Tag)
-		_ = binary.Write(&buf, binary.LittleEndian, t.Value)
+		buf = binary.LittleEndian.AppendUint32(buf, uint32(t.Tag))
+		buf = binary.LittleEndian.AppendUint32(buf, t.Value)
 	}
-	return buf.String()
+	return string(buf)
 }
 
 func (g *GoTextFace) outputCacheKey(text string) goTextOutputCacheKey {
--- a/text/v2/gotext.go
+++ b/text/v2/gotext.go
@@ -41,6 +41,10 @@ var _ Face = (*GoTextFace)(nil)
 // Unlike GoXFace, one GoTextFace instance doesn't have its own glyph image cache.
 // Instead, a GoTextFaceSource has a glyph image cache.
 // You can casually create multiple GoTextFace instances from the same GoTextFaceSource.
+//
+// Mutating a GoTextFace, including assigning to its fields, while the same face
+// is used for drawing in another goroutine is not allowed.
+// Mutate a face before drawing with it, or use another face.
 type GoTextFace struct {
 	// Source is the font face source.
 	Source *GoTextFaceSource
@@ -87,7 +91,7 @@ func (g *GoTextFace) SetVariation(tag Tag, value float32) {
 			return
 		}
 		g.variations[i].Value = value
-		g.variationsString = ""
+		g.variationsString = encodeVariations(g.variations)
 		return
 	}
 
@@ -98,7 +102,7 @@ func (g *GoTextFace) SetVariation(tag Tag, value float32) {
 		Tag:   font.Tag(tag),
 		Value: value,
 	}
-	g.variationsString = ""
+	g.variationsString = encodeVariations(g.variations)
 }
 
 // RemoveVariation removes a variation value.
@@ -113,7 +117,7 @@ func (g *GoTextFace) RemoveVariation(tag Tag) {
 
 		copy(g.variations[i:], g.variations[i+1:])
 		g.variations = g.variations[:len(g.variations)-1]
-		g.variationsString = ""
+		g.variationsString = encodeVariations(g.variations)
 		return
 	}
 }
@@ -134,7 +138,7 @@ func (g *GoTextFace) SetFeature(tag Tag, value uint32) {
 			return
 		}
 		g.features[i].Value = value
-		g.featuresString = ""
+		g.featuresString = encodeFeatures(g.features)
 		return
 	}
 
@@ -145,7 +149,7 @@ func (g *GoTextFace) SetFeature(tag Tag, value uint32) {
 		Tag:   font.Tag(tag),
 		Value: value,
 	}
-	g.featuresString = ""
+	g.featuresString = encodeFeatures(g.features)
 }
 
 // RemoveFeature removes a feature value.
@@ -160,7 +164,7 @@ func (g *GoTextFace) RemoveFeature(tag Tag) {
 
 		copy(g.features[i:], g.features[i+1:])
 		g.features = g.features[:len(g.features)-1]
-		g.featuresString = ""
+		g.featuresString = encodeFeatures(g.features)
 		return
 	}
 }
@@ -197,14 +201,6 @@ func (g *GoTextFace) Metrics() Metrics {
 	return g.Source.metrics(g.Size)
 }
 
-func (g *GoTextFace) ensureVariationsString() string {
-	if g.variationsString != "" {
-		return g.variationsString
-	}
-	g.variationsString = encodeVariations(g.variations)
-	return g.variationsString
-}
-
 // encodeVariations returns a binary string encoding of variations
 // suitable for use as a cache key. It is the canonical form referenced
 // by glyphDataCacheKey and glyphRenderDataCacheKey.
@@ -220,20 +216,18 @@ func encodeVariations(variations []font.Variation) string {
 	return buf.String()
 }
 
-func (g *GoTextFace) ensureFeaturesString() string {
-	if g.featuresString != "" {
-		return g.featuresString
-	}
-	if len(g.features) == 0 {
+// encodeFeatures returns a binary string encoding of features
+// suitable for use as a cache key.
+func encodeFeatures(features []shaping.FontFeature) string {
+	if len(features) == 0 {
 		return ""
 	}
 	var buf bytes.Buffer
-	for _, t := range g.features {
+	for _, t := range features {
 		_ = binary.Write(&buf, binary.LittleEndian, t.Tag)
 		_ = binary.Write(&buf, binary.LittleEndian, t.Value)
 	}
-	g.featuresString = buf.String()
-	return g.featuresString
+	return buf.String()
 }
 
 func (g *GoTextFace) outputCacheKey(text string) goTextOutputCacheKey {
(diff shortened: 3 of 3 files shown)


text/v2/gotextfacesource_test.go
```diff
@@ -0,0 +1,33 @@
+package text_test
```
hajimehoshi, line 15: Rename this to `gotextfacesource_test.go`. Also add copyright comments.

text/v2/glyphimagecache_race_test.go
```diff
@@ -0,0 +1,33 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	face := &text.GoTextFace{Source: src, Size: 16}
+	glyphs := text.AppendLazyGlyphs(nil, "Hello, 世界!", face, nil)
+
+	const goroutines = 8
+	var wg sync.WaitGroup
+	for g := range goroutines {
+		wg.Add(1)
```
hajimehoshi: Use `wg.Go`

text/v2/glyphimagecache_race_test.go
```diff
@@ -0,0 +1,33 @@
...
+		go func(g int) {
```
hajimehoshi: What's the purpose of `g`?

text/v2/gotextfacesource.go
```diff
@@ -1225,13 +1227,17 @@ func (g *GoTextFaceSource) scale(size float64) float64 {
...
 	}
-	if _, ok := g.glyphImageCache[goTextFace.Size]; !ok {
-		g.glyphImageCache[goTextFace.Size] = newCache[goTextGlyphImageCacheKey, *ebiten.Image](128 * glyphVariationCount(goTextFace))
+	c, ok := g.glyphImageCache[goTextFace.Size]
+	if !ok {
+		c = newCache[goTextGlyphImageCacheKey, *ebiten.Image](128 * glyphVariationCount(goTextFace))
+		g.glyphImageCache[goTextFace.Size] = c
 	}
-	return g.glyphImageCache[goTextFace.Size].getOrCreate(key, create)
+	g.glyphImageCacheMu.Unlock()
```
hajimehoshi: Use `defer` for unlocking a mutex whenever possible

text/v2/gotextfacesource_test.go
kumagi: renamed.

text/v2/glyphimagecache_race_test.go
kumagi: replaced.

text/v2/glyphimagecache_race_test.go
kumagi: removed.

text/v2/gotextfacesource.go
kumagi: added defer.

hajimehoshi:
Thanks for the patch. I applied it locally and verified the analysis with `-race`. The diagnosis is right and both races are real, but I'd like to see the `GoTextFace` half reworked before merging. Details and a concrete plan below.

## What I confirmed

Both races reproduce on `main` with the new test, and each half of the fix is independently necessary (verified by applying one half at a time):

- With only the `gotextfacesource.go` fix: race remains in `(*GoTextFace).ensureVariationsString`.
- With only the `gotext.go` fix: race remains in `(*GoTextFaceSource).getOrCreateGlyphImage`.

`Draw` / `AppendGlyphs` / `AppendLazyGlyphs` are all documented as concurrent-safe, so this is a genuine contract violation. Note that the variations race fires for *every* face, not only faces with variations: when there are no variations, `ensureVariationsString` stores `""` back into `g.variationsString` on every call, so it writes unconditionally.

With the full patch applied, `go test -race ./text/v2/...` passes (also with `-count=4`).

## Blocking: a `sync.Mutex` in the exported `GoTextFace` breaks copyability

`GoTextFace` is a public struct that users construct as a literal, and its own doc invites making many instances. Embedding a `sync.Mutex` makes it non-copyable, so `go vet`'s copylocks check now fails on downstream code. Against the patched tree:

```
main.go:8:7: assignment copies lock value to g: .../text/v2.GoTextFace contains sync.Mutex
```

for the natural pattern of deriving a face at another size:

```go
g := *f
g.Size = size
```

That is an API break for user code that compiles today.

The lock in the four setters also doesn't buy the safety it appears to: the shaping code reads `face.variations` and `face.features` directly, without this mutex, in `gotextfacesource.go` (`g.f.SetVariations(face.variations)`, `FontFeatures: face.features`, `slices.Clone(face.variations)`). So `SetVariation` concurrent with `Draw` remains a data race on the slices themselves; only the derived string fields get serialized.

## Improvement plan

1. **Drop `stringCacheMu` and the setter locking entirely.** Instead, compute the cache strings eagerly in the setters and make `ensureVariationsString` / `ensureFeaturesString` pure reads. The setters are the only writers of `variations` / `features`, so concurrent readers become race-free with no lock, and `GoTextFace` stays copyable. This also stops implying that mutating a face concurrently with drawing is supported. The cost is recomputing a short string per setter call instead of once, and the setters are called a handful of times per face.

    <details>
    <summary>Proposed <code>text/v2/gotext.go</code> (diff against <code>main</code>, replacing this PR's changes to that file)</summary>

    ```diff
    @@ func (g *GoTextFace) SetVariation(tag Tag, value float32) {
     		g.variations[i].Value = value
    -		g.variationsString = ""
    +		g.variationsString = encodeVariations(g.variations)
     		return
     	}
    @@
     		Tag:   font.Tag(tag),
     		Value: value,
     	}
    -	g.variationsString = ""
    +	g.variationsString = encodeVariations(g.variations)
     }
    @@ func (g *GoTextFace) RemoveVariation(tag Tag) {
     		g.variations = g.variations[:len(g.variations)-1]
    -		g.variationsString = ""
    +		g.variationsString = encodeVariations(g.variations)
    @@ func (g *GoTextFace) SetFeature(tag Tag, value uint32) {
     		g.features[i].Value = value
    -		g.featuresString = ""
    +		g.featuresString = encodeFeatures(g.features)
     		return
     	}
    @@
     		Tag:   font.Tag(tag),
     		Value: value,
     	}
    -	g.featuresString = ""
    +	g.featuresString = encodeFeatures(g.features)
     }
    @@ func (g *GoTextFace) RemoveFeature(tag Tag) {
     		g.features = g.features[:len(g.features)-1]
    -		g.featuresString = ""
    +		g.featuresString = encodeFeatures(g.features)
    @@
     func (g *GoTextFace) ensureVariationsString() string {
    -	if g.variationsString != "" {
    -		return g.variationsString
    -	}
    -	g.variationsString = encodeVariations(g.variations)
     	return g.variationsString
     }
    @@
     func (g *GoTextFace) ensureFeaturesString() string {
    -	if g.featuresString != "" {
    -		return g.featuresString
    -	}
    -	if len(g.features) == 0 {
    +	return g.featuresString
    +}
    +
    +// encodeFeatures returns a binary string encoding of features
    +// suitable for use as a cache key.
    +func encodeFeatures(features []shaping.FontFeature) string {
    +	if len(features) == 0 {
     		return ""
     	}
     	var buf bytes.Buffer
    -	for _, t := range g.features {
    +	for _, t := range features {
     		_ = binary.Write(&buf, binary.LittleEndian, t.Tag)
     		_ = binary.Write(&buf, binary.LittleEndian, t.Value)
     	}
    -	g.featuresString = buf.String()
    -	return g.featuresString
    +	return buf.String()
     }
    ```
    </details>

    I tried this locally on top of this PR's `gotextfacesource.go` change and its new test: `go test -race -count=3 ./text/v2/` passes, and the copylocks failure is gone.

2. **Keep the `GoTextFaceSource` half as-is.** `glyphImageCacheFor` is correct: map creation and the per-size insert are both under the new mutex, the lock is released before `getOrCreate` runs `create`, and no lock-order inversion is introduced (`glyphVariationCount` -> `Metrics` -> `unscaledMetricsOnce` acquires nothing). Two small things:
   - `shapeMu` right below carries a comment saying what it guards; `glyphImageCacheMu` deserves the same one-liner.
   - Placing `glyphImageCacheMu` immediately next to `glyphImageCache` would make the association obvious.

3. **Nits.** The blank line after `defer ...Unlock()` is present in the setters but not in the `ensure*String` functions. The test discards `gl.Image()`; asserting it is non-nil would make the test do something even without `-race`.

## Adjacent, out of scope for this PR

`GoTextFaceSource.hasGlyphCache` (`runeToBoolMap`) is unguarded as well and is reachable from the same concurrent-safe APIs. A single `MultiFace` happens to be protected by `splitTextCache`'s own mutex, but two `MultiFace`s sharing one source are not:

```
WARNING: DATA RACE
  (*runeToBoolMap).set -> (*GoTextFaceSource).hasGlyph -> (*MultiFace).doSplitText
```

Probably worth a separate issue rather than growing this PR.

---

kumagi:
Thanks for the detailed review — I reworked the GoTextFace half as you proposed.

1. `stringCacheMu` and all the setter locking are gone. The setters now encode eagerly (`g.variationsString = encodeVariations(g.variations)`, same for features via a new `encodeFeatures` extracted from `ensureFeaturesString`), and `ensureVariationsString` / `ensureFeaturesString` are pure reads. Since the setters are the only writers of `variations` / `features`, no lock is needed and `GoTextFace` is copyable again — I re-ran `go vet` with a `g := *f` test file and the copylocks diagnostic is gone. I also documented on `GoTextFace` that mutating a face while it is being drawn in another goroutine is not allowed, since the mutex was implying otherwise.

2. The `GoTextFaceSource` half is unchanged except for the nits: `glyphImageCacheMu` now sits right above `glyphImageCache` with a comment saying it guards the map and the per-size cache creation only, since each per-size cache has its own mutex.

3. The test now fails if `Image()` returns nil for a glyph with non-empty `ImageBounds` (spaces and control characters are skipped), so it asserts something without `-race` too.

Verified: `go test -race -count=3 ./text/v2/` and `./text/...` pass, gofmt/vet clean, and I re-checked both halves are independently necessary by reverting one at a time — without the source mutex the race fires in `getOrCreateGlyphImage`, and with the lazy encoding restored it fires in `ensureVariationsString` on every face, as you found.

Agreed on `hasGlyphCache` / `runeToBoolMap` being a separate issue; I left it out of this PR. Happy to file it if you'd rather not.

hajimehoshi:
Reviewed the latest state (59989a53a). The rework is correct and all three items landed. Verified locally on the head commit:

- `go test -race -count=3 ./text/v2/` and `go test -race ./text/...` pass
- `gofmt -l text/v2/` and `go vet ./text/v2/...` clean
- copylocks: the `g := *f` derive-a-face pattern vets clean again (it failed on the previous revision)
- The eager encoding is complete: `g.variations` / `g.features` are written only by the four setters, so the two `ensure*String` readers have no remaining concurrent writer. All five call sites are reads.

Nothing blocking left. Three small things, one of which I'd fix before merging:

**1. The new godoc paragraph under-states the constraint, and enumerates method names.**

```go
// Mutating a GoTextFace with SetVariation, RemoveVariation, SetFeature, or RemoveFeature
// while the same face is used for drawing in another goroutine is not allowed.
```

The four setters are not the only unsafe mutation: assigning to the exported fields (`Size`, `Direction`, `Language`, `Script`, `Source`) while another goroutine draws with the same face is equally a data race, and `face.Size = 20` mid-frame is the likelier user mistake. Naming only the setters implies field assignment is fine. The method list is also the kind of enumeration that goes stale when methods are added or renamed. A categorical sentence covers both:

```go
// Mutating a GoTextFace, including assigning to its fields, while the same face
// is used for drawing in another goroutine is not allowed.
// Mutate a face before drawing with it, or use another face.
```

**2. `glyphImageCacheMu`'s comment says "mutations", but it guards reads too.**

```go
// glyphImageCacheMu guards mutations of glyphImageCache itself, i.e., the
// creation of the map and of a per-size cache. ...
```

The lookup in `glyphImageCacheFor` is a read, and it is correctly under the lock; a future reader could take "guards mutations" as license to read the map unlocked. "guards access to glyphImageCache" is accurate, and the following sentence about each per-size cache having its own mutex already carries the rest.

**3. Optional: `ensureVariationsString` / `ensureFeaturesString` no longer ensure anything.**

They are one-line field getters now. Either rename them, or drop them and read the fields at the five call sites. Purely cosmetic, and leaving them alone to keep the fix minimal is a defensible call.

---

hajimehoshi:
Resolve conflicts

kumagi:
resolved.

hajimehoshi:
Reviewed the latest state (c40f5594e). After the rebase onto 36e0ee7e7, which already contains e3e564bb1 ("text/v2: guard the glyph image cache with a mutex"), the `GoTextFaceSource` half has dropped out of the diff and what remains is the `GoTextFace` half plus the test. Both remaining nits are addressed, and the optional third one was taken further than suggested: `ensureVariationsString` / `ensureFeaturesString` are gone, and the five call sites read `variationsString` / `featuresString` directly.

Verified on the head commit:

- `go test -race -count=3 ./text/v2/` passes; `go vet ./text/v2/...` and `gofmt -l text/v2/` clean
- The test still earns its place on the new base: reverting just this commit's `gotext.go`, the race fires in `(*GoTextFace).ensureVariationsString` and the test fails
- The eager encoding is still complete after the rebase: `variations` / `features` are written only by the four setters, and the early-return path in `SetVariation` (`v.Value == value`) correctly leaves the string untouched
- `applyFaceState` reading `face.variationsString` before `g.f.SetVariations(face.variations)` stays consistent, since the string is now always in sync with the slice

The doc paragraph reads correctly against the code: the constraint really is about the whole face, not just the setters.

**One thing left: the PR description is now stale.** It still says

> Guard the glyphImageCache map access with a mutex ... and serialize the variations/features state of GoTextFace with a single mutex taken by the setters and the cache-string computations.

Both halves of that sentence describe code no longer in this PR: the map guard landed separately on main, and the `GoTextFace` mutex was replaced by eager encoding. Worth rewriting before merge, since the body tends to become the squash-merge message. Something like:

> `ensureVariationsString` / `ensureFeaturesString` computed the cache strings lazily, so two goroutines realizing glyphs from the same face raced on the write. The variations case fired for every face, since the empty result was written back on every call. Encode eagerly in the setters instead, making the readers pure field reads, and document that a face must not be mutated while it is used for drawing.

Minor, take it or leave it: with the source-side fix now on main, `TestGlyphImageCacheConcurrent` in `gotextfacesource_test.go` tests a fix that lives in `gotext.go`. It does still exercise the image cache end to end, so the placement is defensible; just noting the mismatch.

Otherwise this looks ready to merge.

---

kumagi:
Thanks — rewritten the PR description to match what is actually in the diff now: the lazy `ensure*String` write-back as the race, eager encoding in the setters as the fix, the copyability note, and an explicit line that the `glyphImageCache` half landed separately in e3e564bb1.

Re-verified on the head commit after the doc change: `go test -race ./text/...` and `go test -race -count=3 ./text/v2/` pass, `gofmt -l text/v2/` and `go vet ./text/v2/...` clean, and a `g := *f` copy still vets without a copylocks diagnostic.

Leaving `TestGlyphImageCacheConcurrent` in `gotextfacesource_test.go` as you suggested, since it does exercise the image cache end to end even though the regression it catches now lives in `gotext.go`. No code changes were needed, so nothing new to commit.

hajimehoshi:
One follow-up on the eager encoding, which I suggested without measuring it: it does cost something, and I think the fix belongs in this PR rather than a separate one, since this PR is what makes the cost visible.

The setters now encode on every call, so a face whose axes are animated per frame encodes once per `SetVariation` instead of once per frame. Benchmarks modeled on `examples/fontvariation` (three axes set per frame, `RobotoFlex.ttf`), Apple M3 Pro, `-count=6`, compared with benchstat:

| benchmark | pre-PR (lazy) | this PR (eager) | with the encoder change below |
|---|---|---|---|
| 3 setters, no read | 7.4 ns, 0 allocs | 401 ns, 27 allocs | 43 ns, 3 allocs |
| 3 setters + `Measure` (one frame) | 512 ns, 9 allocs | 774 ns (+51%), 27 allocs | 414 ns (-19%), 3 allocs |
| `Measure`, no variations | 351 ns, 0 allocs | 340 ns, 0 allocs | 351 ns, 0 allocs |

The allocation counts show what happened: 9 -> 27 is exactly 3x, one encode per setter call instead of one per frame.

The interesting part is that the per-encode cost is the real problem, not the extra calls. `binary.Write(&buf, binary.LittleEndian, t.Tag)` takes its value as `any`, so every tag and value is boxed onto the heap; together with `bytes.Buffer` growth that is 9 allocations per encode. That cost already existed before this PR, just amortized over an invalidate-then-read cycle.

Encoding directly removes it, and then this PR is a net improvement over the pre-PR baseline instead of a regression (-19% time and -67% allocations on the per-frame benchmark):

```go
func encodeVariations(variations []font.Variation) string {
	if len(variations) == 0 {
		return ""
	}
	buf := make([]byte, 0, 8*len(variations))
	for _, t := range variations {
		buf = binary.LittleEndian.AppendUint32(buf, uint32(t.Tag))
		buf = binary.LittleEndian.AppendUint32(buf, math.Float32bits(t.Value))
	}
	return string(buf)
}
```

and the same shape for `encodeFeatures`, where `t.Value` is already a `uint32`. This drops `bytes` from the imports and adds `math`. I checked the output is byte-identical to the `binary.Write` form, including NaN and -Inf values, and `go vet ./text/v2/...` plus `go test -race -count=2 ./text/...` pass with it.

In theory this could land as its own change, since the encoder was already this expensive. But without it this PR lands as a measurable per-frame regression for variable fonts, so folding it in seems better than merging the regression and fixing it afterwards.

Two things that bound how much this matters: a face with no variations hits the `len == 0` early return and is completely unaffected, which is the common case; and in the frame benchmark the output cache was hitting, so 774 ns is nearly all bookkeeping. In a real frame that also shapes and rasterizes, the extra ~260 ns is small. The 18 extra allocations per frame in a game loop are the more interesting number.

---

kumagi:
Thanks for measuring this — folded the encoder change into the PR as 561226eae.

`encodeVariations` and `encodeFeatures` now build the key with `binary.LittleEndian.AppendUint32` and `math.Float32bits` on a pre-sized buffer, which drops the `any` boxing from `binary.Write` and removes `bytes` from the imports in favor of `math`. I verified the output is byte-identical to the old form, including NaN and -Inf values.

Re-verified on the new head: `go test -race -count=3 ./text/v2/` and `go test -race ./text/...` pass, `gofmt -l text/v2/` and `go vet ./text/v2/...` are clean, and the `g := *f` pattern still vets without a copylocks diagnostic. With this, the per-frame case should land at or below the pre-PR baseline rather than a regression, as you found.

hajimehoshi (approved):
LGTM

PR #3557, opened 2026-08-28
inpututil: return 0 or false on an out-of-range gamepad button

kumagi:
None

bug

## What this PR does | solves
The gamepad button duration queries indexed the fixed-size duration arrays without validating the button, so an out-of-range button caused an index-out-of-range panic. Validate the button and panic with a descriptive message instead, like the matrix index checks.


hajimehoshi (commented):
Do not add panics. Return `false` or `0`.

kumagi:
OK, changed to `false` or `0`.

inpututil/inpututil_test.go
```diff
@@ -0,0 +1,49 @@
+// Copyright 2026 The Ebitengine Authors
```
hajimehoshi: I don't think we need tests for this

hajimehoshi:
The PR title is no longer valid

inpututil/inpututil_test.go
kumagi: remvoed.

hajimehoshi (approved):
LGTM

kumagi:
PR title fixed.

PR #3558, opened 2026-08-28
audio: fix an int64 overflow when converting a duration and a position

kumagi:
None

bug

## What this PR does | solves

Three duration/position conversions in audio multiplied before dividing, which overflowed int64:

- `timeDurationToPos`: `offset * bytesPerSample * sampleRate` overflowed above about 7.3 hours of duration for 32bit float stereo at 44.1 kHz (about 14.5 hours for 16bit stereo).
- `positionInTimeDuration`: `pos * time.Second` overflowed above about 9.2 GB of stream position, i.e. about 6.5 hours of 32bit float stereo at 44.1 kHz, so a long-running app with an `audio.InfiniteLoop` BGM hits it even without any `SetPosition` call.
- `updatePosition`: `samples * time.Second` overflowed above about 9.22e9 samples (about 58 hours at 44.1 kHz).

All three now share a single unexported `mulDiv` helper, which splits the multiplication so that it cannot overflow int64. The result is bit-identical to the previous expressions wherever the exact result fits in int64.

`SetBufferSize` has the same multiplication shape but needs a buffer longer than about 7 hours to overflow, so it is left as is.

### DIFF
--- a/audio/audio_test.go
+++ b/audio/audio_test.go
@@ -239,6 +239,27 @@ func TestSetPositionBeforeDeviceCreation(t *testing.T) {
 	}
 }
 
+func TestSetPositionLongDuration(t *testing.T) {
+	setup()
+	defer teardown()
+
+	p, err := context.NewPlayerF32(bytes.NewReader(nil))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// 10 milliseconds is exactly 441 samples at 44100 Hz, so the position
+	// round-trips exactly while exercising both the whole-seconds and the
+	// remainder terms of the conversion.
+	const duration = 8*time.Hour + 10*time.Millisecond
+	if err := p.SetPosition(duration); err != nil {
+		t.Fatal(err)
+	}
+	if got, want := p.Position(), duration; got != want {
+		t.Errorf("Position(): got: %v, want: %v", got, want)
+	}
+}
+
 // Issue #3438
 func TestRewindNonSeekableBeforeDeviceCreation(t *testing.T) {
 	setup()
--- a/audio/player.go
+++ b/audio/player.go
@@ -566,7 +566,13 @@ func (p *playerImpl) updatePosition() {
 	}
 
 	// Update the adjusted position every tick. This is necessary to keep the position accurate.
-	p.adjustedPosition = int64(time.Duration(samples)*time.Second/time.Duration(p.factory.sampleRate) + adjustingTime)
+	p.adjustedPosition = mulDiv(samples, int64(time.Second), int64(p.factory.sampleRate)) + int64(adjustingTime)
+}
+
+// mulDiv returns x * mul / div, avoiding the overflow of the intermediate x * mul.
+// mul * div must fit in int64.
+func mulDiv(x, mul, div int64) int64 {
+	return x/div*mul + x%div*mul/div
 }
 
 type timeStream struct {
@@ -631,7 +637,8 @@ func (s *timeStream) Seek(offset int64, whence int) (int64, error) {
 }
 
 func (s *timeStream) timeDurationToPos(offset time.Duration) int64 {
-	o := int64(offset) * int64(s.bytesPerSample) * int64(s.sampleRate) / int64(time.Second)
+	bytesPerSecond := int64(s.bytesPerSample) * int64(s.sampleRate)
+	o := mulDiv(int64(offset), bytesPerSecond, int64(time.Second))
 
 	// Align the byte position with the samples.
 	o -= o % int64(s.bytesPerSample)
@@ -645,5 +652,6 @@ func (s *timeStream) position() int64 {
 }
 
 func (s *timeStream) positionInTimeDuration() time.Duration {
-	return time.Duration(s.pos.Load()) * time.Second / (time.Duration(s.sampleRate * s.bytesPerSample))
+	bytesPerSecond := int64(s.sampleRate) * int64(s.bytesPerSample)
+	return time.Duration(mulDiv(s.pos.Load(), int64(time.Second), bytesPerSecond))
 }


hajimehoshi:
Thanks for the patch! The bug is real and the fix is correct. I confirmed the new test fails on `main` (`got: -8286104072, want: 10160640000`) and passes with the change.

There is one thing I'd like addressed in the same PR, though: the inverse conversion has the identical overflow, so this fixes only half of the problem.

## `positionInTimeDuration` overflows at the same threshold

https://github.com/hajimehoshi/ebiten/blob/main/audio/player.go#L648

```go
return time.Duration(s.pos.Load()) * time.Second / (time.Duration(s.sampleRate * s.bytesPerSample))
```

`pos * 1e9` overflows once `pos > 2^63/1e9 ≈ 9.22e9` bytes, which is the same ~7.3 hours at 44.1 kHz / 32-bit-float stereo where `timeDurationToPos` overflows. This is on the read path via `afterSeek` -> `Player.Position()`, so with this PR applied, the very scenario the new test sets up is still broken end to end:

```
Position() after SetPosition(8h): got: -6h31m26.68955133s, want: 8h0m0s
```

The same decomposition applies. This round-trips 0, 500ms, 8h and 100h correctly (sub-second cases match to within one sample period, which is the pre-existing alignment quantization, not overflow):

```go
func (s *timeStream) positionInTimeDuration() time.Duration {
	bytesPerSecond := int64(s.sampleRate) * int64(s.bytesPerSample)
	pos := s.pos.Load()
	return time.Duration(pos/bytesPerSecond)*time.Second + time.Duration(pos%bytesPerSecond)*time.Second/time.Duration(bytesPerSecond)
}
```

## The test can then be much simpler

Once the above is fixed, `bytes.Reader` seeks past the end happily, so the whole `seekRecordingSource` stub can go and the test covers both directions instead of one, matching the style of the neighboring `TestSetPositionBeforeDeviceCreation`:

```go
p, err := context.NewPlayerF32(bytes.NewReader(nil))
// ...
const duration = 8 * time.Hour
if err := p.SetPosition(duration); err != nil {
	t.Fatal(err)
}
if got, want := p.Position(), duration; got != want {
	t.Errorf("Position(): got: %v, want: %v", got, want)
}
```

## `updatePosition` has the same pattern

https://github.com/hajimehoshi/ebiten/blob/main/audio/player.go#L568

`time.Duration(samples)*time.Second/time.Duration(p.factory.sampleRate)` overflows above ~9.22e9 samples (about 58 hours at 44.1 kHz). It is less pressing and only reachable with a live device, but it is the third instance of the same conversion, so I'd rather have all three consistent.

## Minor

- The `io.SeekEnd` case in the stub does `s.pos += offset`, which is `io.SeekCurrent` semantics. It is never exercised, so it is misleading rather than broken. Since this stub can go away entirely per the above, this is moot, but in general please prefer returning an error for an unsupported whence over implementing it incorrectly.
- Please add a short comment to `timeDurationToPos` explaining that the expression is split because `offset * bytesPerSample * sampleRate` does not fit in `int64`. Without it, this is a prime candidate for a future "simplification" back into the overflowing form. Hoisting `bytesPerSecond := int64(s.bytesPerSample) * int64(s.sampleRate)` into a local would help readability too.
- Please hard-wrap the commit message body at 72 columns. It is currently one 277-character line.
- "a duration of a few hours or longer" in the commit message understates it a little. The threshold is about 7.3 hours for 32-bit-float stereo at 44.1 kHz and about 14.5 hours for 16-bit stereo. Being concrete there is useful, since the threshold is what makes the bug plausible in practice.

---

kumagi:
Thanks for the detailed review! I have addressed all the points in the updated commit (aaaf8d53f):

- **`positionInTimeDuration`**: Applied the same decomposition (`pos/bytesPerSecond * time.Second + pos%bytesPerSecond * time.Second / bytesPerSecond`), so the inverse conversion no longer overflows at the same threshold.
- **Test simplification**: Removed the `seekRecordingSource` stub and the test now uses `bytes.NewReader(nil)`, exercising both directions end-to-end via `SetPosition`/`Position()`. Confirmed it fails on the unfixed code (`got: -6h31m26.68955133s`) and passes with the change.
- **`updatePosition`**: Applied the same decomposition to `samples * time.Second / sampleRate` so all three conversions are consistent.
- **Comment + local hoisting**: Added a short comment explaining the split (the naive product does not fit in int64) and hoisted `bytesPerSecond` into a local in both `timeDurationToPos` and `positionInTimeDuration`.
- **Commit message**: Hard-wrapped at 72 columns and made the threshold concrete (~7.3 h for 32bit float stereo at 44.1 kHz, ~14.5 h for 16bit stereo).
- The `io.SeekEnd` misimplementation in the stub is gone along with the stub itself.

audio/player.go
```diff
@@ -631,7 +636,11 @@ func (s *timeStream) Seek(offset int64, whence int) (int64, error) {
 }
 
 func (s *timeStream) timeDurationToPos(offset time.Duration) int64 {
-	o := int64(offset) * int64(s.bytesPerSample) * int64(s.sampleRate) / int64(time.Second)
+	// The naive offset * bytesPerSample * sampleRate does not fit in int64 for a
+	// long duration, so the offset is split into whole seconds and the remainder.
+	bytesPerSecond := int64(s.bytesPerSample) * int64(s.sampleRate)
+	o := int64(offset/time.Second) * bytesPerSecond
+	o += int64(offset%time.Second) * bytesPerSecond / int64(time.Second)
```
hajimehoshi: All three of these sites compute the same `x * mul / div`, so one unexported helper would remove the triplication and let the rationale live in a single place:

```go
// mulDiv returns x * mul / div without overflowing at the intermediate multiplication.
func mulDiv(x, mul, div int64) int64 {
	return x/div*mul + x%div*mul/div
}
```

- here: `o := mulDiv(int64(offset), bytesPerSecond, int64(time.Second))`
- `positionInTimeDuration`: `time.Duration(mulDiv(pos, int64(time.Second), bytesPerSecond))`
- `updatePosition`: `mulDiv(samples, int64(time.Second), int64(p.factory.sampleRate))`

I checked that this helper is bit-identical to all three of the current expressions over 3M random inputs.

Separately, on the comment: could it state the invariant without naming the rejected form? A future reader doesn't have the naive version in front of them, so something like "the multiplication is split so that it cannot overflow int64" carries the same warning without the comparison.

audio/player.go
```diff
@@ -566,7 +566,12 @@ func (p *playerImpl) updatePosition() {
 	}
 
 	// Update the adjusted position every tick. This is necessary to keep the position accurate.
-	p.adjustedPosition = int64(time.Duration(samples)*time.Second/time.Duration(p.factory.sampleRate) + adjustingTime)
+	// samples * time.Second can overflow int64 for a very long playback, so decompose the
+	// samples into whole seconds and the remainder, which is mathematically equivalent.
+	perSecond := int64(p.factory.sampleRate)
+	s := samples / perSecond
+	r := samples % perSecond
+	p.adjustedPosition = int64(time.Duration(s)*time.Second + time.Duration(r)*time.Second/time.Duration(perSecond) + adjustingTime)
```
hajimehoshi: Two small things here:

- `perSecond` is samples per second, while the analogous value at the other two sites is called `bytesPerSecond`. If the shared helper isn't adopted, `samplesPerSecond` would read better, and `s` / `r` could use real names.
- This one isn't covered by any test — it needs `p.player != nil`, i.e. a real audio device, which the test package can't create. That's fine, just noting that the new test exercises only the other two changes.

audio/audio_test.go
```diff
@@ -239,6 +239,24 @@ func TestSetPositionBeforeDeviceCreation(t *testing.T) {
...
+func TestSetPositionLongDuration(t *testing.T) {
+	setup()
+	defer teardown()
+
+	p, err := context.NewPlayerF32(bytes.NewReader(nil))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	const duration = 8 * time.Hour
```
hajimehoshi: `8 * time.Hour` is a whole number of seconds, so `offset % time.Second` is zero and the newly added remainder term contributes nothing — a sign or rounding mistake in that half of the expression would still pass here.

`8*time.Hour + 10*time.Millisecond` exercises both terms and still round-trips exactly, since 10 ms is exactly 441 samples at 44100 Hz and the sample alignment becomes a no-op. The constant does have to be picked with that quantization in mind: `+123*time.Millisecond`, for instance, comes back as `8h0m0.122993197s`.

hajimehoshi (commented):
Thanks! I verified the math rather than just reading it, and the fix is correct.

**Verification**

- I brute-forced all three rewrites against a `math/big` truncating reference (10M random + edge inputs, including negatives and `math.MaxInt64`/`math.MinInt64`, across five sample-rate / bit-depth combinations): zero mismatches wherever the exact result fits in int64. The decomposition is exactly equivalent, not merely approximate — Go's `/` truncates toward zero and `%` takes the sign of the dividend, so both terms round consistently for negative inputs too.
- `timeDurationToPos` no longer overflows even at a `math.MaxInt64` duration with 192 kHz x 8 bytes/sample.
- The new test is a genuine regression test: reverting only `audio/player.go` to the parent commit while keeping the test gives `audio_test.go:253: bytes.Reader.Seek: negative position`.
- `gofmt`, `go vet`, `go test ./audio/...`, and `GOOS=linux GOARCH=386` / `GOOS=js GOARCH=wasm` builds all pass.

One point worth adding to the description: `positionInTimeDuration` is the most reachable of the three. It overflows above ~9.2 GB of stream position, i.e. about 6.5 hours of 32bit float stereo at 44.1 kHz, so a long-running app with an `audio.InfiniteLoop` BGM hits it without anyone ever calling `SetPosition`. That makes this more than a theoretical fix.

`SetBufferSize` has the same multiplication shape but needs a buffer longer than about 7 hours to overflow, so leaving it alone is the right call.

**Main suggestion:** the decomposition is now open-coded three times in three different shapes, each with its own comment restating the same rationale. A single helper collapses all three and lets the "why" be stated once. Details inline, along with two smaller notes.

Also, the PR title is narrower than the change — it fixes the position-to-duration direction in two places as well. The commit headline, "converting a duration and a position", is the accurate one.

---

kumagi:
Thanks for the review and for verifying the math!

Addressed the latest round in a50b21ecd:

- Adopted the shared helper: `mulDiv(x, mul, div) = x/div*mul + x%div*mul/div`, used at all three sites (`timeDurationToPos`, `positionInTimeDuration`, `updatePosition`), with the signatures matching your snippets.
- The rationale now lives only in `mulDiv`'s doc comment and states the invariant without naming the rejected form; the per-site comments (and the `perSecond` / `s` / `r` locals, moot with the helper) are gone.
- The test now uses `8*time.Hour + 10*time.Millisecond` so both the whole-seconds and the remainder terms are exercised; 10 ms is exactly 441 samples at 44100 Hz, so the round-trip is still exact. Reverting only `audio/player.go` to the parent still fails the test (`bytes.Reader.Seek: negative position`).

Also updated the PR title and description to cover the position-to-duration direction and mention the `positionInTimeDuration` reachability via `audio.InfiniteLoop` (and that `SetBufferSize` is intentionally left alone).

`gofmt`, `go vet`, `go test ./audio/...`, and the `GOARCH=386` / `js/wasm` builds all pass.

audio/player.go
```diff
@@ -566,7 +566,12 @@ func (p *playerImpl) updatePosition() {
 	}
 
 	// Update the adjusted position every tick. This is necessary to keep the position accurate.
-	p.adjustedPosition = int64(time.Duration(samples)*time.Second/time.Duration(p.factory.sampleRate) + adjustingTime)
+	p.adjustedPosition = mulDiv(samples, int64(time.Second), int64(p.factory.sampleRate)) + int64(adjustingTime)
+}
+
+// mulDiv returns x * mul / div without overflowing int64 at the intermediate multiplication.
+func mulDiv(x, mul, div int64) int64 {
+	return x/div*mul + x%div*mul/div
```
hajimehoshi: Optional, and not a bug in this PR: the doc promises slightly more than the helper delivers. `x % div * mul` is itself an intermediate multiplication, and it overflows once `div * mul` exceeds int64:

```
mulDiv(19999999999, 1e10, 1e10) = 10776627962   // exact: 19999999999
```

Every call site here is safe by a wide margin: the largest `div * mul` is 3.5e14 against a 9.2e18 ceiling, so nothing is wrong today. But `mulDiv` reads like a general-purpose helper, and a future caller has no way to know the limit. Stating the precondition would cover that:

```go
// mulDiv returns x * mul / div, avoiding the overflow of the intermediate x * mul.
// mul * div must fit in int64.
```

Happy to merge as is if you would rather not churn the PR for a comment.

hajimehoshi (commented):
Re-verified the amended revision from scratch rather than assuming the refactor preserved the math, and it holds up.

- `mulDiv` is exactly equivalent to the original expressions at all three call sites: 18M inputs (random plus edge cases including `math.MaxInt64`/`math.MinInt64` and negative values, across six sample-rate / bit-depth combinations), checked against a `math/big` truncating reference. Zero mismatches.
- `timeDurationToPos` is now exact even at a `math.MaxInt64` duration, and the remaining first-term overflow at the other two sites sits above roughly 292 years of audio, so it is out of reach.
- The new test constant does what its comment says. I mutation-tested it: dropping the remainder term makes the position come back as `8h0m0s` instead of `8h0m0.01s`, so the test catches it. The previous `8 * time.Hour` would not have.
- `gofmt`, `go vet`, `go test ./audio/...`, `GOOS=linux GOARCH=386` and `GOOS=js GOARCH=wasm` all pass, and the test still fails against pre-fix `player.go`.

One optional note on the helper's godoc, inline. Nothing blocking.

---

hajimehoshi (approved):
LGTM

audio/player.go
kumagi: done

audio/audio_test.go
kumagi: done

PR #3559, opened 2026-08-28
audio: keep a finished player's Position from growing

kumagi:
None

bug

## What this PR does | solves
Play on a player that had already finished started the stopwatch even though the underlying player refused to play, so Position kept growing past the end of the audio. Only adjust the position by the stopwatch while the player is actually playing. Also decompose the samples-to-duration conversion to avoid an int64 overflow for very long playback.


audio/position_test.go
```diff
@@ -0,0 +1,131 @@
+// Copyright 2026 The Ebiten Authors
```
hajimehoshi: Ebitengine

kumagi: fixed.

hajimehoshi:
Thanks for the fix! I confirmed the bug and the mechanism.

In oto v3.5.0-alpha.12, `playerImpl.playImpl` returns early when `p.eof && len(p.buf) == 0`, so `Play` on a finished player leaves `IsPlaying() == false`. Since `stopwatch.start` is a no-op while already running, `lastStarted` is never refreshed, and `stopwatch.current()` ends up measuring wall-clock time from the *first* failed `Play`. `samples` never changes for a finished player, so the position grows without bound. Reverting `player.go` makes both new tests fail, so they are real regression tests.

A few review comments.

### The test should be in `package audio_test`

`audio_test.go` and `loop_test.go` are `package audio_test`; `export_test.go` is the only `package audio` file and it intentionally contains no tests. Please keep the new file consistent with that.

For the main case this is avoidable: the existing `dummyPlayer` already stops playing once its source is exhausted, so the bug reproduces through the public API. I ran this and it fails before the fix (`986.8ms` -> `1.089s`) and passes after, without needing `mockPlayer` at all:

```go
p.Play()
for i := 0; i < 200 && p.IsPlaying(); i++ {
	audio.UpdateForTesting()
	time.Sleep(5 * time.Millisecond)
}
end := p.Position()
for i := 0; i < 20; i++ {
	p.Play()
	audio.UpdateForTesting()
	time.Sleep(5 * time.Millisecond)
}
if got := p.Position(); got > end {
	t.Errorf("position grew after the player finished: %v -> %v", end, got)
}
```

This also has the advantage of exercising the real `Play` path instead of setting `p.stopwatch` directly.

The overflow case genuinely needs internals, since 9.2e9 samples is unreachable through the public API. Extracting the conversion into a small `samplesToDuration(samples int64, sampleRate int) time.Duration` and exposing it via `export_test.go` would make that test pure and instant.

### `Position` now moves backwards on `Pause`

The accumulated adjustment is discarded rather than frozen, so the first tick after `Pause` drops it. I measured a 21ms backward step with a stalled device. It is bounded by roughly one tick in practice, and `Position`'s godoc only promises monotonicity "as long as the player continues to play", so it is not a contract violation, but it is a new behavior.

If you want to avoid it, the root cause can be removed at the source instead, in `Play` and `startIfPending`:

```go
p.player.Play()
if p.player.IsPlaying() {
	p.stopwatch.start()
}
```

and then in `updatePosition`, stop the stopwatch rather than ignoring it, so the last adjustment is retained:

```go
if !p.factory.isVMGuest() {
	if !p.isPlaying() {
		p.stopwatch.stop()
	}
	adjustingTime = p.stopwatch.current()
}
```

### Comments

Neither `&& p.isPlaying()` nor the seconds/remainder split explains itself. The decomposition especially looks gratuitous without a line saying it avoids overflowing `time.Duration` for a long playback. `perSecond` / `s` / `r` would also read better as `sampleRate` / `sec` / `rem`.

### The same overflow remains nearby, at a lower threshold

`timeStream.positionInTimeDuration` overflows at about 7.3 hours of F32 stereo, and `timeStream.timeDurationToPos` overflows for an offset above about 7.3 hours, so `SetPosition(8 * time.Hour)` already seeks to a negative position today. Both are lower thresholds than the 58 hours fixed here. Out of scope for this PR, but worth a follow-up.

### Nits

- `p.stream.pos.Store(44100 * 8)` in `TestUpdatePositionNotPlaying` is dead: the helper already stored that value.
- `p.stopwatch.start()` in `TestUpdatePositionPlayingAdvances` is a no-op, since the preceding `updatePosition` took the `else` branch and already started it. In `TestUpdatePositionNotPlaying` the same call is meaningful, as it simulates the buggy state.
- `var _ io.Seeker = (*mockPlayer)(nil)` is redundant, as assigning `mockPlayer` to the `player` field already enforces it, and `io` is imported only for that line.
- `mockPlayer.buffered` is never set to a non-zero value.
- `stopwatch: stopwatch{}` in the struct literal can be dropped.

`gofmt`, `go vet`, and `internal/vettools` are clean, and the package is stable under `-race -count=3` and `-shuffle=on`.

---

hajimehoshi:
Resolve conflicts

hajimehoshi:
Thanks for the update. Splitting the overflow fix out into #3558 was the right call, and the shared `mulDiv` there covers all three conversions. `player.go` is now a clean one-line change. `gofmt`, `go vet` and the whole `./audio/...` tree pass, and reverting the guard still fails `TestUpdatePositionNotPlaying`, so the regression test holds.

The test file is byte-identical to the previous revision though (only the filename changed), so the test feedback has not been applied yet. Two points are now stronger than before.

### `TestUpdatePositionLongPlaybackNoOverflow` no longer tests this PR

I checked: reverting the `p.isPlaying()` guard leaves that test passing. It only exercises `mulDiv`, which already shipped in #3558 along with `TestSetPositionLongDuration`. Please drop it here. That also removes the only need for the `pos0` parameter on the helper.

### `package audio`

With the overflow test gone, nothing left in the file needs internals, and #3558 already demonstrates the preferred shape: a test in `package audio_test` driving the public API. The remaining case reproduces end to end with the existing `dummyPlayer`. I ran this against cec1a26a4 and it passes with the fix and fails without it (`1s` -> `1.105s`):

```go
p, _ := context.NewPlayer(bytes.NewReader(make([]byte, 44100*4)))
p.Play()
for i := 0; i < 200 && p.IsPlaying(); i++ {
	audio.UpdateForTesting()
	time.Sleep(5 * time.Millisecond)
}
end := p.Position()
for i := 0; i < 20; i++ {
	p.Play()
	audio.UpdateForTesting()
	time.Sleep(5 * time.Millisecond)
}
if got := p.Position(); got > end {
	t.Errorf("position grew after the player finished: %v -> %v", end, got)
}
```

The whole file then collapses to one test, with no `mockPlayer`, no `newTestPlayerImplForPosition`, no `io` import and no `stopwatch: stopwatch{}`, and it exercises the real `Play` path instead of setting `p.stopwatch` directly.

### The commit message is stale

The last sentence still says "Also decompose the samples-to-duration conversion to avoid an int64 overflow for very long playback", but that change is no longer in this commit.

### Still open from the previous round

- No comment explaining the new `&& p.isPlaying()` condition.
- `Position` moves backwards on `Pause`. I re-measured 21ms on this head. It is bounded by roughly one tick and stays within `Position`'s documented "as long as the player continues to play", so I can accept it, but the alternative of not starting the stopwatch in `Play` and `startIfPending` when the underlying player refuses, and calling `stopwatch.stop()` in `updatePosition` instead of ignoring it, avoids the step and removes the root cause.
- Minor, and all moot if the file collapses as above: the dead `p.stream.pos.Store(44100 * 8)`, the no-op `p.stopwatch.start()` in `TestUpdatePositionPlayingAdvances`, the redundant `var _ io.Seeker`, and the unused `mockPlayer.buffered`.

---

kumagi:
Thanks for the thorough review! I pushed bc9196cbc addressing the remaining points:

- Dropped `TestUpdatePositionLongPlaybackNoOverflow`; the overflow fix and its test live in #3558.
- The test is now a single `TestPositionNotGrowingAfterFinished` in `package audio_test` driving only the public API, with no `mockPlayer`, no internals helper, and no `io` import.
- Adopted the root-cause fix: `Play` and `startIfPending` start the stopwatch only when the underlying player actually started playing, and `updatePosition` stops the stopwatch when the player is not playing, so the last adjustment is retained and `Position` no longer moves backwards after `Pause`.
- Added comments explaining both conditions.
- The commit message no longer mentions the samples-to-duration decomposition.

`dummyPlayer` in `export_test.go` now refuses to play once its source is exhausted, as the real players do (and seeking resets that state), so the test reproduces the bug mechanism through the public API deterministically.

The test fails when the `player.go` change is reverted (`position grew after the player finished: 4.03s -> 4.14s`), and `gofmt`, `go vet` and `go test -race ./audio/...` pass.

hajimehoshi:
This revision looks good. Taking the root-cause route in `Play` and `startIfPending`, and stopping the stopwatch in `updatePosition` instead of ignoring it, addresses the backwards step on `Pause` as well. The test now drives the public API, the stale overflow test is gone, and the commit message describes the mechanism accurately.

`gofmt`, `go vet` and the whole `./audio/...` tree pass, stable over `-count=5 -shuffle=on` and `-race -count=2`. Reverting the `Play` guard fails the new test (`4.006s` -> `4.097s`), so the primary case is guarded.

Two things left.

### The dummy player now diverges from oto on pause

`dummyPlayer` sets `eof = true` whenever the reader goroutine drains the source, including when the player was merely paused, since `Pause` does not stop that goroutine. oto refuses to play only when `p.eof && len(p.buf) == 0`, and a paused oto player keeps filling its buffer without consuming it, so `Play` resumes there. I checked the double directly: after `Play`, `Pause`, and a drain, `eof` is true and the next `Play` is refused, where a real player would resume.

No current test hits this, but it is a trap for the next one. This keeps the double correct in both directions, and the whole suite including `TestPositionNotGrowingAfterFinished` stays green with it:

```go
// The source is exhausted only if it was played through: a paused player still has
// unplayed data buffered in a real player.
if p.playing {
	p.eof = true
}
p.playing = false
```

### `updatePosition`'s `stopwatch.stop()` is not reliably covered

Reverting only that hunk leaves the entire suite green at `-count=3`, and `TestPositionNotGrowingAfterFinished` still passes even with a two second gap inserted before the final `Play` loop.

The reason is that `BufferedSize` is always 0 in the double, so `samples` reaches its final value on the same tick the player stops playing, and the `else` branch already resets the stopwatch. The interleaving that `stop()` guards, where `samples` goes final one tick before `IsPlaying` flips, only shows up occasionally: over 25 runs the final position was exactly 4s in 24 of them and 4s + 10.1ms once.

The change itself is right, since a real player drains a buffer and that interleaving is the normal case there. It is just unguarded. Giving the double a non-zero `BufferedSize` that drains would cover it.

### Nits

- `got > want+time.Second` for a four second source is about 100 times looser than the mechanism can produce, given the 0 to 10ms spread above. Some slop is genuinely needed for the reason in the previous section, but around 100ms would still be safe and would actually catch a regression.
- The same three line rationale appears verbatim in `Play` and in `startIfPending`, and the `eof` field comment in `export_test.go` repeats the comment inside `dummyPlayer.Play`.

### Behavior note

The final position can now land up to one tick past the source length, 10ms in the measurement above, instead of exactly on it, in exchange for monotonicity. That matches `Position`'s documented "increased monotonically", so I think it is the right trade. Just noting it as a deliberate consequence.

---

kumagi:
Thanks again! Pushed d5daa2e6d addressing both points and the nits:

- The dummy player now sets `eof` only when the source was actually played through, not when the reading goroutine drained it while paused, so a paused player can still resume, as oto does. I verified with a scratch test that `Play`, `Pause`, drain, `Play` now resumes.
- The dummy player gained a drainable buffered size: `SetBufferedDrainForTesting(n)` makes the simulated device consume `n` bytes per `BufferedSize` call and stop playing one tick after the last buffered byte is consumed, so `samples` reaches its final value while the player is still playing. The new `TestPositionNotGrowingAfterFinishedWithBufferedSource` drives this through the public API: reverting only the `stopwatch.stop()` hunk now fails it deterministically (`2.010s -> 2.114s`), and reverting the `Play` guard fails both tests. Default behavior is unchanged (drain 0), so the rest of the suite is untouched.
- Tightened the position bound from `want+time.Second` to `want+100*time.Millisecond`.
- Extracted the duplicated rationale into `playAndStartStopwatch` (player.go) and dropped the comment in `dummyPlayer.Play` that repeated the `eof` field comment.

`gofmt`, `go vet`, `internal/vettools` and `go test -race -count=3 -shuffle=on ./audio/...` pass.

hajimehoshi:
I verified every point and they all check out:

- Play through to the end, then `Play`, is refused; `Play`, `Pause`, drain, `Play` resumes. Confirmed at the double level.
- Reverting only the `stopwatch.stop()` hunk now fails `TestPositionNotGrowingAfterFinishedWithBufferedSource` 5 times out of 5 (`2.010s` -> `2.115s`), and reverting the `IsPlaying` guard fails both tests. The drainable buffered size does exactly what it needs to: there is now a real tick where `samples` is final while the player is still playing, followed by one where it is not.
- Default behavior is unchanged with a drain of 0, and the rest of the suite is untouched.
- The 100ms bound is stable over 30 consecutive runs; the observed overshoot is 10 to 22ms.
- `gofmt`, `go vet`, `internal/vettools` and `./audio/...` all pass.

`playAndStartStopwatch` states the rationale once and carries the lock precondition. Nice.

One real defect left, plus two small things.

### `dummyBufferedDrain` has a data race

`go test -race ./audio/` fails, 2 runs out of 10 with `-shuffle=on` and 3 out of 10 without:

```
WARNING: DATA RACE
Write at 0x000101053f40 by goroutine 85:
  audio.SetBufferedDrainForTesting()        export_test.go:63
  audio_test.TestPositionNotGrowingAfterFinishedWithBufferedSource()  position_test.go:91

Previous read at 0x000101053f40 by goroutine 76:
  audio.(*dummyContext).NewPlayer()         export_test.go:70
  audio.(*playerImpl).ensurePlayer()        player.go:222
  audio.(*playerImpl).startIfPending()      player.go:241
  audio.(*Context).updatePlayers()          audio.go:295
  audio.NewContext.func4()                  audio.go:161
```

The reading goroutine belongs to an earlier test's context. `ResetContextForTesting` only clears `theContext`, so the goroutine that `NewContext` started keeps running and calls `NewPlayer` while the next test writes the global.

CI runs `-shuffle=on` but not `-race`, so this would not be caught there, but `go test -race` is a normal thing to run locally. An atomic fixes it; I measured 0 failures out of 12 with:

```go
var dummyBufferedDrain atomic.Int64

func SetBufferedDrainForTesting(drainSize int) {
	dummyBufferedDrain.Store(int64(drainSize))
}
// ...
	bufferedDrain: int(dummyBufferedDrain.Load()),
```

### `Seek` does not reset `buffered`

It clears `eof` and `drained` but leaves `buffered`. I measured 180224 bytes both before and after a `Seek`. A real player discards its buffer on a seek, so a stale `buffered` would make `samples` too small afterwards. It is latent today, since no test combines a drain with a seek and a drain of 0 never reads the field, but it is the same kind of trap the pause fix just closed. One line next to the other two resets.

### Nit

`BufferedSize` now clears `playing`, so the point at which the double stops playing depends on someone polling it. That is safe because `updatePosition` is the only caller and runs once per tick before the `IsPlaying` check, and the `drained` field comment half explains it, but the coupling is worth stating on `BufferedSize` itself.

Apart from the race this looks ready.

---

kumagi:
All three fixed in 689ba15b1:

- `dummyBufferedDrain` is now an `atomic.Int64`, with a comment noting why (a goroutine of an earlier test's context can create a player after `ResetContextForTesting`). Before the fix I reproduced the race; after it, 10 runs of `go test -race -count=2 -shuffle=on ./audio/` report no race.
- `Seek` now clears `buffered` next to `eof` and `drained`.
- `BufferedSize` documents the coupling itself: it clears the playing state when the device has drained everything, and the context calls it exactly once per tick before checking `IsPlaying`.

`gofmt`, `go vet`, `internal/vettools` and `go test -race -count=3 -shuffle=on ./audio/...` pass.

audio/export_test.go
```diff
@@ -136,6 +205,13 @@ func (p *dummyPlayer) SetBufferSize(bufferSize int) {
 }
 
 func (p *dummyPlayer) Seek(offset int64, whence int) (int64, error) {
+	// Seeking discards the buffered data and resets the finished state as real players do, so
+	// that the source can be played again.
+	p.mu.Lock()
+	p.buffered = 0
+	p.eof = false
+	p.drained = false
+	p.mu.Unlock()
```
hajimehoshi: Use defer for unlocking

audio/position_test.go
```diff
@@ -0,0 +1,134 @@
...
+	src := bytes.NewReader(make([]byte, 44100*8*4))
+	p, err := context.NewPlayerF32(src)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	p.Play()
+	// Wait until the player finishes its source.
+	finished := false
+	for i := 0; i < 200; i++ {
```
hajimehoshi: Use for-range

```go
for range 200
```

audio/position_test.go
```diff
@@ -0,0 +1,134 @@
...
+	// 44100 [Hz] * 8 [bytes/sample] is one second of 32bit float stereo audio.
+	src := bytes.NewReader(make([]byte, 44100*8*4))
+	p, err := context.NewPlayerF32(src)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	p.Play()
+	// Wait until the player finishes its source.
+	finished := false
```
hajimehoshi:
```go
var finished bool
```

audio/position_test.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+		time.Sleep(5 * time.Millisecond)
+	}
+	if !finished {
+		t.Fatal("time out: the player did not finish")
+	}
+
+	// The context does not update a finished player anymore, so play and update a few more times
+	// to let the position settle with the exhausted source.
+	for i := 0; i < 5; i++ {
```
hajimehoshi: for-range

audio/position_test.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+	end := p.Position()
+	if got, want := end, 4*time.Second; got < want || got > want+100*time.Millisecond {
+		t.Errorf("Position() after the player finished: got: %v, want: %v or a bit larger", got, want)
+	}
+
+	// Play on a finished player does nothing, so the position must not grow anymore.
+	for i := 0; i < 20; i++ {
```
hajimehoshi: for-range

audio/position_test.go
```diff
@@ -0,0 +1,134 @@
...
+	src := bytes.NewReader(make([]byte, 44100*8*2))
+	p, err := context.NewPlayerF32(src)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	p.Play()
+	// Wait until the player drains both its source and its buffer.
+	finished := false
+	for i := 0; i < 500; i++ {
```
hajimehoshi: for-range

audio/position_test.go
```diff
@@ -0,0 +1,134 @@
...
+		t.Fatal("time out: the player did not finish")
+	}
+
+	end := p.Position()
+	if got, want := end, 2*time.Second; got < want || got > want+100*time.Millisecond {
+		t.Errorf("Position() after the player finished: got: %v, want: %v or a bit larger", got, want)
+	}
+
+	// Play on a finished player does nothing, so the position must not grow anymore.
+	for i := 0; i < 20; i++ {
```
hajimehoshi: for-range

audio/position_test.go
```diff
@@ -0,0 +1,134 @@
+// Copyright 2026 The Ebitengine Authors
```
hajimehoshi: Move tests to `audio_test.go`

audio/position_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+import (
+	"bytes"
+	"testing"
+	"time"
+
+	"github.com/hajimehoshi/ebiten/v2/audio"
+)
+
+// The position must not grow after the player finished its source, even if Play is called again.
```
hajimehoshi: Remove the comment

audio/position_test.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	if got := p.Position(); got != end {
+		t.Errorf("position grew after the player finished: %v -> %v", end, got)
+	}
+}
+
+// The position must not grow after the player finished, even when the underlying player keeps
+// playing its buffered data after its source is exhausted, as real players do. In that case the
+// consumed samples reach their final value before the player stops playing, and the stopwatch must
+// be stopped in the update then to freeze the position.
```
hajimehoshi: Remove comments

kumagi:
Thanks for the review! Addressed the remaining inline comments in 4fc4030bc:

- Moved the position tests into `audio_test.go` and removed `position_test.go`.
- Removed the test-name comments.
- Used `var finished bool` and `for range N` in all the loops.
- `dummyPlayer.Seek` now unlocks with `defer`.

`gofmt`, `go vet`, `internal/vettools` and `go test -race -count=2 -shuffle=on ./audio/...` pass. Reverting the player.go fix still fails both tests (4.000s -> 4.072s and 2.011s -> 2.086s), so the regression coverage is intact.

audio/export_test.go
kumagi: Fixed in 4fc4030bc.

audio/position_test.go
kumagi: Fixed in 4fc4030bc.

audio/position_test.go
kumagi: Fixed in 4fc4030bc.

audio/position_test.go
kumagi: Fixed in 4fc4030bc.

audio/position_test.go
kumagi: Fixed in 4fc4030bc.

audio/position_test.go
kumagi: Fixed in 4fc4030bc.

audio/position_test.go
kumagi: Fixed in 4fc4030bc.

audio/position_test.go
kumagi: Fixed in 4fc4030bc.

audio/position_test.go
kumagi: Fixed in 4fc4030bc.

audio/position_test.go
kumagi: Fixed in 4fc4030bc.

audio/export_test.go
```diff
@@ -65,13 +97,17 @@ func (p *dummyPlayer) Pause() {
 
 func (p *dummyPlayer) Play() {
 	p.mu.Lock()
+	if p.eof {
+		p.mu.Unlock()
```
hajimehoshi: Can we avoid unlocking without defer?

kumagi: Yes, fixed in 0c320aa59: `Play` now holds the mutex with `defer`, the buffered byte accounting is moved into `readOnce` which already locks with `defer`, and the goroutine locks with `defer` again at its end, so no manual unlock is left in the dummy player. (Note: the branch was also rebased onto the latest main, so the earlier commit hashes changed.)

hajimehoshi (approved):
LGTM

PR #3560, opened 2026-08-28
audio/wav: avoid allocating a buffer for a huge chunk size

kumagi:
None

bug

## What this PR does | solves

The chunk size in a WAV header is unvalidated, and decode allocated a buffer of that size, so a corrupt file could exhaust memory. Read the fmt chunk's first 16 bytes into a fixed buffer and skip the rest with io.CopyN instead of allocating.

### DIFF
--- a/audio/wav/decode_test.go
+++ b/audio/wav/decode_test.go
@@ -1,0 +1,119 @@
+// Copyright 2026 The Ebitengine Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package wav_test
+
+import (
+	"bytes"
+	"encoding/binary"
+	"io"
+	"testing"
+
+	"github.com/hajimehoshi/ebiten/v2/audio/wav"
+)
+
+func TestDecodeInvalidHugeChunkSize(t *testing.T) {
+	buf := []byte("RIFF")
+	buf = binary.LittleEndian.AppendUint32(buf, 0)
+	buf = append(buf, "WAVE"...)
+	buf = append(buf, "fmt "...)
+	buf = binary.LittleEndian.AppendUint32(buf, 0x7fffffff)
+	buf = binary.LittleEndian.AppendUint16(buf, 1)
+	buf = binary.LittleEndian.AppendUint16(buf, 1)
+	buf = binary.LittleEndian.AppendUint32(buf, 8000)
+	buf = binary.LittleEndian.AppendUint32(buf, 8000)
+	buf = binary.LittleEndian.AppendUint16(buf, 1)
+	buf = binary.LittleEndian.AppendUint16(buf, 16)
+
+	if _, err := wav.DecodeWithoutResampling(bytes.NewReader(buf)); err == nil {
+		t.Errorf("DecodeWithoutResampling: got no error, want an error")
+	}
+}
+
+func TestDecodeInvalidHugeUnknownChunkSize(t *testing.T) {
+	buf := []byte("RIFF")
+	buf = binary.LittleEndian.AppendUint32(buf, 0)
+	buf = append(buf, "WAVE"...)
+	buf = append(buf, "JUNK"...)
+	buf = binary.LittleEndian.AppendUint32(buf, 0x7fffffff)
+	buf = append(buf, 0, 0)
+
+	if _, err := wav.DecodeWithoutResampling(bytes.NewReader(buf)); err == nil {
+		t.Errorf("DecodeWithoutResampling: got no error, want an error")
+	}
+}
+
+// TestDecodeFmtChunkLargerThan16 tests a 'fmt ' chunk larger than 16 bytes.
+func TestDecodeFmtChunkLargerThan16(t *testing.T) {
+	data := []byte{1, 2, 3, 4, 5, 6, 7, 8}
+	buf := []byte("RIFF")
+	buf = binary.LittleEndian.AppendUint32(buf, uint32(len(data)+38))
+	buf = append(buf, "WAVE"...)
+	buf = append(buf, "fmt "...)
+	buf = binary.LittleEndian.AppendUint32(buf, 18)
+	buf = binary.LittleEndian.AppendUint16(buf, 1)
+	buf = binary.LittleEndian.AppendUint16(buf, 2)
+	buf = binary.LittleEndian.AppendUint32(buf, 8000)
+	buf = binary.LittleEndian.AppendUint32(buf, 32000)
+	buf = binary.LittleEndian.AppendUint16(buf, 4)
+	buf = binary.LittleEndian.AppendUint16(buf, 16)
+	buf = binary.LittleEndian.AppendUint16(buf, 0)
+	buf = append(buf, "data"...)
+	buf = binary.LittleEndian.AppendUint32(buf, uint32(len(data)))
+	buf = append(buf, data...)
+
+	s, err := wav.DecodeWithoutResampling(bytes.NewReader(buf))
+	if err != nil {
+		t.Fatal(err)
+	}
+	if got, want := s.Length(), int64(len(data)); got != want {
+		t.Errorf("Length(): got: %d, want: %d", got, want)
+	}
+	if _, err := s.Seek(0, io.SeekStart); err != nil {
+		t.Fatal(err)
+	}
+	got, err := io.ReadAll(s)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !bytes.Equal(got, data) {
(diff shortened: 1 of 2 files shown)


audio/wav/decode_test.go
```diff
@@ -0,0 +1,78 @@
+// Copyright 2026 The Ebiten Authors
```
hajimehoshi: Ebitengine

kumagi: Fixed.

hajimehoshi:
Thanks for the patch. The core change looks correct to me:

* `size` is assembled from four bytes into an `int64`, so it is always in `[0, 0xFFFFFFFF]`; together with the existing `size < 16` guard, `size-16` can never go negative.
* `io.CopyN(io.Discard, src, ...)` is genuinely bounded: `*io.LimitedReader` has no `WriteTo`, so `io.Copy` takes `io.Discard`'s `ReadFrom`, which uses an 8 KiB pooled buffer.
* Renaming the inner buffer to `fmtBuf` also removes the shadowing of the outer 8-byte chunk header `buf`.

Two things I would like to see addressed:

**1. The new tests pass without the fix.**

I reverted `decode.go` to `main` and kept `decode_test.go`, and both new tests still pass:

```
=== RUN   TestDecodeInvalidHugeChunkSize
--- PASS: TestDecodeInvalidHugeChunkSize (0.00s)
=== RUN   TestDecodeInvalidHugeUnknownChunkSize
--- PASS: TestDecodeInvalidHugeUnknownChunkSize (0.31s)
```

That is expected: the old code also returned an error, it just allocated ~2 GiB first, which succeeds on a 64-bit host because the pages are never touched. So the PR does not currently have a regression test for the behavior it changes. Asserting on the allocation would pin it. With the `JUNK` chunk input I measured `TotalAlloc` deltas of 2,147,518,624 bytes before the fix and 10,160 bytes after:

```go
var before, after runtime.MemStats
runtime.ReadMemStats(&before)
_, _ = DecodeWithoutResampling(bytes.NewReader(buf))
runtime.ReadMemStats(&after)
if delta := after.TotalAlloc - before.TotalAlloc; delta > 1<<20 {
	t.Errorf("DecodeWithoutResampling allocated %d bytes, want a small constant", delta)
}
```

**2. `package wav` should be `package wav_test`.**

`decode_test.go` uses only the exported API, so it can be an external test package. That is the convention in this repository: `audio/vorbis/vorbis_test.go` and `audio/vorbis/int16_test.go` are `package vorbis_test`, and only `export_test.go` shares the package name.

Nits, take them or leave them:

* `wn` is an unusual name here; the rest of the file uses `n`. In the `default` branch `n` is free, and in the `fmt ` branch it can be a plain reassignment.
* The title says "reject", but a file that really does contain a huge `fmt ` chunk is still accepted, just streamed instead of buffered. Something like "avoid allocating a buffer for a huge chunk size" describes the change more precisely.

---

kumagi:
Thanks for the review! Fixed in 0db78c4:

1. **Regression tests**: Added the `runtime.MemStats` allocation assertion to both huge-chunk tests. I verified they fail against `decode.go` at `main`: both report ~2 GiB allocated (`allocated 2147489320 bytes, want a small constant` and `allocated 2147489336 bytes, want a small constant`), and pass with the fix.
2. **Package**: `decode_test.go` is now `package wav_test`.
3. **Nit (`wn`)**: The `default` branch now uses `n`. In the `fmt ` branch a plain reassignment doesn't type-check since `io.CopyN` returns an `int64` while `n` from `io.ReadFull` is an `int`, so the count is discarded there: `if _, err := io.CopyN(io.Discard, src, size-16); err != nil`.
4. **Nit (title)**: Renamed the PR to "audio/wav: avoid allocating a buffer for a huge chunk size".

hajimehoshi:
Thanks! The fix itself looks right to me, and I confirmed it is behavior-preserving for valid files: in the `fmt ` branch, 16 bytes read plus `size-16` skipped consumes exactly `size`, so `headerSize` and the resulting `sectionReader` offset are unchanged. `io.CopyN` also returns a non-nil error whenever fewer than `n` bytes are copied, so the truncation checks are sound.

A few requests before merging.

## Please don't use `runtime.ReadMemStats` in tests

`TotalAlloc` is process-global and cumulative across every goroutine, so the assertion only works because `audio/wav` happens to have no other goroutines today, and nothing keeps that true. The value also moves under `-race`, `-cover`, `GOARCH=386` and `GOOS=js`, all of which CI runs, and `1<<20` encodes nothing about the actual contract. A test whose assertion depends on the allocator's accounting is not one we can debug when it fails. `runtime.ReadMemStats` is not used anywhere else in this repository, and I'd like to keep it that way.

Please drop the `runtime` import, the `MemStats` values, the `ReadMemStats` calls and the `TotalAlloc` delta checks from both tests. What remains in each is the `err == nil` check, which is the real behavioral claim: a chunk header declaring more bytes than the file contains must be rejected.

## Please add a test for a `fmt ` chunk larger than 16 bytes

This is the only new code path a well-formed file takes, and it is currently untested. It matters more than it looks: `sectionReader.Read` never touches `offset`, so a wrong `headerSize` is invisible to `Read` and shows up only through `Seek`. Any test of this path has to seek.

```go
// TestDecodeFmtChunkLargerThan16 tests a 'fmt ' chunk larger than 16 bytes.
func TestDecodeFmtChunkLargerThan16(t *testing.T) {
	data := []byte{1, 2, 3, 4, 5, 6, 7, 8}
	buf := []byte("RIFF")
	buf = binary.LittleEndian.AppendUint32(buf, uint32(len(data)+38))
	buf = append(buf, "WAVE"...)
	buf = append(buf, "fmt "...)
	buf = binary.LittleEndian.AppendUint32(buf, 18)
	buf = binary.LittleEndian.AppendUint16(buf, 1)
	buf = binary.LittleEndian.AppendUint16(buf, 2)
	buf = binary.LittleEndian.AppendUint32(buf, 8000)
	buf = binary.LittleEndian.AppendUint32(buf, 32000)
	buf = binary.LittleEndian.AppendUint16(buf, 4)
	buf = binary.LittleEndian.AppendUint16(buf, 16)
	buf = binary.LittleEndian.AppendUint16(buf, 0)
	buf = append(buf, "data"...)
	buf = binary.LittleEndian.AppendUint32(buf, uint32(len(data)))
	buf = append(buf, data...)

	s, err := wav.DecodeWithoutResampling(bytes.NewReader(buf))
	if err != nil {
		t.Fatal(err)
	}
	if got, want := s.Length(), int64(len(data)); got != want {
		t.Errorf("Length(): got: %d, want: %d", got, want)
	}
	if _, err := s.Seek(0, io.SeekStart); err != nil {
		t.Fatal(err)
	}
	got, err := io.ReadAll(s)
	if err != nil {
		t.Fatal(err)
	}
	if !bytes.Equal(got, data) {
		t.Errorf("data: got: %v, want: %v", got, data)
	}
}
```

Stereo 16bit, so no conversion wrapper sits between the section reader and the assertion.

I mutation-checked this against your `decode.go`:

| mutation | caught by |
| --- | --- |
| `headerSize += size` -> `headerSize += 16` | only this test (`got [0 0 1 2 3 4 5 6]`) |
| `io.CopyN(..., size-16)` -> `io.CopyN(..., size)` | this test and `TestDecodeValid` |

The first one is the realistic mistake, and nothing in the PR as it stands catches it.

## A dead check in the unknown-chunk branch

```go
n, err := io.CopyN(io.Discard, src, size)
if n != size {
	return nil, fmt.Errorf("wav: invalid header")
}
if err != nil {
	return nil, err
}
```

`io.CopyN` returns a nil error if and only if `n == size`, so the second check is unreachable. Collapsing it to the same shape as the `fmt ` branch just above makes the two skips read alike:

```go
if _, err := io.CopyN(io.Discard, src, size); err != nil {
	return nil, fmt.Errorf("wav: invalid header")
}
```

## Commit message

The first commit's subject is `audio/wav: reject a malformed chunk with a huge size`, but the change doesn't reject a huge size, it stops allocating for one. The PR title (`audio/wav: avoid allocating a buffer for a huge chunk size`) is the accurate wording.

---

For reference, with the changes above applied locally: `gofmt`, `go vet` and `internal/vettools` pass, the test binary cross-builds for `linux/386`, `js/wasm` and `windows/amd64`, and the tests pass under `-count=20 -shuffle=on` and `-race`.

kumagi:
Thanks for the thorough review! Addressed all four points in the latest force-push:

1. **`runtime.ReadMemStats` removed** — dropped the `runtime` import, the `MemStats` values, and the `TotalAlloc` delta checks from both huge-chunk tests. Each now asserts only the behavioral claim: a chunk header declaring more bytes than the file contains must be rejected.
2. **Added `TestDecodeFmtChunkLargerThan16`** — a stereo 16-bit file with an 18-byte `fmt ` chunk, asserting `Length()` and round-tripping the data after a `Seek(0, io.SeekStart)`. I confirmed it catches the `headerSize += 16` mutation (`got: [0 0 1 2 3 4 5 6]`) and fails against the `io.CopyN(..., size)` mutation as well.
3. **Collapsed the dead check** in the unknown-chunk branch to `if _, err := io.CopyN(io.Discard, src, size); err != nil`, matching the `fmt ` branch.
4. **Commit messages**: the fix commit is now `audio/wav: avoid allocating a buffer for a huge chunk size`.

Verified locally: `gofmt`, `go vet` and `internal/vettools` pass, the test binary cross-builds for `linux/386`, `js/wasm` and `windows/amd64`, and the tests pass under `-count=5 -shuffle=on` and `-race`.

hajimehoshi (approved):
LGTM

PR #3561, opened 2026-08-28
internal/mipmap: avoid a division by zero for an extreme scale

kumagi:
None

bug

## What this PR does | solves
mipmapLevelFromDistance computed 1<<level to adjust the mipmap level. On 32-bit platforms a large level overflowed to zero, causing a division by zero. Use sizeForLevel, which divides iteratively without shifting.

### DIFF
--- a/internal/mipmap/mipmap_test.go
+++ b/internal/mipmap/mipmap_test.go
@@ -1,0 +1,66 @@
+// Copyright 2026 The Ebitengine Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


internal/mipmap/mipmap_test.go
```diff
@@ -0,0 +1,66 @@
+// Copyright 2026 The Ebiten Authors
```
hajimehoshi: Ebitengine

kumagi: fixed.

hajimehoshi (approved):
LGTM

PR #3562, opened 2026-08-28
text/v2: make LimitedFace.AdvanceAt and MultiFace.AdvanceAt stop at the first line break

kumagi:
None

bug

## What this PR does | solves
`LimitedFace.advanceAt` and `MultiFace.advanceAt` filtered/split the whole text before measuring, so a line break that was not in the allowed range or not in the first chunk was no longer seen as a line break. The underlying measurement then advanced the entire remaining text as one line instead of stopping at the first line, returning a much larger value than the base face for a text with multiple lines.

Restrict the text to its first line before filtering or splitting, matching the base face's AdvanceAt semantics.

Add TestLimitedFaceAdvanceAtNewline and TestMultiFaceAdvanceAtNewline, which fail without these fixes, and extend TestAdvanceAtLineBreak with LimitedFace and MultiFace cases.


hajimehoshi:
Thanks! The `LimitedFace` change itself looks right, but it isn't the whole fix: `MultiFace.advanceAt` has the same bug and isn't touched here.

`doSplitText` splits the whole text by face with no line-break awareness, and `advanceAt` then sums every chunk whose `textEndIndex <= indexInBytes`, so chunks that live past the first line break get added in.

```go
l := text.NewLimitedFace(base)
l.AddUnicodeRange('A', 'A')
m, _ := text.NewMultiFace(l, base)
text.AdvanceAt("A\nA", 3, m) // 21.34375, want 10.671875 - exactly 2x one 'A'
```

The chunker assigns `A` to face0, `\n` to face1, `A` to face0, and all three chunks are summed. A `LimitedFace` isn't needed to trigger this: any face switch after the line break does it, e.g. a Latin first line and a CJK second line in a Latin+CJK `MultiFace`. The sub-faces can't catch it either, since each chunk is measured in isolation and looks like a complete line to them.

`GoXFace` and `GoTextFace` are both fine (they already stop at the first line break, including the RTL and vertical paths), so `MultiFace` is the only remaining one.

Could you give `MultiFace.advanceAt` the same treatment as `LimitedFace.advanceAt` (clamp `indexInBytes` to `textutil.FirstLineLen(text)` and split only the first line), with a test alongside `TestLimitedFaceAdvanceAtNewline`?

---

(Reviewed by Claude (Claude Code), posted on behalf of @hajimehoshi.)

kumagi:
Thanks for the detailed review. You're right — the LimitedFace change alone
wasn't the whole fix. MultiFace.advanceAt had the same bug, and I've now given
it the same treatment.

The fix in MultiFace.advanceAt:
- Clamp indexInBytes to textutil.FirstLineLen(text), matching the base faces'
  AdvanceAt semantics (first line only).
- Split only the first line (m.splitText(text[:firstLineLen])) instead of the
  whole text, so chunks past the first line break can no longer enter the sum.
- Added the textutil import to multi.go.

Verified with your reproduction:

    text.AdvanceAt("A\nA", 3, m)  // was 21.34375, now 10.671875 == one 'A'

Added TestMultiFaceAdvanceAtNewline alongside TestLimitedFaceAdvanceAtNewline,
which asserts AdvanceAt("A\nA", 3, m) equals AdvanceAt("A", 1, base). Without
the fix it returns 21.34375 and fails; with the fix it passes. GoXFace and
GoTextFace were confirmed fine as you noted, so MultiFace is now covered too.

The branch has been rebased onto the current main and the commit amended to
describe both fixes. All text/v2 tests pass.

Reviewed by DeepSeek V4 Flash (opencode-go/deepseek-v4-flash), on behalf of @kumagi.

text/v2/face_test.go
```diff
@@ -0,0 +1,63 @@
+// Copyright 2026 The Ebitengine Authors
```
hajimehoshi, line 1: The test file name is odd. `face_test.go`, maybe

kumagi: renamed.

text/v2/limited.go
```diff
@@ -51,17 +51,22 @@ func (l *LimitedFace) Metrics() Metrics {
 
 // advanceAt implements Face.
 func (l *LimitedFace) advanceAt(text string, indexInBytes int) float64 {
-	filtered := l.unicodeRanges.Filter(text)
-	if filtered == text {
+	firstLineLen := textutil.FirstLineLen(text)
+	if indexInBytes > firstLineLen {
+		indexInBytes = firstLineLen
+	}
```
hajimehoshi: indexInBytes = min(indexInBytes, firstLineLen)

hajimehoshi:
The `MultiFace` fix looks right. I applied the branch locally: both tests fail without it and pass with it, and the first-line contract now holds for `LimitedFace` and `MultiFace` across every line-break codepoint, `\r\n`, empty first lines, nested `MultiFace`, and the deprecated `Advance`.

Three follow-ups, on top of the `min` comment above (which applies to `multi.go` as well).

### Tests

`TestLimitedFaceAdvanceAtNewline` only exercises the fast path. `'A'`-`'D'` covers `"AB"`, so `filtered == firstLine` and the U+FFFD index translation loop, which this PR also touches, never runs. A range that excludes a rune inside the first line reaches it, e.g. `AddUnicodeRange('a', 'd')` with `"aZb\ncd"`, compared against `AdvanceAt("aZb", min(i, 3), l)`.

Separately, `TestAdvanceAtLineBreak` in `text_test.go` already sweeps `\n`, `\r`, `\v`, `\f`, U+0085, U+2028, U+2029 and `\r\n` over every byte index, with a face table. Adding the two wrappers there covers the clamping completely, with no new helper or font dependency:

```go
goTextFace := &text.GoTextFace{
    Source:    source,
    Size:      24,
    Direction: text.DirectionLeftToRight,
}
limitedFace := text.NewLimitedFace(goTextFace)
limitedFace.AddUnicodeRange('a', 'd')
multiFace, err := text.NewMultiFace(limitedFace, goTextFace)
if err != nil {
    t.Fatal(err)
}
```

plus `{name: "LimitedFace", face: limitedFace}` and `{name: "MultiFace", face: multiFace}` in `cases`. I checked this: it fails without the fix (`AdvanceAt("a\r\nbcd", 6, _) = 53.640625, want 13.15625`) and passes with it.

The two suggestions are complementary: the table entries cover the clamp for both wrappers, the `"aZb\ncd"` case covers the filtering path. If you move the tests into the table, `face_test.go` is no longer needed; keeping the separate file and widening it there is also fine.

### `advanceAt` contract

Not part of this PR, and nothing to do here: the `advanceAt` method of the `Face` interface has no doc comment, unlike its `appendLazyGlyphsForLine` / `appendVectorPathForLine` siblings whose names carry the per-line rule. Two implementations diverged from it, so I will document it on the interface separately.

**EDIT (2026-08-29)**: This is already done

### Title and description

The PR title and description still describe `LimitedFace` only. The commit message headline covers both. Could you update the title and description to match?

---

(Reviewed by Claude (Claude Code), posted on behalf of @hajimehoshi.)

kumagi:
Thanks for the thorough review! All three follow-ups are addressed in 7e9c1ae98:

- `indexInBytes = min(indexInBytes, firstLineLen)` is now used in both `limited.go` and `multi.go`.
- `TestLimitedFaceAdvanceAtNewline` gained the `AddUnicodeRange('a', 'd')` + `"aZb\ncd"` case compared against `AdvanceAt("aZb", min(i, 3), l)`, so the U+FFFD index translation path is exercised.
- `TestAdvanceAtLineBreak` now has `LimitedFace` and `MultiFace` entries using the suggested setup, covering the clamp across every line-break codepoint.

I kept `face_test.go` (widened) as you noted that's fine. The PR title and description now describe both `LimitedFace.AdvanceAt` and `MultiFace.AdvanceAt` as well.

All `text/v2` tests pass with the fix and fail without it.

hajimehoshi:
> covering the clamp across every line-break codepoint.

Hmm? I failed to find this. Do the tests cover all the new line chars?

kumagi:
Good catch — my wording was misleading. The full sweep was only implicit: the new cases run through TestAdvanceAtLineBreak's existing `tails` list, which is unchanged code and outside the diff context, so it's easy to miss. The dedicated tests themselves only used `\n`.

I've widened TestLimitedFaceAdvanceAtNewline and TestMultiFaceAdvanceAtNewline to sweep `\n`, `\v`, `\f`, `\r`, U+0085, U+2028, U+2029 and `\r\n` at every byte index, so the coverage is visible in the new file too. They fail for every codepoint without the fix and pass with it.

hajimehoshi (approved):
LGTM

PR #3566, opened 2026-08-29
ebiten: document ReadPixels panic on disposed images

### LINKED ISSUE #3556: ebiten: (*Image).ReadPixels panics for a disposed image, contradicting its documentation
### Ebitengine Version

v2.10.0-alpha.13 (main, cbe2d1dec)

### Operating System

- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

(Verified on macOS, but the code path is platform-independent: the panic happens before any graphics backend is touched.)

go version go1.27.0 darwin/arm64

Call `(*ebiten.Image).ReadPixels` on an image that has been disposed:

```go
func TestDisposedReadPixels(t *testing.T) {
	img := ebiten.NewImage(16, 16)
	img.Dispose()

	pix := make([]byte, 4*16*16)
	img.ReadPixels(pix) // panics
}
```

`ReadPixels` fills `pixels` with transparent color and returns, as its documentation states:

> ReadPixels always sets a transparent color if the image is disposed.

`ReadPixels` even has code for exactly this:

```go
	if i.isDisposed() {
		for i := range pixels {
			pixels[i] = 0
		}
		return
	}
```

### What happens instead?

It panics:

```
panic: ebiten: the image is already disposed
```

The cause is the statement order in `ReadPixels` (image.go): the length check runs first, and it calls `Bounds`, which panics for a disposed image.

```go
func (i *Image) ReadPixels(pixels []byte) {
	b := i.Bounds() // panics when the image is disposed
	if got, want := len(pixels), 4*b.Dx()*b.Dy(); got != want {
		panic(fmt.Sprintf("ebiten: len(pixels) must be %d but %d at ReadPixels", want, got))
	}

	if i.isDisposed() { // unreachable
		...
```

```go
func (i *Image) Bounds() image.Rectangle {
	if i.isDisposed() {
		panic("ebiten: the image is already disposed")
	}
	return i.bounds
}
```

So the `isDisposed` branch in `ReadPixels` is dead code, and the documented behavior has never been observable. This has been the case since `ReadPixels` was added in 81bd5b488 (v2.4).

`ReadPixels` is the only accessor with this ordering. The siblings all check `isDisposed` before touching `Bounds`, so they do behave as documented:

* `At` / `RGBA64At` go through `at`, which returns early when disposed (`TestImageDispose` covers this).
* `Fill` (and thus `Clear`) and `WritePixels` return early when disposed.

Two ways to resolve it, depending on which side is considered correct:

1. Move the `isDisposed` check above the `Bounds` call in `ReadPixels`, matching the documentation and the sibling methods. Note that the length check would then be skipped for a disposed image, unless `i.bounds` is read directly instead of through `Bounds`.
2. Change the documentation to say that `ReadPixels` panics for a disposed image, and delete the dead branch. That would make `ReadPixels` inconsistent with `At`, which is documented and implemented as returning transparent.

---

Filed by Claude (Claude Code), on behalf of @hajimehoshi.

pttydou:
Closes #3556

bug

## What this PR does | solves
`ReadPixels` now documents its existing panic for disposed images and removes the unreachable transparent-fill branch. This keeps the API from treating a disposed image as if it still had bounds.

The regression test asserts that `ReadPixels` panics after `Dispose`.

Tests:
- `go test -run '^TestImageReadPixelsDispose$' .`
- `go test .`
- `go vet .`

### DIFF
--- a/image.go
+++ b/image.go
@@ -1243,7 +1243,7 @@ func (i *Image) ColorModel() color.Model {
 //
 // ReadPixels loads pixels from GPU to system memory if necessary, which means that ReadPixels can be slow.
 //
-// ReadPixels always sets a transparent color if the image is disposed.
+// ReadPixels panics if the image is disposed.
 //
 // len(pixels) must be 4 * (bounds width) * (bounds height).
 // If len(pixels) is not correct, ReadPixels panics.
@@ -1255,16 +1255,9 @@ func (i *Image) ColorModel() color.Model {
 //
 // ReadPixels can't be called outside the main loop (ebiten.Run's updating function) starts.
 func (i *Image) ReadPixels(pixels []byte) {
-	b := i.bounds
+	b := i.Bounds()
 	if got, want := len(pixels), 4*b.Dx()*b.Dy(); got != want {
 		panic(fmt.Sprintf("ebiten: len(pixels) must be %d but %d at ReadPixels", want, got))
-	}
-
-	if i.isDisposed() {
-		for i := range pixels {
-			pixels[i] = 0
-		}
-		return
 	}
 
 	i.invokeUsageCallbacks()
--- a/image.go
+++ b/image.go
@@ -1255,7 +1255,7 @@ func (i *Image) ColorModel() color.Model {
 //
 // ReadPixels can't be called outside the main loop (ebiten.Run's updating function) starts.
 func (i *Image) ReadPixels(pixels []byte) {
-	b := i.Bounds()
+	b := i.bounds
 	if got, want := len(pixels), 4*b.Dx()*b.Dy(); got != want {
 		panic(fmt.Sprintf("ebiten: len(pixels) must be %d but %d at ReadPixels", want, got))
 	}
--- a/image_test.go
+++ b/image_test.go
@@ -333,33 +333,15 @@ func TestImageDispose(t *testing.T) {
 }
 
 func TestImageReadPixelsDispose(t *testing.T) {
-	img := ebiten.NewImageWithOptions(image.Rect(-1, -2, 2, 2), nil)
-	img.Fill(color.White)
-
-	pix := make([]byte, 4*3*4)
-	img.ReadPixels(pix)
-	for i, got := range pix {
-		if got != 0xff {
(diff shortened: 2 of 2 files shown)


hajimehoshi:
I have not determined what to do for this issue yet. I might leave panicking.

hajimehoshi:
I've decided to fill 0. Let me review this

hajimehoshi:
Sorry for going back and forth on this. I said above that I had decided to fill 0, but after looking at this patch I have changed my mind: I am going to leave the panic and fix the documentation instead.

The reason is the length rule. ReadPixels requires len(pixels) to equal the bounds area, so its contract depends on the image having a size. Bounds panics for a disposed image, which means a disposed image has no size as far as the API is concerned. Filling zeros therefore needs bounds that the API says are gone, and this patch gets them by reading the bounds field directly. That makes ReadPixels the only method for which a disposed image still has a shape, which is what I want to avoid.

At stays as it is, since a point query needs no size to be well defined, so it can return transparent without contradicting anything.

Dispose has also been deprecated since v2.7 in favor of Deallocate, and Deallocate has no such problem: the bounds remain valid and ReadPixels returns zeros through the normal path. I would rather not add a special case to the deprecated path.

So what I would like is:

Change the documentation of ReadPixels to state that it panics if the image is disposed, replacing the current sentence about a transparent color.
Delete the unreachable isDisposed branch.
Keep a test, inverted to assert the panic. The rest of your test can go away with the branch it covered.

If you would like to update this PR along those lines, that would be very welcome. Thank you for the patch and for your patience with my indecision.

pttydou:
Updated in bbcc4c715: `ReadPixels` now documents the disposed-image panic, the unreachable transparent-fill branch is removed, and the test only asserts the panic. I ran `go test -run '^TestImageReadPixelsDispose$' .`, `go test .`, and `go vet .`.

hajimehoshi (approved):
LGTM, thanks!

PR #3567, opened 2026-08-29
internal/gamepaddb: reject a GUID not exactly 16 bytes in the Android default mappings

kumagi:
None

bug

## What this PR does | solves
addAndroidDefaultMappings decodes the id with hex.DecodeString and reads bytes 12..15 of the result without checking its length. A shorter id that still decodes (e.g. 12 bytes) panicked with an index out of range.

Check that the decoded id is at least 16 bytes and reject it otherwise.

Add TestAndroidDefaultMappingsShortID, which panics without this fix.

### DIFF
--- a/internal/gamepaddb/gamepaddb.go
+++ b/internal/gamepaddb/gamepaddb.go
@@ -556,6 +556,9 @@ func addAndroidDefaultMappings(id string) bool {
 	if err != nil {
 		return false
 	}
+	if len(idBytes) < 16 {
+		return false
+	}
 	buttonMask := uint16(idBytes[12]) | (uint16(idBytes[13]) << 8)
 	axisMask := uint16(idBytes[14]) | (uint16(idBytes[15]) << 8)
 	if buttonMask == 0 && axisMask == 0 {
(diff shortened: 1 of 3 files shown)


internal/gamepaddb/gamepaddb.go
```diff
@@ -556,6 +556,9 @@ func addAndroidDefaultMappings(id string) bool {
 	if err != nil {
 		return false
 	}
+	if len(idBytes) < 16 {
```
hajimehoshi: `len(idBytes) != 16`? Too long ID should be invalid too

kumagi: changed to `!= 16`

hajimehoshi:
I think this check should be useful for any platforms, but why does this fix only for Android?

kumagi:
You're right, the length check is useful regardless of the platform, so I moved it into parseLine in ae7f956c9: an id whose hex-decoded length is not 16 bytes is now ignored on any platform.

I kept the check in addAndroidDefaultMappings as well, since the id passed there comes from the platform (Android) directly and doesn't necessarily go through parseLine.

I also checked the bundled gamecontrollerdb: all entries use 16-byte hex GUIDs except the two `xinput` lines, which SDL uses as an API selector rather than a real GUID. Ebitengine never matches them (Windows XInput devices are keyed by `78696e70757401000000000000000000`, which passes the check), so ignoring them doesn't change any behavior.

hajimehoshi:
As we discussed in another channel, let me revisit this issue and a fix

kumagi:
thanks.

PR #3568, opened 2026-08-29
internal/gamepaddb: take the write lock when reading may add the Anroid default mappings

kumagi:
None

bug

## What this PR does | solves
`buttonMappings` and `axisMappings` call `addAndroidDefaultMappings` when an id is not mapped yet, which writes `gamepadButtonMappings` and `gamepadAxisMappings`. The exported lookup functions held `mappingsM.RLock()` around these calls, so a read raced with a write from another goroutine also holding the read lock on Android.

`mappingsM` becomes a `sync.Mutex`, and every function that can reach `addAndroidDefaultMappings` (`buttonMapping`, `axisMapping`, `HasStandardLayoutMapping`, `HasStandardAxis`, `HasStandardButton`) holds it exclusively. `Name` only reads `gamepadNames`, but a plain mutex has no read lock anymore.

The functions that need a `GamepadState` (`StandardAxisValue`, `StandardButtonValue`, `IsStandardButtonPressed`) resolve the mapping under the lock and then read the device state after unlocking. `internal/gamepad` reaches this package in both orders: `Gamepad.IsStandardAxisAvailable` and friends hold the gamepad's own lock before calling in, while `Gamepad.StandardAxisValue` and friends pass the gamepad as the state, so holding `mappingsM` across a callback would let the two orders wait for each other and hang. `mapping` is a value type and both `Update` and `addAndroidDefaultMappings` install newly built maps rather than mutating published ones, so handing a mapping out is safe.

`IsStandardButtonPressed` used to read `gamepadButtonMappings[id]` directly, so it now goes through `buttonMapping` for consistency: it can register the Android default mappings like its siblings. Its callers already did that via `HasStandardLayoutMapping`, so nothing changes in practice.

Tests: `TestConcurrentAccess` hammers the package from several goroutines in both lock orders, with a concurrent `Update`, and reports a deadlock even without `-race` (which CI does not run) if a gamepad state is used while `mappingsM` is held.


internal/gamepaddb/android_mappings_test.go
```diff
@@ -0,0 +1,53 @@
+// Copyright 2026 The Ebitengine Authors
```
hajimehoshi: Why is this test file name Android specific?

`android_mappings_test.go`

kumagi: not specific. moved to the existing test file.

hajimehoshi:
I started to think the original implementation seems a little odd. Let me think

hajimehoshi:
The fix itself is good but I don't think we need tests

hajimehoshi:
Also, let's change mappingsM from `sync.RWMutex` to `sync.Mutex`

kumagi:
PTAL. I removed the tests and the test-only platform override, and changed mappingsM from sync.RWMutex to sync.Mutex so that all the lookup functions, including Name, take the write lock.

hajimehoshi (approved):
LGTM

hajimehoshi (changes requested):
Oops, my suggestion introduces deadlocks unfortunately. Let me think

hajimehoshi (commented):
Thanks for the report. The data race is real: `buttonMappings` and `axisMappings` call `addAndroidDefaultMappings`, which writes `gamepadButtonMappings` and `gamepadAxisMappings` while the caller holds only a read lock.

Taking the exclusive lock in those functions does fix the race, but it introduces a deadlock. `internal/gamepad` reaches `gamepaddb` in two opposite lock orders:

* `g.m` -> `mappingsM`: `Gamepad.IsStandardLayoutAvailable`, `IsStandardAxisAvailable` and `IsStandardButtonAvailable` hold the gamepad's own mutex and then call into `gamepaddb`.
* `mappingsM` -> `g.m`: `Gamepad.StandardAxisValue`, `StandardButtonValue` and `IsStandardButtonPressed` deliberately do *not* hold `g.m` (see the comment "StandardAxisValue invokes g.Axis, g.Button, or g.Hat so this cannot be locked") and pass the gamepad in as the `GamepadState`. `gamepaddb` then calls `state.IsAxisReady`, `Axis`, `Button` or `Hat` while holding `mappingsM`, and each of those takes `g.m`.

While `mappingsM` was an `RWMutex` and both sides used `RLock`, the two orders coexisted because read locks are shared. Once both sides take the lock exclusively they block each other. All of these entry points are exported as concurrent safe (`ebiten.IsStandardGamepadAxisAvailable`, `ebiten.StandardGamepadAxisValue`, ...), so a game that queries a gamepad from two goroutines can hang. This reproduces on both revisions of this PR (80003dce with `RWMutex` + `Lock`, and 8de4520 with `sync.Mutex`):

```go
// internal/gamepaddb, external test package.

// gamepadLike mimics internal/gamepad.Gamepad: every state query takes the gamepad's own mutex.
type gamepadLike struct {
	mu      sync.Mutex
	ready   chan struct{}
	proceed chan struct{}
	once    sync.Once
}

func (g *gamepadLike) IsAxisReady(index int) bool {
	g.mu.Lock()
	g.mu.Unlock()
	// Hand over to the other goroutine while gamepaddb's lock is held.
	g.once.Do(func() {
		close(g.ready)
		<-g.proceed
	})
	return true
}

func (g *gamepadLike) Axis(index int) float64 { g.mu.Lock(); defer g.mu.Unlock(); return 0 }
func (g *gamepadLike) Button(index int) bool  { g.mu.Lock(); defer g.mu.Unlock(); return false }
func (g *gamepadLike) Hat(index int) int      { g.mu.Lock(); defer g.mu.Unlock(); return 0 }

func TestLockOrderInversion(t *testing.T) {
	const id = "00000000000000000000000000000001"
	if err := gamepaddb.Update([]byte(id + ",Test Pad,a:b0,leftx:a0,\n")); err != nil {
		t.Fatal(err)
	}

	g := &gamepadLike{ready: make(chan struct{}), proceed: make(chan struct{})}
	done := make(chan struct{})

	// Gamepad.StandardAxisValue: gamepaddb takes its lock, then calls back into the gamepad.
	go func() {
		defer close(done)
		gamepaddb.StandardAxisValue(id, gamepaddb.StandardAxisLeftStickHorizontal, g)
	}()

	<-g.ready

	// Gamepad.IsStandardAxisAvailable: holds the gamepad's lock, then calls into gamepaddb.
	go func() {
		g.mu.Lock()
		defer g.mu.Unlock()
		close(g.proceed)
		gamepaddb.HasStandardAxis(id, gamepaddb.StandardAxisLeftStickHorizontal)
	}()

	<-done
}
```

```
panic: test timed out after 20s

goroutine 21 [sync.Mutex.Lock]:
  ...gamepaddb_test.(*gamepadLike).Axis(...)
  ...gamepaddb.StandardAxisValue(...)   gamepaddb.go:392   <- holds mappingsM, wants g.m

goroutine 7 [sync.Mutex.Lock]:
  ...gamepaddb.HasStandardAxis(...)     gamepaddb.go:362   <- holds g.m, wants mappingsM
```

The test passes on `main`.

Note that the inversion is not new: because Go's `RWMutex` blocks new readers once a writer is waiting, a concurrent `Update` (i.e. `ebiten.UpdateStandardGamepadLayoutMappings`) could already wedge the same cycle today. This PR removes the "usually two readers" luck that hides it.

## Suggestion

The invariant worth establishing is that `mappingsM` is never held across a `GamepadState` callback. Resolve the mapping under the write lock, where adding the Android defaults is legal, then read the device state after unlocking:

```go
// axisMapping returns the mapping for the axis, and reports whether it exists.
func axisMapping(id string, axis StandardAxis) (mapping, bool) {
	mappingsM.Lock()
	defer mappingsM.Unlock()

	m := axisMappings(id)
	if m == nil {
		return mapping{}, false
	}
	v, ok := m[axis]
	return v, ok
}

func StandardAxisValue(id string, axis StandardAxis, state GamepadState) float64 {
	m, ok := axisMapping(id, axis)
	if !ok {
		return 0
	}

	// The existing switch on m.Type, with no lock held.
	...
}
```

with the `buttonMapping` counterpart for `StandardButtonValue` and `IsStandardButtonPressed`, and `standardButtonValue` taking a `mapping` instead of looking it up by `id` again. `HasStandardLayoutMapping`, `HasStandardAxis` and `HasStandardButton` can keep the plain write lock exactly as this PR has them, since they invoke no callback, and `Name` can keep the read lock.

Handing the mapping out from under the lock is safe: `Update` and `addAndroidDefaultMappings` both install newly built maps rather than mutating published ones, and `mapping` is a value type.

Two details while doing this:

* `IsStandardButtonPressed` currently reads `gamepadButtonMappings[id]` directly, so unlike its siblings it never triggers the Android default mappings. Routing it through `buttonMapping` would fix that inconsistency, but it is a behavior change on Android and should be a deliberate one.
* `buttonMappings` and `axisMappings` then require `mappingsM` to be held; that is worth a comment.

On the test that 80003dce added: it should be in `package gamepaddb_test` (the `export_test.go` bridge is already there for the platform override), and `SetPlatformForTesting(0)` reads better as `platformUnknown`.

hajimehoshi (commented):
Thanks, 60c7702 addresses it. I applied the current diff and checked it rather than only reading it:

* the lock-order test above now passes (it deadlocks on both 80003dce and 8de4520);
* a two second stress run with nine goroutines mixing all three access patterns, i.e. the gamepad lock held while calling into this package, this package calling back into the gamepad, and a concurrent `Update`, is clean under `-race`;
* `gofmt -l`, `go vet ./internal/gamepaddb/`, the existing tests and a `GOOS=android` build are clean.

The property that makes this correct is that every `buttonMappings` / `axisMappings` call, i.e. every path that can reach `addAndroidDefaultMappings`, now sits in a function holding `mappingsM` exclusively (`buttonMapping`, `axisMapping`, `HasStandardLayoutMapping`, `HasStandardAxis`, `HasStandardButton`), while every `state.IsAxisReady` / `Axis` / `Button` / `Hat` call site is outside any lock. `mappingsM` is a leaf lock now, so the cycle with the gamepad's own mutex cannot re-form. Passing `mapping` by value is sound: it is a flat struct of scalars.

A few remarks, none of them blocking:

1. `IsStandardButtonPressed` used to read `gamepadButtonMappings[id]` directly and therefore never registered the Android default mappings; through `buttonMapping` it now can. Nothing changes in practice, because `Gamepad.IsStandardButtonPressed` calls `HasStandardLayoutMapping` first and that already registers them, but the change deserves a line in the commit message rather than being silent.

2. The same rationale is now stated on `buttonMapping`, `axisMapping`, `standardButtonValue` and `GamepadState`. Stating the invariant once, on `GamepadState`, would be enough; the helpers can be one line or nothing.

3. "None of the methods must be called while mappingsM is held in this package" reads ambiguously. "The methods must not be called while mappingsM is held" is what is meant.

4. The PR description is stale: it still describes the write-lock approach and says "Name keeps the read lock", which is no longer true now that `mappingsM` is a `sync.Mutex`, and it mentions a `TestAndroidDefaultMappingsConcurrent` that is no longer part of the diff.

5. The tests dropped in 8de4520 are worth restoring. The Android concurrency test with the platform override in `package gamepaddb_test` (the `export_test.go` bridge is the right shape for the override), plus a lock-order regression test, which is the one that catches a reintroduction without `-race`, since CI does not run `-race` today (#3569).

6. Minor: `Name` now serializes against every other lookup, which is unavoidable with a plain `Mutex` and negligible now that the lock holds no callbacks.

7. The branch carries four commits, two of which undo earlier ones; worth squashing on merge.

kumagi:
PTAL. The lock-order inversion is fixed in 60c7702 by establishing the invariant you suggested: `mappingsM` is never held across a `GamepadState` callback.

* `axisMapping` and `buttonMapping` resolve the mapping under the write lock, where adding the Android default mappings is legal, and return it by value. `StandardAxisValue`, `StandardButtonValue` and `IsStandardButtonPressed` then read the device state with no lock held, so only the `g.m` -> `mappingsM` order remains.
* `standardButtonValue` takes a `mapping` instead of looking it up by `id` again.
* `HasStandardLayoutMapping`, `HasStandardAxis` and `HasStandardButton` keep the plain write lock since they invoke no callback, and `Name` only takes the lock too (it is a `sync.Mutex` now, so there is no read lock left).
* `buttonMappings` and `axisMappings` are documented as requiring `mappingsM`, and `GamepadState` documents why its methods must not be called while the lock is held.

On the two details:

* I did route `IsStandardButtonPressed` through `buttonMapping`, so it now adds the Android default mappings like its siblings. I think this is the deliberate choice worth making: every in-tree caller (`internal/gamepad`) already reaches it only after `HasStandardLayoutMapping`, which populates those mappings, so nothing changes in practice, and keeping the direct `gamepadButtonMappings[id]` read is exactly the kind of inconsistency that caused the original race. Say the word if you would rather keep the old behavior here.
* I did not add a test, per your earlier note. I verified locally with your `TestLockOrderInversion` (extended to `StandardButtonValue` and `IsStandardButtonPressed` via an axis-mapped button): it hangs on 8de4520 with the two stacks you posted, and passes with this change. `go test -race` for `internal/gamepaddb` and `internal/gamepad`, and cross builds for android, js/wasm, darwin and windows are clean.

hajimehoshi:
> PTAL.

Have you really pushed?

kumagi:
Yes, and sorry for the confusion: [60c7702](https://github.com/hajimehoshi/ebiten/pull/3568/commits/60c7702483f9559f0938242b5dc5651f36e0b9d0) was already the head when I commented above (pushed 2026-08-29 16:13 UTC), I just wrote the SHA as plain text. Two more commits are pushed now:

* 9c02739 states the invariant once, on `GamepadState`, with the wording you suggested ("The methods must not be called while `mappingsM` is held"), and reduces the helpers to one line each.
* a07153d adds the tests back: `TestAndroidDefaultMappingsConcurrent` (data races with `-race` without the write lock) and `TestLockOrderInversion` (fails without `-race` when a gamepad state is used while `mappingsM` is held, since CI does not run `-race`). Both are in `package gamepaddb_test`, with the platform override in `export_test.go`. I checked that each one fails where it should: applying the override to df9a556 makes the former report `WARNING: DATA RACE`, and the latter fails on 8de4520 with "deadlocked: mappingsM must not be held while a gamepad state is used". To keep a regression from hanging the suite forever, the latter reports after 10s per case instead of blocking forever.

The PR description is updated, and the commit message for 60c7702 already has a paragraph about `IsStandardButtonPressed` now registering the Android default mappings; if the branch is squashed I'll keep that line in the squashed message. Say the word if you would rather I squash the four commits myself.

`gofmt -l`, `go vet ./internal/gamepaddb/`, `go test -race -count=2 ./internal/gamepaddb/ ./internal/gamepad/`, `go test -race .` and `GOOS=android`/`js`/`windows`/`darwin` builds of the package are clean.

hajimehoshi (commented):
Thanks, this addresses everything. I applied the current diff and checked that the new tests are real regression guards rather than only that they pass:

* `gofmt -l`, `go vet`, and `go test ./internal/gamepaddb/ ./internal/gamepad/` are clean, as is `-race -count=3`.
* Reintroducing the old locking in `StandardAxisValue` makes `TestLockOrderInversion/StandardAxisValue` fail with "deadlocked: mappingsM must not be held while a gamepad state is used", and the two later subtests report the cascade through the other message, as intended.
* Removing the lock from `HasStandardButton` makes `TestAndroidDefaultMappingsConcurrent` report `WARNING: DATA RACE` and `fatal error: concurrent map read and map write` under `-race`.

So both tests fail for the right reason on the right regression, and the lock-order one does so without `-race`, which is the case that matters for CI.

A few things that are nicely handled: the `sync.Once` hand-over fires on whichever state method the mapping type reaches first, so all three subtests really do exercise a callback; `close(g.proceed)` happens inside the gamepad's critical section, which makes the inversion deterministic rather than timing dependent; and `SetPlatformForTesting` returning the previous value makes the deferred restore correct.

Remaining nits, none of them blocking:

1. "This fails without `-race`" on `TestLockOrderInversion` reads as "this test fails unless you pass `-race`". The intended meaning is the opposite: it detects the deadlock even without `-race`.

2. Both new tests leave their entries in the package-global mapping maps once they finish. Harmless as long as the ids stay distinct, but it is shared state between tests.

3. The branch is six commits, two of which undo earlier ones. The rewritten description reads well as the squashed commit message.

kumagi:
Thanks for verifying the tests fail for the right reasons.

1. Fixed in 192526b: the comment now reads "It detects the deadlock even without `-race`, which CI does not run."

2. Agreed that the entries stay in `gamepadButtonMappings` / `gamepadAxisMappings`. There is no way to remove a mapping (`Update` only adds), so keeping the maps clean would mean a reset hook just for tests. I left the ids test-only (`00000000000000000000000000000001` and the `030000004c0500006802000001xx0000` series, neither of which is a real device) and kept them away from the ids the existing tests use. I would rather not add the reset, but I'll do it if you prefer.

3. Understood as the message to use when this is squash-merged; I did not rewrite the already pushed and reviewed history. Say the word if you want the branch squashed before merging.

internal/gamepaddb/gamepaddb_test.go
```diff
@@ -89,3 +92,154 @@ func TestGLFWGamepadMappings(t *testing.T) {
...
+	const goroutines = 16
+	ids := make([]string, goroutines)
+	for g := range goroutines {
+		ids[g] = fmt.Sprintf("030000004c0500006802000001%02x0000", g)
+	}
+
+	start := make(chan struct{})
+	var wg sync.WaitGroup
+	for g := range goroutines {
+		wg.Add(1)
```
hajimehoshi: Use wg.Go

internal/gamepaddb/gamepaddb_test.go
```diff
@@ -89,3 +92,154 @@ func TestGLFWGamepadMappings(t *testing.T) {
 		t.Errorf("got: %q, want: %q", got, want)
 	}
 }
+
+func TestAndroidDefaultMappingsConcurrent(t *testing.T) {
```
hajimehoshi: This test seems intrusive. Remove this.

hajimehoshi (commented):
One more thought on `TestLockOrderInversion`, now that the wording is fixed: it pins one scripted interleaving, and the choreography (`handOver`, `sync.Once`, the three subtests, `mustReceive`) exists only to force it. Hammering both lock orders from several goroutines is a better fit for this kind of bug, and it turns out to be shorter as well.

I measured both against the current head with the old locking put back into `StandardAxisValue`:

| | `TestLockOrderInversion` | stress test below |
| --- | --- | --- |
| catches the regression | deterministic | 5/5 runs at a 200ms window, still 5/5 at a 20ms window |
| time when green | 0.00s | 0.2s, 1.6s under `-race` |
| exported API only | yes | yes |
| exercises `Update` concurrently | no | yes |
| finds data races under `-race` | no | yes |

The deadlock window here is the whole callback, so the stress version hits it immediately rather than probabilistically. What it gives up is the guarantee: a future regression that holds the lock across a callback on some rare branch could be missed on a given run, where the scripted test would still force it. That seems like the right trade for roughly half the lines and no choreography to read.

This replaces `gamepadLike`, `handOver`, `mustReceive` and `TestLockOrderInversion`:

```go
// gamepadLike mimics internal/gamepad.Gamepad:
// every state query takes the gamepad's own mutex.
type gamepadLike struct {
	mu sync.Mutex
}

func (g *gamepadLike) IsAxisReady(index int) bool { g.mu.Lock(); defer g.mu.Unlock(); return true }
func (g *gamepadLike) Axis(index int) float64     { g.mu.Lock(); defer g.mu.Unlock(); return 0 }
func (g *gamepadLike) Button(index int) bool      { g.mu.Lock(); defer g.mu.Unlock(); return false }
func (g *gamepadLike) Hat(index int) int          { g.mu.Lock(); defer g.mu.Unlock(); return 0 }

// queryWithGamepadLock calls this package while the gamepad's own lock is held,
// like Gamepad.IsStandardAxisAvailable does.
func queryWithGamepadLock(g *gamepadLike, id string) {
	g.mu.Lock()
	defer g.mu.Unlock()

	_ = gamepaddb.HasStandardLayoutMapping(id)
	_ = gamepaddb.HasStandardAxis(id, gamepaddb.StandardAxisLeftStickHorizontal)
	_ = gamepaddb.HasStandardButton(id, gamepaddb.StandardButtonRightBottom)
	_ = gamepaddb.Name(id)
}

// TestConcurrentAccess uses this package from multiple goroutines in both lock orders:
// one takes the gamepad's own lock before calling in, and the other lets this package
// call back into the gamepad. Holding mappingsM while a gamepad state is used makes the
// two orders wait for each other.
//
// This must be the last test in this file, as a deadlock leaves mappingsM held
// and any test running after it would hang.
func TestConcurrentAccess(t *testing.T) {
	const id = "00000000000000000000000000000001"
	mappings := []byte(id + ",Test Pad,a:b0,leftx:a0,\n")
	if err := gamepaddb.Update(mappings); err != nil {
		t.Fatal(err)
	}

	g := &gamepadLike{}
	deadline := time.Now().Add(200 * time.Millisecond)

	var wg sync.WaitGroup
	for range 4 {
		wg.Go(func() {
			for time.Now().Before(deadline) {
				queryWithGamepadLock(g, id)
			}
		})

		wg.Go(func() {
			for time.Now().Before(deadline) {
				_ = gamepaddb.StandardAxisValue(id, gamepaddb.StandardAxisLeftStickHorizontal, g)
				_ = gamepaddb.StandardButtonValue(id, gamepaddb.StandardButtonRightBottom, g)
				_ = gamepaddb.IsStandardButtonPressed(id, gamepaddb.StandardButtonRightBottom, g)
			}
		})
	}

	wg.Go(func() {
		for time.Now().Before(deadline) {
			_ = gamepaddb.Update(mappings)
		}
	})

	done := make(chan struct{})
	go func() {
		wg.Wait()
		close(done)
	}()

	select {
	case <-done:
	case <-time.After(10 * time.Second):
		t.Fatal("deadlocked: a gamepad state was probably used while mappingsM was held")
	}
}
```

`gofmt` and `go vet` are clean, the package passes with and without `-race`, and with the old locking reintroduced it fails 3 out of 3 runs with the message above rather than hanging until the package timeout.

internal/gamepaddb/gamepaddb_test.go
```diff
@@ -89,3 +92,154 @@ func TestGLFWGamepadMappings(t *testing.T) {
...
+	g.mu.Lock()
+	defer g.mu.Unlock()
+	return 0
+}
+
+// TestLockOrderInversion checks that gamepaddb does not hold its lock
+// while calling a gamepad state, which would deadlock with the other order,
+// where a gamepad holds its own lock and then calls into gamepaddb.
+// It detects the deadlock even without -race, which CI does not run.
+func TestLockOrderInversion(t *testing.T) {
```
hajimehoshi: Remove this and instead add a stress test we suggested

internal/gamepaddb/gamepaddb_test.go
kumagi: Removed in 092388714, along with the test-only platform override and `export_test.go`, so nothing intrusive is left.

internal/gamepaddb/gamepaddb_test.go
kumagi: Done in 092388714: `TestLockOrderInversion`, `gamepadLike`, `handOver` and `mustReceive` are replaced with your stress test as written (`TestConcurrentAccess`). I re-checked it is a real guard: putting the old locking back into `StandardAxisValue` makes it fail with the deadlock message in 10s rather than hanging until the package timeout. `gofmt`, `go vet`, `go test` and `-race -count=2` for `internal/gamepaddb` and `internal/gamepad`, and android/js/windows/darwin builds are clean. The PR description is updated to drop the mentions of the removed tests.

internal/gamepaddb/gamepaddb_test.go
```diff
@@ -89,3 +91,79 @@ func TestGLFWGamepadMappings(t *testing.T) {
 		t.Errorf("got: %q, want: %q", got, want)
 	}
 }
+
+// gamepadLike mimics internal/gamepad.Gamepad:
+// every state query takes the gamepad's own mutex.
+type gamepadLike struct {
```
hajimehoshi: mockGamepad?

kumagi:
PTAL. [0923887](https://github.com/hajimehoshi/ebiten/pull/3568/commits/0923887142d9f2ef598e79fbd7291886d66f8355) addresses the latest comments:

* `TestAndroidDefaultMappingsConcurrent` is removed, together with the test-only platform override and `export_test.go`, so nothing intrusive is left.
* `TestLockOrderInversion`, `gamepadLike`, `handOver` and `mustReceive` are replaced with the suggested stress test `TestConcurrentAccess`, which hammers both lock orders and a concurrent `Update` from several goroutines. It is a real guard: putting the old locking back into `StandardAxisValue` makes it fail with the deadlock message in 10s rather than hanging until the package timeout.

`gofmt`, `go vet`, `go test` and `-race -count=2` for `internal/gamepaddb` and `internal/gamepad`, and android/js/windows/darwin builds are clean. The PR description no longer mentions the removed tests.

internal/gamepaddb/gamepaddb_test.go
kumagi: Renamed in the latest commit.

hajimehoshi (approved):
LGTM, thanks!

PR #3570, opened 2026-08-29
audio: reject a loop length shorter than one frame in InfiniteLoop

kumagi:
None

bug

## What this PR does | solves
InfiniteLoopWithIntro aligns introLength and loopLength down to frame boundaries, so a loopLength shorter than one frame (e.g. 3 bytes for 16-bit stereo) became llength == 0 with no error. With such a stream, Seek wrapped positions with a modulo by zero and panicked, and Read consumed the intro and then returned (0, nil) forever, violating the io.Reader contract.

Panic at construction time with a descriptive message instead, which removes both downstream failure modes. Add TestInfiniteLoopZeroLoopLength.

### DIFF
--- a/audio/loop_test.go
+++ b/audio/loop_test.go
@@ -495,3 +495,23 @@ func TestInfiniteLoopSeekCurrentAfterPartialFrameRead(t *testing.T) {
 		t.Errorf("got: %v, want: %v", buf[:n], want)
 	}
 }
+
+func TestInfiniteLoopZeroLoopLength(t *testing.T) {
+	for _, tc := range []struct {
+		name        string
+		introLength int64
+		loopLength  int64
+	}{
+		{"ShortLoop", 8, 3},
+		{"ZeroLoop", 8, 0},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			defer func() {
+				if r := recover(); r == nil {
+					t.Errorf("expected a panic for a zero loop length, but got none")
+				}
+			}()
+			audio.NewInfiniteLoopWithIntro(bytes.NewReader(make([]byte, 64)), tc.introLength, tc.loopLength)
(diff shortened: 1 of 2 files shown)


audio/loop_zerolength_test.go
```diff
@@ -0,0 +1,32 @@
+package audio_test
```
hajimehoshi: Use `loop_test.go`

hajimehoshi:
Revert the test but at an existing file

audio/loop_test.go
```diff
@@ -151,6 +151,30 @@ func TestInfiniteLoopWithIntro(t *testing.T) {
 	}
 }
 
+// TestInfiniteLoopZeroLoopLength tests that a loop length shorter than one frame
+// (which would be aligned down to zero) is rejected at construction instead of
+// causing an integer division by zero at Seek or an endless (0, nil) from Read.
```
hajimehoshi: Remove comments

audio/loop_test.go
```diff
@@ -151,6 +151,30 @@ func TestInfiniteLoopWithIntro(t *testing.T) {
...
+func TestInfiniteLoopZeroLoopLength(t *testing.T) {
+	// 3 bytes is shorter than one stereo 16-bit frame (4 bytes).
+	for _, tc := range []struct {
+		name        string
+		introLength int64
+		loopLength  int64
+	}{
+		{"ShortLoop", 8, 3},
+		{"ZeroLoop", 8, 0},
```
hajimehoshi: Use explicit member names and multiple lines

audio/loop_test.go

audio/loop_test.go

hajimehoshi (approved):
LGTM

PR #3571, opened 2026-08-29
audio: reject a negative seek result

kumagi:
None

bug

## What this PR does | solves
Both the int16 and the float32 streams accepted a seek resolving to a negative position without an error. Following reads then failed with "vorbis: decoding error" until a valid seek happened again. The prefix is broadened to `audio:` because `convert.Resampling` is public surface (returned by `audio.Resample*` and wrapped by `audio/mp3` and `audio/wav` whenever they resample), so this is a user-visible change beyond Ogg/Vorbis.

* `audio/internal/convert`: make `Resampling.Seek` return an error for a negative position instead of clamping it to 0, so the behavior does not depend on whether resampling is needed, and so a rejected seek leaves the state (`pos`, `eof`) untouched. Also fix `io.SeekEnd`, which was relative to the current position rather than the end.
* `audio/vorbis`: return an error for a negative seek result (before any state mutation).
* `audio/vorbis`: return the actual frame-aligned position from the int16 stream's `Seek`, and update the position fields (`posInBytes` / `pos`) only after `SetPosition` succeeds so a failed seek does not desync the stream.
* `audio/vorbis`: align the float32 offset after applying the whence, so a negative `io.SeekEnd` offset rounds the same direction as the int16 stream.

Note: this makes `audio.Player.SetPosition` with a negative duration return an error instead of silently rewinding to 0 for any resampled source, which matches the documented behavior. This belongs in the release notes.

Add tests which fail without these fixes (including a mono/stereo x i16/f32/resampling matrix with non-zero-position `io.SeekEnd` cases).

### DIFF
--- a/audio/internal/convert/resampling.go
+++ b/audio/internal/convert/resampling.go
@@ -319,23 +319,25 @@ func (r *Resampling) Seek(offset int64, whence int) (int64, error) {
 		return 0, fmt.Errorf("convert: source must be io.Seeker: %w", errors.ErrUnsupported)
 	}
 
+	pos := r.pos
 	switch whence {
 	case io.SeekStart:
-		r.pos = offset
+		pos = offset
 	case io.SeekCurrent:
-		r.pos += offset
+		pos += offset
 	case io.SeekEnd:
 		if r.size <= 0 {
 			return 0, fmt.Errorf("convert: seeking from the end is not possible when the length is unknown: %w", errors.ErrUnsupported)
 		}
-		r.pos = r.Length() + offset
+		pos = r.Length() + offset
 	default:
 		return 0, fmt.Errorf("convert: whence must be io.SeekStart, io.SeekCurrent, or io.SeekEnd but was %d", whence)
 	}
-	r.eof = false
-	if r.pos < 0 {
-		r.pos = 0
+	if pos < 0 {
+		return 0, fmt.Errorf("convert: position must be >= 0 but was %d", pos)
 	}
+	r.eof = false
+	r.pos = pos
 	// The position can be clamped by the length only when the length is known.
 	if r.size > 0 && r.Length() <= r.pos {
 		r.pos = r.Length()
--- a/audio/internal/convert/resampling_test.go
+++ b/audio/internal/convert/resampling_test.go
@@ -166,6 +166,68 @@ func TestResampling(t *testing.T) {
 	}
 }
 
+func TestResamplingSeekNegative(t *testing.T) {
+	for _, bitDepthInBytes := range []int{2, 4} {
+		t.Run(fmt.Sprintf("bitDepthInBytes=%d", bitDepthInBytes), func(t *testing.T) {
+			inB := newSoundBytes(44100, bitDepthInBytes)
+			r := convert.NewResampling(bytes.NewReader(inB), int64(len(inB)), 44100, 48000, bitDepthInBytes)
+
+			if _, err := r.Seek(-1, io.SeekStart); err == nil {
+				t.Error("Seek(-1, io.SeekStart): expected an error but got none")
+			}
+			if _, err := r.Seek(0, io.SeekStart); err != nil {
+				t.Fatal(err)
+			}
+			if _, err := r.Seek(-1, io.SeekCurrent); err == nil {
+				t.Error("Seek(-1, io.SeekCurrent): expected an error but got none")
+			}
+
+			// io.SeekEnd must be relative to the end, not to the current position.
+			if _, err := r.Seek(0, io.SeekStart); err != nil {
+				t.Fatal(err)
+			}
+			fromStart, err := r.Seek(-1000, io.SeekEnd)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if _, err := r.Seek(4000, io.SeekStart); err != nil {
+				t.Fatal(err)
+			}
+			fromMiddle, err := r.Seek(-1000, io.SeekEnd)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if fromStart != fromMiddle {
+				t.Errorf("Seek(-1000, io.SeekEnd) must not depend on the current position: from 0: %d, from 4000: %d", fromStart, fromMiddle)
+			}
+
+			// A negative result from io.SeekEnd must be rejected even from a non-zero position.
+			if _, err := r.Seek(4000, io.SeekStart); err != nil {
+				t.Fatal(err)
+			}
+			if _, err := r.Seek(-r.Length()-1, io.SeekEnd); err == nil {
+				t.Error("Seek(-Length()-1, io.SeekEnd): expected an error but got none")
+			}
+
+			// The stream must not be broken by the failed seeks.
+			pos, err := r.Seek(0, io.SeekStart)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if got, want := pos, int64(0); got != want {
+				t.Errorf("Seek(0, io.SeekStart): got: %d, want: %d", got, want)
+			}
+			n, err := r.Read(make([]byte, 64))
+			if err != nil {
+				t.Fatalf("Read after Seek(0): %v", err)
+			}
+			if got, want := n, 64; got != want {
+				t.Errorf("Read: got: %d, want: %d", got, want)
+			}
+		})
+	}
+}
+
 // Issue #3352
 func TestResamplingLen(t *testing.T) {
 	buf := make([]byte, 8*48000)
--- a/audio/vorbis/float32.go
+++ b/audio/vorbis/float32.go
@@ -84,6 +84,9 @@ func (r *float32BytesReadSeeker) Seek(offset int64, whence int) (int64, error) {
 	default:
 		return 0, fmt.Errorf("vorbis: whence must be io.SeekStart, io.SeekCurrent, or io.SeekEnd but was %d", whence)
 	}
+	if offset < 0 {
+		return 0, fmt.Errorf("vorbis: position must be >= 0 but was %d", offset)
+	}
 	r.pos = offset
 	if err := r.r.SetPosition(r.pos / sampleSize); err != nil {
 		return 0, err
--- a/audio/vorbis/float32.go
+++ b/audio/vorbis/float32.go
@@ -73,7 +73,6 @@ func (r *float32BytesReadSeeker) Seek(offset int64, whence int) (int64, error) {
 	}
 
 	sampleSize := int64(r.r.Channels()) * 4
-	offset = offset / sampleSize * sampleSize
 
 	switch whence {
 	case io.SeekStart:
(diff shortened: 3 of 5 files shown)


hajimehoshi (commented):
Thanks for the fix. I applied the branch locally and verified the behavior with and without it.

## Verification

The bug is real and the test genuinely catches it. On `main` with only the test hunk applied:

```
TestSeekNegativePosition/i16: Seek(-100, io.SeekStart): expected an error but got position -100
TestSeekNegativePosition/f32: Seek(-100, io.SeekStart): expected an error but got position -96
```

and the follow-on read fails with `vorbis: decoding error`. With the full change, `go test ./audio/...` and `go vet ./audio/...` are clean.

Both checks are placed before any state mutation (`r.pos` / `s.posInBytes`), so a rejected seek leaves the stream untouched. That is the right ordering.

One correction to the description: "never recovered" is not accurate. A subsequent valid seek does restore the stream. On unpatched `main`, `Seek(0, io.SeekStart)` followed by `Read` returns 64 bytes and a nil error. The impact is still real, since a player just keeps reading and never re-seeks, but the stream is not permanently broken.

## Main issue: inconsistent with `convert.Resampling`

`audio/internal/convert/resampling.go:331` clamps a negative position to 0 instead of returning an error, and `DecodeWithSampleRate` wraps the stream in `Resampling` only when the sample rates differ. So this change makes the same call behave differently depending on the source file:

```
DecodeWithSampleRate(44100, <44.1kHz ogg>).Seek(-100, io.SeekStart)
  -> 0, "vorbis: position must be >= 0 but was -50"
DecodeWithSampleRate(88200, <44.1kHz ogg>).Seek(-100, io.SeekStart)
  -> 0, <nil>      (and the following Read succeeds)
```

Both measured. Whichever policy we pick, `Resampling` should match it. Returning an error is the better one: that is what `io.Seeker` specifies, and `audio/wav/sectionreader.go:75` already errors on an out-of-range position, so `Resampling` is the outlier. Since `DecodeWithSampleRate` is the path most games take, I would like this resolved before merging.

## The second change is bundled in and untested

`return s.posInBytes` in place of `return next` (`audio/vorbis/vorbis.go`) is a correct fix, but it is a different bug from rejecting a negative seek, and the new test does not cover it. It only shows up for stereo, because mono routes through `StereoI16ReadSeeker`, which pre-aligns the offset. Measured on `test_stereo.ogg`: `Seek(5, io.SeekStart)` returns 4 with the change and 5 without. Please either split it into its own commit or add a stereo assertion for it.

## Test comments

- `_ = got` after the `t.Fatalf` is dead code; `got` is already consumed by the format arguments.
- The `seekStream` interface is not needed. Both `DecodeWithoutResampling` and `DecodeF32` return `*vorbis.Stream`, so `f func(io.Reader) (*vorbis.Stream, error)` works and the type can go away. If it stays, please move it above the test, matching the rest of the file.
- Only `test_mono_ogg` is exercised, which means neither `i16Stream.Seek` nor `float32BytesReadSeeker.Seek` is reached directly; both go through a `convert` wrapper. Adding `test_stereo_ogg` covers the direct path.
- There is no `io.SeekEnd` case, e.g. `Seek(-s.Length()-1, io.SeekEnd)`.
- The recovery assertions check only the error, not the returned position or byte count. The neighboring tests such as `TestMonoF32Seek` compare a `got` against a `want`.

## Adjacent, pre-existing, not blockers

- `i16Stream.Seek` sets `s.posInBytes` before calling `SetPosition`. If that call fails, the position is updated but `s.i16Reader` is left stale. Moving the assignment below the `SetPosition` call would close that.
- `float32BytesReadSeeker.Seek` aligns `offset` before the whence switch, while `i16Stream.Seek` aligns after it. Both truncate toward zero, so a negative `io.SeekEnd` offset rounds in opposite directions in the two streams: f32 rounds toward the end, i16 away from it. Worth a follow-up.

The error wording matches the existing style in the package, and a bare `fmt.Errorf` is consistent with what `wav` does today.

Overall the fix is correct and the test is a real regression guard. The `Resampling` inconsistency is the one thing I would like addressed first.

---

kumagi:
Thanks for the thorough review. I rebased the branch into focused commits and addressed every point:

**`convert.Resampling` inconsistency (the blocker)** — `c55f5f9` makes `Resampling.Seek` return an error for a negative position instead of clamping to 0, matching `io.Seeker` and the `wav` section reader, and it checks before mutating any state. So `DecodeWithSampleRate` now behaves the same whether or not resampling is actually needed.

**Bundled second change** — split into its own commit `3e970916` (`i16Stream.Seek` returns the frame-aligned position), plus a stereo assertion `TestStereoI16SeekUnalignedPosition` (`Seek(5, io.SeekStart)` -> 4).

**Adjacent, non-blocker items I also fixed:**
- `3e970916` moves the `posInBytes` assignment below the `SetPosition` call so a failed seek no longer leaves the position out of sync with the reader.
- `8f899d3c` aligns the float32 offset *after* applying the whence and after the negative check, so a negative `io.SeekEnd` offset now rounds the same direction in both streams. Covered by `TestStereoF32SeekUnalignedNegativeOffset`.

**Test feedback:**
- Removed the dead `_ = got` and dropped the `seekStream` interface; the cases now use `func(io.Reader) (*vorbis.Stream, error)` directly.
- `TestSeekNegativePosition` exercises mono **and** stereo across i16, f32, and resampling, so both `i16Stream.Seek` and `float32BytesReadSeeker.Seek` are reached directly.
- Added an `io.SeekEnd` case (`Seek(-Length()-1024, io.SeekEnd)`).
- Recovery assertions now check the returned position and byte count, not just the error.

The new tests all fail on `main` (verified by reverting only the source files) and pass with the fixes; `go test ./audio/...` and `go vet ./audio/...` are clean, and each commit builds and tests on its own.

hajimehoshi:
Resolve the conflict

kumagi:
done

hajimehoshi (commented):
Thanks, this is much better. I checked out `f82fa4cb5` and re-ran everything.

## Resolved

All five points are addressed, and well.

- **`Resampling` inconsistency** — `7c90e29c` returns an error instead of clamping, and correctly leaves `r.eof` and `r.pos` untouched on the error path; previously `r.eof = false` ran before the check. `DecodeWithSampleRate` now behaves the same whether or not resampling is needed.
- **Bundled change split out** — `c8fb6451` isolates the frame-aligned return and adds `TestStereoI16SeekUnalignedPosition`. It also moves `s.posInBytes = pos` below `SetPosition`.
- **Rounding divergence** — `f82fa4cb` moves the float32 alignment after the whence switch. I confirmed `TestStereoF32SeekUnalignedNegativeOffset` really guards it: the old code returned `Length()` for `Seek(-5, io.SeekEnd)` and the new one returns `Length()-8`.
- **Tests** — the dead `_ = got` and the `seekStream` interface are gone, the matrix covers mono and stereo across I16, F32, and resampling, there is an `io.SeekEnd` case, and positions and byte counts are asserted.

Both test files really are 44100Hz, so the resampling subtest does exercise the wrapper rather than silently skipping it. `go test ./audio/...`, `go vet ./audio/...`, and `gofmt -l ./audio/` are clean. The commit messages are accurate.

Two things remain.

## 1. The float32 stream still writes `r.pos` before `SetPosition`

`c8fb6451` fixed exactly this in `i16Stream.Seek`, but `float32BytesReadSeeker.Seek` still assigns `r.pos = pos` above the `SetPosition` call. With a source whose `Read` fails partway, seeking from a good position at 4096:

```
f32: failing Seek(8192) = 0, boom   ->  next Seek(0, io.SeekCurrent) = 8192
i16: failing Seek(8192) = 0, boom   ->  next Seek(0, io.SeekCurrent) = 4096   (correct)
```

A failed float32 seek still moves the reported position. Moving `r.pos = pos` below the `SetPosition` call fixes it.

## 2. `Resampling.Seek` with `io.SeekEnd` is relative to the current position

This is pre-existing, but the PR rewrites the line and the new test's guarantee depends on it:

```go
pos := r.pos
case io.SeekEnd:
    pos += r.Length() + offset   // should be: pos = r.Length() + offset
```

Since `pos` starts at `r.pos`, `io.SeekEnd` accumulates. Measured with `Length()` = 192000:

```
from pos 0:    Seek(-1000, io.SeekEnd)        = 191000   (correct)
from pos 4000: Seek(-1000, io.SeekEnd)        = 192000   (want 191000)
from pos 4000: Seek(-Length()-1, io.SeekEnd)  = 3996, nil   (want an error)
```

The last line is the exact assertion `TestResamplingSeekNegative` makes; it passes only because the test seeks from position 0. So the new negative guard is position dependent for `io.SeekEnd`. Both `i16Stream.Seek` and `float32BytesReadSeeker.Seek` assign rather than accumulate here, so `Resampling` is the odd one out. Please use `pos = r.Length() + offset`, and add a non-zero `Seek(..., io.SeekStart)` before the `io.SeekEnd` assertions in both new tests so the case is actually covered.

## Note

`Resampling` is public surface: `audio.ResampleReader`, `audio.ResampleReaderF32`, `audio.Resample`, and `audio.ResampleF32` return it directly, and `audio/mp3` and `audio/wav` wrap it whenever they resample. So `audio.Player.SetPosition` with a negative duration now returns an error instead of silently rewinding to 0 for any resampled source, not just Ogg/Vorbis. That is the correct behavior and is already covered by `Player.SetPosition`'s documentation, but it is a user visible change I will want in the release notes. Since one commit and the widest reaching effect live in `audio/internal/convert`, please broaden the PR title from `audio/vorbis:`.

Minor: the description's bullets do not mention the fourth commit.

Neither finding is a regression in what is proposed. The first is a leftover of a fix the PR already makes in the sibling function, and the second is a pre-existing bug that undercuts the new test.

---

kumagi:
Both points fixed and folded into the relevant commits (rebased onto current `main`).

**1. float32 wrote `r.pos` before `SetPosition`** — the assignment now runs after `SetPosition` succeeds, matching `i16Stream.Seek`, so a failing seek no longer moves the reported position. Folded into `audio/vorbis: align the float32 seek position after applying whence`.

**2. `Resampling.Seek` `io.SeekEnd` was relative to the current position** — changed `pos += r.Length() + offset` to `pos = r.Length() + offset`, matching the int16 and float32 streams. `TestResamplingSeekNegative` now seeks from a non-zero position (`Seek(4000, io.SeekStart)`) and asserts `Seek(-1000, io.SeekEnd)` is position-independent and that `Seek(-Length()-1, io.SeekEnd)` errors from that non-zero position; the negative `io.SeekEnd` assertion in `TestSeekNegativePosition` likewise runs after `Seek(4096, io.SeekStart)`. Both fail with the accumulating form and pass with the fix.

**Title / description:** broadened the title to `audio:` and updated the body to cover all four commits, including the `convert` change's public-surface impact (`audio.Resample*`, and `mp3`/`wav` when resampling) — `audio.Player.SetPosition` with a negative duration now errors instead of rewinding to 0, which I've flagged for the release notes.

`go build ./...`, `go vet ./audio/...`, `gofmt -l audio/`, and `go test ./audio/...` are clean, and each of the four commits builds and tests on its own.

hajimehoshi (approved):
LGTM, thanks!

hajimehoshi:
Resolve conflicts

PR #3572, opened 2026-08-29
audio/wav: reject a non-positive sample rate

kumagi:
None

bug

## What this PR does | solves
A WAV whose fmt chunk declared a sample rate of 0 was accepted, and DecodeWithSampleRate then created a resampler whose Length() divided by zero: the float-to-int64 conversion overflowed and returned the minimum int64 value.

Reject a non-positive sample rate as an invalid header.

Add TestWavZeroSampleRate, which fails without this fix.

### DIFF
--- a/audio/wav/decode_test.go
+++ b/audio/wav/decode_test.go
@@ -137,25 +137,25 @@ func appendChunk(dst []byte, id string, data []byte) []byte {
 }
 
 // stereoI16FmtChunkData returns the content of a 'fmt ' chunk for 2 channels signed 16bit little endian PCM.
-func stereoI16FmtChunkData() []byte {
+func stereoI16FmtChunkData(sampleRate uint32) []byte {
 	var buf []byte
-	buf = binary.LittleEndian.AppendUint16(buf, 1)                  // Format tag (linear PCM)
-	buf = binary.LittleEndian.AppendUint16(buf, 2)                  // Channel count
-	buf = binary.LittleEndian.AppendUint32(buf, testSampleRate)     // Sample rate
-	buf = binary.LittleEndian.AppendUint32(buf, testSampleRate*2*2) // Byte rate
-	buf = binary.LittleEndian.AppendUint16(buf, 4)                  // Block align
-	buf = binary.LittleEndian.AppendUint16(buf, 16)                 // Bits per sample
+	buf = binary.LittleEndian.AppendUint16(buf, 1)              // Format tag (linear PCM)
+	buf = binary.LittleEndian.AppendUint16(buf, 2)              // Channel count
+	buf = binary.LittleEndian.AppendUint32(buf, sampleRate)     // Sample rate
+	buf = binary.LittleEndian.AppendUint32(buf, sampleRate*2*2) // Byte rate
+	buf = binary.LittleEndian.AppendUint16(buf, 4)              // Block align
+	buf = binary.LittleEndian.AppendUint16(buf, 16)             // Bits per sample
 	return buf
 }
 
 // wavFile returns a WAV file whose 'data' chunk holds data.
 // beforeFmt and afterFmt are put before and after the 'fmt ' chunk respectively.
-func wavFile(beforeFmt []chunk, afterFmt []chunk, data []byte) []byte {
+func wavFile(sampleRate uint32, beforeFmt []chunk, afterFmt []chunk, data []byte) []byte {
 	var body []byte
 	for _, c := range beforeFmt {
 		body = appendChunk(body, c.id, []byte(c.data))
 	}
-	body = appendChunk(body, "fmt ", stereoI16FmtChunkData())
+	body = appendChunk(body, "fmt ", stereoI16FmtChunkData(sampleRate))
 	for _, c := range afterFmt {
 		body = appendChunk(body, c.id, []byte(c.data))
 	}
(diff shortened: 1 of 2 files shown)


audio/wav/decode_test.go
```diff
@@ -235,3 +235,49 @@ func TestDecodeChunkPadding(t *testing.T) {
...
+}
+
+func TestDecodeInvalidSampleRate(t *testing.T) {
+	testCases := []struct {
+		name       string
+		sampleRate uint32
+		wantErr    bool
+	}{
+		{name: "zero", sampleRate: 0, wantErr: true},
+		{name: "valid", sampleRate: 44100, wantErr: false},
```
hajimehoshi: Use multiple lines

kumagi: done.

hajimehoshi:
Thanks! The fix itself looks correct to me, and I think the WAV decoder is the right layer for it: this is untrusted file data, so rejecting it at parse time next to the existing `channelCount` / `bitsPerSample` checks is consistent. `<= 0` rather than `== 0` is also right, since a rate >= 2^31 wraps negative on a 32-bit `int`.

I confirmed the new test fails without the change. A few comments, all on the test and the commit message.

### Reuse the existing helpers in `decode_test.go`

`wavBytesWithSampleRate` builds a WAV file byte by byte with manual little-endian shifts, but this file already has `appendChunk`, `stereoI16FmtChunkData`, `wavFile` and `testSampleRate`, and every other test uses `binary.LittleEndian.AppendUint32`.

Threading a sample rate through the existing pair removes the need for a new helper:

```go
func stereoI16FmtChunkData(sampleRate uint32) []byte {
	var buf []byte
	buf = binary.LittleEndian.AppendUint16(buf, 1)              // Format tag (linear PCM)
	buf = binary.LittleEndian.AppendUint16(buf, 2)              // Channel count
	buf = binary.LittleEndian.AppendUint32(buf, sampleRate)     // Sample rate
	buf = binary.LittleEndian.AppendUint32(buf, sampleRate*2*2) // Byte rate
	buf = binary.LittleEndian.AppendUint16(buf, 4)              // Block align
	buf = binary.LittleEndian.AppendUint16(buf, 16)             // Bits per sample
	return buf
}

func wavFile(sampleRate uint32, beforeFmt, afterFmt []chunk, data []byte) []byte { ... }
```

There is only one existing call site, which would pass `testSampleRate`. The new test then becomes `wavFile(tc.sampleRate, nil, nil, []byte{1, 2, 3, 4})`.

### The hand-rolled header declares a wrong RIFF size

`wavSize := uint32(44 - 8)` writes 36, but the file it builds is 48 bytes, so the field should be 40. The decoder ignores the RIFF size, so the test passes either way, but a test about header validation should not contain a malformed header. Reusing `wavFile` fixes this for free, as it computes `len(body)+4`.

### Commit message

The body mentions `TestWavZeroSampleRate`, but the test is named `TestDecodeInvalidSampleRate`. Could you also squash the two commits?

### The description understates the bug

"the float-to-int64 conversion overflowed and returned the minimum int64 value" holds on amd64, but the conversion saturates the other way on arm64, where I measured `Length()` returning 9223372036854775804. Both are garbage, but neither value is portable enough to state as the result.

More importantly, a wrong `Length()` is not the only consequence. In `Resampling.at`, `tInSrc = float64(t) * float64(r.from) / float64(r.to)` is 0 for every `t` when `from` is 0, so the stream decodes to a constant value of effectively unbounded length. A player fed such a file keeps emitting that constant rather than just reporting a wrong size. That is worth saying in the commit message.

### Note on scope

`convert.Resampling` still divides by zero, and it is reachable via the public `audio.ResampleReader(src, size, 0, 44100)`. I would leave that out of this PR: `from == 0` there is a caller bug, while the WAV path takes the value from a file, so guarding the decoder is the correct fix here.

---

kumagi:
Thanks for the review! I've addressed your comments in the force-pushed branch:

- Reused the existing helpers: `stereoI16FmtChunkData` and `wavFile` now take the sample rate, and `wavBytesWithSampleRate` is removed, so the test no longer hand-rolls a header with a wrong RIFF size.
- Squashed the two commits into one and rewrote the message: it now refers to `TestDecodeInvalidSampleRate` and describes the consequences as you noted (the architecture-dependent `Length()`, and the constant stream of effectively unbounded length because `tInSrc` is 0 for every `t` when `from` is 0).
- Left `convert.Resampling` out of this PR as suggested.

I also confirmed the test still fails without the fix.

audio/wav/decode_test.go
```diff
@@ -235,3 +235,37 @@ func TestDecodeChunkPadding(t *testing.T) {
...
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			data := wavFile(tc.sampleRate, nil, nil, []byte{1, 2, 3, 4})
+
+			if _, err := wav.DecodeWithoutResampling(bytes.NewReader(data)); (err != nil) != tc.wantErr {
+				t.Errorf("wav.DecodeWithoutResampling: err %v, wantErr %v", err, tc.wantErr)
+			}
+			if _, err := wav.DecodeWithSampleRate(44100, bytes.NewReader(data)); (err != nil) != tc.wantErr {
```
hajimehoshi: `testSampleRate`?

kumagi: Done.

hajimehoshi (approved):
LGTM

hajimehoshi:
Resolve conflicts

PR #3573, opened 2026-08-29
audio/internal/convert: fix StereoI16ReadSeeker's byte counts

kumagi:
None

bug

## What this PR does | solves
`Read` scaled the raw source byte count after the conversion, so a source
returning a byte count with a trailing partial frame made `Read` report
more bytes than it actually wrote, sometimes a count that is not a
multiple of 4, and the caller played its own stale buffer contents as
audio. `Seek` had the same bug converting the source position back, so a
source whose length is not a whole number of frames made it report a
position that is not a multiple of 4 either.

Derive both counts from the number of whole source frames, and carry the
remainder of an incomplete frame over to the next `Read` instead of
dropping it, which desynced the stream whenever the source returned a
short read. `Read` now fills at least one whole frame before returning so
that it does not return `(0, nil)` for a source that dribbles bytes,
reports `io.ErrShortBuffer` for a non-empty buffer shorter than a frame
while a zero-length read reports `(0, nil)` as `io.Reader` expects, and
accounts for the carried-over remainder when seeking relative to the
current position.

Add tests for frame-unaligned source lengths and read sizes. They fail
without this fix.

### DIFF
--- a/audio/internal/convert/stereoi16.go
+++ b/audio/internal/convert/stereoi16.go
@@ -31,6 +31,9 @@ type StereoI16ReadSeeker struct {
 	mono   bool
 	format Format
 	buf    []byte
+	// rest is a partial source frame that was read from the source but is not
+	// converted yet, so the source is ahead of the consumer by len(rest).
+	rest []byte
 }
 
 func NewStereoI16ReadSeeker(source io.ReadSeeker, mono bool, format Format) *StereoI16ReadSeeker {
@@ -41,28 +44,58 @@ func NewStereoI16ReadSeeker(source io.ReadSeeker, mono bool, format Format) *Ste
 	}
 }
 
-func (s *StereoI16ReadSeeker) Read(b []byte) (int, error) {
-	l := len(b) / 4 * 4
-	if s.mono {
-		l /= 2
-	}
+const dstFrameSize = 4
+
+func (s *StereoI16ReadSeeker) srcFrameSize() int {
 	switch s.format {
 	case FormatU8:
-		l /= 2
+		if s.mono {
+			return 1
+		}
+		return 2
 	case FormatS16:
+		if s.mono {
+			return 2
+		}
+		return 4
 	case FormatS24:
-		l *= 3
-		l /= 2
+		if s.mono {
+			return 3
+		}
+		return 6
 	}
+	panic("not reached")
+}
+
+func (s *StereoI16ReadSeeker) Read(b []byte) (int, error) {
+	if len(b) == 0 {
+		return 0, nil
+	}
+	if len(b) < dstFrameSize {
+		return 0, io.ErrShortBuffer
+	}
+	srcFrameSize := s.srcFrameSize()
+	l := len(b) / dstFrameSize * srcFrameSize
 
 	if cap(s.buf) < l {
 		s.buf = make([]byte, l)
 	}
+	s.buf = s.buf[:l]
 
-	n, err := s.source.Read(s.buf[:l])
-	if err != nil && err != io.EOF {
-		return 0, err
+	n := copy(s.buf, s.rest)
+	var err error
+	for n < srcFrameSize {
+		var dn int
+		dn, err = s.source.Read(s.buf[n:])
+		n += dn
+		if err != nil || dn == 0 {
+			break
+		}
 	}
+
+	s.rest = append(s.rest[:0], s.buf[n/srcFrameSize*srcFrameSize:n]...)
+	n = n / srcFrameSize * srcFrameSize
+
 	if s.mono {
 		switch s.format {
 		case FormatU8:
@@ -110,49 +143,23 @@ func (s *StereoI16ReadSeeker) Read(b []byte) (int, error) {
 			}
 		}
 	}
-	if s.mono {
-		n *= 2
-	}
-	switch s.format {
-	case FormatU8:
-		n *= 2
-	case FormatS16:
-	case FormatS24:
-		n *= 2
-		n /= 3
-	}
-	return n, err
+
+	return n / srcFrameSize * dstFrameSize, err
 }
 
 func (s *StereoI16ReadSeeker) Seek(offset int64, whence int) (int64, error) {
-	offset = offset / 4 * 4
-	if s.mono {
-		offset /= 2
+	srcFrameSize := int64(s.srcFrameSize())
+	o := offset / dstFrameSize * srcFrameSize
+	if whence == io.SeekCurrent {
+		o -= int64(len(s.rest))
 	}
-	switch s.format {
-	case FormatU8:
-		offset /= 2
-	case FormatS16:
-	case FormatS24:
-		offset *= 3
-		offset /= 2
-	}
-	pos, err := s.source.Seek(offset, whence)
+	pos, err := s.source.Seek(o, whence)
 	if err != nil {
 		return 0, err
 	}
+	s.rest = s.rest[:0]
+
 	// Convert the returned position from the source format's byte space to the
 	// stereo-i16 byte space this wrapper presents.
-	if s.mono {
-		pos *= 2
-	}
-	switch s.format {
-	case FormatU8:
-		pos *= 2
-	case FormatS16:
-	case FormatS24:
-		pos *= 2
-		pos /= 3
-	}
-	return pos, nil
+	return pos / srcFrameSize * dstFrameSize, nil
 }
--- a/audio/internal/convert/stereoi16_test.go
+++ b/audio/internal/convert/stereoi16_test.go
@@ -16,6 +16,7 @@ package convert_test
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"math/rand/v2"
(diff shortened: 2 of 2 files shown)


hajimehoshi (commented):
The bug is real and the direction is right, but the fix is incomplete in a way that trades one failure mode for another. I'd like one more change before merging.

I verified the bug exists: applying this PR's test against `main`'s implementation fails 4 of the 8 subcases (`got 10, want 8`, `got 5, want 4`, ...). With the patch all pass, and `go test ./audio/...` / `go vet ./audio/...` are clean.

The worst pre-existing symptom is worth stating explicitly, because it's stronger than the PR description says: `Read` could return a count that isn't a multiple of 4 (e.g. 5 or 6) on a stream that is by definition 4 bytes per frame, with the tail never written, so the caller's stale buffer contents got played.

## 1. The desync isn't actually fixed, and `(0, nil)` is new

The trailing partial frame is still read out of the source and then discarded. The PR body says the fix stops the stream desyncing; it doesn't.

Driving mono/S16 through a source that returns 3 bytes per `Read` (4 samples `0x1111 0x2222 0x3333 0x4444`), with the patch applied:

```
read #0: n=4 out=11 11 11 11     <- correct; source byte 0x22 consumed and dropped
read #1: n=4 out=22 33 22 33     <- garbage sample 0x3322: high byte of one, low byte of the next
read #2: n=4 out=44 44 44 44
```

One sample is lost and one garbage sample is emitted. `main` produces the same garbage sample, just padded with stale bytes, so this path is broken either way.

Worse, with a source that dribbles 1 byte per `Read`:

```
read #0: n=0 err=<nil>    (and one source byte silently discarded)
read #1: n=0 err=<nil>
read #2: n=0 err=<nil>
read #3: n=0 err=<nil>
read #4: n=0 err=EOF
```

Every call returns `(0, nil)`, which https://pkg.go.dev/io#Reader explicitly discourages, while consuming the entire stream and yielding pure silence. On `main` each call at least returned data. Short reads are legal for any `io.Reader` and entirely plausible for the `wav` path, where `newSectionReader` wraps whatever `io.Reader` the caller supplied (an HTTP body, a pipe).

The fix that matches the PR description is to carry the remainder instead of dropping it, roughly:

```go
copy(s.buf, s.rest)
n, err := s.source.Read(s.buf[len(s.rest):l])
n += len(s.rest)
// ... conversion loops ...
frames := n / srcFrameSize
s.rest = append(s.rest[:0], s.buf[frames*srcFrameSize:n]...)
n = frames * 4
```

`Seek` would need to reset `s.rest`. If you'd rather keep this PR to the byte count alone, that's defensible, but then the commit message shouldn't claim the desync is fixed.

## 2. The frame-size block can be half the size

The nested `if`/`switch` adds a fourth copy of the format-to-size table that already appears in `Read`'s `l` computation and twice in `Seek`. Channels and sample width are independent:

```go
var srcFrameSize int
switch s.format {
case FormatU8:
	srcFrameSize = 1
case FormatS16:
	srcFrameSize = 2
case FormatS24:
	srcFrameSize = 3
}
if !s.mono {
	srcFrameSize *= 2
}
n = n / srcFrameSize * 4
```

Note that either form makes an out-of-range `Format` a division-by-zero panic where it previously did nothing. That isn't reachable from today's callers, but it's a new panic path, and a reason to prefer a single `srcFrameSize()` helper that the other three sites share, if you want that as a follow-up rather than in this PR.

## 3. `Seek` has the identical bug, untouched

Same class, same file, a few lines below:

```
stereo S24, source len 8: Seek(0, io.SeekEnd) = 5
mono S16,   source len 5: Seek(0, io.SeekEnd) = 10
```

Neither is a multiple of 4. `pos = pos / srcFrameSize * 4` fixes it the same way. `TestStereoI16SeekEnd` misses this because it only uses frame-aligned sizes. Your call whether it belongs here or in a separate change.

## 4. Test nits

- The `if n > len(buf)` check can never fire: `buf` is 64 bytes and the largest `want` is 12. Dead assertion.
- The test asserts only on `n`, never on `b[:n]`. Since the reported symptom is stale bytes surfacing as audio, a content check would pin the actual defect rather than its proxy.
- `PartialFrameAtEOF` covers four aligned cases. They're good as regression guards, but the name doesn't describe them.
- `FormatS24` is currently unreachable, since `audio/wav/decode.go` rejects 24-bit (TODO #2215), so those cases are speculative. Harmless, but worth knowing.

---

kumagi:
Thanks for the thorough review, and for spelling out the stale-buffer symptom — the non-multiple-of-4 return value was indeed worse than the PR body said.

All four points are addressed:

**1. The remainder is carried over, not dropped.** `Read` now keeps the trailing partial source frame in `s.rest` and prepends it to the next call, and it loops until at least one whole source frame is filled, so a source that dribbles bytes never yields `(0, nil)` or drops bytes. Your mono/S16 3-bytes-per-read repro now returns the four samples intact instead of losing one and emitting the garbage sample, and the 1-byte-per-read source returns data on every call. `Seek` resets `s.rest`. The commit-message/PR-body claim about the desync is now accurate.

**2. Single `srcFrameSize()` helper.** Read's `l` computation, Read's return-value scaling and both `Seek` sites share it, so the format-to-size table exists once. Out-of-range `Format` panics with the `panic("not reached")` idiom already used in this package instead of dividing by zero.

**3. `Seek` fixed the same way.** The returned position is truncated to a whole source frame, so it is always a multiple of 4: stereo S24 over an 8-byte source now reports 4 (was 5), mono S16 over a 5-byte source reports 4 (was 10). `TestStereoI16SeekEndPartialFrame` covers the unaligned lengths `TestStereoI16SeekEnd` missed.

**4. Test nits.** The dead `n > len(buf)` assertion is gone, the cases compare the converted bytes (`want` is now `[]byte`), and the test was renamed to `TestStereoI16ReadSeekerPartialFrame`. I kept the S24 cases and the aligned ones as cheap guards.

One thing I hit while doing this and fixed on top of it: a *relative* seek was applied to the source position, which is ahead of the position the consumer reached by exactly the carried-over remainder, so a plain tell — `(*InfiniteLoop).ensurePos` does `Seek(0, io.SeekCurrent)` — silently skipped those bytes and left a gap in the audio. `Seek` now backs the source up over the remainder before a `io.SeekCurrent` seek, which also makes seeking by less than a frame the no-op the existing tests already document. `TestStereoI16ReadSeekerShortReads` now checks the position a tell reports against the bytes returned so far.

`go test -race ./audio/...`, `go vet ./audio/...` and gofmt are clean. Against `main`, `TestStereoI16ReadSeekerPartialFrame`, `TestStereoI16ReadSeekerShortReads` and `TestStereoI16SeekEndPartialFrame` all fail, so they are real regression guards.

I've left the `FormatS24` wav path (TODO #2215) alone as you suggested.

hajimehoshi (commented):
All four points are genuinely addressed, and the core fix is correct. I'd merge it after squashing.

I checked the substance rather than taking the reply at face value. A property test — every format x source lengths 0-40 x source chunk sizes 1-9 x destination buffer sizes {4, 5, 8, 12, 64}, with poisoned destination buffers — confirms the converted bytes are byte-identical to a single-big-read conversion in every combination, that a `Seek(0, io.SeekCurrent)` tell always matches the bytes returned so far, and that `Read` never stalls at `(0, nil)`. My earlier 3-bytes-per-read repro now returns all four samples intact; the 1-byte-per-read source returns data on every call.

Also verified:

- The three new tests fail against `main` (35 failing subcases in total), so they are real regression guards.
- `go test -race ./audio/...`, `go vet ./audio/...` and `gofmt -l` are all clean.
- `panic("not reached")` is indeed the local idiom (`audio/internal/convert/resampling.go:171`, `:308`).
- A non-EOF source error now returns the already-converted bytes alongside the error instead of discarding them (`n=8`, then `(0, boom)`). That is an unadvertised improvement over `main`'s `return 0, err`, and the right call per the `io.Reader` contract.

Everything below is minor.

## 1. `Read` with a buffer smaller than one frame now spins (new)

The early return at the top of `Read` swallows EOF:

```
3-byte buffer at EOF     main: (0, io.EOF)      PR: (0, nil), forever
```

A caller looping until a non-nil error hangs instead of terminating. No current caller passes a sub-4-byte buffer, so this is a trap rather than a live bug, but it is the one behavior the patch makes strictly worse. Note that the guard cannot simply be deleted: with `l == 0` the `s.rest` recompute would silently drop the remainder. Returning `io.ErrShortBuffer` there, or consulting the source's error before returning, would both close it.

## 2. `Seek` splits one seek into two

```go
if whence == io.SeekCurrent && len(s.rest) > 0 {
    if _, err := s.source.Seek(-int64(len(s.rest)), io.SeekCurrent); err != nil { ... }
}
s.rest = s.rest[:0]
pos, err := s.source.Seek(offset/dstFrameSize*srcFrameSize, whence)
```

Folding the rewind into the offset (`offset/dstFrameSize*srcFrameSize - int64(len(s.rest))`) makes it one call and removes the half-applied state: as written, if the rewind succeeds and the second seek fails, the source is left backed up by `len(rest)` with `rest` already cleared, so those bytes get delivered twice. Relatedly, `s.rest = s.rest[:0]` runs before the main seek, so a failed `SeekStart`/`SeekEnd` also discards buffered data. Both are error-path only.

## 3. Pre-existing, not for this PR: `SeekEnd` on an unaligned source

```
mono S16, 5-byte source:  Seek(-4, io.SeekEnd) -> 4    (main: 6)
  then Read -> 22 33 22 33     (want 22 22 22 22)
```

The PR fixes the half it set out to fix: the reported position is a multiple of 4 now. But the source itself still lands mid-frame when the source length is not a whole number of frames, so the following read straddles two samples. Identical on `main`. Worth a follow-up issue rather than more scope here.

## 4. The comment strip went one step too far

fa0f758 removes every explanatory comment added by the two commits before it. That is mostly right, since the `dstFrameSize` and `srcFrameSize()` comments restated their names. But the `Seek` rewind is the least obvious line in the whole change, and it now has no rationale at all: nothing says the source runs ahead of the consumer by exactly `len(s.rest)`. I would restore one line, either there or on the `rest` field, and leave the rest stripped.

## 5. Commit hygiene

- 8ca2dd76's body has 258- and 253-column lines. 7cf14d83 wraps at 72; please match that.
- fa0f758 only undoes comments introduced two commits earlier. Please squash the series.
- The title and PR body still describe only the byte-count fix; the change now also rewrites `Seek`.
- The `(*InfiniteLoop).ensurePos` citation in 8ca2dd76 does not quite hold: `ensurePos` returns early once `i.pos >= 0`, so its tell runs exactly once, before any `Read` reaches the wrapper, when `rest` is empty. The fix is right for any consumer that tells mid-stream; the message just should not name a call site where it cannot fire. (There is also a "does silently dropped" slip in the same sentence.)

---

kumagi:
Thanks — all five addressed, and the series is squashed into a single commit.

**1. Sub-frame buffers.** `Read` now returns `io.ErrShortBuffer` when `len(b) < 4`
instead of `(0, nil)`, so a caller looping until an error terminates. The guard
stays for the reason you gave: with `l == 0` the `s.rest` recompute would drop the
remainder. `TestStereoI16ReadSeekerTooShortBuffer` checks the error for buffers of
0, 1 and 3 bytes, and that such a read consumes nothing — the carried-over
remainder included.

**2. One seek.** The rewind is folded into the offset — `offset/dstFrameSize*srcFrameSize`,
minus `len(s.rest)` when `whence == io.SeekCurrent` — so `Seek` makes a single call
and there is no half-applied state, and `s.rest` is cleared only after that call
succeeds, so a failed `SeekStart`/`SeekEnd` no longer discards buffered data.
`TestStereoI16ReadSeekerFailedSeekKeepsRemainder` drives a source whose `Seek`
always fails and checks the stream still drains to the right bytes afterwards.

**3.** Agreed, out of scope. The unaligned `SeekEnd` desync reproduces identically
on `main`; I'll file it as a follow-up issue with the `Seek(-4, io.SeekEnd)` →
`22 33 22 33` case rather than widen this PR.

**4.** Restored one comment, on the `rest` field, since "the source is ahead of the
consumer by `len(rest)`" is the invariant both `Read` and `Seek` depend on. The
rest stay stripped.

**5.** Squashed, body wrapped at 72 columns, and the `(*InfiniteLoop).ensurePos`
citation is gone together with the "does silently dropped" sentence — you're right
that it can't fire there, since `ensurePos` returns early once `i.pos >= 0` and its
tell happens before any read reaches the wrapper. Title and body are updated to
cover the `Seek` rewrite as well.

`go test -race ./audio/...`, `go vet ./audio/...` and gofmt are clean, and 38
sub-cases of the new tests fail against `main`, so they're still real guards.

hajimehoshi (commented):
Ready to merge. All five points are closed, and I confirmed each independently rather than taking the reply's word for it.

**Strongest evidence, end-to-end through the public API rather than the unit alone.** Decoding a mono 16-bit WAV whose data chunk is 65 bytes (two whole frames plus a trailing byte), with the underlying reader delivering 3, 7 and 4096 bytes per call:

| | bytes decoded | SHA-256 |
|---|---|---|
| `main` | 260 | **three different hashes**, one per chunk size |
| PR | 256 | **one hash**, identical across all three, and equal to the 64-byte-data case |

On `main` the decoded audio depends on how the source happens to chunk its reads. That is gone.

Also verified: `go test -race ./audio/...`, `go vet ./audio/...` and `gofmt -l` are clean; a property sweep (every format x source lengths 0-40 x chunk sizes 1-9 x destination sizes {4, 5, 8, 12, 64}, with poisoned destination buffers) passes with tells always matching the bytes returned; and exactly 38 sub-cases of the new tests fail against `main`, matching your count.

Point by point:

1. **`io.ErrShortBuffer`.** I traced every wrapper that reads from this type — `float32BytesReader`, `Resampling.src`, `InfiniteLoop.Read` and its afterLoop fill — and each keeps its read slice frame-aligned precisely because `Read` now always returns a multiple of 4, so no real caller can trip the new error. One nit you can take or leave: it also fires for `len(b) == 0`, where `io.Reader` asks for `n == 0` and allows an error only for a known condition such as EOF. `Read(nil)` returning `ErrShortBuffer` is defensible; `(0, nil)` is the more conventional answer.
2. **One seek, `rest` cleared only on success.** Correct, and `TestStereoI16ReadSeekerFailedSeekKeepsRemainder` pins the failure path properly.
3. Agreed, follow-up issue.
4. **The one restored comment** is on `rest` and states the invariant both `Read` and `Seek` depend on. Right content, right place.
5. Squashed, body wrapped at 72, the `ensurePos` citation and the grammar slip are gone, title updated. One leftover: the GitHub PR description above is still the original text. It still advertises `TestStereoI16ReadSeekerPartialFrameAtEOF`, which no longer exists, and does not mention `Seek`. Worth syncing to the commit message, since that is what people read first.

## Two things I found while testing, both pre-existing and neither for this PR

**(a) `wav.Stream.Length()` now overstates by one frame** when the data chunk is not a whole number of source frames. Mono 16-bit with a 65-byte data chunk: `Length()` reports 260, but only 256 bytes are readable. `audio/wav/decode.go` scales `dataSize` without truncating it to a whole frame. On `main` the two agreed only because `Read` over-reported by exactly the same amount, so the two bugs cancelled and fixing one exposes the other. Benign in practice, since `InfiniteLoop` restarts on `io.EOF` regardless, and only reachable with a malformed WAV, but the wav side could truncate `dataSize` to match.

**(b) `wav.DecodeF32` on a stereo 16-bit WAV returns `(0, nil)`** when the underlying reader delivers 3 bytes at a time and the data chunk is not frame-aligned, which spins any `for pos < len(buf)` caller. That path bypasses this converter entirely, since `decode.go` only wraps when `mono || bitsPerSample != 16`, and it reproduces identically on `main`. Unrelated to this change; worth its own issue alongside the `SeekEnd` one.

---

kumagi:
Thanks for the end-to-end verification — much stronger evidence than the unit test alone.

The one nit is addressed: `Read` now returns `(0, nil)` for `len(b) == 0`, per the `io.Reader` contract, while a non-empty buffer shorter than a frame still reports `io.ErrShortBuffer`. `TestStereoI16ReadSeekerTooShortBuffer` now table-tests the expected error per buffer size (`0` → `nil`, `1` and `3` → `ErrShortBuffer`) and still checks that such a read consumes nothing, carried-over remainder included.

The PR description is now synced to the commit message, so it covers the `Seek` rewrite and the zero-length behavior. This is pushed as an amend to the single squashed commit; `go test -race ./audio/...`, `go vet ./audio/...` and `gofmt -l` are clean, and 38 sub-cases of the new tests still fail against `main`.

On the two wav findings: agreed that neither belongs here. I'll file them as separate issues alongside the unaligned `SeekEnd` one.

hajimehoshi (commented):
LGTM. The delta from the previous head is three lines of code plus the matching table row, and it is exactly right.

```go
if len(b) == 0 {
    return 0, nil
}
if len(b) < dstFrameSize {
    return 0, io.ErrShortBuffer
}
```

I checked that the test actually asserts something: `errors.Is(err, nil)` reduces to `err == target` for a nil target, so the `{0, nil}` row really does reject a non-nil error rather than passing vacuously. And I re-ran the property sweep with a `Read(nil)` interleaved before every real read: the zero-length path is a true no-op that consumes no source bytes and leaves the carried remainder intact.

`go test -race ./audio/...`, `go vet ./audio/...` and `gofmt -l` are clean, and there are still exactly 38 failing sub-cases against `main`.

## One correction to what I said last round

My end-to-end comparison last time only covered frame-unaligned data chunks (65 and 67 bytes), so I described the bug as needing a malformed WAV. With aligned data included, `main` fails on 6 of 9 configurations, including `dataLen=64`, a perfectly well-formed file:

```
main:  ch=1 bits=16 dataLen=64: decoded audio differs by source chunk size
       ch=1 bits=16 dataLen=65: differs
       ch=1 bits=16 dataLen=67: differs
       ch=2 bits= 8 dataLen=64: differs
       ch=2 bits= 8 dataLen=65: differs
       ch=2 bits= 8 dataLen=67: differs
PR:    all nine identical across chunk sizes 3, 7 and 4096
```

The trigger is simply a source whose read sizes are not a multiple of the source frame size: an odd chunk boundary from an HTTP body, a pipe, a `bufio` seam. `ch=1 bits=8` is immune only because its source frame is one byte. So this is not a malformed-input edge case. On `main`, a well-formed mono 16-bit WAV decodes to different audio depending on how the reader underneath happens to chunk its reads.

That does not change the verdict, but it does raise how much the fix is worth, and it is an argument for picking this into 2.9 rather than leaving it for the next minor.

Nothing else outstanding. The two wav-side findings are yours to file as agreed.

---

kumagi:
Filed the three follow-ups the review suggested, all pre-existing and out of scope here (re-filed to follow the issue template):

- https://github.com/hajimehoshi/ebiten/issues/3583 — `SeekEnd` lands mid-frame when the source length is not a whole number of source frames (the `Seek(-4, io.SeekEnd)` → `22 33 22 33` case).
- https://github.com/hajimehoshi/ebiten/issues/3584 — `wav.Stream.Length()` over-reports by one frame for an unaligned data chunk; exposed by fixing the `Read` over-report.
- https://github.com/hajimehoshi/ebiten/issues/3585 — `wav.DecodeFloat32` returns `(0, nil)` and spins on a stereo 16-bit WAV with an unaligned chunk read through a short-reading source; bypasses this converter.

The PR description is now synced to the commit message, including the zero-length `Read` behavior.

hajimehoshi:
I found that this issue is not only for StereoI16 but also other readers. I'll take a look. Let me close this

PR #3574, opened 2026-08-29
audio/internal/convert: fix Resampling.Seek's io.SeekEnd result

kumagi:
None

bug

## What this PR does | solves
Resampling.Seek resolved io.SeekEnd relative to the current position: it added the resolved offset to r.pos instead of assigning it, so the seek result depended on how much data had been read before the seek. For example, Seek(-4, io.SeekEnd) after reading 2000 bytes returned Length() instead of Length()-4.

Resolve the position to Length()+offset as the io.Seeker contract requires.

Add TestResamplingSeekEndAbsolute, which fails without this fix.


hajimehoshi (approved):
LGTM

hajimehoshi:
Resolve the conflict

PR #3575, opened 2026-08-29
text: fix a data race on the monotonic clock

kumagi:
None

bug

## What this PR does | solves
The monotonic clock was incremented by the before-update hook without synchronization while Draw read it from other goroutines. Make the clock an atomic.Int64.

### DIFF
--- a/text/text_test.go
+++ b/text/text_test.go
@@ -17,7 +17,6 @@ package text_test
 import (
 	"image"
 	"image/color"
-	"sync"
 	"testing"
 
 	"github.com/hajimehoshi/bitmapfont/v4"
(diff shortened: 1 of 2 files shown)


text/text_test.go
```diff
@@ -59,6 +61,28 @@ func TestTextColor(t *testing.T) {
 
 const testFaceSize = 6
 
+func TestMonotonicClockConcurrent(t *testing.T) {
```
hajimehoshi: We don't need this test

kumagi: removed.

hajimehoshi (approved):
LGTM

A race condition can happen when now is read and written from different groutines at the same time. As `text` package is already deprecated, this is not a serious issue, but the fix itself is correct

PR #3576, opened 2026-08-29
internal/shader: reject recursive function calls at compile time

### LINKED ISSUE #3536: internal/shader: recursive function calls are accepted but the graphics backends can't reliably run them
### Ebitengine Version

main (6ba90ccbd)

### Operating System

- [x] Windows
- [x] macOS
- [x] Linux
- [x] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [x] Web Browsers

Compile a Kage shader with a recursive function:

```kage
package main

func fact(n int) int {
	if n <= 1 {
		return 1
	}
	return n * fact(n-1)
}

func Fragment(dstPos vec4, srcPos vec2, color vec4) vec4 {
	return vec4(vec3(float(fact(5)) / 255), 1)
}
```

`ebiten.NewShader` accepts it without an error. Mutual recursion (two functions calling each other)
is accepted too.

Either the compiler reports an error (recursion is not supported by the shading languages that
matter), or the backends are documented and guaranteed to handle it.

The shading languages do not support recursion consistently:

- GLSL and GLSL ES forbid recursion in the spec ("Function calls cannot be recursive"); WebGL2 /
  ANGLE reject it at the driver's shader compile when the shader is first used.
- Metal (MSL) allows recursion.
- HLSL has only limited support.

### What happens instead?

The compiler generates a recursive shader. The GLSL output for the fragment function is:

```glsl
int F0(in int l0) {
	if ((l0) <= (1)) {
		return 1;
	}
	return (l0) * (F0((l0) - (1)));
}
```

Where the driver rejects it, the failure surfaces at draw time, far away from the actual mistake.
Worse, it can fail silently: on Mesa (Linux desktop GL, Xvfb) the program below reports no error
anywhere, but `fact(5)` (which is 120) renders `{0 0 0 0}` instead of `{120 120 120 255}` — the
recursion is miscompiled into garbage, while the return value's alpha is a literal `1`. A
non-recursive control that returns a constant `120` renders the correct `{120 120 120 255}`.

Because "not supported" means undefined behavior on some drivers, there is no single failure mode:
it can be a draw-time error, or silently wrong pixels, and which one depends on the driver.

Possible directions: detect cycles in the function call graph (over the IR `Call` expressions)
and reject recursive calls at compile time, or explicitly document recursion as unsupported.

```go
// This program shows that ebiten.NewShader accepts a recursive Kage function. The generated
// shader is recursive, which the graphics drivers handle inconsistently: on WebGL2/ANGLE the draw
// fails, and on some desktop drivers the recursion is miscompiled into wrong output.
package main

import (
	"fmt"
	"image/color"
	"log"

)

const shaderSrc = `//kage:unit pixels

package main

func fact(n int) int {
	if n <= 1 {
		return 1
	}
	return n * fact(n-1)
}

func Fragment(dstPos vec4, srcPos vec2, color vec4) vec4 {
	return vec4(vec3(float(fact(5)) / 255), 1)
}
`

type game struct {
	checked bool
}

func (g *game) Update() error {
	if g.checked {
		return ebiten.Termination
	}
	g.checked = true

	s, err := ebiten.NewShader([]byte(shaderSrc))
	if err != nil {
		log.Fatalf("NewShader returned an error (already fixed?): %v", err)
	}
	fmt.Println("NewShader: no error")

	dst := ebiten.NewImage(8, 8)
	defer dst.Deallocate()
	func() {
		defer func() {
			if r := recover(); r != nil {
				fmt.Printf("draw: PANIC: %v\n", r)
			}
		}()
		op := &ebiten.DrawRectShaderOptions{}
		dst.Fill(color.RGBA{R: 0xff, A: 0xff})
		fmt.Printf("before draw, pixel: %v\n", dst.At(4, 4))
		dst.DrawRectShader(8, 8, s, op)
		fmt.Println("draw: no error reported by Ebitengine")
	}()
	fmt.Printf("pixel at (4,4): %v (fact(5)=120, so ~{120,120,120,255} when the shader is correct)\n", dst.At(4, 4))

	return ebiten.Termination
}

func (g *game) Draw(screen *ebiten.Image) {}

func (g *game) Layout(outsideWidth, outsideHeight int) (int, int) {
	return outsideWidth, outsideHeight
}

func main() {
	if err := ebiten.RunGame(&game{}); err != nil {
		log.Fatal(err)
	}
}
```

kumagi:
Closes #3536

bug

## What this PR does | solves
A Kage shader with a recursive function (direct or mutual) was accepted by the compiler, but the shading languages Ebiten targets do not support recursion consistently: GLSL/GLSL ES forbid it and reject the shader at the driver when it is first used, while on some drivers (e.g. Mesa) the recursion is miscompiled into silently wrong output. The failure surfaces far from the actual mistake and can be silent.

Detect cycles in the function call graph (over the IR Call expressions) and report an error at compile time instead.

### DIFF
--- a/internal/shader/shader.go
+++ b/internal/shader/shader.go
@@ -40,6 +40,7 @@ type constant struct {
 
 type function struct {
 	name string
+	pos  token.Pos
 
 	ir shaderir.Func
 }
@@ -202,6 +203,11 @@ func Compile(src []byte, vertexEntry, fragmentEntry string, textureCount int) (*
 		return nil, &ParseError{s.errs}
 	}
 
+	s.checkRecursion()
+	if len(s.errs) > 0 {
+		return nil, &ParseError{s.errs}
+	}
+
 	// TODO: Resolve identifiers?
 	// TODO: Resolve constants
 
@@ -293,6 +299,7 @@ func (cs *compileState) parse(f *ast.File) {
 		}
 		cs.funcs = append(cs.funcs, function{
 			name: n,
+			pos:  fd.Pos(),
 			ir: shaderir.Func{
 				Index:     len(cs.funcs),
 				InParams:  inT,
@@ -480,6 +487,7 @@ func (cs *compileState) parseDecl(b *block, fname string, d ast.Decl) ([]shaderi
 				if cs.funcs[i].name == d.Name.Name {
 					// Index is already determined by the provisional parsing.
 					f.ir.Index = cs.funcs[i].ir.Index
+					f.pos = cs.funcs[i].pos
 					cs.funcs[i] = f
 					break
 				}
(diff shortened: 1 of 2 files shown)


hajimehoshi:
I'll take a look after we release v2.10

PR #3577, opened 2026-08-29
ebiten: fix the legacy AntiAlias and FillRule paths on a sub-image

kumagi:
None

bug

## What this PR does | solves
The internal offscreen images for the legacy AntiAlias and FillRule paths of DrawTriangles were created with the destination's bounds as their own bounds. For a destination whose bounds' origin is not (0, 0), e.g. a sub-image, the rendering was offset by the bounds' origin and vanished.

Create the offscreen images with the origin at (0, 0), shift the doubled vertices by the bounds' origin, and shift the result back when rendering the offscreen onto the destination.

Add TestImageDrawTrianglesWithStencilBufferOnSubImage, which fails without this fix.

### DIFF
--- a/stencilbuffer_test.go
+++ b/stencilbuffer_test.go
@@ -1,0 +1,182 @@
+// Copyright 2026 The Ebitengine Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ebiten_test
+
+import (
+	"image"
+	"image/color"
+	"testing"
+
+	"github.com/hajimehoshi/ebiten/v2"
+)
+
+func TestImageDrawTrianglesWithStencilBufferOnSubImage(t *testing.T) {
+	whiteImage := ebiten.NewImage(3, 3)
+	whiteImage.Fill(color.White)
+	whiteSubImage := whiteImage.SubImage(image.Rect(1, 1, 2, 2)).(*ebiten.Image)
+
+	shader, err := ebiten.NewShader([]byte(`//kage:unit pixels
+
+package main
+
+func Fragment(dstPos vec4, src0Pos vec2, color vec4) vec4 {
+	return vec4(1, 0, 0, 1)
+}
+`))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	const size = 20
+	// The sub-image's origin is at least as large as its own size, so that a
+	// draw lost by keeping the destination's coordinate space cannot
+	// accidentally fall inside the sub-image.
+	dstMin := image.Pt(40, 30)
+
+	vertices := func(ox, oy int) []ebiten.Vertex {
+		v := func(x, y int) ebiten.Vertex {
+			return ebiten.Vertex{
+				DstX:   float32(ox + x),
+				DstY:   float32(oy + y),
+				SrcX:   1,
+				SrcY:   1,
+				ColorR: 1,
+				ColorG: 1,
+				ColorB: 1,
+				ColorA: 1,
+			}
+		}
+		return []ebiten.Vertex{
+			v(2, 2),
+			v(18, 2),
+			v(2, 18),
+			v(4, 4),
+			v(10, 4),
+			v(4, 10),
+		}
+	}
+	// The second triangle is inside the first one with the same winding
+	// order, so that FillRuleNonZero and FillRuleEvenOdd result in different
+	// pixels.
+	is := []uint32{0, 1, 2, 3, 4, 5}
+
+	dt := func(options *ebiten.DrawTrianglesOptions) func(*ebiten.Image, int, int) {
+		return func(dst *ebiten.Image, ox, oy int) {
+			dst.DrawTriangles32(vertices(ox, oy), is, whiteSubImage, options)
+		}
+	}
+	dts := func(options *ebiten.DrawTrianglesShaderOptions) func(*ebiten.Image, int, int) {
+		return func(dst *ebiten.Image, ox, oy int) {
+			options.Images[0] = whiteSubImage
+			dst.DrawTrianglesShader32(vertices(ox, oy), is, shader, options)
+		}
+	}
+
+	for _, tc := range []struct {
+		name string
+		draw func(dst *ebiten.Image, ox, oy int)
+	}{
+		{
+			name: "FillRuleNonZero",
+			draw: dt(&ebiten.DrawTrianglesOptions{
+				FillRule: ebiten.FillRuleNonZero,
+			}),
+		},
+		{
+			name: "FillRuleEvenOdd",
+			draw: dt(&ebiten.DrawTrianglesOptions{
+				FillRule: ebiten.FillRuleEvenOdd,
+			}),
+		},
+		{
+			name: "AntiAlias",
+			draw: dt(&ebiten.DrawTrianglesOptions{
+				AntiAlias: true,
+			}),
+		},
+		{
+			name: "AntiAliasFillRuleNonZero",
+			draw: dt(&ebiten.DrawTrianglesOptions{
+				AntiAlias: true,
+				FillRule:  ebiten.FillRuleNonZero,
+			}),
+		},
+		{
+			name: "AntiAliasBlendCopy",
+			draw: dt(&ebiten.DrawTrianglesOptions{
+				AntiAlias: true,
+				Blend:     ebiten.BlendCopy,
+			}),
+		},
+		{
+			name: "ShaderFillRuleNonZero",
+			draw: dts(&ebiten.DrawTrianglesShaderOptions{
+				FillRule: ebiten.FillRuleNonZero,
+			}),
+		},
+		{
+			name: "ShaderFillRuleEvenOdd",
+			draw: dts(&ebiten.DrawTrianglesShaderOptions{
+				FillRule: ebiten.FillRuleEvenOdd,
+			}),
+		},
+		{
+			name: "ShaderAntiAlias",
+			draw: dts(&ebiten.DrawTrianglesShaderOptions{
+				AntiAlias: true,
+			}),
+		},
+		{
+			name: "ShaderAntiAliasFillRuleNonZero",
+			draw: dts(&ebiten.DrawTrianglesShaderOptions{
+				AntiAlias: true,
+				FillRule:  ebiten.FillRuleNonZero,
+			}),
+		},
+		{
+			name: "ShaderAntiAliasBlendCopy",
+			draw: dts(&ebiten.DrawTrianglesShaderOptions{
+				AntiAlias: true,
+				Blend:     ebiten.BlendCopy,
+			}),
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			background := color.RGBA{0, 0, 0, 0xff}
+
+			dstPlain := ebiten.NewImage(size, size)
+			dstPlain.Fill(background)
+
+			base := ebiten.NewImage(dstMin.X+size, dstMin.Y+size)
+			base.Fill(background)
+			dstSub := base.SubImage(image.Rectangle{Min: dstMin, Max: dstMin.Add(image.Pt(size, size))}).(*ebiten.Image)
+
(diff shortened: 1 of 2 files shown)


hajimehoshi (commented):
Thanks for tracking this down. The diagnosis is right, and I confirmed the antialias half of the fix works. But the non-antialias `FillRule` path is still broken after this change — the new test only passes because of the particular offset it picks.

## `doDrawTrianglesShaderWithStencilBuffer` half-converts the coordinate space

`bounds` is moved to the origin unconditionally, but the vertices are shifted, and `op.GeoM.Translate(dstMin)` is applied, only inside the `AntiAlias` branch:

```go
vs := vertices
dstMin := dst.Bounds().Min
bounds := dst.Bounds().Sub(dst.Bounds().Min)
if ...AntiAlias... {
    // vertices shifted here only
}
```

So for `FillRule != FillRuleFillAll && !AntiAlias`, the offscreen images now live at `(0, 0)-(w, h)` while `vs` is still in the destination's coordinate space (`dstMin` .. `dstMin+size`). The triangles are drawn outside the offscreen and clipped away, and the final `dst.DrawImage(finalOS, op)` has no translation back either. When the sub-image's origin is at least its size, nothing is drawn at all.

I compared a sub-image destination against an equivalent plain destination over a whole 20x20 region, with the sub-image at offset (40, 30):

| case | main | this PR |
| --- | --- | --- |
| `FillRuleNonZero` | 120/400 px wrong | **120/400 px wrong** |
| `FillRuleEvenOdd` | 120/400 px wrong | **120/400 px wrong** |
| `AntiAlias` | 136/400 px wrong | 0 (fixed) |
| `AntiAlias` + `FillRuleNonZero` | 136/400 px wrong | 0 (fixed) |
| `AntiAlias` + `BlendCopy` | 136/400 px wrong | 0 (fixed) |

This makes all five pass:

```go
	dstMin := dst.Bounds().Min
	bounds := dst.Bounds().Sub(dstMin)
	var scale float32 = 1
	if dtOptions != nil && dtOptions.AntiAlias || dtsOptions != nil && dtsOptions.AntiAlias {
		bounds.Max.X *= 2
		bounds.Max.Y *= 2
		scale = 2
	}
	tmpVerticesForStencilBuffer = slices.Grow(tmpVerticesForStencilBuffer, len(vertices))
	vs := tmpVerticesForStencilBuffer[:len(vertices)]
	copy(vs, vertices)
	for i := range vs {
		vs[i].DstX = scale * (vs[i].DstX - float32(dstMin.X))
		vs[i].DstY = scale * (vs[i].DstY - float32(dstMin.Y))
	}
```

together with moving `op.GeoM.Translate(float64(dstMin.X), float64(dstMin.Y))` out of both `if ...AntiAlias` blocks at the end of the function, so that it always runs.

One caveat: this always copies the vertices and loses the current `vs := vertices` fast path. Guarding the copy on `scale != 1 || dstMin != (image.Point{})` keeps the common full-image, non-antialias case free of the copy.

## The test doesn't guard the fix

`TestImageDrawTrianglesWithStencilBufferOnSubImage` samples three points with the sub-image origin at (10, 10) on a 20x20 region. With the `FillRule` path still broken, two of those points fall inside the accidentally surviving `[10, 20)` window and the third is background in both images, so the test passes. It does fail on unpatched main, so it isn't useless, but it doesn't pin down the behavior. Some suggestions:

* Compare the whole region pixel by pixel against a plain destination instead of three sample points.
* Put the sub-image origin at least at its own size (e.g. (40, 30) for a 20x20 region), so a completely lost draw can't hide.
* Cover `FillRuleEvenOdd` and the `DrawTrianglesShader32` variants: the shader path shares this code and this bug.
* Cover a non-`SourceOver` blend, which exercises the `uncommonBlend` `os1.DrawImage(dst, ...)` copy in `doDrawTrianglesWithAntialias`. This PR does fix it, but nothing tests it.

## Minor

* `dst.Bounds().Sub(dst.Bounds().Min)` calls `dst.Bounds()` twice in both functions; `dst.Bounds().Sub(dstMin)` reads better now that `dstMin` exists.
* `2*vs[i].DstX - 2*float32(dstMin.X)` states the transform more directly as `2 * (vs[i].DstX - float32(dstMin.X))`.
* Moving the offscreens to the origin is a nice benefit beyond the bug fix: `ensureOffscreenImage*` unions with the previous bounds, so a sub-image far into a large destination used to inflate the cached offscreen images to cover the whole offset. That's worth mentioning in the commit message.

---

*This review was written by Claude (Claude Code), on behalf of @hajimehoshi.*

kumagi:
Thank you for the thorough review! I've addressed all the points:

* `doDrawTrianglesShaderWithStencilBuffer` now converts the coordinate space unconditionally: the vertices are shifted by the destination's bounds' origin with a `scale` (1 or 2), and `op.GeoM.Translate(dstMin)` is moved out of the `AntiAlias` branches so that it always runs. As you suggested, the vertex copy is guarded by `scale != 1 || dstMin != (image.Point{})`, so the common full-image, non-antialias case keeps the `vs := vertices` fast path.
* The test now compares the whole 20x20 region pixel by pixel against a plain destination, with the sub-image origin at (40, 30), so a completely lost draw can't hide. It also covers `FillRuleEvenOdd`, the `DrawTrianglesShader32` variants, and `AntiAlias` + `BlendCopy` to exercise the `uncommonBlend` path. The geometry uses a second triangle with the opposite winding order so that NonZero and EvenOdd actually differ.
* Applied the minor cleanups: `dst.Bounds().Sub(dstMin)` and the `2 * (vs[i].DstX - float32(dstMin.X))` form.
* Mentioned the offscreen-caching side benefit (no more inflation of the cached offscreen images for offset destinations) in the commit message.

I confirmed that the new test fails on the previous revision (the non-antialias FillRule cases catch the half-converted coordinate space) and fails on unpatched main for all cases, and it passes with this fix. The full `go test .` also passes.

PTAL, thanks!

hajimehoshi (commented):
Thanks, this looks good now. I verified it rather than taking it on trust.

The fix is correct: `doDrawTrianglesShaderWithStencilBuffer` converts the coordinate space unconditionally, `op.GeoM.Translate(dstMin)` is hoisted out of both `AntiAlias` branches, and the `vs := vertices` fast path is preserved behind the `scale != 1 || dstMin != (image.Point{})` guard.

| | main | previous revision | this revision |
| --- | --- | --- | --- |
| the 10 new subtests | all 10 fail | 4 fail (the non-antialias `FillRule` ones) | all pass |

Full `go test .` passes, and `gofmt` and `go vet` are clean.

I also checked two cases the test doesn't cover, and both are handled correctly: a destination created by `NewImageWithOptions` with a non-zero origin that is *not* a sub-image (which takes a different branch in `Image.adjustPosition`), and a sub-image of an image that itself has a non-zero origin. Using `dst.Bounds().Min` rather than the original image's origin is the right choice there.

## The winding comment is backwards

```go
// The second triangle is inside the first one with the opposite winding
// order, so that FillRuleNonZero and FillRuleEvenOdd result in different
// pixels.
is := []uint32{0, 1, 2, 3, 5, 4}
```

The opposite winding order is exactly the case where the two rules *agree*: inside both triangles the winding number is `+1 - 1 = 0` (not filled) and the crossing count is 2, which is even (not filled), so both rules leave the same hole. Measured on a 20x20 destination:

| inner triangle | NonZero vs EvenOdd |
| --- | --- |
| opposite winding (`3, 5, 4`, as in this PR) | **0 px differ** |
| same winding (`3, 4, 5`) | 15 px differ |

So the comment states something that isn't true, and the geometry doesn't actually distinguish the two fill rules: swapping the two stencil shaders would go unnoticed. `is := []uint32{0, 1, 2, 3, 4, 5}` achieves the stated intent, and the comment then needs to say "the same winding order".

This doesn't weaken what the PR is actually about, since each fill rule is still compared sub-image against plain and those comparisons are what fail on main. It's a fix-up rather than a blocker.

## Optional nit

The per-pixel `t.Errorf` is uncapped, so a regression prints up to 400 lines per subtest, 1206 lines in total against main. There is precedent for this in `image_test.go`, so take it or leave it, but breaking after the first mismatch in each subtest would keep failures readable.

---

*This review was written by Claude (Claude Code), on behalf of @hajimehoshi.*

kumagi:
Thanks for the second round! Both points are addressed.

## Winding comment / geometry

You're right, and I verified it independently rather than assuming: with the opposite winding (`3, 5, 4`) the inner overlap gets winding number `+1 - 1 = 0` and crossing count `2`, so NonZero and EvenOdd both leave the hole and **0 px differ** — swapping the two stencil shaders would have gone unnoticed. With the same winding they diverge by **15 px**, matching your measurement.

Fixed the indices and the comment:

```diff
-	// The second triangle is inside the first one with the opposite winding
+	// The second triangle is inside the first one with the same winding
 	// order, so that FillRuleNonZero and FillRuleEvenOdd result in different
 	// pixels.
-	is := []uint32{0, 1, 2, 3, 5, 4}
+	is := []uint32{0, 1, 2, 3, 4, 5}
```

## Optional nit

Took it. Each subtest now reports only the first mismatch:
```diff
 					if got != want {
+						// Report only the first mismatch to keep the log readable.
 						t.Errorf("pixel (%d, %d) of the sub-image: got %v, want %v", i, j, got, want)
+						return
 					}
```
Against unpatched main this prints exactly one line per failing subtest (10 lines total instead of 1206), while still failing all 10.

On the corrected geometry the full test still passes on this revision and still fails all 10 subtests on main. `gofmt` and `go vet` are clean. Amended and pushed.

stencilbuffer_test.go
```diff
@@ -0,0 +1,116 @@
...
+		{"FillRuleEvenOdd", dt(&ebiten.DrawTrianglesOptions{FillRule: ebiten.FillRuleEvenOdd})},
+		{"AntiAlias", dt(&ebiten.DrawTrianglesOptions{AntiAlias: true})},
+		{"AntiAliasFillRuleNonZero", dt(&ebiten.DrawTrianglesOptions{AntiAlias: true, FillRule: ebiten.FillRuleNonZero})},
+		{"AntiAliasBlendCopy", dt(&ebiten.DrawTrianglesOptions{AntiAlias: true, Blend: ebiten.BlendCopy})},
+		{"ShaderFillRuleNonZero", dts(&ebiten.DrawTrianglesShaderOptions{FillRule: ebiten.FillRuleNonZero})},
+		{"ShaderFillRuleEvenOdd", dts(&ebiten.DrawTrianglesShaderOptions{FillRule: ebiten.FillRuleEvenOdd})},
+		{"ShaderAntiAlias", dts(&ebiten.DrawTrianglesShaderOptions{AntiAlias: true})},
+		{"ShaderAntiAliasFillRuleNonZero", dts(&ebiten.DrawTrianglesShaderOptions{AntiAlias: true, FillRule: ebiten.FillRuleNonZero})},
+		{"ShaderAntiAliasBlendCopy", dts(&ebiten.DrawTrianglesShaderOptions{AntiAlias: true, Blend: ebiten.BlendCopy})},
+	} {
```
hajimehoshi, line 155: For readability, use member names and multiple lines

stencilbuffer_test.go
```diff
@@ -0,0 +1,116 @@
...
+	dstMin := image.Pt(40, 30)
+
+	vertices := func(ox, oy int) []ebiten.Vertex {
+		return []ebiten.Vertex{
+			{DstX: float32(ox + 2), DstY: float32(oy + 2), SrcX: 1, SrcY: 1, ColorR: 1, ColorG: 1, ColorB: 1, ColorA: 1},
+			{DstX: float32(ox + 18), DstY: float32(oy + 2), SrcX: 1, SrcY: 1, ColorR: 1, ColorG: 1, ColorB: 1, ColorA: 1},
+			{DstX: float32(ox + 2), DstY: float32(oy + 18), SrcX: 1, SrcY: 1, ColorR: 1, ColorG: 1, ColorB: 1, ColorA: 1},
+			{DstX: float32(ox + 4), DstY: float32(oy + 4), SrcX: 1, SrcY: 1, ColorR: 1, ColorG: 1, ColorB: 1, ColorA: 1},
+			{DstX: float32(ox + 10), DstY: float32(oy + 4), SrcX: 1, SrcY: 1, ColorR: 1, ColorG: 1, ColorB: 1, ColorA: 1},
+			{DstX: float32(ox + 4), DstY: float32(oy + 10), SrcX: 1, SrcY: 1, ColorR: 1, ColorG: 1, ColorB: 1, ColorA: 1},
```
hajimehoshi: For readability, use multiple lines

hajimehoshi (commented):
The change seems good.

stencilbuffer_test.go
kumagi: done

stencilbuffer_test.go
kumagi: done.

hajimehoshi (approved):
LGTM

PR #3578, opened 2026-08-29
ebiten: accept a nil source image at index 0 in texels mode

kumagi:
None

bug

## What this PR does | solves
DrawTrianglesShader derived the reference image size only from Images[0], so a texel-unit draw whose first slot was nil panicked even when a later slot held a valid image. Use the first non-nil image's size instead.

### DIFF
--- a/image_test.go
+++ b/image_test.go
@@ -2296,7 +2296,7 @@ func TestImageDrawTrianglesShaderInterpolatesValues(t *testing.T) {
 		Images: [4]*ebiten.Image{src, nil, nil, nil},
 	}
 	is := []uint16{0, 1, 2, 1, 2, 3}
-	shader, err := ebiten.NewShader([]byte(`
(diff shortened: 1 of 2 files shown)


image_test.go
```diff
@@ -4791,7 +4791,7 @@ func TestImageDrawTrianglesShader32(t *testing.T) {
 	}
 	is := []uint32{0, 1, 2, 1, 2, 3}
 	op := &ebiten.DrawTrianglesShaderOptions{}
-	shader, err := ebiten.NewShader([]byte(`//kage:unit pixels
```
hajimehoshi, line 4794: I don't understand why you changed the modes

hajimehoshi:
Ah, I understand what you are trying to fix. DrawTrianglesShader always use the 0th source size as the base, so the fix is not correct. This might be odd, but this feature is valid only for a legacy texel mode, so I leave this as it is.

PR #3579, opened 2026-08-29
text/v2: report glyph indices in the original text at LimitedFace

kumagi:
None

bug

## What this PR does | solves
LimitedFace.appendLazyGlyphsForLine passed the filtered string to the inner face as-is, so the LazyGlyph byte indices referred to positions in the filtered string instead of the original text. Whenever an unsupported rune's byte length differs from U+FFFD's 3 bytes, every glyph after it got wrong indices.

Remap the appended glyphs' indices to the original string.

Add TestLimitedFilterMapping, which fails without this fix.


hajimehoshi (commented):
*Edited: this review previously contained text belonging to a different pull request, posted here in error. Replaced below with the intended review of this PR. Apologies for the confusion.*

---

Reviewed at 7970c03.

Thanks for the patch. **The bug is real and worth fixing.** I confirmed it end-to-end on `main`: with a `LimitedFace` allowing `'a'`–`'d'`, `AppendLazyGlyphs("aébc", …)` returns glyph ranges `[0,1) [1,4) [4,5) [5,6)` on a 5-byte string. The last one is out of range, so any caller slicing the original text panics. The patched code returns `[0,1) [1,3) [3,4) [4,5)`, which is correct. I also checked multi-line, invalid UTF-8, 4-byte-rune, and leading-replacement inputs; all come out right.

The lockstep-decode approach is also the right instinct. The "obvious" alternative — rebuilding the mapping from `unicodeRanges.Contains` the way `advanceAt` does — is actually *wrong*, because `Filter` copies `str[:replaceStartIndex]` verbatim and only re-encodes the suffix, so an invalid byte's width in `filtered` depends on where it sits. I wrote that alternative first and it failed. Worth keeping in mind if a simplification is proposed later.

Findings below, roughly by importance.

### 1. `limitedFilterMapping` leaves zeroed holes on invalid UTF-8, and the rune comparison is unnecessary

The `r == fr` branch writes `size` entries but advances `fi` by `fsize`. For an invalid byte these differ (`DecodeRuneInString` gives U+FFFD with `size == 1`, while `Filter` re-encodes it to 3 bytes), so the skipped slots keep their zero value:

```
line="a\xffbc" range 'a'-'d'  ->  filtered="a�bc"
mapping = [0 1 0 0 2 3 4]     // entries 2,3 should be 1 -- non-monotonic
```

This is currently latent. Over 20,000 random inputs I found 27,138 divergences from a faithful model of `Filter`, **all** at byte positions interior to a rune, and **zero** at the rune boundaries a glyph index can actually land on. So there is no user-visible symptom today, but it is a wrong table sitting in a hot path.

The comparison can go away entirely, which fixes this as a side effect. Both branches consume exactly one rune from each string, and `fsize` is already 3 in the replacement case:

```go
for oi < len(line) && fi < len(filtered) {
	_, size := utf8.DecodeRuneInString(line[oi:])
	_, fsize := utf8.DecodeRuneInString(filtered[fi:])
	for k := range fsize {
		if k < size {
			mapping[fi+k] = oi + k
		} else {
			mapping[fi+k] = oi
		}
	}
	oi += size
	fi += fsize
}
```

I ran this against a faithful reimplementation of `Filter` over 50,000 random strings (mixed 1-4 byte runes, literal U+FFFD, invalid bytes, three different range sets): identical at *every* position, and the full `text/v2` suite passes.

### 2. The new test does not test the fix

Mutation check: I replaced the whole body of `limitedFilterMapping` with

```go
mapping := make([]int, len(filtered)+1)
mapping[len(filtered)] = len(line)
return mapping
```

Every glyph then reports `StartIndexInBytes == 0`, i.e. the exact bug this PR fixes, in a worse form. `TestLimitedFilterMapping` passes, and so does the entire `text/v2` package. The test only checks that entries are in range and that the last one is `len(line)`, not that the mapping is correct.

Relatedly, this line is vacuous:

```go
if !utf8.ValidRune(rune(tc.line[oi])) && oi != len(tc.line) {
```

`tc.line[oi]` is a `byte`, so `rune(...)` is always 0-255 and `utf8.ValidRune` is always true; the branch never fires. The intent was presumably `utf8.RuneStart(tc.line[oi])`. The `&& oi != len(tc.line)` clause is dead too, since the preceding `if oi >= len(tc.line)` already `continue`d for that case.

A test that actually pins the behavior goes through the public API and asserts the substrings:

```go
l := text.NewLimitedFace(base)
l.AddUnicodeRange('a', 'd')
line := "aébc"
var got []string
for _, g := range text.AppendLazyGlyphs(nil, line, l, nil) {
	got = append(got, line[g.StartIndexInBytes:g.EndIndexInBytes])
}
// want: []string{"a", "é", "b", "c"}
```

That one panics on `main` (index out of range), which is a much stronger demonstration than "the helper doesn't exist yet".

### 3. Test package convention

`limited_test.go` declares `package text`. Every other `_test.go` under `text/v2` is `package text_test`; `export_test.go` is the only `package text` file and it holds no tests. Also, `face_test.go` already has `LimitedFace` coverage (`limitedFaceTestFace`, `TestLimitedFaceAdvanceAtNewline`), so an end-to-end test would sit naturally there.

### 4. Per-draw allocation on the text hot path

`make([]int, len(filtered)+1)` runs once per line per `Draw` / `AppendGlyphs` / `AppendLazyGlyphs`. Benchmark (161-byte line, one filtered rune, `GoTextFace`, M3 Pro):

```
before  4494 ns/op   3140 B/op   4 allocs/op
after   5139 ns/op   4548 B/op   5 allocs/op
```

That is 8 bytes per *byte* of line; a 1 KB line costs 8 KB per draw. `text/v2` already pools exactly this kind of scratch buffer (`theDrawGlyphsPool`, `theAppendGlyphsLazyBufPool`, `rgbapool`), so a `sync.Pool` (or `[]int32`) would keep it off the hot path.

### 5. Minor

- `limitedFilterMapping` has no doc comment. What the slice means (index in `filtered` -> index in `line`, `len(filtered)+1` entries, last entry `len(line)`) is not inferable from the signature.
- `if size <= 0 { size = 1 }` is unreachable: `DecodeRuneInString` returns 0 only for an empty string, which the loop condition excludes. (`advanceAt` above has the same dead `runeLen < 0` guard, so this may be deliberate mirroring.)
- `mapping[fi+fffdLen] = oi + size` writes the next iteration's slot, which that iteration, or the trailing loop, immediately rewrites with the same value.
- The out-of-range guard in `appendLazyGlyphsForLine` cannot trigger: glyph indices are always within `[0, len(filtered)]` and `len(mapping)` is `len(filtered)+1`. Silently `continue`ing there would also leave the wrong index behind, so if it is kept as defensive code it is worth a comment saying it is unreachable.

### Adjacent, pre-existing (not this PR's job)

`advanceAt` has the same class of bug for invalid UTF-8: it adds `runeLen` for runes inside an allowed range, but `Filter` re-encodes everything after `replaceStartIndex` through `WriteRune`, so an invalid byte that decodes to U+FFFD *and* falls inside an allowed range is 1 byte in the source and 3 in `filtered`. With `line = "Z\xffb"` and range `'a'`-`0xFFFF`, `advanceAt` computes filtered index 4 for `'b'`, which actually sits at 6. Mentioning it because a shared original<->filtered mapping helper would fix both, and that may be the cleaner shape for this patch.

---

Overall: correct diagnosis, correct result for valid UTF-8. I would like the test to actually pin the mapping before merging; as it stands it does not guard the fix.

---

*Review performed by Claude (Claude Code) on behalf of @hajimehoshi.*

hajimehoshi (commented):
_

kumagi:
Both inline comments addressed:

* Vertex literals are now multiple lines. Since every vertex shared the same `Src` and color fields, I factored that into a small helper so the table stays readable and only the coordinates are spelled out:

```go
	vertices := func(ox, oy int) []ebiten.Vertex {
		v := func(x, y int) ebiten.Vertex {
			return ebiten.Vertex{
				DstX:   float32(ox + x),
				DstY:   float32(oy + y),
				SrcX:   1,
				SrcY:   1,
				ColorR: 1,
				ColorG: 1,
				ColorB: 1,
				ColorA: 1,
			}
		}
		return []ebiten.Vertex{
			v(2, 2),
			v(18, 2),
			v(2, 18),
			v(4, 4),
			v(10, 4),
			v(4, 10),
		}
	}
```

* The test-case table now uses member names (`name:` / `draw:`) and one entry per multiple lines, with the options literals expanded the same way.

The geometry and logic are unchanged, so the test still passes all 10 subtests on this revision and still fails all 10 against unpatched main. `gofmt`, `go vet`, and the full `go test .` are clean. Amended and pushed.

hajimehoshi (commented):
Reviewed at 60ba488. This addresses everything from the previous round. I verified rather than assumed:

**The mapping rewrite is correct.** I ran it against a faithful reimplementation of `UnicodeRanges.Filter` over 200,000 random strings (mixed 1-4 byte runes, literal U+FFFD, invalid bytes, four different range sets), including with a reused buffer: identical at **every** byte position, not just at rune boundaries. The interior-hole defect from the previous version is gone.

**The test now guards the fix.** Same mutation as before, forcing the mapping to all zeros:

```
"aébc": glyphs: got ["" "" "" "aébc"], want ["a" "é" "b" "c"]
```

The old test passed that; this one fails on four of the six cases. Good regression guard.

**Pooling removed the allocation.** `AppendLazyGlyphs` on a 161-byte line with one filtered rune: `4548 B/op, 5 allocs/op` -> `3123 B/op, 4 allocs/op`, back to the pre-fix baseline. `gofmt`, `go vet` and the full `text/v2` suite pass.

Two small things left, neither blocking.

## 1. `limitedFilterMapping` does not append to `dst`

The doc says "appends to dst", but `slices.Grow(dst, len(filtered)+1)[:len(filtered)+1]` reslices from index 0, so `dst`'s contents are overwritten rather than preserved:

```
dst = []int{111, 222, 333}
limitedFilterMapping(dst, "aébc", filtered)  ->  [0 1 2 1 3 4 5]   // len 7, not 3+7
```

Harmless as written, since the only caller passes `(*mappingP)[:0]`. But the parameter name `dst` and the word "appends" both signal append semantics to the next caller. Something like "buf is reused as the backing array; its contents are discarded", with the parameter renamed to `buf`, would match what it does.

Minor, in the same doc comment: "byte indices in the middle of a rune in filtered have no correspondences in line" is now slightly off, since the loop does assign those slots (`oi+k`, or `oi` past the rune's length). The advice not to rely on them is right; it is the "no correspondences" phrasing that no longer matches the code.

## 2. Two gaps in the test table

* `{"�abc", ...}` is commented "The first rune is replaced", but U+FFFD filters to itself, so `filtered == line` and the case takes the early-return fast path without ever reaching the mapping. `"éabc"` is the input that actually exercises a replaced first rune.
* Nothing covers an invalid byte *after* a replacement, e.g. `"é\xffbc"` -> `"��bc"`. That 1-byte-to-3-byte re-encode is exactly what produced the wrong mapping entries in the previous version, so it is the natural regression case for what this commit fixed.

Both behave correctly today, they are just untested:

```
"éabc"    -> ["é" "a" "b" "c"]
"é\xffbc" -> ["é" "\xff" "b" "c"]
"ab\xffé" -> ["a" "b" "\xff" "é"]
```

## Note on the removed bounds check

Dropping it is the right call, but it does promote `[0, len(filtered)]` to a hard invariant that every `Face` implementation has to honour, with a panic rather than a silent skip if one ever does not. I traced `GoTextFace`, `GoXFace` and `MultiFace`, including `LimitedFace` nested inside `MultiFace` and vice versa, and all of them hold. The comment documents it; just noting that the contract is now load-bearing.

LGTM once the two nits above are addressed, or as-is if you would rather land it and follow up.

---

*Review performed by Claude (Claude Code) on behalf of @hajimehoshi.*

kumagi:
Thanks for the careful review! Both nits are addressed in aeb4dbe:

1. The `dst` parameter of `limitedFilterMapping` is renamed to `buf`, and the doc comment now says buf is reused as the backing array and its contents are discarded, instead of claiming it appends. The middle-of-rune sentence now states that those entries are assigned arbitrary values and should not be relied on, which matches the code.

2. Test table gaps:
   * `{"éabc", …}` is added as the input that actually exercises a replaced first rune and reaches the mapping. The literal `"\ufffdabc"` entry is kept, but its comment now correctly describes what it pins: U+FFFD filters to itself, so the filtered string equals the original one and the fast path is taken.
   * Invalid bytes around replacements are covered: `{"é\xffbc", …} -> ["é", "\xff", "b", "c"]` and the reverse order `{"ab\xffé", …} -> ["a", "b", "\xff", "é"]`, matching the behavior you verified.

I re-ran the mutation check (mapping forced to a broken one): the test fails with `"aébc": glyph indices [5, 6) are out of range`. gofmt, go vet and the full text/v2 suite, including -race, pass.

hajimehoshi (commented):
Reviewed at aeb4dbe. Both remaining points are addressed correctly, and I have a correction to make on one of them.

**Doc comment.** `dst` -> `buf`, "appends to dst" -> "returns a mapping", plus the explicit note that `buf` is reused as the backing array and its contents discarded. The interior-index sentence is fixed too: it now says those slots are assigned but the values are arbitrary and should not be relied on, which matches the code. Leaving them formally unspecified rather than documenting the `oi+k` / `oi` values is the better contract.

**Test table.** `"éabc"`, `"é\xffbc"` and `"ab\xffé"` added, and the "first rune is replaced" comment moved off `"�abc"` onto the case it actually describes. `"�abc"` now correctly notes that the filtered string equals the original, which documents the early-return path.

`gofmt`, `go vet` and the full `text/v2` suite pass. The mapping logic is byte-identical to 60ba488 apart from the parameter rename; I re-ran the differential against a reimplementation of `Filter` at this commit anyway: 107,155 filtered cases out of 200,000 random inputs, exact at every byte position.

## Correction: I was wrong about `"é\xffbc"`

In the previous round I described that case as "the natural regression case for what this commit fixed". I checked that claim and it does not hold. Reverting `limitedFilterMapping` to the earlier hole-producing form (write `size` entries, still advance `fi` by `fsize`) leaves the expanded table passing:

```
mutation: mapping forced to all zeros        -> 7 failures, caught
mutation: reintroduce the interior holes     -> ok, NOT caught
```

That is the same property that made the original defect latent in the first place: a glyph index is always at a rune boundary, so the interior slots are never read. My apologies for sending you after a test case on a premise that turns out to be wrong.

The cases are still worth keeping. They cover `Filter`'s verbatim-prefix behaviour and the 1-byte-to-3-byte re-encode, and they would catch a genuine misalignment. They just do not discriminate the interior-hole fix specifically.

I do not think anything further is warranted here. That behaviour is unobservable through the public API by construction, so the only thing that could guard it is a direct unit test on the unexported helper, which is exactly what this PR moved away from when the test was relocated to `package text_test`. Reintroducing an internal test to pin values that can never be read would not be a good trade.

No remaining findings from me. LGTM.

---

*Review performed by Claude (Claude Code) on behalf of @hajimehoshi.*

kumagi:
Thanks for the re-review and the correction — no hard feelings, the added cases are useful coverage regardless, and your differential checks gave me a lot of confidence in the mapping.

Since there are no remaining findings, I'll leave the branch as-is at aeb4dbe. PTAL for merge.

text/v2/face_test.go
```diff
@@ -68,6 +69,49 @@ func TestLimitedFaceAdvanceAtNewline(t *testing.T) {
 	}
 }
 
+// TestLimitedFaceAppendLazyGlyphsIndicesInOriginalText checks that the indices
+// of glyphs appended for a LimitedFace are indices in the original text, not in
+// the text filtered by the unicode ranges.
```
hajimehoshi: Remove the comment

text/v2/limited.go
```diff
@@ -87,7 +89,79 @@ func (l *LimitedFace) hasGlyph(r rune) bool {
...
+// The returned mapping has len(filtered)+1 entries so that the end index of
+// the last rune can be mapped, and thus its last entry is len(line).
+//
+// Filter replaces every unsupported rune with U+FFFD, whose byte length can
+// differ from the replaced rune's length. As Filter never adds nor removes a
+// rune, a rune in line corresponds to a rune in filtered at the same order,
+// and the indices are translated rune by rune. Byte indices in the middle of
+// a rune in filtered are assigned values in the mapping, but those values are
+// arbitrary and should not be relied on: a glyph index is always at a rune
+// boundary.
```
hajimehoshi: Can we shorten comments?

text/v2/face_test.go
kumagi: Done in 928493b: the comment is removed. The test name says it all.

text/v2/limited.go
kumagi: Shortened in 928493b: 17 lines to 11, keeping the points that were worth pinning down (buf is reused and its contents are discarded, len(filtered)+1 entries with the last entry len(line), and the interior values being arbitrary). The sentence about Filter never adding nor removing a rune is gone as it is implied by the rune-by-rune translation.

hajimehoshi (approved):
LGTM

Issue #3318, opened 2025-09-18
an empty Ebitengine application uses 6-8% CPU


hajimehoshi:
### Ebitengine Version

4bb016f262b1f162a2e8ba86e33419abea47d62b

### Operating System

- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

### Go Version (`go version`)

go version go1.24.3 darwin/arm64

### What steps will reproduce the problem?

1. Run this program:

```go
package main

import (
	"log"
	"net/http"
	_ "net/http/pprof"

	"github.com/hajimehoshi/ebiten/v2"
)

type Game struct{}

func (self *Game) Layout(w, h int) (int, int) {
	return w, h
}

func (self *Game) Update() error {
	return nil
}

func (self *Game) Draw(*ebiten.Image) {}

func main() {
	go func() {
		log.Println(http.ListenAndServe("localhost:6060", nil))
	}()

	game := &Game{}
	ebiten.SetScreenClearedEveryFrame(false)
	if err := ebiten.RunGame(game); err != nil {
		log.Fatal(err)
	}
}
```

2. See the CPU usage at Activity Monitor
3. Run 

### What is the expected result?

CPU usages should be much low like 1%

### What happens instead?

CPU usages were around 6-8%

### Anything else you feel useful to add?

`go tool pprof -http=:8081 "http://localhost:6060/debug/pprof/profile?seconds=5"` shows that pthread_cond is called many times

<img width="3600" height="1520" alt="Image" src="https://github.com/user-attachments/assets/6f276874-0e71-4f04-88b3-dceaaec6253f" />

tjpalmer:
This also affects linux and windows. It's currently one of my biggest concerns with Ebitengine, so I'm glad to see it's on the radar!

hajimehoshi:
Apparently `Thread.Call` and `UserInterface.currentMonitor` are more expensive than I expected.

A) takes about 2.5% CPU on my machine

```diff
diff --git a/internal/ui/ui_glfw.go b/internal/ui/ui_glfw.go
index 34d91a911..5ac108d72 100644
--- a/internal/ui/ui_glfw.go
+++ b/internal/ui/ui_glfw.go
[...]

PR #3586, opened 2026-08-29
internal/ui: cache the window focus and visibility

kumagi:
#3318

feature

## What this PR does | solves
The glfw backend queried GetAttrib(Focused) and GetAttrib(Visible) on the main thread every tick. Each is a window-system round trip that contributes to the idle CPU usage of an empty application (#3318).

Cache both states and refresh them from GLFW's focus and iconify callbacks and from the places that show or hide the window themselves, re-querying once a second as a safety net against a missed callback. This mirrors how the current monitor is already cached.


hajimehoshi:
> bug

This is not a bug.

Resolve conflicts btw

kumagi:
updated to `feature`.

hajimehoshi:
Reviewed e07bea64c84e274dc65d603d7aa55ce85701abbb. The callback approach looks reasonable, but I found several issues and further opportunities to remove per-frame queries. This is a source review; GUI behavior has not been runtime-tested.

1. **A successful show/hide call does not guarantee the requested visibility.** `setWindowVisible(false)` unconditionally sets `cachedVisible = false`, but GLFW's `Hide()` returns nil without hiding the window when `w.monitor != nil`. Calling this in GLFW fullscreen therefore leaves the window visible while presentation is suppressed until the cache expires. Query the actual visibility after show/hide, or otherwise preserve GLFW's no-op semantics.

2. **The fallback refresh does not work at TPS zero.** Returning an interval of 1 does not force a refresh: `u.Tick()` remains constant, so `lastQuery > tick-1` stays true after a read or callback. Low achieved TPS also stretches the stated one-second bound. A monotonic elapsed-time deadline would keep the fallback independent of game ticks, including while updates are paused.

3. **The initial seeding calls return the zero-valued cache.** Both timestamps and `u.Tick()` initially equal zero, so at normal TPS the freshness check is `0 > -60`. Neither seeding call queries GLFW. The initially hidden window masks much of this, but the caches should be invalidated before seeding, or seeded with direct queries.

4. **The default per-frame focus/visibility queries remain in `update()`.** The loop guarded by `!u.isRunnableOnUnfocused() && u.bufferOnceSwapped` still queries both attributes on every visible, focused frame, as well as while waiting for focus. The PR removes duplicate queries but does not make almost every tick query-free. `IsFocused()` and `glfwWindow.IsVisible()` also bypass the new helpers. These reads could share the corrected cache, with refresh still working while game ticks are paused.

There are more cacheable properties worth considering:

- **Iconified:** `layoutSizes()` queries it every frame. The new iconify callback already receives the actual boolean, but currently discards it.
- **Window and framebuffer sizes:** both are queried every frame, and on X11 both ultimately call `XGetWindowAttributes`. Cache actual sizes from callbacks, accounting for temporarily disabled callbacks and asynchronous fullscreen transitions.
- **macOS occlusion:** queried every frame; the existing `windowDidChangeOcclusionState:` delegate method could update a cache.
- **macOS native fullscreen:** repeated `StyleMask` reads could use state maintained by the existing enter/exit fullscreen notifications, with transition behavior handled explicitly.

`Maximized` is another candidate, though less relevant to the idle loop. `Decorated`, `Floating`, `Resizable`, and `MousePassthrough` already read GLFW fields, so another cache would not remove OS queries for those.

Authored by Codex on behalf of @hajimehoshi.

hajimehoshi:
I think this requires rebasing. Perhaps, reworking might be better

kumagi:
Thanks for the detailed review. I reworked the caching along the lines you described:

1. `setWindowVisible` no longer caches the requested value. After `Show`/`Hide` it invalidates the cached states so that the next access re-queries the window system, which preserves GLFW's no-op semantics (e.g. `Hide` doing nothing in GLFW fullscreen) instead of suppressing presentation.
2. The safety-net refresh is now a deadline measured on Go's monotonic clock (`time.Time` with `time.Now`) rather than a tick distance, so it keeps working at TPS 0 and while the ticks do not advance, and the bound is a real one second regardless of the achieved TPS.
3. The caches start from the zero deadline, which forces a query on the first access, so the seeding after the callbacks are registered actually queries GLFW now.
4. The focus/visibility reads in `update()`'s wait-for-focus loop, in `updateGame()`, and in `IsFocused()` and `glfwWindow.IsVisible()` all go through the same cached helpers, so a normal focused, visible tick no longer queries the window system at all.

I also cached iconification, since the iconify callback reports the actual boolean and `layoutSizes()` used to query the attribute every frame; the iconify callback now records it and invalidates the visibility cache (GLFW's visible attribute follows iconification on some platforms), and `layoutSizes()`, the size/position callbacks and `IsMinimized()` read the cached value.

The polling loops in `iconifyWindow`, `restoreWindow` and `maximizeWindow` still query GLFW directly, since they need the live value to know when the window system has finished.

The window/framebuffer sizes, the macOS occlusion state, the macOS native-fullscreen style mask and `Maximized` are also cacheable as you listed, but they need care with the temporarily disabled callbacks and the asynchronous fullscreen transitions, so I left them out of this change; I can do them here or in a follow-up, whichever you prefer.

This is on top of the current branch; I'll rebase onto main once the shape looks right to you.

hajimehoshi:
There are still conflicts

kumagi:
Resolved the conflicts against the latest `main`. Upstream had moved the window show/hide handling out of `update()` into `setWindowVisible` and renamed the desktop-window visibility accessor, so the caching was adapted to fit: a programmatic show or hide now invalidates the caches there, and the cached reads go through the updated accessors. The behavior is unchanged — a normal focused, visible tick no longer queries the window system, the safety-net re-query uses a monotonic one-second deadline that works at TPS 0 and while the ticks are paused, and a no-op show/hide (e.g. `Hide` in GLFW fullscreen) is preserved by re-querying instead of caching the requested value. The focus/visibility/iconification caching from the previous round is unchanged and squashed into a single change. `go build` (linux, darwin, windows) and `go test ./internal/ui` pass.

hajimehoshi:
Not resolved yet

hajimehoshi:
I found a potential regression in the position callback at `internal/ui/ui_glfw.go:673`:

**[P2] Keep the direct iconification query in the position callback.**

The Win32 iconify callback runs from `WM_SIZE`, but `WM_MOVE` can arrive first during minimization. With a valid cached `false` value, `isWindowIconified()` accepts the minimized window coordinates and the callback overwrites `windowXInDIP`/`windowYInDIP`, causing `WindowPosition` to report the off-screen minimized position. During restoration, the cached `true` value can likewise discard the restoring position event. The periodic refresh cannot repair the stored coordinates.

Keeping the direct `GetAttrib(glfw.Iconified)` query in this callback would preserve the guard; this event callback is outside the per-tick hot path.

The move-before-size ordering is explicit in [Wine's Win32 implementation](https://github.com/wine-mirror/wine/blob/master/dlls/win32u/defwnd.c#L465-L485). This finding is based on source tracing; I have not reproduced it on native Windows.

Posted by Codex (OpenAI), on behalf of @hajimehoshi.

kumagi:
Fixed. The window position callback queries `glfw.Iconified` from the window system directly again.

I applied the same to the framebuffer size callback: a window state change does not guarantee that the iconify callback is delivered before the other event callbacks of that change (WM_MOVE before WM_SIZE on Windows, and on X11 the position and size come from a ConfigureNotify while the iconification comes from the WM_CHANGE_STATE client message), so neither callback can rely on the cached value. Both are event callbacks outside the per-tick path, so the query costs nothing per frame. The per-tick readers, `layoutSizes` and `IsMinimized`, still use the cache, and the iconify callback still keeps it up to date.

The direct query is factored into `isWindowIconifiedUncached`, which documents the ordering, so the two callbacks cannot silently drift back to the cached helper.

`go build` (linux, darwin, windows, js/wasm), `gofmt`, the vettool check and `go test ./internal/ui` pass. I could not reproduce the move-before-size ordering natively either; since the change only restores the pre-PR direct query in these two callbacks, their behavior is exactly the previous one. The branch is also conflict-free against the current `main`.

internal/ui/ui_glfw.go
```diff
@@ -677,6 +702,39 @@ func (u *glfwBackend) registerWindowPosCallback() error {
...
+		u.iconifyCallback = func(_ *glfw.Window, iconified bool) {
+			u.setCachedIconified(iconified)
+		}
+	}
+	if _, err := u.window.SetIconifyCallback(u.iconifyCallback); err != nil {
+		return err
+	}
+
+	// Seed the caches so that the states are known before the first focus or iconify callback.
+	if _, err := u.isWindowFocused(); err != nil {
```
hajimehoshi: Using a setter method to initialize internal cache sounds odd. Define a separate method to refresh the state and use it.

kumagi: Fixed. The seeding no longer relies on the cached accessors: a dedicated refreshCachedWindowStates method now queries Focused, Visible and Iconified from the window system directly and updates the caches with the actual values and a fresh deadline, and registerWindowFocusAndIconifyCallbacks calls it after registering the callbacks. The callbacks themselves still record the values they are given.

kumagi:
Addressed the review comment on the cache initialization: the caches are now seeded by a dedicated refreshCachedWindowStates method, which queries Focused, Visible and Iconified from the window system directly and records the actual values with a fresh one-second deadline, instead of going through the cached accessors. The focus and iconify callbacks still record their reported values directly. The branch is conflict-free against the current main, and go build (linux, darwin, windows, js/wasm), gofmt, the vettool check and go test ./internal/ui pass.

internal/ui/ui_glfw.go
```diff
@@ -1030,6 +1081,9 @@ func (u *glfwBackend) initOnMainThread(options *RunOptions) error {
 	if err := u.registerWindowCloseCallback(); err != nil {
 		return err
 	}
+	if err := u.registerWindowFocusAndIconifyCallbacks(); err != nil {
```
hajimehoshi: I think we should separate this into two (focus and iconify).

refreshCachedWindowStates can be invoked outside.

kumagi: Done. The registration is split into registerWindowFocusCallback and registerWindowIconifyCallback, each registering only its own callback, and neither touches the caches. initOnMainThread now calls the two registrations and then invokes refreshCachedWindowStates directly, so seeding the caches stays outside the registration.

kumagi:
Addressed the new review comment: the combined registerWindowFocusAndIconifyCallbacks is split into two methods, registerWindowFocusCallback and registerWindowIconifyCallback, each registering only its own callback, matching the pattern of the other per-callback registration methods. refreshCachedWindowStates is no longer invoked from the registration; initOnMainThread now calls the two registrations and then refreshes the caches itself, so the caches are still known before the first focus or iconify callback. go build (linux, darwin, windows, js/wasm), gofmt, the vettool check and go test ./internal/ui pass, and the branch is conflict-free against the current main.

hajimehoshi (approved):
LGTM

Issue #3202, opened 2025-02-26
Icon not being set on Ubuntu via ebiten.SetWindowIcon()


radeeyate:
### Ebitengine Version

v2.8.6

### Operating System

- [ ] Windows
- [ ] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

### Go Version (`go version`)

1.23.4

### What steps will reproduce the problem?

Add the following code to generate an icon and set it:
```go
icon := ebiten.NewImage(32, 32)
icon.Fill(color.RGBA{0x33, 0x33, 0x55, 0xff})
ebiten.SetWindowIcon([]image.Image{icon})
```

### What is the expected result?

The application icon to be set.

### What happens instead?

The application icon just seems to stay as the default icon on GNOME:
![Image](https://github.com/user-attachments/assets/a9d58e8d-d8c7-4487-9352-ccc2d1b6ede5)

### Anything else you feel useful to add?

I'm not sure if this is just specific to Ubuntu or whether it would affect other distros as well. I'm aware that I should in reality be setting the icon with a `.desktop` file for the game - it just seems like it should be working as it does on Windows as I presume.

hajimehoshi:
Did you call ebiten.RunGame?

hajimehoshi:
Could you try

1. Run `go run github.com/hajimehoshi/ebiten/v2/windowsize`
2. Press 'I' some times to see icon changes

?

radeeyate:
Icon changes don't seem to apply using that either.

https://github.com/user-attachments/assets/a1ec6089-f8f1-4a51-9cca-12bbb7620822

hajimehoshi:
What desktop environment are you using?

radeeyate:
GNOME 46 (on Wayland) on Ubuntu 24.04

hajimehoshi:
Thanks, I'll take a look later

hajimehoshi:
> Wayland

I'm afraid Wayland is not supported well so far.

hajimehoshi:
Probably related

* https://github.com/glfw/glfw/issues/2597
* https://github.com/fyne-io/fyne/issues/5983

hajimehoshi:
[...]

PR #3590, opened 2026-08-30
Fix SetWindowIcon on Linux by setting WM_HINTS with icon flags

kumagi:
#3202 

feature

## What this PR does | solves
On X11, platformSetWindowIcon only set the _NET_WM_ICON property. Some window managers and shells (e.g., Ubuntu/GNOME) also inspect the ICCCM WM_HINTS property to determine whether an icon is provided, and without the icon-related flags set in WM_HINTS the window icon was ignored and the default icon remained (issue #3202).

To fix this, after setting _NET_WM_ICON, allocate a WM_HINTS structure with XAllocWMHints, set its flags field to IconPixmapHint (1) | IconMaskHint (2) with null IconPixmap and IconMask values, which signals that the actual icon data is delivered via _NET_WM_ICON, and apply it with XSetWMHints. The allocated structure is then released with XFree. This follows the same WM_HINTS handling pattern already used during window creation.

### DIFF
--- a/internal/glfw/x11_window_linbsd.go
+++ b/internal/glfw/x11_window_linbsd.go
@@ -2054,6 +2054,20 @@ func (w *Window) platformSetWindowIcon(images []*Image) error {
 			_glfw.platformWindow.NET_WM_ICON)
 	}
 
+
+	// Set ICCCM WM_HINTS with icon flags for Ubuntu/GNOME compatibility
+	{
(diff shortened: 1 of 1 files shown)


hajimehoshi:
I'm afraid this PR didn't work. Let me close this as we discussed in another place. I'll take a look at the issue.

Issue #2571, opened 2023-02-12
internal/processtest: sometimes tests time out


hajimehoshi:
### Ebitengine Version

05f86b3974f0df168b7f0c85c3cec9c9f7f7cbb3

### Operating System

- [ ] Windows
- [ ] macOS
- [X] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] Xbox
- [ ] Web Browsers

### Go Version (`go version`)

go 1.18 (but can be reproduced with any version probably)

### What steps will reproduce the problem?

Run tests on GitHub CI.

### What is the expected result?

Tests pass

### What happens instead?

```
panic: test timed out after 10m0s

goroutine 23 [running]:
testing.(*M).startAlarm.func1()
	/opt/hostedtoolcache/go/1.18.10/x64/src/testing/testing.go:2029 +0x8e
created by time.goFunc
	/opt/hostedtoolcache/go/1.18.10/x64/src/time/sleep.go:176 +0x32

goroutine 1 [chan receive]:
testing.(*T).Run(0xc00009aea0, {0x531064?, 0x4be2e5?}, 0x53a828)
	/opt/hostedtoolcache/go/1.18.10/x64/src/testing/testing.go:1487 +0x37a
testing.runTests.func1(0xc000090390?)
	/opt/hostedtoolcache/go/1.18.10/x64/src/testing/testing.go:1839 +0x6e
testing.tRunner(0xc00009aea0, 0xc000086cd8)
	/opt/hostedtoolcache/go/1.18.10/x64/src/testing/testing.go:1439 +0x102
testing.runTests(0xc0000ae140?, {0x60dd60, 0x1, 0x1}, {0x2?, 0x40?, 0x616b40?})
	/opt/hostedtoolcache/go/1.18.10/x64/src/testing/testing.go:1837 +0x457
testing.(*M).Run(0xc0000ae140)
	/opt/hostedtoolcache/go/1.18.10/x64/src/testing/testing.go:1719 +0x5d9
main.main()
	_testmain.go:47 +0x1aa

goroutine 18 [chan receive]:
testing.(*T).Run(0xc00009b040, {0xc0000ac140?, 0x3a?}, 0xc000090540)
	/opt/hostedtoolcache/go/1.18.10/x64/src/testing/testing.go:1487 +0x37a
[...]

PR #3591, opened 2026-08-30
internal/processtest: build test programs in parallel

kumagi:
None

feature

## What this PR does | solves
TestPrograms built and ran each testdata program inside a single subtest while holding a mutex, so each of the 20 programs was built one at a time (~11 s total). Building is CPU/IO bound and independent between the programs, so serializing it only wastes wall time.

Build all the programs in parallel first, then run the subtests one by one as before (#2571).

### DIFF
--- a/internal/processtest/processtest_test.go
+++ b/internal/processtest/processtest_test.go
@@ -60,9 +60,13 @@ func TestPrograms(t *testing.T) {
 
 	tmpdir := t.TempDir()
 
-	// Run sub-tests one by one, not in parallel (#2571).
-	var m sync.Mutex
-
+	type buildResult struct {
+		name string
+		bin  string
+		err  error
+		out  []byte
+	}
+	results := make([]buildResult, 0, len(ents))
 	for _, e := range ents {
 		if e.IsDir() {
 			continue
@@ -71,20 +75,40 @@ func TestPrograms(t *testing.T) {
 		if !strings.HasSuffix(n, ".go") {
 			continue
 		}
+		results = append(results, buildResult{name: n, bin: filepath.Join(tmpdir, n)})
+	}
 
-		t.Run(n, func(t *testing.T) {
+	var wg sync.WaitGroup
+	for i := range results {
+		wg.Add(1)
+		go func(i int) {
+			defer wg.Done()
+			if out, err := exec.Command("go", "build", "-o", results[i].bin, filepath.Join(dir, results[i].name)).CombinedOutput(); err != nil {
+				results[i].err = err
+				results[i].out = out
+			}
+		}(i)
+	}
+	wg.Wait()
+	for _, r := range results {
+		if r.err != nil {
+			t.Fatalf("%v\n%s", r.err, r.out)
+		}
+	}
+
+	// Run sub-tests one by one, not in parallel (#2571).
+	var m sync.Mutex
+
+	for _, r := range results {
+		r := r
+		t.Run(r.name, func(t *testing.T) {
 			m.Lock()
 			defer m.Unlock()
-
-			bin := filepath.Join(tmpdir, n)
-			if out, err := exec.Command("go", "build", "-o", bin, filepath.Join(dir, n)).CombinedOutput(); err != nil {
-				t.Fatalf("%v\n%s", err, string(out))
-			}
 
 			ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 			defer cancel()
 
-			cmd := exec.CommandContext(ctx, bin)
+			cmd := exec.CommandContext(ctx, r.bin)
 			stderr := &bytes.Buffer{}
 			cmd.Stderr = stderr
 			if err := cmd.Run(); err != nil {
--- a/internal/processtest/processtest_test.go
+++ b/internal/processtest/processtest_test.go
@@ -17,10 +17,11 @@ package processtest_test
 import (
 	"bytes"
 	"context"
-	"fmt"
+	"flag"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 	"sync"
@@ -39,6 +40,24 @@ func isWSL() (bool, error) {
 		return false, err
 	}
 	return strings.HasPrefix(abs, `\\wsl$\`), nil
+}
+
+// isSelected reports whether a subtest named name is selected by the -run
+// flag, so that only the programs that will run are built.
+func isSelected(name string) bool {
+	f := flag.Lookup("test.run")
+	if f == nil {
+		return true
+	}
+	parts := strings.Split(f.Value.String(), "/")
+	if len(parts) < 2 || parts[1] == "" {
+		return true
+	}
+	re, err := regexp.Compile(parts[1])
+	if err != nil {
+		return true
+	}
+	return re.MatchString(name)
 }
 
 func TestPrograms(t *testing.T) {
@@ -63,11 +82,11 @@ func TestPrograms(t *testing.T) {
 
 	tmpdir := t.TempDir()
 
-	type buildResult struct {
+	type program struct {
 		name string
 		bin  string
 	}
-	results := make([]buildResult, 0, len(ents))
+	programs := make([]program, 0, len(ents))
 	for _, e := range ents {
 		if e.IsDir() {
 			continue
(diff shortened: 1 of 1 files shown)


internal/processtest/processtest_test.go
```diff
@@ -71,25 +75,45 @@ func TestPrograms(t *testing.T) {
 		if !strings.HasSuffix(n, ".go") {
 			continue
 		}
+		results = append(results, buildResult{name: n, bin: filepath.Join(tmpdir, n)})
+	}
+
+	var wg sync.WaitGroup
+	for i := range results {
+		wg.Add(1)
```
hajimehoshi: wg.Go

kumagi: Done. Switched to errgroup.Group and used wg.Go as suggested - thanks for the review!

hajimehoshi:
Thanks for the PR! The idea is reasonable, but I measured it and there is a case where it makes things considerably worse, so I would like a few things addressed first.

### Cold build cache: this is ~7x slower

Concurrent `go build` processes do not share in-flight compile actions, so each of the 20 invocations independently recompiles the whole Ebitengine dependency tree. Measured on a 12-core machine with `GOCACHE` emptied beforehand:

|                                            | wall                       | CPU              |
| ------------------------------------------ | -------------------------- | ---------------- |
| serial (current `main`)                    | 14.3s                      | 37.8s            |
| this PR (20-wide fan-out)                  | 104.2s / 105.4s (two runs) | 418s / 433s      |
| build 1 first, then fan out the other 19   | 8.8s                       | 49.3s            |

(Warm cache, for reference: 7.2s serial vs 1.7s parallel, so the speedup you saw is real when the cache is already warm.)

A cold cache is not hypothetical: a fresh clone or container, `go clean -cache`, a Go toolchain upgrade, or a different `GOARCH` all hit it. In particular the `go test (Linux 386)` and `go test (Windows 386)` CI steps build for an architecture that the preceding `go build -v ./...` never warms. CI runners also have fewer cores than my machine, so it would likely be worse there.

The third row is the suggestion: build one program serially first to populate the shared dependencies, then fan out the remaining ones, bounded to `runtime.NumCPU()` rather than unbounded. That beats the current serial code even on a cold cache, while keeping the warm-cache win.

### A build failure now aborts every subtest

`t.Fatalf` is now called on the parent `TestPrograms` before the `t.Run` loop, so it calls `runtime.Goexit` and none of the 20 subtests run. Today each build lives in its own subtest, so one program failing to compile fails exactly that subtest and the other 19 still report their status. Could you keep the error in the struct and report it inside the corresponding `t.Run` instead? That preserves per-program attribution and still lets the builds run concurrently.

### gofmt

The file lost its trailing newline, so `gofmt -l .` reports it and the gofmt CI step fails.

### Minor

* `r := r` is not needed: go.mod requires `go 1.25.0`, and loop variables have been per-iteration since Go 1.22. The same applies to passing `i` as a parameter to the goroutine.
* `go test -run 'TestPrograms/issue2079.go'` now builds all 20 programs instead of just the selected one.
* `buildResult` mixes inputs (`name`, `bin`) with results (`err`, `out`).

---

kumagi:
Thanks for the detailed measurements! Pushed 35667826f addressing all the points:

- **Cold cache**: the first program is now built serially to populate the shared dependencies in the build cache, then the remaining ones fan out bounded to `runtime.NumCPU()` via `errgroup.SetLimit`.
- **Per-program error attribution**: build errors/outputs are kept in slices and reported with `t.Fatalf` inside the corresponding `t.Run`, so a failing build only fails that subtest and the other programs still run.
- **gofmt**: fixed (the missing trailing newline was from my edit tool, not intentional).
- **Minor**: dropped `r := r`, split the input struct (`program`) from the result slices (`errs`/`outs`), and added an `isSelected` helper that parses the subtest part of `-run` so `go test -run 'TestPrograms/issue2079.go'` only builds the selected program (flag.Lookup("test.run") - happy to drop this part if it's too hacky).

hajimehoshi:
Thanks, this looks much better. I re-measured 35667826f with the same cold-cache setup (12 cores, `GOCACHE` emptied, 20 programs):

|                                            | wall      | CPU         |
| ------------------------------------------ | --------- | ----------- |
| serial (current `main`)                    | 14.3s     | 37.8s       |
| v1 (20-wide fan-out)                       | 104-105s  | 418s / 433s |
| **v2 (warm-first + `NumCPU` bound)**       | **10.8s** | 58.9s       |

Warm cache is unchanged at ~1.7s vs 7.2s serial, so v2 is now faster than `main` on both a cold and a warm cache. The regression is gone. I also confirmed gofmt is clean, `go vet ./internal/processtest/` passes, and the per-subtest error attribution works. `golang.org/x/sync` is already a direct dependency with five existing users, so `errgroup` adds nothing new. The `for i := 1; i < len(programs); i++` capture in `wg.Go` is fine, since Go 1.22 gave three-clause `:=` loops per-iteration variables as well.

### Please drop `isSelected`: it can silently skip tests

It re-implements a subset of `testing`'s unexported matcher, and it diverges because `splitRegexp` handles a top-level `|` alternation *before* splitting on `/`. Running the helper side by side with the real matcher:

```
$ go test -v -run 'TestPrograms|TestOther/x'
    isSelected("issue2079.go") = false
    isSelected("issue2089.go") = false
    isSelected("shader.go")    = false
=== RUN   TestPrograms/issue2079.go     <- real matcher runs it
=== RUN   TestPrograms/issue2089.go     <- real matcher runs it
=== RUN   TestPrograms/shader.go        <- real matcher runs it
```

Go parses that pattern as two branches, `["TestPrograms"]` and `["TestOther", "x"]`. The first branch matches with no subtest filter, so every subtest runs. The naive `strings.Split(s, "/")` instead yields `["TestPrograms|TestOther", "x"]` and applies `x` as the subtest filter. In this PR that means every program is filtered out, `programs` ends up empty, nothing is built, no `t.Run` is called, and `TestPrograms` passes green having run zero programs.

Malformed patterns fall back to `true`, which is safe, but this case fails silently in the dangerous direction. There is a second latent case too: `t.Run` rewrites spaces in subtest names to `_`, so a testdata file with a space in its name would never match.

For what it is worth, I measured what the helper buys: on a cold cache a single-program `-run` is 4.0s with it and 10.8s without, and about 1.6s warm. That is real but modest, and a test that can silently run nothing is not worth seven seconds. An alternative that keeps the win correctly (prefetch every build in the background and have each subtest wait on its own) is considerably more machinery, so I would just drop the helper along with the `flag` and `regexp` imports.

### Minor

* `wg.Wait()` can never return a non-nil error, since the closure unconditionally returns `nil`. The `t.Fatal(err)` branch is dead, so a plain `wg.Wait()` is enough.
* `var wg errgroup.Group` reads oddly for a `Group`; `eg` or `g` is the more usual name.
* `isSelected`'s doc comment carries caller-side rationale ("so that only the programs that will run are built"), which we would normally keep in the caller. Moot if the helper goes.

With `isSelected` removed I think this is good to merge.

---

kumagi:
Good catch on the `|` alternation case - that silent skip was exactly the kind of failure mode a helper like this shouldn't have. Dropped `isSelected` along with the `flag`/`regexp` imports in 37e7f7394, and applied the minor cleanups: renamed `wg` to `eg` and made the `Wait()` call plain since the closures never return an error. Thanks for the re-measurements!

hajimehoshi:
Thanks, `isSelected` being gone resolves the silent-skip problem, and the rename and the `Wait()` cleanup both look right.

One thing left, and it is my fault: my "a plain `wg.Wait()` is enough" advice was wrong. This repo runs errcheck as part of the `go vet (vettool)` CI step, and it wants the discard to be explicit. Building `./internal/vettools` and running the same command CI runs:

```
$ go vet -vettool=$(which vettools) ./internal/processtest/
internal/processtest/processtest_test.go:107:9: unchecked error
```

`main` is clean under the same tool, so this would newly turn that job red. The fix is one character:

```go
_ = eg.Wait()
```

With that, the vettool reports nothing. Sorry for sending you down that path.

Everything else checks out on 37e7f7394: `gofmt -l` is clean, plain `go vet` is clean, and the build logic is unchanged from the revision I measured at 10.8s on a cold cache (versus 14.3s on `main`), so that result still stands. The concurrency is sound too, since `errs` and `outs` are written at distinct indices and only read after `Wait()`, and the `i` and `p` captures in both loops are per-iteration on Go 1.22 and later.

With `_ = eg.Wait()` this is good to merge from my side.

---

kumagi:
No worries at all - fixed in 4cb8e240c. Reproduced the vettool error locally with `go install ./internal/vettools && go vet -vettool=... ./internal/processtest/`, confirmed `_ = eg.Wait()` makes it clean. Thanks for the thorough review!

hajimehoshi (approved):
LGTM

hajimehoshi:
Unfortunately a flaky test never passed. The flakiness was already fixed in the latest commit. Please rebase the PR

PR #3592, opened 2026-08-30
exp/vmhost: cache guest binaries between tests

kumagi:
None

feature

## What this PR does | solves
buildGuest recompiled the same guest program for every test that used it. Several guests are shared across tests (e.g. ./testdata/atlas is used by three tests), so each of those tests rebuilt an identical binary (~0.5 s each), which is unnecessary work for the investigation.

Cache the built guest binary per pkgPath and activation so each distinct guest is compiled only once per test run.

### DIFF
--- a/exp/vmhost/guest_test.go
+++ b/exp/vmhost/guest_test.go
@@ -15,16 +15,42 @@
 package vmhost_test
 
 import (
+	"fmt"
 	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


hajimehoshi:
Thanks, but I'll pass on this one. The lifetime is the problem: `t.TempDir` removes each guest binary automatically today, while the new shared directory is never removed at all, so every `go test ./exp/vmhost/` abandons close to 200MB of binaries in `TMPDIR`. There is no good place to clean it up either, since `etesting.MainWithRunLoop` calls `os.Exit` and skips defers when a test fails.

The gain does not justify giving that up. Only 3 of the 16 guests are ever built more than once, so a full run gets 7 cache hits out of 23 lookups, worth about two seconds, and a single `go test -run ...` gets none.

PR #3593, opened 2026-08-30
exp/vmhost: return an error on out-of-range draw data

kumagi:
None

bug

## What this PR does | solves
renderOne replayed guest-recorded indices, vertices, and pixel buffers without checking their bounds, so a misbehaving guest could panic the host. Validate them and return an error like the other reference checks.


hajimehoshi:
Panic is fine, since out-of-range indicates there is a mistake in an internal vm protocol, not in a user input.

Issue #3317, opened 2025-09-16
Missed keyboard key presses wasm browser build. You can replicate it in your example: https://ebitengine.org/en/examples/keyboard.html


Palessan:
### Ebitengine Version

v2.8.8

### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [x] Web Browsers

### Go Version (`go version`)

1.24.4

### What steps will reproduce the problem?

Missed keyboard key presses wasm browser build. You can replicate it in your example: https://ebitengine.org/en/examples/keyboard.html

when you press keys fast, some of them are not registered

It started happening with both ebiten.IsKeyPressed and inpututil.IsKeyJustPressed

It started happening recently recently on my hobby game too:
https://palessan.github.io

Some months ago i didn't have this issue. Its only happening on the keyboard, not the joystick.

Tried playing with ebiten.SetTPS() didn't help

### What is the expected result?

For browsers with wasm not to have missed keyboard key presses

### What happens instead?

when you press keys fast, some of them are not registered

Palessan:
Also if you notice the issue doesn't happen with escape and the up down right left arrows.

hajimehoshi:
I failed to reproduce the issue but I know this can happen in theory. This will be taken care of #1704

hajimehoshi:
You can also try [`AppendInputChars`](https://pkg.go.dev/github.com/hajimehoshi/ebiten/v2#AppendInputChars)

hajimehoshi:
> Some months ago i didn't have this issue. Its only happening on the keyboard, not the joystick.

Just in case, does this happen without updating Ebitengine?

Palessan:
I am able to replicate on your original keyboard example that uses the inpututil.AppendPressedKeys()
You need to press for a small amount of time any printable character.

I could not find which update broke this.

I tried changing the keyboard example to AppendInputChars and it fixed the issue.
But you can't use it for non printable character keys such as F1 and escapeEscape. 
So i will prefer to use the fix when its available.

hajimehoshi:
This is similar to #3249, where the mouseup and mousedown events come at the same frame. Let me take a look

hajimehoshi:
#2898 should already fix this issue. Try `v2.9.0-alpha.16`

[...]

PR #3595, opened 2026-08-30
internal/ui: queue browser key events and stamp them when read

kumagi:
#3317

feature

## What this PR does | solves
A browser fires its keyboard and mouse handlers on its own event loop, so an event can arrive at any moment — even after the game loop has already read the input snapshot for the current tick. The handlers stamped such an event with the live tick, which attributed its press and release edges to a tick that had already been sampled, so a key pressed and released between two samples went unnoticed. Pressing keys fast made this frequent on the wasm browser build (#3317).

Queue the key and mouse-button state changes without a timestamp and fold them into the input state with the current input time when the game loop reads it, so an edge is always attributed to the tick that observes it. The GLFW and mobile backends already serialize the input state with a lock; the browser path wrote it unsynchronized, so guard it with an input mutex now, covering the queued events and the buffers the handlers also touch (runes, wheel, lock keys, cursor and touches). Defer the focus-loss release to the game loop as well.

Add a regression test for a press and release delivered within one tick, plus a wasm test for the queue.

### DIFF
--- a/internal/ui/input_js.go
+++ b/internal/ui/input_js.go
@@ -96,38 +96,95 @@ func eventToKeys(e js.Value) (key0, key1 Key) {
 	return -1, -1
 }
 
+// An inputButtonEvent is a key or mouse-button state change reported by a browser event handler,
+// queued without a timestamp.
+//
+// A browser fires its event handlers on its own event loop, so an event can arrive at any moment —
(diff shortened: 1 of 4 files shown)


hajimehoshi:
Wait for the reporter's answer.

PR #3598, opened 2026-08-30
internal/ui: ignore unknown mouse button codes on browsers

kumagi:
None

bug

## What this PR does | solves
codeToMouseButton maps a MouseEvent.button value to an Ebitengine MouseButton. An unknown code (e.g. an auxiliary button or an out-of-spec value) was looked up with a missing-key access, which returned the zero value MouseButton0 (left button). The glfw backend guards the equivalent lookup with an ok check; do the same for js.


hajimehoshi (approved):
LGTM

PR #3599, opened 2026-08-30
internal/gamepad: write the JS gamepad value under the per-gamepad lock

kumagi:
None

bug

## What this PR does | solves
nativeGamepadImpl.value is read by Gamepad.Axis, Button, and similar methods under g.m. The JS backend wrote it under theGamepads.m only, so a user goroutine calling Axis while the frame loop runs update raced on the js.Value. Take g.m on the write side as desktop backends do.

### DIFF
--- a/internal/gamepad/gamepad_js.go
+++ b/internal/gamepad/gamepad_js.go
@@ -94,7 +94,9 @@ func (g *nativeGamepadsImpl) update(gamepads *gamepads) error {
 				mapping: gp.Get("mapping").String(),
 			}
 		}
+		gamepad.m.Lock()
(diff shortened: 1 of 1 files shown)


hajimehoshi:
Apparently there are other race conditions in this package, so I'll take a look at them first

kumagi:
Thanks for taking a look! I'll keep this PR scoped to the JS `value` race then, and leave the other race conditions in the package to your investigation. I'll rebase and update this PR on top of your findings as needed.

hajimehoshi:
046661575551895e1be7f9eb536d667a2838a94e supersedes this PR, so let me close this

PR #3600, opened 2026-08-30
internal/gamepad: reject a negative virtual gamepad ID

kumagi:
None

bug

## What this PR does | solves
setVirtualGamepads indexed the gamepad slice with the user-supplied ID without checking that it is non-negative. A negative ID caused an index out of range panic. The exported Get method already guards against this; apply the same check to the slice indexing in setVirtualGamepads.

### DIFF
--- a/internal/gamepad/gamepad_virtual_test.go
+++ b/internal/gamepad/gamepad_virtual_test.go
@@ -294,3 +294,14 @@ func TestVirtualGamepadCopiesState(t *testing.T) {
 		t.Errorf("StandardAxisValue(LeftStickHorizontal) = %v; want %v (input mutation leaked)", got, want)
 	}
 }
+
+func TestVirtualGamepadNegativeID(t *testing.T) {
+	updateVirtualGamepads(t, []gamepad.VirtualGamepadState{
+		{ID: -1, SDLID: "neg", Name: "Negative"},
+	})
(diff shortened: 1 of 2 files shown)


internal/gamepad/gamepad_virtual_test.go
```diff
@@ -294,3 +294,14 @@ func TestVirtualGamepadCopiesState(t *testing.T) {
 		t.Errorf("StandardAxisValue(LeftStickHorizontal) = %v; want %v (input mutation leaked)", got, want)
 	}
 }
+
+func TestVirtualGamepadNegativeID(t *testing.T) {
+	updateVirtualGamepads(t, []gamepad.VirtualGamepadState{
+		{ID: -1, SDLID: "neg", Name: "Negative"},
```
hajimehoshi: For a better test, we should use a dummy SDLID as a valid GUID (all 0s are fine)

Also use multiple lines

kumagi: Done: the test now uses an all-zero dummy SDLID as a valid GUID, and the snapshot has multiple entries on multiple lines (the negative ID is ignored while the valid ID 0 stays connected). PTAL, thanks!

hajimehoshi (approved):
LGTM

PR #3601, opened 2026-08-30
internal/buffered: do not use the pixel cache for a whole-region read

### LINKED ISSUE #3208: internal/buffered: do not use the pixel cache for a whole-region read
### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

Reading pixels populates the pixel cache in `internal/buffered`, so the pixels are copied twice: from the GPU to the cache, and from the cache to the caller's buffer. The copy in the cache is then kept until the image is drawn to or deallocated, even though the caller already has the same pixels in its own buffer.

A flag could be added to read a whole region at once, and turned on when `ebiten.Image.ReadPixels` is invoked. On that path, the pixels would be read from the GPU directly into the caller's buffer without populating or consulting the cache. `At` would keep using the cache, as repeated one-pixel reads are what the cache is for. The pending writes in the dots buffer and in the cached tiles have to be written back before such a read.

Drawing to a full HD (1920x1080) image and then reading the whole image, measured on an Apple M3 Pro with Metal:

| Reading the whole image | Time |
| --- | --- |
| Into the cache, then into the caller's buffer (current) | 1.48 ms |
| Into the caller's buffer directly | 0.84 ms |

This does not cover the pixels cached by `At`, which are kept for the lifetime of an image that is never drawn to. Removing `Image.At` in v3 would make that cache, and the tiles in it, unnecessary.

This replaces the earlier proposals in this issue: releasing the cached pixels a while after the last access, which needed a central manager for `*buffered.Image` objects, and dividing a cache into tiles regardless of the image size, which made reading a whole full HD image about 20% slower.

### Why is this needed?

The pixel cache keeps a second copy of an image that is read at once, until the image is drawn to or deallocated. This can matter in environments with restricted memory, and reading without the cache is also faster.

kumagi:
#3208 

bug

## What this PR does | solves
ebiten.Image.ReadPixels reads a whole region at once, so copying the pixels into the CPU-side pixel cache is wasteful: the pixels are copied twice (GPU to cache, then cache to the caller's buffer) and the extra copy is kept until the image is drawn to or deallocated.

Add a useCache flag threaded from ebiten.Image.ReadPixels down to internal/buffered. When false, the pixels are read from the GPU directly into the caller's buffer without consulting or populating the cache, after writing back the pending writes in the dots buffer and the cached tiles. At keeps using the cache, as repeated one-pixel reads are what the cache is for.

Fixes #3208

### DIFF
--- a/exp/vmhost/render.go
+++ b/exp/vmhost/render.go
@@ -261,7 +261,7 @@ func (f *frameRenderer) readPixels(pixels [][]byte, id graphicsdriver.ImageID, r
 		return fmt.Errorf("vmhost: ReadPixels references unknown image %d", id)
 	}
 	for i, r := range regions {
-		hi.img.ReadPixels(pixels[i], r)
+		hi.img.ReadPixels(pixels[i], r, true)
 	}
 	return nil
 }
--- a/image.go
+++ b/image.go
@@ -1262,7 +1262,9 @@ func (i *Image) ReadPixels(pixels []byte) {
 
 	i.invokeUsageCallbacks()
 
-	i.image.ReadPixels(pixels, i.adjustedBounds())
+	// ReadPixels reads a whole region at once, so the pixel cache is not used to avoid keeping a
+	// second copy of the pixels (see #3208).
+	i.image.ReadPixels(pixels, i.adjustedBounds(), false)
 }
 
 // At returns the color of the image at (x, y).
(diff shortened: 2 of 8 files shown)


hajimehoshi:
We need more discussion

PR #3602, opened 2026-08-30
cmd/ebitenmobile: don't use GLSurfaceView on Android

### LINKED ISSUE #3371: cmd/ebitenmobile: avoid using GLSurfaceView
### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [x] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

A context is lost due to `GLSurfaceView`, rather than OpenGL ES itself. In theory, it is possible to avoid `GLSurfaceView` but keep using OpenGL ES. In this case, we have to create a class inheriting `SurfaceView` and implement bunch of code which `GLSurfaceView` did.

See also: 

* https://developer.android.com/ndk/samples/sample_na
* https://github.com/google/grafika/blob/master/app/src/main/java/com/android/grafika/TextureViewCanvasActivity.java

### Why is this needed?

We'd like to avoid context lost on Android apps.

kumagi:
#3371

feature

## What this PR does | solves
An Ebitengine app on Android lost its OpenGL ES context when GLSurfaceView decided so, although a context does not have to be lost where only the surface is recreated: GLSurfaceView recreates the context when it is not preserved by the platform, and Ebitengine could restore the GPU objects only when the game had asked to save them, killing the app otherwise.

EbitenSurfaceView now inherits SurfaceView and manages EGL on its own render thread, which is shared by all the EbitenSurfaceView instances. The EGL display and context are created once and are kept while the process lives, only the EGL surface is recreated when the Android surface is recreated, and the buffers are swapped after Ebitenmobileview.update as GLSurfaceView did. The render thread renders continuously or only when a frame is requested, which corresponds to the FPS modes, suspends rendering while the game is suspended or the window is invisible, and stops using the Android surface before surfaceDestroyed returns. An actual context lost, which eglSwapBuffers reports, recreates the context and restores GPU resources as before.

Closes #3371

### DIFF
--- a/cmd/ebitenmobile/_files/EbitenSurfaceView.java
+++ b/cmd/ebitenmobile/_files/EbitenSurfaceView.java
@@ -15,104 +15,43 @@
 package $Placeholder_JavaPkg$.$Placeholder_PrefixLower$;
 
 import android.content.Context;
-import android.opengl.GLSurfaceView;
+import android.graphics.Rect;
+import android.opengl.EGL14;
+import android.opengl.EGLConfig;
+import android.opengl.EGLContext;
+import android.opengl.EGLDisplay;
+import android.opengl.EGLSurface;
 import android.os.Handler;
 import android.os.Looper;
 import android.util.AttributeSet;
 import android.util.Log;
-
-import javax.microedition.khronos.egl.EGL10;
-import javax.microedition.khronos.egl.EGLConfig;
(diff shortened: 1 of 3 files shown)


hajimehoshi:
We need more discussion

PR #3603, opened 2026-08-30
vector: clamp the Acos arguments in ArcTo and LineJoinMiter

kumagi:
None

bug

## What this PR does | solves
ArcTo and the miter join compute angles with math.Acos on float32 dot products of normalized vectors. The dot product can slightly exceed 1.0 due to rounding, making Acos return NaN.

In ArcTo, the NaN propagates into the arcs center and the path silently disappears. In LineJoinMiter, the NaN comparison always reports that the miter limit is not exceeded, so a miter spike can be huge.

Clamp the dot products to [-1, 1] before Acos. Includes the fix from #3604 as requested.

### DIFF
--- a/vector/path.go
+++ b/vector/path.go
@@ -483,8 +483,8 @@ func (p *Path) ArcTo(x1, y1, x2, y2, radius float32) {
 	d0 = d0.norm()
 	d1 = d1.norm()
 
-	// theta is the angle between two vectors d0 and d1.
-	theta := math.Acos(float64(d0.x*d1.x + d0.y*d1.y))
+	dot := min(max(float64(d0.x*d1.x+d0.y*d1.y), -1.0), 1.0)
+	theta := math.Acos(dot)
 	// TODO: When theta is bigger than π/2, the arc should be split into two.
 	if theta == 0 {
 		p.LineTo(x2, y2)
--- a/vector/path.go
+++ b/vector/path.go
@@ -484,6 +484,7 @@ func (p *Path) ArcTo(x1, y1, x2, y2, radius float32) {
 	d1 = d1.norm()
 
 	dot := min(max(float64(d0.x*d1.x+d0.y*d1.y), -1.0), 1.0)
+	// theta is the angle between two vectors d0 and d1.
 	theta := math.Acos(dot)
 	// TODO: When theta is bigger than π/2, the arc should be split into two.
 	if theta == 0 {
--- a/vector/path_test.go
+++ b/vector/path_test.go
@@ -501,6 +501,14 @@ func TestBounds(t *testing.T) {
 			},
 			want: image.Rect(10, 20, 30, 40),
 		},
+		{
+			name: "shallow angle arc",
+			path: func(p *vector.Path) {
+				p.MoveTo(676.47327, 1502.2303)
+				p.ArcTo(257.7812, 1856.779, 1046.7478, 1188.3281, 100)
+			},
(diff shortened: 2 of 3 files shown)


hajimehoshi:
Add a test to reproduce the case

hajimehoshi:
Let's include the fix #3604

kumagi:
Thanks for the review!

- **Test to reproduce the case**: 44527b0 adds the "shallow angle arc" case to TestBounds with the exact failing input. I confirmed it fails without the clamp (`got: (0,0)-(0,0)`, the path disappears) and passes with it.
- **Include the fix from #3604**: cherry-picked as ee63e63 (with its reproducing test, which I also confirmed fails without the miter clamp: the spike reaches y=3979 instead of staying within the path bounds). Updated the PR title/description accordingly, and closed #3604 as superseded.

vector/path.go
```diff
@@ -483,8 +483,8 @@ func (p *Path) ArcTo(x1, y1, x2, y2, radius float32) {
 	d0 = d0.norm()
 	d1 = d1.norm()
 
-	// theta is the angle between two vectors d0 and d1.
```
hajimehoshi, line 486: Do not remove this comment

vector/path_test.go
```diff
@@ -501,6 +501,14 @@ func TestBounds(t *testing.T) {
 			},
 			want: image.Rect(10, 20, 30, 40),
 		},
+		{
+			name: "shallow angle arc",
```
hajimehoshi: `"nearly collinear"` would be better

vector/path_test.go
```diff
@@ -513,3 +521,26 @@ func TestBounds(t *testing.T) {
...
+	var sp vector.Path
+	sp.AddStroke(&p, op)
+
+	got := sp.Bounds()
+	if got.Min.X < 0 || got.Min.Y < 0 {
+		t.Errorf("Bounds().Min = %v; want non-negative (a miter spike must not extend far beyond the path)", got.Min)
+	}
+	if got, want := got.Max, (image.Point{X: 1048, Y: 1858}); got.X > want.X || got.Y > want.Y {
+		t.Errorf("Bounds().Max = %v; want within %v (a miter spike must not extend far beyond the path)", got, want)
+	}
```
hajimehoshi, line 542: These bounds are exactly `(257,1187)-(1048,1858)`, so both checks are looser than they look: the lower bound of 0 is arbitrary (a spike that reached only `x=+100` would still pass), and the upper bound happens to sit exactly on the real value. `TestBounds` just above asserts exact rectangles, which is both shorter and stricter here:

```go
if got, want := sp.Bounds(), image.Rect(257, 1187, 1048, 1858); got != want {
	t.Errorf("got: %v, want: %v", got, want)
}
```

vector/path.go
kumagi: Restored the comment in 00de3ac77. Sorry about that!

vector/path_test.go
kumagi: Renamed the case (and the stroke test for consistency) to "nearly collinear" in 00de3ac77.

vector/path_test.go
kumagi: Good point. Replaced the loose checks with the exact bounds assertion in 00de3ac77. The test passes with image.Rect(257, 1187, 1048, 1858).

hajimehoshi (approved):
LGTM

PR #3604, opened 2026-08-30
vector: clamp the Acos argument in LineJoinMiter

kumagi:
None

bug

## What this PR does | solves
The miter join computes the joint angle with math.Acos on a float32 dot product. For a nearly straight joint the dot product can exceed 1.0 due to rounding, making Acos return NaN. The NaN comparison then always reports that the miter limit is not exceeded, so the miter point is computed for nearly parallel lines and can be a huge spike. Clamp the dot product to [-1, 1] before Acos.

### DIFF
--- a/vector/path_test.go
+++ b/vector/path_test.go
@@ -513,3 +513,26 @@ func TestBounds(t *testing.T) {
 		})
 	}
 }
+
+func TestStrokeMiterJoinShallowAngle(t *testing.T) {
+	var p vector.Path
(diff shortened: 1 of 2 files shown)


hajimehoshi:
What about doing this in #3603 at the same time?

kumagi:
Superseded by #3603 as requested in https://github.com/hajimehoshi/ebiten/pull/3603#issuecomment-5470585303 (the fix is cherry-picked there).

PR #3607, opened 2026-08-31
text/v2: fix AdvanceX loss when keepGlyph skips a glyph in GoXFace

kumagi:
None

bug

## What this PR does | solves
appendLazyGlyphsForLine for GoXFace computed AdvanceX as the advance of the current rune relative to the previous rune. When keepGlyph skipped a glyph, the previous output glyph's AdvanceX was still just the current rune's advance, so the skipped glyph's advance was lost. The chain OriginX + AdvanceX = next OriginX was broken whenever a glyph was skipped, violating the LazyGlyph contract.

Track the last output glyph's origin and fill its AdvanceX when the next output glyph is appended, so that skipped glyphs' advances are accounted for. At the end of the line, fill the last output glyph's AdvanceX up to the line's total advance.

### DIFF
--- a/text/v2/gox.go
+++ b/text/v2/gox.go
@@ -201,11 +201,8 @@ func (g *GoXFace) appendLazyGlyphsForLine(glyphs []LazyGlyph, line string, index
 	// imager is allocated lazily on the first glyph that produces an image.
 	var imager *goXLineImager
 
-	// lastOutputIndex is the index of the last appended glyph in glyphs, and
-	// lastOutputOriginX is its origin.X. AdvanceX of a glyph is filled in when
-	// the next output glyph is appended so that skipped glyphs' advances are
-	// not lost: OriginX + AdvanceX of a glyph must equal OriginX of the next
-	// output glyph.
+	// The previous output glyph's AdvanceX is filled when the next glyph is
+	// appended, so that skipped glyphs' advances are not lost.
 	lastOutputIndex := -1
 	var lastOutputOriginX fixed.Int26_6
 
--- a/text/v2/gox.go
+++ b/text/v2/gox.go
@@ -201,11 +201,6 @@ func (g *GoXFace) appendLazyGlyphsForLine(glyphs []LazyGlyph, line string, index
 	// imager is allocated lazily on the first glyph that produces an image.
 	var imager *goXLineImager
 
-	// The previous output glyph's AdvanceX is filled when the next glyph is
-	// appended, so that skipped glyphs' advances are not lost.
-	lastOutputIndex := -1
-	var lastOutputOriginX fixed.Int26_6
-
 	for i, r := range line {
 		if i > 0 {
 			origin.X = ox + originXs[advanceIndex]
@@ -214,11 +209,6 @@ func (g *GoXFace) appendLazyGlyphsForLine(glyphs []LazyGlyph, line string, index
 
 		if keepGlyph != nil && !keepGlyph(fixed26_6ToFloat64(origin.X), fixed26_6ToFloat64(origin.Y)) {
 			continue
-		}
-
-		// Fill the previous output glyph's advance up to this glyph's origin.
-		if lastOutputIndex >= 0 {
-			glyphs[lastOutputIndex].AdvanceX = fixed26_6ToFloat64(origin.X - lastOutputOriginX)
 		}
 
 		// The image position is integer so that the nearest filter can be used.
@@ -234,6 +224,12 @@ func (g *GoXFace) appendLazyGlyphsForLine(glyphs []LazyGlyph, line string, index
 			size = 1
 		}
 
+		var prevOriginX fixed.Int26_6
+		if advanceIndex > 0 {
+			prevOriginX = originXs[advanceIndex-1]
+		}
+		advanceX := fixed26_6ToFloat64(originXs[advanceIndex] - prevOriginX)
+
 		lg := LazyGlyph{
 			StartIndexInBytes: indexOffset + i,
 			EndIndexInBytes:   indexOffset + i + size,
@@ -242,6 +238,8 @@ func (g *GoXFace) appendLazyGlyphsForLine(glyphs []LazyGlyph, line string, index
 			OriginY:           fixed26_6ToFloat64(origin.Y),
 			OriginOffsetX:     0,
 			OriginOffsetY:     0,
+			AdvanceX:          advanceX,
+			AdvanceY:          0,
 		}
 		if hasImage {
 			if imager == nil {
@@ -254,13 +252,6 @@ func (g *GoXFace) appendLazyGlyphsForLine(glyphs []LazyGlyph, line string, index
 		// Append a glyph even if it has no image (control characters etc.).
 		// This is necessary to return index information for control characters.
 		glyphs = append(glyphs, lg)
-		lastOutputIndex = len(glyphs) - 1
-		lastOutputOriginX = origin.X
-	}
-
-	// Fill the last output glyph's advance up to the end of the line.
-	if lastOutputIndex >= 0 {
-		glyphs[lastOutputIndex].AdvanceX = fixed26_6ToFloat64(ox + originXs[len(originXs)-1] - lastOutputOriginX)
 	}
 
 	return glyphs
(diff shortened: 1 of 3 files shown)


text/v2/text_test.go
```diff
@@ -1438,3 +1438,38 @@ func TestGoTextFaceSourceConcurrentDrawWithDifferentSizes(t *testing.T) {
...
+	const eps = 1.0 / (1 << 6)
+
+	face := text.NewGoXFace(&testGoXFace{})
+
+	// Keep the first and last glyphs but skip the middle one. The skipped
+	// glyph's advance must still be reflected in the previous output glyph's
+	// AdvanceX so that OriginX + AdvanceX of one output glyph equals OriginX
+	// of the next, and the last output glyph reaches the line's total advance.
+	// testGoXFace gives every rune a fixed advance of testGoXFaceSize, so the
+	// glyph origins are 0, testGoXFaceSize, and 2*testGoXFaceSize.
```
hajimehoshi: Shorten comments

text/v2/gox.go
```diff
@@ -201,6 +201,14 @@ func (g *GoXFace) appendLazyGlyphsForLine(glyphs []LazyGlyph, line string, index
 	// imager is allocated lazily on the first glyph that produces an image.
 	var imager *goXLineImager
 
+	// lastOutputIndex is the index of the last appended glyph in glyphs, and
+	// lastOutputOriginX is its origin.X. AdvanceX of a glyph is filled in when
+	// the next output glyph is appended so that skipped glyphs' advances are
+	// not lost: OriginX + AdvanceX of a glyph must equal OriginX of the next
+	// output glyph.
```
hajimehoshi: Shorten comments

text/v2/text_test.go
```diff
@@ -1438,3 +1438,38 @@ func TestGoTextFaceSourceConcurrentDrawWithDifferentSizes(t *testing.T) {
...
+	line := "acd"
+	glyphs := text.AppendLazyGlyphsForLine(face, nil, line, 0, 0, 0, func(originX, originY float64) bool {
+		return originX < testGoXFaceSize/2 || originX > 3*testGoXFaceSize/2
+	})
+	if len(glyphs) != 2 {
+		t.Fatalf("got %d glyphs, want 2 (a and d)", len(glyphs))
+	}
+
+	for i := 0; i < len(glyphs)-1; i++ {
```
hajimehoshi: Use for-range

text/v2/text_test.go
kumagi: Done.

text/v2/gox.go
kumagi: Done.

text/v2/text_test.go
kumagi: Done.

hajimehoshi:
Thanks for the patch, and sorry for the long comment. I applied the head commit onto `main` and ran `go test ./text/v2/...` (plus `gofmt`/`go vet`) — everything passes, and I couldn't find a defect in the new code itself. My hesitation is about whether the change should go in as it stands.

## The condition being fixed is not reachable from the public API

`keepGlyph` is non-nil from exactly one caller: `Draw`, via `keepGlyphFilter` (`layout.go:158`). `Draw` reads only `ImageBounds`, `Image()` and `Colored()` — never `AdvanceX`. The two public entry points, `AppendGlyphs` and `AppendLazyGlyphs`, both pass `nil`.

With `keepGlyph == nil` the existing code is already correct: `advanceIndex` is incremented before the `continue`, so `originXs[k] - originXs[k-1]` is the glyph's own advance whether or not anything was skipped. I checked that the new code produces the identical value in that case (`ox` cancels exactly in fixed-point), which matches the test results — nothing observable changes today.

## It makes GoXFace and GoTextFace disagree

`GoTextFace` always sets `AdvanceX` to the glyph's own `shapingGlyph.Advance` (`gotext.go:314-319`), so it breaks the same chain under the same culling. Skipping the middle glyph of `"acd"` at size 16:

```
GoTextFace glyph 0: OriginX=0          AdvanceX=8.765625  sum=8.765625
GoTextFace glyph 1: OriginX=17.109375  AdvanceX=9.25      sum=26.359375
```

After this PR the two faces would mean different things by `AdvanceX` under culling, and `MultiFace` mixes both in a single glyph slice. Whichever semantics is right, both faces should agree.

## The new value is no longer "this glyph's advance"

`LazyGlyph.AdvanceX` is documented as "the X advance applied after rendering this glyph" (`text.go:251`). Under culling the new value is the gap to the next *kept* glyph, and for the last kept glyph the distance to the end of the line — so trailing culled glyphs are all absorbed into one inflated advance. That is a different quantity.

The PR description says the current behavior violates "the LazyGlyph contract", but `LazyGlyph.AdvanceX` does not document a chain property. Only `Glyph.AdvanceX` does (`text.go:194-195`), and `Glyph` is only ever produced with `keepGlyph == nil`.

## Smaller points

- `export_test.go`: this exports an internal `Face` method solely to test behavior no public API can reach. Also, the other helpers in that file don't carry a "is a test helper that calls X" comment.
- The test line `"acd"` avoids `testGoXFace.Kern`, which only fires for `r1 == 'b'`. Skipping a *kerned* glyph is the case where "origin difference" and "sum of advances" actually diverge, so a case built around `'b'` would exercise more.
- In the `LazyGlyph` literal, `OriginOffsetX: 0, OriginOffsetY: 0` are still spelled out while `AdvanceX`/`AdvanceY` are now omitted and assigned in two other places. That is harder to follow than the single-site assignment.

## Where this leaves us

If you'd like to take this further, I think the shape would be: decide what `AdvanceX` means when glyphs are culled, apply that consistently to `GoTextFace` as well, and document it on `LazyGlyph.AdvanceX`. As a `GoXFace`-only change to an unreachable path, I'd rather not merge it.

---

kumagi:
Thank you for the thorough review — all points confirmed on my side too.

You're right that `keepGlyph` is non-nil only via `Draw`/`keepGlyphFilter`, which never reads `AdvanceX`; `AppendGlyphs` and `AppendLazyGlyphs` pass `nil`, and with `keepGlyph == nil` the previous code already gave each glyph its own advance. You're also right that `LazyGlyph.AdvanceX` doesn't document the `OriginX + AdvanceX` chain property (that's on `Glyph.AdvanceX`, which is only built with `keepGlyph == nil`), and that filling forward to the next *kept* glyph would make `GoXFace` disagree with `GoTextFace` (which always sets `AdvanceX` to the glyph's own shaping advance) inside `MultiFace`.

Since the fixed condition is unreachable from the public API and the semantics of `AdvanceX` under culling are undefined for now, I've reverted the GoXFace change (including the exported test helper and the test) in the head commit — text/v2 is back to the pre-PR state.

If culled-glyph advance semantics are worth defining later (consistently for both faces and documented on `LazyGlyph.AdvanceX`), I'm happy to open a fresh PR for that discussion.

hajimehoshi:
Thanks — but I think the revert didn't get pushed. The PR head is still `bdf9eea0d`, authored at 07:40:49Z, i.e. before my review, with 2 commits and 3 changed files (+52/−8). Nothing has landed on `kumagi/ebiten:fix-text-gox-keepglyph-advancex` since.

The head tree still carries all three pieces:

- `text/v2/gox.go` — `lastOutputIndex` / `lastOutputOriginX`, the fill-forward assignment, and the end-of-line fill
- `text/v2/export_test.go` — the `AppendLazyGlyphsForLine` helper
- `text/v2/text_test.go` — `TestGoXFaceKeepGlyphAdvanceX`

So `text/v2` is not back to its pre-PR state yet.

If the intent is to drop the change entirely, reverting would leave this PR with an empty diff — closing it is simpler. And thank you for digging into this; if the semantics of `AdvanceX` under culling get defined later, a fresh PR covering both faces would be very welcome.

---

hajimehoshi:
Thanks — the revert is on the branch now (`df93f7bbb`), and I've confirmed the tree is byte-identical to the fork point, so `text/v2` is exactly back to its pre-PR state.

That leaves this PR with an empty diff, so I'm closing it. No wasted effort from my side: the analysis in this thread is a useful record of why `AdvanceX` under culling is currently undefined, and I appreciate you chasing it down rather than defending the patch.

If you'd like to pick up the follow-up — deciding what `AdvanceX` means when glyphs are culled, applying it consistently to both `GoXFace` and `GoTextFace`, and documenting it on `LazyGlyph.AdvanceX` — a fresh PR for that would be very welcome.

---

PR #3608, opened 2026-08-31
all: fix doc comments that diverge from the behavior

kumagi:
None

bug

## What this PR does | solves
Several doc comments promised panics or behaviors the code does not have, or described the opposite of what the code does:

  * WindowPosition, SetScreenTransparent, SetInitFocused, and RunOnMainThread documented panics or blocking that do not happen; document what actually happens instead.
  * NewImage and NewImageWithOptions panic for sizes over MaxImageSize only when the image is first used, not at creation.
  * DrawTrianglesShader and DrawTrianglesShader32 documented a nonexistent (width, height); the check compares against the image at index 0, and fix the panic message likewise.
  * The mouse button comments in internal/ui swapped the right and the middle buttons, and the monitor content-scale retry comment stated the loop condition backwards.
  * thread.Loop referenced a Stop method that does not exist.
  * The generated shader bridge documented destination quantities as the source texture unit, and the atlas counter as never reset.
  * vector.ArcTo described its tangent corner and end point as two control points.
  * colorm.ColorM's zero-value note was detached from the doc comment by a blank line.
  * text/v2: LimitedFace renders U+FFFD rather than nothing; CacheGlyphs does treat multiple lines; MultiFace's known issue about disagreeing directions is now rejected by NewMultiFace.
  * ebitenutil: DebugPrint and DebugPrintAt draw with a one-pixel margin on the left, and NewImageFromFileSystem named the wrong function in its doc.
  * ebitenmobile: EBITENMOBILE_GOMOBILE selects a local checkout, not a version.

### DIFF
--- a/ebitenutil/debugprint.go
+++ b/ebitenutil/debugprint.go
@@ -41,14 +41,16 @@ func init() {
 	debugPrintTextImage = ebiten.NewImageFromImage(img)
 }
 
-// DebugPrint draws the string str on the image at (0, 0) position (the upper-left corner in most cases).
+// DebugPrint draws the string str on the image at (1, 0) position (the upper-left corner in most cases).
+// The string is drawn with a one-pixel margin on the left.
 //
 // The available runes are in U+0000 to U+00FF, which is C0 Controls and Basic Latin and C1 Controls and Latin-1 Supplement.
 func DebugPrint(image *ebiten.Image, str string) {
 	DebugPrintAt(image, str, 0, 0)
 }
 
-// DebugPrintAt draws the string str on the image at (x, y) position.
+// DebugPrintAt draws the string str on the image at (x+1, y) position.
+// The string is drawn with a one-pixel margin on the left.
 //
 // The available runes are in U+0000 to U+00FF, which is C0 Controls and Basic Latin and C1 Controls and Latin-1 Supplement.
 func DebugPrintAt(image *ebiten.Image, str string, x, y int) {
--- a/ebitenutil/debugprint.go
+++ b/ebitenutil/debugprint.go
@@ -41,16 +41,14 @@ func init() {
 	debugPrintTextImage = ebiten.NewImageFromImage(img)
 }
 
-// DebugPrint draws the string str on the image at (1, 0) position (the upper-left corner in most cases).
-// The string is drawn with a one-pixel margin on the left.
+// DebugPrint draws the string str on the image at (0, 0) position (the upper-left corner in most cases).
 //
 // The available runes are in U+0000 to U+00FF, which is C0 Controls and Basic Latin and C1 Controls and Latin-1 Supplement.
 func DebugPrint(image *ebiten.Image, str string) {
 	DebugPrintAt(image, str, 0, 0)
 }
 
-// DebugPrintAt draws the string str on the image at (x+1, y) position.
-// The string is drawn with a one-pixel margin on the left.
+// DebugPrintAt draws the string str on the image at (x, y) position.
 //
 // The available runes are in U+0000 to U+00FF, which is C0 Controls and Basic Latin and C1 Controls and Latin-1 Supplement.
 func DebugPrintAt(image *ebiten.Image, str string, x, y int) {
--- a/text/v2/multi.go
+++ b/text/v2/multi.go
@@ -28,7 +28,7 @@ var _ Face = (*MultiFace)(nil)
 // The face in the first index is used in the highest priority, and the last the lowest priority.
 //
 // There is a known issue: if the writing directions of the faces don't agree, the rendering result might be messed up.
-// NewMultiFace rejects such faces, but a face's direction can be changed after the creation (e.g. [GoTextFace.Direction]).
+// [NewMultiFace] rejects such faces, but a face's direction can be changed after the creation (e.g. [GoTextFace.Direction]).
 type MultiFace struct {
 	faces []Face
 
--- a/text/v2/multi.go
+++ b/text/v2/multi.go
@@ -28,9 +28,7 @@ var _ Face = (*MultiFace)(nil)
 // The face in the first index is used in the highest priority, and the last the lowest priority.
 //
 // There is a known issue: if the writing directions of the faces don't agree, the rendering result might be messed up.
-// NewMultiFace rejects the faces whose directions don't agree at its creation,
-// but a face's direction can be changed after the creation (e.g., [GoTextFace.Direction]),
-// and then the rendering result might still be messed up.
+// NewMultiFace rejects such faces, but a face's direction can be changed after the creation (e.g. [GoTextFace.Direction]).
 type MultiFace struct {
 	faces []Face
 
--- a/text/v2/multi.go
+++ b/text/v2/multi.go
@@ -26,6 +26,11 @@ var _ Face = (*MultiFace)(nil)
 
 // MultiFace is a Face that consists of multiple Face objects.
 // The face in the first index is used in the highest priority, and the last the lowest priority.
+//
+// There is a known issue: if the writing directions of the faces don't agree, the rendering result might be messed up.
+// NewMultiFace rejects the faces whose directions don't agree at its creation,
+// but a face's direction can be changed after the creation (e.g., [GoTextFace.Direction]),
+// and then the rendering result might still be messed up.
 type MultiFace struct {
 	faces []Face
 
--- a/text/v2/multi.go
+++ b/text/v2/multi.go
@@ -26,8 +26,6 @@ var _ Face = (*MultiFace)(nil)
 
 // MultiFace is a Face that consists of multiple Face objects.
 // The face in the first index is used in the highest priority, and the last the lowest priority.
-//
-// There is a known issue: if the writing directions of the faces don't agree, the rendering result might be messed up.
 type MultiFace struct {
 	faces []Face
 
--- a/window.go
+++ b/window.go
@@ -162,7 +162,7 @@ func SetWindowIcon(iconImages []image.Image) {
 // The origin position is the upper-left corner of the current monitor.
 // The unit is device-independent pixels.
 //
-// If the main loop does not start yet, WindowPosition returns the initial window position set by SetWindowPosition.
+// If the main loop does not start yet, WindowPosition returns the initial window position set by [SetWindowPosition].
 //
 // WindowPosition returns the original window position in fullscreen mode.
 //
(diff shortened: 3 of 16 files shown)


ebitenutil/debugprint.go
```diff
@@ -41,14 +41,16 @@ func init() {
 	debugPrintTextImage = ebiten.NewImageFromImage(img)
 }
 
-// DebugPrint draws the string str on the image at (0, 0) position (the upper-left corner in most cases).
+// DebugPrint draws the string str on the image at (1, 0) position (the upper-left corner in most cases).
```
hajimehoshi: I think this is an internal detail, and users don't have to care about the internal offset which is for cosmetic.

ebitenutil/debugprint.go
```diff
@@ -41,14 +41,16 @@ func init() {
...
+// The string is drawn with a one-pixel margin on the left.
 //
 // The available runes are in U+0000 to U+00FF, which is C0 Controls and Basic Latin and C1 Controls and Latin-1 Supplement.
 func DebugPrint(image *ebiten.Image, str string) {
 	DebugPrintAt(image, str, 0, 0)
 }
 
-// DebugPrintAt draws the string str on the image at (x, y) position.
+// DebugPrintAt draws the string str on the image at (x+1, y) position.
```
hajimehoshi: ditto

hajimehoshi (commented):
Thanks for going through these. Most of the changes are correct, and a few fix real mistakes: the swapped mouse button comments and the content scale retry loop in particular were plainly wrong.

I checked each claim against the code. Below are the ones I would like changed.

### 1. `vector.ArcTo`: the new description is incorrect

`(x2, y2)` is not the end point of the arc. It only gives the direction of the second tangent line. The arc ends at the tangent point, which is at distance `radius / tan(θ/2)` from `(x1, y1)` along the ray toward `(x2, y2)`.

For example, with

```go
p.MoveTo(0, 0)
p.ArcTo(100, 0, 100, 100, 10)
```

the arc ends at `(100, 10)`, not at `(100, 100)`.

The tangency sentence is fine. The second sentence needs rewording.

### 2. `ebitenutil.DebugPrint` and `DebugPrintAt`

See my inline comments. The one-pixel offset is an internal cosmetic detail. Please drop this hunk.

### 3. `text.MultiFace`: the known issue is still reachable

`NewMultiFace` checks the directions only at construction time, but `GoTextFace.Direction` is an exported field that can be changed afterwards. The situation the note describes can still happen, so I would rather reword the note than remove it.

### 4. `SetScreenTransparent`: "has no effect" is not accurate

The value is still stored and is read back by `IsScreenTransparent`. Calling `SetScreenTransparent` after the main loop starts makes `IsScreenTransparent` report a transparency that the window does not actually have.

`SetInitFocused`'s "has no effect" is accurate.

### 5. `RunOnMainThread`: this documents an accidental nil dereference

Before `RunGame`, the internal backend is nil, and calling a method on it fails with `runtime error: invalid memory address or nil pointer dereference`. That is a bug rather than a contract, and I do not want to write it into the godoc. Please drop this sentence from this PR; the code should be fixed instead.

### 6. `NewImageFromImage` and `NewImageFromImageWithOptions` have the same sentence

They still say "less than 1 or more than [MaxImageSize] ... panics". Leaving them makes the four constructors inconsistent with each other. Could you update them too?

### 7. `WindowPosition`: please do not name `math.MinInt`

That value is an internal sentinel, and documenting it makes it part of the API. It also reads oddly next to the existing "WindowPosition returns (0, 0) if the platform is not a desktop." The first clause alone is enough.

### 8. `DrawTrianglesShader` and `DrawTrianglesShader32`

The new sentence is right for the usual case, but when `Images[0]` is nil, the size to compare against stays the zero value, and any non-nil image at index 1 or later panics. Also the line is about 155 characters; could you wrap it?

### 9. `internal/atlas`: minor

"or when its backend is gone" is the same case as deallocation. Also, the original "never reset" was meant as a contrast with `usedAsSourceCount`, which is reset during normal use. Something like "usedAsDestinationCount is not reset while the image is alive" would keep that.

---

The rest looks good: the mouse buttons, the content scale retry loop, `thread.Loop`, the destination unit comments in the shader bridge, `EBITENMOBILE_GOMOBILE`, `colorm.ColorM`, `NewImageFromFileSystem`, the `false.//` typo, the panic message (and it is right that `DrawRectShader`'s copy stays as it is), `NewImage` and `NewImageWithOptions`, `LimitedFace`, and `CacheGlyphs`.

---

ebitenutil/debugprint.go
kumagi: Done. I reverted this hunk in the follow-up commit: the one-pixel offset is dropped from the doc.

ebitenutil/debugprint.go
kumagi: Done. Reverted this hunk as well.

kumagi:
Thank you for the thorough review! I addressed all the points in the follow-up commit:

1. **vector.ArcTo**: Reworded. (x2, y2) now says it gives the direction of the second tangent line, not the end point, and the doc describes that the arc ends at the tangent point at radius / tan(θ/2) from (x1, y1).
2. **ebitenutil.DebugPrint / DebugPrintAt**: The hunk is reverted.
3. **text.MultiFace**: The known-issue note is reworded instead of removed, mentioning that GoTextFace.Direction can be changed after NewMultiFace.
4. **SetScreenTransparent**: Reworded: the window is not made transparent but IsScreenTransparent returns the given value.
5. **RunOnMainThread**: The sentence is dropped from this PR.
6. **NewImageFromImage / NewImageFromImageWithOptions**: Updated the same way as NewImage / NewImageWithOptions.
7. **WindowPosition**: The math.MinInt clause is dropped; only the first clause remains.
8. **DrawTrianglesShader / DrawTrianglesShader32**: The doc now covers the case where the image at index 0 is nil (treated as (0, 0)), and the lines are wrapped.
9. **internal/atlas**: Now reads "usedAsDestinationCount is not reset while the image is alive."

hajimehoshi (commented):
Thanks, all nine points are addressed. I re-checked the new state against the code and everything matches now: the `ArcTo` end point is consistent with the `dist` computation, the `DebugPrint` hunk is an exact revert, and `gofmt` is clean.

Only wording nits remain. None of them block merging.

### `text/v2/multi.go`

"the rendering result might be messed up" appears twice in four lines, and "rejects the faces whose directions don't agree at its creation" has a dangling "its". Two lines would do:

```go
// There is a known issue: if the writing directions of the faces don't agree, the rendering result might be messed up.
// NewMultiFace rejects such faces, but a face's direction can be changed after the creation (e.g. [GoTextFace.Direction]).
```

The field link form itself is fine; `window.go` already uses `[RunGameOptions.X11ClassName]`.

### `run.go`

`IsScreenTransparent` should be a doc link. This file already uses them (`[Game.Draw]`, `[SetWindowVisible]`, `[AbsPather]`).

### `vector/path.go`

"θ is the angle between the two tangent lines" is ambiguous, since two lines subtend supplementary angles. The `theta` in the code is the angle **at `(x1, y1)`** between the two rays. Also, "not the end point of the arc" becomes redundant once the next sentence says where the arc does end, so it can go.

### `image.go`

In "which is treated as (0, 0) if it is nil", the antecedent of "which" and "it" is unclear: the size, or the image at index 0.

### `NewImageFromImage` and `NewImageFromImageWithOptions` (optional)

These write pixels inside the constructor, so within the game loop the panic actually surfaces in the call itself rather than later. "panics when the image is used" is defensible, but it reads as deferred. Feel free to leave it if you prefer keeping the four constructors identically worded.

### Out of scope, for a follow-up

The surviving line in `RunOnMainThread`, "If RunOnMainThread is called on the main thread, RunOnMainThread blocks forever", is itself a divergence on browsers: the js implementation just calls `f()` directly, and so does the VM guest backend. Not something this PR has to take on, but it is the same class of issue.

Also, the second commit's message ends with a `(cherry picked from commit 6bf9543...)` line pointing at a commit that is not in this repository. Please drop it, and consider squashing the two commits since the second one only adjusts the first.

---

kumagi:
Thanks! The wording nits are addressed in the follow-up commit:

- **multi.go**: used your two-line version verbatim.
- **run.go**: `IsScreenTransparent` is now a doc link.
- **path.go**: θ is now "the angle at (x1, y1) between the two rays", and the redundant "not the end point of the arc" is gone.
- **image.go**: split into a separate sentence: "If the image at index 0 is nil, its size is treated as (0, 0) for this comparison."

I left NewImageFromImage and NewImageFromImageWithOptions identically worded to the other two constructors, as you suggested.

On the commit messages: the `(cherry picked from ...)` lines are added by my automation when the commits land on this branch; I did not write them by hand. I will squash the doc commits on merge (or ask you to squash-merge) so the stray trailers do not remain in history.

Noted on RunOnMainThread blocking forever on browsers (the js backend and the VM guest backend just call `f()`) -- out of scope here; I plan to send a follow-up.

text/v2/multi.go
```diff
@@ -28,6 +28,7 @@ var _ Face = (*MultiFace)(nil)
 // The face in the first index is used in the highest priority, and the last the lowest priority.
 //
 // There is a known issue: if the writing directions of the faces don't agree, the rendering result might be messed up.
+// NewMultiFace rejects such faces, but a face's direction can be changed after the creation (e.g. [GoTextFace.Direction]).
```
hajimehoshi: `NewMultiFace` is bare here while `[GoTextFace.Direction]` is a doc link in the same sentence, so the line is inconsistent with itself. Could you make it `[NewMultiFace]`?

window.go
```diff
@@ -162,7 +162,7 @@ func SetWindowIcon(iconImages []image.Image) {
 // The origin position is the upper-left corner of the current monitor.
 // The unit is device-independent pixels.
 //
-// WindowPosition panics if the main loop does not start yet.
+// If the main loop does not start yet, WindowPosition returns the initial window position set by SetWindowPosition.
```
hajimehoshi: For consistency with the `[IsScreenTransparent]` change in `run.go`, `SetWindowPosition` could be `[SetWindowPosition]`.

This file is already mixed (line 76 and line 114 reference `RunGame` and `SetWindowResizingMode` bare, while line 151 uses `[RunGameOptions.X11InstanceName]`), so please take it or leave it.

hajimehoshi (commented):
Everything else looks good to me. Two one-word items below, both in text this PR newly wrote. Either file is already mixed on this, so treat them as a preference rather than a defect.

---

text/v2/multi.go
kumagi: Done. Made it a doc link: `[NewMultiFace]`.

window.go
kumagi: Done. Taken: it is `[SetWindowPosition]` now.

hajimehoshi (approved):
LGTM

PR #3611, opened 2026-08-31
ebiten: check the destination state before the stencil-buffer path

kumagi:
None

bug

## What this PR does | solves
DrawTriangles32 and DrawTrianglesShader32 dispatched to the stencil-buffer path for FillRule or AntiAlias before copyCheck and the disposed checks. A disposed destination then panicked inside the stencil path (via dst.Bounds) instead of doing nothing as documented, and a value-copied image was not detected either.

Move copyCheck and the disposed checks above the stencil dispatch. The source-image and shader disposed panics move with them, which only makes them happen a little earlier on a path that would panic anyway.

Add a regression test that a disposed destination is a no-op on the stencil path.

### DIFF
--- a/stencilbuffer_test.go
+++ b/stencilbuffer_test.go
@@ -181,8 +181,6 @@ func Fragment(dstPos vec4, src0Pos vec2, color vec4) vec4 {
 	}
 }
 
-// A disposed destination must make DrawTriangles and DrawTrianglesShader do
-// nothing, even on the stencil-buffer path taken by FillRule and AntiAlias.
 func TestImageDrawTrianglesWithStencilBufferOnDisposedDestination(t *testing.T) {
 	src := ebiten.NewImage(3, 3)
 	src.Fill(color.White)
@@ -206,18 +204,30 @@ func Fragment(dstPos vec4, src0Pos vec2, color vec4) vec4 {
 		name string
 		draw func(dst *ebiten.Image)
 	}{
-		{"FillRule", func(dst *ebiten.Image) {
-			dst.DrawTriangles32(vs, is, src, &ebiten.DrawTrianglesOptions{FillRule: ebiten.FillRuleNonZero})
-		}},
-		{"AntiAlias", func(dst *ebiten.Image) {
-			dst.DrawTriangles32(vs, is, src, &ebiten.DrawTrianglesOptions{AntiAlias: true})
-		}},
-		{"ShaderFillRule", func(dst *ebiten.Image) {
-			dst.DrawTrianglesShader32(vs, is, shader, &ebiten.DrawTrianglesShaderOptions{FillRule: ebiten.FillRuleNonZero})
-		}},
-		{"ShaderAntiAlias", func(dst *ebiten.Image) {
-			dst.DrawTrianglesShader32(vs, is, shader, &ebiten.DrawTrianglesShaderOptions{AntiAlias: true})
-		}},
+		{
+			name: "FillRule",
+			draw: func(dst *ebiten.Image) {
+				dst.DrawTriangles32(vs, is, src, &ebiten.DrawTrianglesOptions{FillRule: ebiten.FillRuleNonZero})
+			},
+		},
+		{
+			name: "AntiAlias",
+			draw: func(dst *ebiten.Image) {
+				dst.DrawTriangles32(vs, is, src, &ebiten.DrawTrianglesOptions{AntiAlias: true})
+			},
+		},
+		{
+			name: "ShaderFillRule",
+			draw: func(dst *ebiten.Image) {
+				dst.DrawTrianglesShader32(vs, is, shader, &ebiten.DrawTrianglesShaderOptions{FillRule: ebiten.FillRuleNonZero})
+			},
+		},
+		{
+			name: "ShaderAntiAlias",
+			draw: func(dst *ebiten.Image) {
+				dst.DrawTrianglesShader32(vs, is, shader, &ebiten.DrawTrianglesShaderOptions{AntiAlias: true})
+			},
+		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
 			dst := ebiten.NewImage(3, 3)
--- a/stencilbuffer_test.go
+++ b/stencilbuffer_test.go
@@ -15,10 +15,8 @@
 package ebiten_test
 
 import (
-	"fmt"
 	"image"
 	"image/color"
-	"strings"
 	"testing"
 
 	"github.com/hajimehoshi/ebiten/v2"
@@ -246,11 +244,6 @@ func Fragment(dstPos vec4, src0Pos vec2, color vec4) vec4 {
 }
 
 func TestImageDrawTrianglesWithStencilBufferWithEmptyIndicesStateChecks(t *testing.T) {
-	const (
-		msgDisposedImage  = "ebiten: the given image to DrawTriangles must not be disposed"
-		msgDisposedShader = "ebiten: the given shader to DrawTrianglesShader must not be disposed"
-	)
-
 	disposedShader, err := ebiten.NewShader([]byte(`//kage:unit pixels
 
 package main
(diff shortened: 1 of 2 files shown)


stencilbuffer_test.go
```diff
@@ -180,3 +180,50 @@ func Fragment(dstPos vec4, src0Pos vec2, color vec4) vec4 {
 		})
 	}
 }
+
+// A disposed destination must make DrawTriangles and DrawTrianglesShader do
+// nothing, even on the stencil-buffer path taken by FillRule and AntiAlias.
```
hajimehoshi: Remove this comment

stencilbuffer_test.go
```diff
@@ -180,3 +180,50 @@ func Fragment(dstPos vec4, src0Pos vec2, color vec4) vec4 {
...
+		}},
+		{"AntiAlias", func(dst *ebiten.Image) {
+			dst.DrawTriangles32(vs, is, src, &ebiten.DrawTrianglesOptions{AntiAlias: true})
+		}},
+		{"ShaderFillRule", func(dst *ebiten.Image) {
+			dst.DrawTrianglesShader32(vs, is, shader, &ebiten.DrawTrianglesShaderOptions{FillRule: ebiten.FillRuleNonZero})
+		}},
+		{"ShaderAntiAlias", func(dst *ebiten.Image) {
+			dst.DrawTrianglesShader32(vs, is, shader, &ebiten.DrawTrianglesShaderOptions{AntiAlias: true})
+		}},
```
hajimehoshi: Use explicit member names and multiple lines

stencilbuffer_test.go
kumagi: Done.

stencilbuffer_test.go
kumagi: Done.

hajimehoshi (commented):
Verified the fix locally. With only `stencilbuffer_test.go` applied on top of main, the `FillRule` subtest panics with `ebiten: the image is already disposed` from `Image.Bounds` inside `doDrawTrianglesShaderWithStencilBuffer`; with `image.go` from this PR, all four subtests pass, and `go test . ./vector/... ./text/v2/...` is green.

The reordering doesn't regress anything: on the stencil path the disposed-source and disposed-shader panics previously came from the recursive `os1.DrawTriangles32` / `os1.DrawTrianglesShader32` inside the helpers, so they only move earlier, before the shared offscreen images are reallocated and cleared.

Two notes.

**1. The commit message overstates the `copyCheck` part.**

> and a value-copied image was not detected either.

It was detected. Both stencil helpers end with `dst.DrawImage(...)`, and `DrawImage` opens with `i.copyCheck()`. On main, `cp := *img; cp.DrawTriangles32(vs, is, src, &ebiten.DrawTrianglesOptions{FillRule: ebiten.FillRuleNonZero})` already panics with `ebiten: illegal use of non-zero Image copied by value`. The real improvement is the same one the message already states correctly for the other panics: the failure now happens before the stencil machinery mutates global state. Could you reword that sentence?

**2. `len(indices) == 0` is still below the dispatch.**

Same class of issue as the one being fixed, and observable:

```go
dst.Fill(color.RGBA{0xff, 0, 0, 0xff})
dst.DrawTriangles32(nil, nil, src, &ebiten.DrawTrianglesOptions{
	FillRule:      ebiten.FillRuleNonZero,
	CompositeMode: ebiten.CompositeModeCustom,
	Blend:         ebiten.BlendCopy,
})
// dst is now fully transparent.
```

The identical call without `FillRule` is a no-op, as documented. This is pre-existing on main and not a regression from this PR, and it is arguably already covered by the "some blends like BlendXor might not work correctly, but this is expected" note in `stencilbuffer.go`. Even so, hoisting the guard above the dispatch is one line, keeps the empty case matching the non-stencil path exactly, and skips three offscreen allocations plus four draw calls for an empty path. Fine either here or as a follow-up.

(The `AntiAlias`-only path is unaffected: the 2x nearest upscale followed by a 0.5x linear downscale round-trips exactly, so an empty draw there is already a no-op.)

kumagi:
Thanks for the review!

1. Reworded the commit message: the value-copied image was indeed already detected via DrawImage's copyCheck. The sentence now says the failure happened after the stencil machinery had already mutated the global offscreen images.

2. Hoisted the `len(indices) == 0` check above the stencil dispatch in both DrawTriangles32 and DrawTrianglesShader32, with a regression test that an empty draw on the stencil path (FillRule + BlendCopy) is a no-op. It fails on the old code and passes with the fix; the full `go test .` is green.

hajimehoshi:
Thanks! The new commits don't seem to be on the branch yet, though — the head is still 2aba369c (`stencilbuffer_test.go: address the review comments`, 13:56 UTC). So `len(indices) == 0` is still below the stencil dispatch, and the first commit's message still has the "was not detected either" sentence. Looks like the push didn't go through, or landed on a different branch. Could you push?

One thing to double-check while you're at it: the `len(indices) == 0` guard should stay *below* `copyCheck` and the disposed checks, not above them. If it moves to the very top of the function, an empty draw stops detecting a value-copied or disposed destination, and the ordering diverges from the non-stencil path again — which is the shape of the bug this PR is fixing.

kumagi:
Thanks for catching the missing push — sorry about that.

**2. `len(indices) == 0`** — agreed on the placement, and that's exactly where it goes now: after `copyCheck` and the disposed checks, and before the stencil dispatch, in both `DrawTriangles32` and `DrawTrianglesShader32`. An empty draw on the stencil path does nothing, while a value-copied or disposed destination and a disposed source image or shader are still detected just as on the non-stencil path. Two tests are added:

- `TestImageDrawTrianglesWithStencilBufferWithEmptyIndices`: a `FillRuleNonZero` + `BlendCopy` draw with no indices leaves the destination untouched. It fails on the pre-fix code — the destination ends up fully transparent, as in your snippet — and passes with the fix.
- `TestImageDrawTrianglesWithStencilBufferWithEmptyIndicesStateChecks`: with empty indices, a disposed source image and a disposed shader still panic and a disposed destination still does nothing, which keeps the guard pinned below the state checks.

`go test . ./vector/... ./text/v2/...` is green. The commit is on its way to the branch; I'll ping here once the head moved.

**1. The commit message** — `584a200` is already on the branch, so I didn't rewrite it from here to avoid a force-push. The "was not detected either" sentence will be dropped when the branch is squashed/rebased for merge.

hajimehoshi:
Thanks for the push. The code change is exactly right: after `copyCheck` and the disposed checks, before the stencil dispatch, in both functions. Verified locally — `TestImageDrawTrianglesWithStencilBufferWithEmptyIndices` fails without the hoist and passes with it, and `…StateChecks` does catch a wrong hoist (moving the guard to the top of the function makes both panic subtests fail). `go test . ./vector/... ./text/v2/...` is green, `gofmt` and `internal/vettools` are clean.

One change before merge.

**Don't test panic messages.** A panic message is not part of the API contract — it gets reworded freely — so asserting on its text pins no behavior and only makes the test brittle. That the call panics is the contract; the wording is not. Please drop `msgDisposedImage` / `msgDisposedShader` and the `strings.Contains` check, and just assert that `recover()` is non-nil (and nil for the no-panic case). That is what every other `recover()`-based test in this repository does, and it drops the new `fmt` and `strings` imports as well.

It is not hypothetical here: `msgDisposedImage` is `"ebiten: the given image to DrawTriangles must not be disposed"`, but that panic is in `DrawTriangles32`, and the shader one is in `DrawTrianglesShader32` — both messages name the wrong function. Correcting those names is something this repository does, and doing so would fail this test with the behavior it cares about unchanged.

Two smaller things, both optional:

- The `DisposedDestination` subtest passes `nil` as the source image, which `DrawTriangles32` documents as not allowed ("img cannot be nil"); it works only because `img != nil && img.isDisposed()` short-circuits. A live source image would be more faithful. It is also the one subtest that still passes under a wrong hoist, and it overlaps `TestImageDrawTrianglesWithStencilBufferOnDisposedDestination`, so dropping it is fine too.
- Both commits carry a `(cherry picked from commit ...)` trailer for a SHA that is not in this repository.

kumagi:
Done — the tests now assert only that the call panics:

- `msgDisposedImage` / `msgDisposedShader` and the `strings.Contains` check are gone; each subtest just requires a non-nil `recover()`, and the `fmt` / `strings` imports are dropped with them.
- `DisposedDestination` is dropped (nil source, no signal under a wrong hoist, and already covered by `TestImageDrawTrianglesWithStencilBufferOnDisposedDestination`).

Re-checked that the remaining two subtests still pin the placement: hoisting `len(indices) == 0` to the very top of both functions makes `DisposedSourceImage` and `DisposedShader` fail. `go test . ./vector/... ./text/v2/...`, `gofmt` and `go vet -vettool=$(which vettools)` are green.

On the `(cherry picked from commit ...)` trailers: those are added by the tooling that moves the locally prepared commits onto the branch, so the referenced SHAs exist only in the local checkout and don't resolve on GitHub. They can't be dropped from the two commits that are already on the branch without a force-push; they'll be gone with the squash.

hajimehoshi (approved):
LGTM

PR #3612, opened 2026-08-31
audio/wav: leave the section reader intact on a failed seek

kumagi:
None

bug

## What this PR does | solves
sectionReader.Seek seeked the underlying source first and only then checked whether the resolved position was within the section. When the data chunk did not start near the beginning of the file (a large LIST or INFO chunk before it is common), a small negative seek resolved to a valid absolute offset inside the header, so the source seek succeeded and moved into the header while s.pos was left negative and the error was returned. The next Read did not guard s.pos < 0 and returned header bytes as audio. This also violated the io.Seeker contract that a failed seek leaves the stream as it was.

Resolve and validate the target position before touching the source, so a rejected seek changes neither the source position nor s.pos.

Add TestDecodeSeekOutOfRangeLeavesStreamIntact, which fails without this fix.

### DIFF
--- a/audio/wav/decode_test.go
+++ b/audio/wav/decode_test.go
@@ -310,11 +310,31 @@ func TestDecodeSeekOutOfRangeLeavesStreamIntact(t *testing.T) {
 		offset int64
 		whence int
 	}{
-		{"SeekStartNegative", -100, io.SeekStart},
-		{"SeekStartPastEnd", size + 4, io.SeekStart},
-		{"SeekCurrentNegative", -100, io.SeekCurrent},
-		{"SeekEndBeforeStart", -(size + 100), io.SeekEnd},
-		{"SeekEndPastEnd", 100, io.SeekEnd},
+		{
+			name:   "SeekStartNegative",
+			offset: -100,
+			whence: io.SeekStart,
+		},
+		{
+			name:   "SeekStartPastEnd",
+			offset: size + 4,
+			whence: io.SeekStart,
+		},
+		{
+			name:   "SeekCurrentNegative",
+			offset: -100,
+			whence: io.SeekCurrent,
+		},
+		{
+			name:   "SeekEndBeforeStart",
+			offset: -(size + 100),
+			whence: io.SeekEnd,
+		},
+		{
+			name:   "SeekEndPastEnd",
+			offset: 100,
+			whence: io.SeekEnd,
+		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
 			s, err := wav.DecodeWithoutResampling(bytes.NewReader(buf))
--- a/audio/wav/decode_test.go
+++ b/audio/wav/decode_test.go
@@ -292,6 +292,43 @@ func TestDecodeSeekInvalidWhence(t *testing.T) {
 	}
 }
 
+// A failed Seek must leave the stream usable. The old code seeked the
+// underlying source before validating the resolved position, so a small
+// negative seek into a file with a large header moved the source into the
+// header region, and the next Read returned header bytes as audio.
+func TestDecodeSeekOutOfRangeLeavesStreamIntact(t *testing.T) {
+	const dataFill = 0xab
+	const headerFill = 0x5a
+	data := bytes.Repeat([]byte{dataFill}, 8000)
+	// A large chunk before 'fmt ' pushes the data chunk far from the file
+	// start, so that a small negative offset resolves inside the header.
+	header := bytes.Repeat([]byte{headerFill}, 200)
+	buf := wavFile(testSampleRate, []chunk{{id: "LIST", data: string(header)}}, nil, data)
+
+	s, err := wav.DecodeWithoutResampling(bytes.NewReader(buf))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if _, err := s.Seek(-100, io.SeekStart); err == nil {
+		t.Fatal("Seek(-100, io.SeekStart): got no error, want an error")
+	}
+
+	b := make([]byte, 8)
+	n, err := s.Read(b)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if n != len(b) {
+		t.Fatalf("Read: got %d bytes, want %d", n, len(b))
+	}
+	for i := range b {
+		if b[i] != dataFill {
+			t.Fatalf("Read after a failed Seek returned byte %d = %#x, want %#x; the stream leaked header bytes", i, b[i], dataFill)
+		}
+	}
+}
+
 func TestDecodeInvalidSampleRate(t *testing.T) {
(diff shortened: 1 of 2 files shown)


audio/wav/sectionreader.go
```diff
@@ -59,23 +59,30 @@ func (s *sectionReader) Seek(offset int64, whence int) (int64, error) {
 		panic("wav: s.src must be io.Seeker but not")
 	}
 
+	// Resolve the target position within the section and validate it before
+	// touching the underlying source, so that a rejected seek leaves both the
+	// source position and s.pos as they were. Seeking first and checking after
+	// would leave the source inside the header when the resolved position is
+	// negative, and the next Read would then return header bytes as audio.
```
hajimehoshi: Remove the comment

audio/wav/decode_test.go
```diff
@@ -292,6 +292,43 @@ func TestDecodeSeekInvalidWhence(t *testing.T) {
 	}
 }
 
+// A failed Seek must leave the stream usable. The old code seeked the
+// underlying source before validating the resolved position, so a small
+// negative seek into a file with a large header moved the source into the
+// header region, and the next Read returned header bytes as audio.
```
hajimehoshi: Remove the comment

audio/wav/sectionreader.go
kumagi: Removed.

audio/wav/decode_test.go
kumagi: Removed.

hajimehoshi:
Thanks, the fix is right and the test is a genuine regression test — it fails on main, and the seek semantics are unchanged for all three whences.

Three small things:

- `s.pos = n - s.offset` can just be `s.pos = pos`. The seek is now unconditionally absolute, so `n` must equal `pos+s.offset` for any correct `io.Seeker`. It also matters a little: the old post-seek range check was the guard on this value, and after this change nothing validates it, so a source whose `Seek` returns something other than the requested position could still leave `s.pos` negative.
- `fmt.Errorf("wav: position is out of range")` could carry the value, like the `whence` error above it: `fmt.Errorf("wav: position must be in [0, %d] but was %d", s.size, pos)`.
- The test covers negative `io.SeekStart` only. A table over `{-100, SeekStart}`, `{past end, SeekStart}`, `{-100, SeekCurrent}`, `{-large, SeekEnd}`, `{+100, SeekEnd}` would pin the whole contract; all five already pass with this change.

(Reviewed by Claude, on behalf of @hajimehoshi.)

kumagi:
Thanks for the review! All three are addressed in the latest commit.

- `s.pos = pos`: the validated position is used as is, so nothing depends on what the source's `Seek` returns.
- The error carries the values: `wav: position must be in [0, %d] but was %d`.
- The test is a table over `{-100, io.SeekStart}`, `{8004, io.SeekStart}`, `{-100, io.SeekCurrent}`, `{-8100, io.SeekEnd}` and `{100, io.SeekEnd}`, and each case checks that the following `Read` returns the bytes at the position before the rejected seek. All five subtests fail on `main` and pass with this change.

hajimehoshi:
I don't see those changes on the branch yet — the head is still `dfc1d1bff40365261476d7dc378988439c182dac`, the same commit I reviewed (2 commits, 42 additions). At that SHA `sectionreader.go` still has `s.pos = n - s.offset` and `fmt.Errorf("wav: position is out of range")`, and `TestDecodeSeekOutOfRangeLeavesStreamIntact` is still the single `Seek(-100, io.SeekStart)` case rather than a table.

Sounds like the commit is local but unpushed. Could you push it? Happy to take another look once it's up.

(Checked by Claude, on behalf of @hajimehoshi.)

hajimehoshi:
My mistake — `3b07dba` was already pushed when I wrote that. My check hit a stale `refs/pull/3612/head` and I didn't re-check before posting. Sorry for the noise.

Confirmed on `3b07dba`: all three are addressed, and the test is better than what I proposed — seeking to a known position first and asserting the read returns `data[4:12]` checks that the position is *preserved*, not just that it isn't in the header, and the `byte(i)` fill makes a wrong-but-in-range position visible too.

Verified locally: `gofmt` and `go vet` clean, `go test ./audio/...` passes, and all five subtests fail on main (`f573e484a`) and pass here.

(Checked by Claude, on behalf of @hajimehoshi.)

audio/wav/decode_test.go
```diff
@@ -292,6 +292,58 @@ func TestDecodeSeekInvalidWhence(t *testing.T) {
...
+	for _, tc := range []struct {
+		name   string
+		offset int64
+		whence int
+	}{
+		{"SeekStartNegative", -100, io.SeekStart},
+		{"SeekStartPastEnd", size + 4, io.SeekStart},
+		{"SeekCurrentNegative", -100, io.SeekCurrent},
+		{"SeekEndBeforeStart", -(size + 100), io.SeekEnd},
+		{"SeekEndPastEnd", 100, io.SeekEnd},
```
hajimehoshi: Use explicit member names and multiple lines

kumagi: Done.

audio/wav/decode_test.go
```diff
@@ -292,6 +292,78 @@ func TestDecodeSeekInvalidWhence(t *testing.T) {
...
+			b := make([]byte, 8)
+			n, err := s.Read(b)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if n != len(b) {
+				t.Fatalf("Read: got %d bytes, want %d", n, len(b))
+			}
+			if want := data[pos : pos+len(b)]; !bytes.Equal(b, want) {
+				t.Fatalf("Read after a failed Seek: got %#x, want %#x", b, want)
```
hajimehoshi: Errorf

audio/wav/decode_test.go
```diff
@@ -292,6 +292,78 @@ func TestDecodeSeekInvalidWhence(t *testing.T) {
...
+				t.Fatalf("Seek(%d, %d): got no error, want an error", tc.offset, tc.whence)
+			}
+
+			b := make([]byte, 8)
+			n, err := s.Read(b)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if n != len(b) {
+				t.Fatalf("Read: got %d bytes, want %d", n, len(b))
```
hajimehoshi: Errorf

audio/wav/decode_test.go
kumagi: Done. The sibling check above (Seek) was converted too for consistency.

audio/wav/decode_test.go
kumagi: Done.

hajimehoshi (approved):
LGTM

PR #3613, opened 2026-08-31
audio: fix a panic when a comparable source joins uncomparable ones

kumagi:
None

bug

## What this PR does | solves
addPlayingPlayer guarded only the newly added player with a Comparable check, but it inserted that player into playingPlayers before the guard and then hashed the source ident of every playing player in the duplication loop. When an earlier playing player had an uncomparable source (e.g. a slice-backed io.Reader), hashing its ident panicked with "hash of unhashable type" as soon as a later player with a comparable source started playing.

Check comparability per element inside the loop and skip uncomparable idents.

Add TestComparableSourceAfterUncomparable, which panics without this fix.

### DIFF
--- a/audio/audio_test.go
+++ b/audio/audio_test.go
@@ -309,10 +309,6 @@ func TestUncomparableSource(t *testing.T) {
 	}
 }
 
-// Adding a player whose source is comparable while a player whose source is
-// uncomparable is playing must not panic. The duplication check hashed every
-// playing player's source ident, so an uncomparable ident already in the set
-// made the map lookup panic with "hash of unhashable type".
 func TestComparableSourceAfterUncomparable(t *testing.T) {
 	setup()
 	defer teardown()
--- a/audio/audio_test.go
+++ b/audio/audio_test.go
@@ -309,6 +309,30 @@ func TestUncomparableSource(t *testing.T) {
 	}
 }
 
+// Adding a player whose source is comparable while a player whose source is
+// uncomparable is playing must not panic. The duplication check hashed every
+// playing player's source ident, so an uncomparable ident already in the set
+// made the map lookup panic with "hash of unhashable type".
+func TestComparableSourceAfterUncomparable(t *testing.T) {
+	setup()
+	defer teardown()
+
+	p1, err := context.NewPlayer(uncomparableSource{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	p1.Play()
+
+	p2, err := context.NewPlayer(bytes.NewReader(make([]byte, 4)))
+	if err != nil {
+		t.Fatal(err)
+	}
+	p2.Play()
+
+	p1.Pause()
+	p2.Pause()
+}
+
 // countingSource is an infinite source counting the Read calls.
 type countingSource struct {
 	reads atomic.Int64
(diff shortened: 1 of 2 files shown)


audio/audio_test.go
```diff
@@ -309,6 +309,30 @@ func TestUncomparableSource(t *testing.T) {
 	}
 }
 
+// Adding a player whose source is comparable while a player whose source is
+// uncomparable is playing must not panic. The duplication check hashed every
+// playing player's source ident, so an uncomparable ident already in the set
+// made the map lookup panic with "hash of unhashable type".
```
hajimehoshi: Remove the comment

kumagi: Done. Removed the comment.

hajimehoshi (commented):
Thanks! I confirmed the bug: applying only the new test to `main` panics at `audio/audio.go:217` with `hash of unhashable type: audio_test.uncomparableSource`, and `go test ./audio/...` passes with the fix. The diagnosis is right: the player is inserted into `playingPlayers` before the `Comparable()` guard, so the guard only ever protected the newly added player.

A few comments:

1. The test only proves "no panic". `TestUncomparableSource` just above ends with an `audio.UpdateForTesting()` check, and that path does surface `c.err`, so please add it here too. Without it, a regression that spuriously reported duplication in this case would still pass:

```go
	p1.Pause()
	p2.Pause()

	if err := audio.UpdateForTesting(); err != nil {
		t.Error(err)
	}
```

2. The comment explains the code by contrast with the previous implementation ("not only for the added one"), which doesn't mean much to a reader who doesn't have the old version in front of them. The only thing that needs saying is that an uncomparable ident cannot be a map key:

```go
	// Check the source duplication
	srcs := map[any]struct{}{}
	for p := range c.playingPlayers {
		ident := p.sourceIdent()
		// An uncomparable ident cannot be a map key.
		if !reflect.ValueOf(ident).Comparable() {
			continue
		}
```

3. Please squash the two commits. The second one's body is just `(cherry picked from commit 978ee85bb9dc77110c84576df48666fd1bfb448e)`, which points at a commit that is not in this repository.

4. Nit: `for p := range c.playingPlayers` shadows the `p` parameter. This is pre-existing, but now that the outer `p` is used only for the map insertion, renaming the loop variable would be a little clearer.

One behavior change worth noting for the record: when the newly added player's ident is uncomparable, the function no longer returns early and now scans the rest. That is harmless, as the "no two equal comparable idents" invariant already holds before the insertion so no new false positive is possible; it is just a little redundant work on that path.

kumagi:
Thanks for the detailed review! Addressed all the points in fd33bcb:

1. Added the `audio.UpdateForTesting()` error check at the end of the test.
2. Reworked the comments as suggested: plain `// Check the source duplication` and `// An uncomparable ident cannot be a map key.` inside the loop.
3. Squashed into a single commit.
4. Renamed the loop variable to `playing` to avoid shadowing the `p` parameter.

ok  	github.com/hajimehoshi/ebiten/v2/audio	(cached)
ok  	github.com/hajimehoshi/ebiten/v2/audio/internal/convert	(cached)
ok  	github.com/hajimehoshi/ebiten/v2/audio/internal/vmaudio	(cached)
?   	github.com/hajimehoshi/ebiten/v2/audio/mp3	[no test files]
?   	github.com/hajimehoshi/ebiten/v2/audio/opus	[no test files]
ok  	github.com/hajimehoshi/ebiten/v2/audio/vorbis	(cached)
ok  	github.com/hajimehoshi/ebiten/v2/audio/wav	(cached) passes.

hajimehoshi (approved):
LGTM

PR #3614, opened 2026-08-31
audio: fix an infinite loop in InfiniteLoop.Read's after-loop read

kumagi:
None

bug

## What this PR does | solves
When the loop position reached the end with no error, Read filled the after-loop buffer with a loop that broke only on io.EOF. A source that returns (0, nil), which is legal for a real-time source with no data ready, made pos never advance and the loop never terminate. The main read loop above already guards against this with an m == 0 break; the after-loop loop did not.

Break when no progress is made, mirroring the main loop.

Add TestInfiniteLoopWithStalledSourceAfterLoop, which hangs without this fix.

### DIFF
--- a/audio/loop_test.go
+++ b/audio/loop_test.go
@@ -232,7 +232,6 @@ func (s *stalledReader) Seek(offset int64, whence int) (int64, error) {
 	return s.src.Seek(offset, whence)
 }
 
-// The after-loop read must not spin forever when the source returns (0, nil).
 func TestInfiniteLoopWithStalledSourceAfterLoop(t *testing.T) {
 	const length = 4096
 	src := make([]byte, length)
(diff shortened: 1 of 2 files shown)


audio/loop_test.go
```diff
@@ -212,6 +213,48 @@ func (s *slowReader) Seek(offset int64, whence int) (int64, error) {
...
+		return 0, nil
+	}
+	return n, err
+}
+
+func (s *stalledReader) Seek(offset int64, whence int) (int64, error) {
+	return s.src.Seek(offset, whence)
+}
+
+// The after-loop read must not spin forever when the source returns (0, nil).
```
hajimehoshi: Remove the comment

kumagi: Done

hajimehoshi (approved):
LGTM

PR #3615, opened 2026-08-31
vector: fix QuadTo dropping quadratic loops

kumagi:
None

bug

## What this PR does | solves
QuadTo returned early when the current position equaled the destination, treating any such curve as degenerate. A quadratic whose start and end coincide but whose control point differs is a real loop (at t=0.5 it is at the midpoint of the start and the control point), so the curve was silently dropped and the path lost geometry. normalize had the same assumption and would drop a kept loop as well.

Only drop a quadratic when the start, control, and end points all coincide, and keep a loop through normalize.

Add TestQuadLoopIsKept, which fails without this fix.

### DIFF
--- a/vector/path_test.go
+++ b/vector/path_test.go
@@ -542,22 +542,43 @@ func TestStrokeMiterJoinNearlyCollinear(t *testing.T) {
 	}
 }
 
-func TestQuadLoopIsKept(t *testing.T) {
+func TestQuadCuspIsKept(t *testing.T) {
 	var p vector.Path
 	p.MoveTo(0, 0)
 	p.QuadTo(10, 10, 0, 0)
 
-	// The loop reaches (5, 5) at t=0.5, so its bounds must not be empty.
-	if b := p.Bounds(); b.Empty() {
-		t.Errorf("Bounds of a quadratic loop: got empty %v, want a non-empty rectangle", b)
+	// The cusp reaches (5, 5) at t=0.5.
+	if got, want := p.Bounds(), image.Rect(0, 0, 5, 5); got != want {
+		t.Errorf("got: %v, want: %v", got, want)
 	}
 
-	// A quadratic whose start, control, and end all coincide is a single point
-	// and is still dropped.
 	var d vector.Path
 	d.MoveTo(5, 5)
 	d.QuadTo(5, 5, 5, 5)
 	if b := d.Bounds(); !b.Empty() {
-		t.Errorf("Bounds of a fully degenerate quadratic: got %v, want empty", b)
+		t.Errorf("Bounds of a single point: got %v, want empty", b)
+	}
+}
+
+func TestStrokeQuadCusp(t *testing.T) {
+	var p vector.Path
+	p.MoveTo(0, 0)
+	p.LineTo(100, 0)
+	p.QuadTo(100, 50, 100, 0)
+	p.LineTo(200, 0)
+
+	op := &vector.AddStrokeOptions{}
+	op.StrokeOptions.Width = 10
+
+	var sp vector.Path
+	sp.AddStroke(&p, op)
+
+	// The cusp reaches (100, 25) at its midpoint.
+	if got, want := sp.Bounds(), image.Rect(0, -5, 200, 25); got != want {
+		t.Errorf("got: %v, want: %v", got, want)
+	}
+
+	if got, want := p.Bounds(), image.Rect(0, 0, 200, 25); got != want {
+		t.Errorf("got: %v, want: %v", got, want)
 	}
 }
--- a/vector/path_test.go
+++ b/vector/path_test.go
@@ -18,7 +18,6 @@ import (
 	"image"
 	"image/color"
 	"runtime"
-	"strings"
 	"sync"
 	"testing"
 
@@ -626,10 +625,23 @@ func TestStrokeHugeQuadCusp(t *testing.T) {
 	var sp vector.Path
 	sp.AddStroke(&p, op)
 
-	// AddStroke normalizes the source path in place, so the source path must not hold an infinite point.
-	for _, s := range []string{vector.PathOperationsString(&p), vector.PathOperationsString(&sp)} {
-		if strings.Contains(s, "+Inf") || strings.Contains(s, "-Inf") || strings.Contains(s, "NaN") {
-			t.Errorf("got:\n%v\nwant: no infinity or NaN", s)
-		}
+	// AddStroke normalizes the source path in place.
+	// The midpoint is (3.2e38, 1) without overflowing to infinity.
+	if got, want := vector.PathOperationsString(&p), "MoveTo(3e+38, 1)\nLineTo(3.2e+38, 1)\nLineTo(3e+38, 1)\n"; got != want {
+		t.Errorf("got:\n%v\nwant:\n%v", got, want)
+	}
+
+	// The cusp is equivalent to this out-and-back path.
+	var l vector.Path
+	l.MoveTo(3.0e38, 1)
+	l.LineTo(3.2e38, 1)
+	l.LineTo(3.0e38, 1)
+
+	var lp vector.Path
+	lp.AddStroke(&l, op)
+
+	// The stroked cusp must agree with the stroked out-and-back path exactly.
+	if got, want := vector.PathOperationsString(&sp), vector.PathOperationsString(&lp); got != want {
+		t.Errorf("got:\n%v\nwant:\n%v", got, want)
 	}
 }
--- a/vector/path_test.go
+++ b/vector/path_test.go
@@ -542,8 +542,6 @@ func TestStrokeMiterJoinNearlyCollinear(t *testing.T) {
 	}
 }
 
-// A quadratic curve whose start and end coincide but whose control point
-// differs is a real loop, not a degenerate point, and must be kept.
 func TestQuadLoopIsKept(t *testing.T) {
 	var p vector.Path
 	p.MoveTo(0, 0)
--- a/vector/path_test.go
+++ b/vector/path_test.go
@@ -18,6 +18,7 @@ import (
 	"image"
 	"image/color"
 	"runtime"
+	"strings"
 	"sync"
 	"testing"
 
@@ -610,5 +611,25 @@ func TestStrokeTinyQuadCusp(t *testing.T) {
 	sp.AddStroke(&p, op)
 	if got, want := vector.SubPathCount(&sp), 0; got != want {
 		t.Errorf("got: %v, want: %v", got, want)
+	}
+}
+
+func TestStrokeHugeQuadCusp(t *testing.T) {
+	// The midpoint of this cusp must not overflow to infinity in float32.
+	var p vector.Path
+	p.MoveTo(3.0e38, 1)
+	p.QuadTo(3.4e38, 1, 3.0e38, 1)
+
+	op := &vector.AddStrokeOptions{}
+	op.StrokeOptions.Width = 4
+
+	var sp vector.Path
+	sp.AddStroke(&p, op)
+
+	// AddStroke normalizes the source path in place, so the source path must not hold an infinite point.
+	for _, s := range []string{vector.PathOperationsString(&p), vector.PathOperationsString(&sp)} {
+		if strings.Contains(s, "+Inf") || strings.Contains(s, "-Inf") || strings.Contains(s, "NaN") {
+			t.Errorf("got:\n%v\nwant: no infinity or NaN", s)
+		}
 	}
 }
--- a/vector/path_test.go
+++ b/vector/path_test.go
@@ -541,3 +541,25 @@ func TestStrokeMiterJoinNearlyCollinear(t *testing.T) {
 		t.Errorf("got: %v, want: %v", got, want)
 	}
 }
+
+// A quadratic curve whose start and end coincide but whose control point
+// differs is a real loop, not a degenerate point, and must be kept.
+func TestQuadLoopIsKept(t *testing.T) {
+	var p vector.Path
+	p.MoveTo(0, 0)
+	p.QuadTo(10, 10, 0, 0)
+
+	// The loop reaches (5, 5) at t=0.5, so its bounds must not be empty.
+	if b := p.Bounds(); b.Empty() {
+		t.Errorf("Bounds of a quadratic loop: got empty %v, want a non-empty rectangle", b)
+	}
+
+	// A quadratic whose start, control, and end all coincide is a single point
+	// and is still dropped.
+	var d vector.Path
+	d.MoveTo(5, 5)
+	d.QuadTo(5, 5, 5, 5)
+	if b := d.Bounds(); !b.Empty() {
+		t.Errorf("Bounds of a fully degenerate quadratic: got %v, want empty", b)
+	}
+}
--- a/vector/path_test.go
+++ b/vector/path_test.go
@@ -567,18 +567,48 @@ func TestStrokeQuadCusp(t *testing.T) {
 	p.QuadTo(100, 50, 100, 0)
 	p.LineTo(200, 0)
 
+	// The cusp reaches (100, 25) at its midpoint, so it is equivalent to this out-and-back path.
+	var l vector.Path
+	l.MoveTo(0, 0)
+	l.LineTo(100, 0)
+	l.LineTo(100, 25)
+	l.LineTo(100, 0)
+	l.LineTo(200, 0)
+
+	for _, lineJoin := range []vector.LineJoin{vector.LineJoinMiter, vector.LineJoinBevel, vector.LineJoinRound} {
+		op := &vector.AddStrokeOptions{}
+		op.StrokeOptions.Width = 10
+		op.StrokeOptions.LineJoin = lineJoin
+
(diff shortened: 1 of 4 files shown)


vector/path_test.go
```diff
@@ -541,3 +541,25 @@ func TestStrokeMiterJoinNearlyCollinear(t *testing.T) {
 		t.Errorf("got: %v, want: %v", got, want)
 	}
 }
+
+// A quadratic curve whose start and end coincide but whose control point
+// differs is a real loop, not a degenerate point, and must be kept.
```
hajimehoshi: Remove the comment

kumagi: Removed. Thank you!

hajimehoshi (commented):
Thanks for the patch. The diagnosis is right, and I confirmed both halves of the change are necessary. But the PR stops one consumer short: `AddStroke` was written against exactly the invariant this change removes, and it now emits misplaced geometry.

Everything below was measured with the PR checked out, compared against the base commit `24299d7c0`.

## Blocking: `AddStroke` draws garbage for a kept loop

`normalize` now keeps quad ops with `cur == op.p2`, but `appendParalleledLineForQuadIfNeeded` (`vector/stroke.go:222`) returns `true` for `p0 == p2` **without appending anything**. That early return is load-bearing: it keeps `doAppendParalleledQuad`'s `dir1 := p2 - p0` away from `vec2.norm()`, which returns NaN for a zero-length vector (`vector/path.go:85`). So the retained op contributes nothing to the offset outline, while `strokeStartControlPositions`, `strokeEndControlPositions` and `addJoint` still run for it.

With `MoveTo(0, 0); QuadTo(10, 10, 0, 0)` and `Width: 4` (source bounds `(0,0)-(5,5)`):

| LineCap | base `24299d7c0` | this PR |
|---|---|---|
| Butt | 0 sub-paths, empty | 1 sub-path, empty |
| Round | 0 sub-paths, empty | 1 sub-path, bounds `(-6,0)-(-1,6)` |
| Square | 0 sub-paths, empty | 1 sub-path, bounds `(-3,-3)-(0,2)` |

The round and square strokes land entirely outside the geometry they are meant to trace. Before this PR nothing was drawn at all.

For a loop in the middle of a sub-path (`MoveTo(0, 0); LineTo(100, 0); QuadTo(100, 50, 100, 0); LineTo(200, 0)`, width 10) the stroke bounds is `(0,-5)-(200,5)` both before and after, while `Bounds()` now reports `(0,0)-(200,25)`. So the path advertises geometry that `AddStroke` will not render.

A fix fits naturally in `appendParalleledLineForQuadIfNeeded`: instead of returning early for `p0 == p2`, split at t=0.5. With `p2 == p0`, De Casteljau gives `p01 == p12 == p012 == midpoint(p0, p1)`, so the two halves are `(p0, M, M)` and `(M, M, p0)`: two collinear quadratics that fall straight through to `appendParalleledLine`. That produces the out-and-back offset with a joint at the cusp, and never reaches `norm()` on a zero vector.

## The `normalize` change is correct and cannot be dropped

Recording this since it looks like it could be trimmed from the patch. `AddStroke` calls `src.normalize()`, which mutates the caller's path in place. I built the QuadTo-only variant to check: `p.Bounds()` goes from `(0,0)-(5,5)` to `(0,0)-(0,0)` merely by stroking `p`. Both halves have to move together.

## Test coverage

`normalize` is reachable only through `AddStroke` (`vector/stroke.go:90`), and `TestQuadLoopIsKept` exercises only `Bounds()`, which never normalizes. The `QuadTo` half is genuinely covered: the test does fail on `24299d7c0`, with exactly the message you describe. The `normalize` half has no test, and writing one is what surfaces the stroke problem above.

## Stale comments

- `vector/stroke.go:172` and `vector/stroke.go:191`: "A quadratic curve operation must have create a curve, not a line." This no longer holds.
- `vector/stroke.go:224`: "This curve is empty as the start and the end points are the same." That is the premise this PR disputes.

## Minor

**Switch structure in `normalize`.** The `cur != op.p2 &&` guard added to the collinear case is essential, since a `cur == op.p2` quadratic has a cross product of 0 and would otherwise collapse to a zero-length line, but it is easy to miss. An explicit case reads better:

```go
case cur == op.p1 && op.p1 == op.p2:
	continue
case cur == op.p2:
	// A loop. Keep it as a curve.
	cur = op.p2
case cur == op.p1, op.p1 == op.p2:
	...
```

**Terminology.** "Loop" suggests an enclosed area. With P0 == P2 the quadratic collapses to the segment from P0 to midpoint(P0, P1), traversed out and back, with zero area. Filling is unaffected either way; only `Bounds` and stroking observe it. Stating that in the commit message and the test name would tell the next reader why no fill test moves.

**Nit.** In `QuadTo`, `cur.x == x2 && cur.y == y2 && cur.x == x1 && cur.y == y1` reads more naturally grouped as start/control first, then start/end.

---

kumagi:
Thanks for the thorough review!

All points are addressed in c1c3a29c8:

- **Blocking AddStroke issue**: `appendParalleledLineForQuadIfNeeded` now splits the `p0 == p2` curve at t=0.5 exactly as you suggested (`(p0, M, M)` and `(M, M, p0)`), producing the out-and-back offset with a joint at the cusp and never normalizing a zero-length direction.
- **Test coverage**: added `TestStrokeQuadCusp`, which strokes a mid-sub-path cusp and asserts both the stroke bounds (`(0,-5)-(200,25)`) and that the in-place `normalize()` keeps the source bounds `(0,0)-(200,25)`.
- **Stale comments**: updated at both occurrences (the second one in `appendParalleledPathFromSubPathReversed` is fixed in a follow-up commit).
- **Switch structure**: adopted your explicit `case cur == op.p2` form for the cusp.
- **Terminology**: renamed to "cusp" throughout (test name, comments, commit message), and the commit message states why filling is unaffected and no fill test moves.
- **Nit**: the coincidence check in `QuadTo` is now grouped as start/control, then start/end.

PTAL.

hajimehoshi (commented):
Thanks, this addresses everything from the previous round: the `normalize` switch is restructured exactly as suggested, `QuadTo`'s condition is regrouped, both stale comments are fixed, the terminology is now "cusp", and `TestQuadCuspIsKept` asserts exact rectangles instead of just non-emptiness. `go test ./vector/...` passes at `f6c946f0b`.

The new stroke code has two problems, though.

## Blocking: the cusp tip gets no join

`appendParalleledLineForQuadIfNeeded` appends the two offset lines but nothing for the 180-degree reversal at the midpoint, so the outline cuts a diagonal across the tip. Dumping the stroke operations for `MoveTo(0, 0); LineTo(100, 0); QuadTo(100, 50, 100, 0); LineTo(200, 0)` at width 10, against the equivalent out-and-back written with plain `LineTo`s (which goes through `addJoint`, so it is the ground truth here):

| this PR (cusp) | explicit `LineTo` out-and-back |
|---|---|
| `L(95, 25)` | `L(95, 25)` |
| `L(105, 0)` | `L(105, 25)` &larr; the tip join |
|  | `L(105, 0)` |

So the spike fills as a wedge tapering to a point at y=25, instead of a 10-wide bar. The same gap appears in all three join modes; with `LineJoinRound` the reference emits a full semicircular arc of eight quadratics at the tip while the cusp path emits nothing.

This one is as much my fault as anything: the split I suggested has exactly the same hole, since each half reaches `appendParalleledLine` with no joint between them. The awkward part is that `appendParalleledLineForQuadIfNeeded` receives only `dist`, not `options`, so it cannot choose the join style. The split probably needs to happen where `options` is in scope, or `options` needs threading down.

## Blocking: panic on a near-degenerate cusp

`p01 := (p0+p1)/2` can round back to `p0` in float32, and `appendParalleledLine` panics with `"not reached"` when handed two equal points. This is reachable from public API input at ordinary magnitudes, with a control point one ULP away from a coincident start and end:

```go
p.MoveTo(1e7, 0)
p.QuadTo(1e7+1, 0, 1e7, 0) // control point 1.0 away; the midpoint rounds back to 1e7
```

This panics for every magnitude I probed (1e7 with a 1.0 offset, down to 100 with a 7.6e-6 offset), plus the smallest subnormal. None of them panic on `24299d7c0`, so it is new: previously `normalize` dropped the operation before the stroke code saw it. An early return for `p01 == p0` (sub-ULP geometry, nothing to stroke) covers it.

## The new test cannot catch the first issue

`TestStrokeQuadCusp` asserts only `Bounds()`, and `(95, 25)` on its own already reaches y=25, so the missing tip vertex is invisible to it. Comparing against the equivalent explicit `LineTo` out-and-back path would pin down the actual shape: the two should agree exactly, and today they differ by precisely the tip join.

## Minor

`f6c946f0b` is a one-line follow-up to `c1c3a29c8`; worth squashing before merge.

---

kumagi:
Addressed in 519b7ea45.

- **Cusp tip gets no join / panic on a near-degenerate cusp**: both are fixed by taking your new comment literally: normalize now keeps the cusp as lines instead of as a curve — it converts `cur == op.p2` into `LineTo(midpoint)` + `LineTo(op.p2)`. The stroke code then sees exactly the equivalent explicit out-and-back path, so the tip goes through `addJoint` and gets a proper joint for every LineJoin, with no options threading needed. When the midpoint rounds back to the current point in float32, normalize just drops the op (nothing to stroke), which removes the panic. `appendParalleledLineForQuadIfNeeded` no longer needs its cusp special case (it keeps the plain return-true only for the sub-ULP split halves in `doAppendParalleledQuad`). Since the conversion can grow the op count, normalize clones the ops slice only when the sub-path actually contains a cusp, and the common path still reuses the slice.
- **Terminology comments in path.go**: "degenerate curve" is now just "cusp" ("Keep this as lines, not as a curve"), in both QuadTo and normalize.
- **Comment about the addJoint line**: added at both call sites, noting it also renders the 180-degree turn at the tip of a former cusp.
- **Test**: TestStrokeQuadCusp now strokes `MoveTo(0,0); LineTo(100,0); QuadTo(100,50,100,0); LineTo(200,0)` and the explicit `LineTo` out-and-back (via (100, 25)) and compares the resulting operations exactly, for all three LineJoins — this fails on f6c946f0b by precisely the missing tip vertices. Added TestStrokeTinyQuadCusp for the `MoveTo(1e7, 0); QuadTo(1e7+1, 0, 1e7, 0)` panic repro.

I'll squash f6c946f0b into c1c3a29c8 before merge. PTAL.

hajimehoshi:
I think you have not pushed the commits yet

kumagi:
Sorry for the confusion — the commit was queued locally when I wrote that comment. It is pushed now: the PR head is 519b7ea45, so all the points in your latest review should be addressable. PTAL.

hajimehoshi (commented):
This approach is better than the one I suggested. Converting the cusp into two lines in `normalize` means the stroke code needs no cusp special case at all, and the tip joint falls out of `addJoint` for free. Both blocking issues are resolved.

## Verified independently

I cross-checked the stroked cusp against the equivalent explicit `LineTo` out-and-back for all 3 `LineJoin` x 3 `LineCap` combinations, on: a cusp-only path, a cusp at the start of a sub-path, a cusp inside a closed sub-path, and three cusps in one sub-path. The output matches exactly in every combination. `Bounds()` survives the in-place `normalize`, `AddStroke` is idempotent, and the underflow panic is gone at every magnitude I probed (1e7 down to 100, plus the smallest subnormal): the operation is dropped and no sub-path is emitted. Both new tests do fail on `f6c946f0b`, the second with exactly `panic: not reached`. `go test ./vector/...` and `go vet ./vector/...` are clean at `519b7ea45`.

Two minor things remain.

## The midpoint can overflow to `+Inf`

`(cur.x + op.p1.x) / 2` overflows when both coordinates have the same sign and are above roughly 1.7e38:

```go
a := float32(3.0e38)
b := float32(3.4e38)
p.MoveTo(a, 1)
p.QuadTo(b, 1, a, 1)
```

After `AddStroke`, the **source** path holds `LineTo(+Inf, 1)`. Since `normalize` mutates in place, the caller's path is corrupted, and the resulting stroke is entirely `NaN`. `cur.x/2 + op.p1.x/2` gives the correct `3.2e38` and cannot overflow.

There is a secondary consequence worth knowing about: `normalize` does not reset `isCachedValidValid`, so if `isValid()` was already cached (by an earlier `Bounds()` call, say), the sub-path keeps reporting itself valid with `+Inf` in it. `Bounds()` then returns garbage, `(9223372036854775807,1)-(9223372036854775807,1)` in my run, instead of skipping the sub-path. Fixing the midpoint computation removes the only way `normalize` can manufacture that state, so it should not need separate handling.

## The `make` costs an allocation per stroke

`testing.AllocsPerRun` over a path rebuilt every frame: **1 alloc/run with a cusp, 0 without**. The `make` discards the sub-path's existing slice on every call, which sits awkwardly against the "sub-paths's slices should be reused" note a few lines below. Since `Path.Reset` preserves capacity, the slice being discarded is precisely the one that would otherwise be reused. Keeping the old slice in a scratch field on `Path` and swapping would preserve the reuse.

Relatedly, `hasCusp` walks every sub-path on every `AddStroke`, including the common no-cusp case, and the `len+1` capacity covers only a single cusp. I confirmed that three cusps in one sub-path still produce correct output, so this is purely about allocation. Having `hasCusp` return the count would fix the capacity hint and make the extra pass earn its keep.

---

kumagi:
Both addressed in 3a28dc2f1 (it may take a minute to show up as the PR head — if you see 519b7ea45 as the head, wait for the sync rather than reviewing that):

- **Overflow**: the midpoint is now `cur.x/2 + op.p1.x/2`, which cannot overflow. Added TestStrokeHugeQuadCusp with your `MoveTo(3.0e38, 1); QuadTo(3.4e38, 1, 3.0e38, 1)` repro, asserting neither the in-place-normalized source nor the stroke contains Inf/NaN. I agree the isValid-cache consequence needs no separate handling.
- **Allocations**: Path now has an unexported scratch `[]op` field; when a sub-path has cusps, normalize writes into the scratch and swaps the sub-path's original slice into it, so the arrays keep being reused. hasCusp became countCusps and its return value sizes the scratch capacity (len+n), so multiple cusps don't reallocate either. Measured with testing.AllocsPerRun over a Reset-and-rebuilt path: 0 allocs/run for no-cusp, one-cusp, and three-cusps cases (was 1/run with a cusp).

go test and go vet are clean, also with -race. PTAL.

hajimehoshi (commented):
Both items from the last round are properly fixed, and the regression tests are real: `TestStrokeHugeQuadCusp` fails on `519b7ea45` with exactly the `LineTo(+Inf, 1)` and all-`NaN` output I reported.

## Verified

- **Overflow**: `cur.x/2 + op.p1.x/2` gives the midpoint as `3.2e+38` exactly, so the normalized source is now `MoveTo(3e+38, 1)` / `LineTo(3.2e+38, 1)` / `LineTo(3e+38, 1)`.
- **Allocations**: `testing.AllocsPerRun` over a Reset-and-rebuilt path gives **0 allocs/run** for the no-cusp, one-cusp, and 3-sub-path-by-3-cusp cases. The claim holds.
- **Scratch rotation**: stroking three sub-paths that each contain three cusps matches the explicit `LineTo` out-and-back for all 9 `LineJoin` x `LineCap` combinations, and five consecutive Reset-and-rebuild rounds reusing the same `Path` objects produce identical output, so the slice hand-off does not drift or alias across sub-paths.
- **Self-stroke**: `p.AddStroke(&p, opts)` is explicitly supported, and it now runs while `normalize` has parked one of `p`'s own arrays in the scratch field and reassigned sub-path slices. Checked against the two-path reference with cusps in two sub-paths: correct.
- `go test ./vector/...`, `go vet ./vector/...` and `go test -race ./vector/` are all clean at `3a28dc2f1`.

## Nit: field naming

`opsScratch` should be `opsBuf`. This repository uses the `Buf` suffix uniformly for exactly this `slices.Grow(buf[:0], n)` reuse pattern (`verticesBuf`, `indicesBuf`, `uniformsBuf`, `pixelsBuf`, `controlsBuf`, `gamepadIDsBuf`), and `Scratch` appears nowhere else in the tree.

## Optional

`TestStrokeHugeQuadCusp` asserts by scanning the formatted path for `"+Inf"` and `"NaN"` substrings. Comparing against the expected operations directly would pin down the midpoint's actual value rather than only its finiteness, and would match how `TestStrokeQuadCusp` already asserts:

```
MoveTo(3e+38, 1)
LineTo(3.2e+38, 1)
LineTo(3e+38, 1)
```

Nothing else outstanding from my side, beyond the `f6c946f0b` squash you already mentioned.

---

kumagi:
Both done in 722da6f7d:

- Renamed `opsScratch` to `opsBuf`, following the `verticesBuf`/`controlsBuf` convention.
- `TestStrokeHugeQuadCusp` now asserts the normalized source operations exactly (`MoveTo(3e+38, 1)` / `LineTo(3.2e+38, 1)` / `LineTo(3e+38, 1)`) and additionally compares the stroke against the explicit `LineTo` out-and-back, as TestStrokeQuadCusp does, so the midpoint value is pinned down.

The `f6c946f0b` squash is on my list; I'll take care of it when this is ready to merge. Thanks a lot for the very thorough reviews!

hajimehoshi (approved):
LGTM

PR #3616, opened 2026-08-31
text/v2: detect Arabic Presentation Forms and RLM in the chunk fast path

kumagi:
None

bug

## What this PR does | solves
mayContainStrongRTLInFirstLine is an upper bound that must return true for every first line containing a strong-RTL rune so the bidi pass runs. Its lead-byte set missed two strong-RTL cases:

  * Arabic Presentation Forms A and B (U+FB50..U+FDFF, U+FE70..U+FEFF, class AL), whose UTF-8 lead byte is 0xEF.
  * RLM (U+200F, class R), whose 0xE2 0x80 prefix was only inspected for the LS/PS line separators.

Text made only of those runes took the pure-LTR fast path and every chunk got level 0, so the composition layer laid RTL sentences out left-to-right and AdvanceAt returned wrong caret positions.

Add 0xEF to the lead-byte set and detect RLM alongside the line separators.

Add TestChunks_Levels cases for both, which fail without this fix.

### DIFF
--- a/text/v2/internal/chunk/chunk.go
+++ b/text/v2/internal/chunk/chunk.go
@@ -319,12 +319,15 @@ func mayContainStrongRTLInFirstLine(text string) bool {
 	//     (Hebrew, Arabic, Syriac, Arabic Supplement, Thaana, NKo).
 	//   - 0xE0       — 3-byte UTF-8 covering U+0800..U+08FF
 	//     (Samaritan, Mandaic, Syriac Supplement, Arabic Extended-A/B).
+	//   - 0xEF       — 3-byte UTF-8 covering U+FB50..U+FDFF and
+	//     U+FE70..U+FEFF (Arabic Presentation Forms A and B, class AL).
 	//   - 0xF0       — 4-byte UTF-8 covering plane 1
 	//     (Mende Kikakui, Adlam, and other SMP RTL scripts).
 	//
-	// 0xF0 admits false positives for non-RTL SMP content (emoji,
-	// mathematical alphanumerics); those texts go through the bidi
-	// pass and chunk correctly anyway.
+	// 0xEF and 0xF0 admit false positives for non-RTL content (the
+	// private use area and CJK compatibility for 0xEF; emoji and
+	// mathematical alphanumerics for 0xF0); those texts go through the
+	// bidi pass and chunk correctly anyway.
 	//
 	// Line-break bytes are detected inline so the scan stops at the
 	// first line break instead of walking past it.
@@ -332,6 +335,8 @@ func mayContainStrongRTLInFirstLine(text string) bool {
 	//   NEL  U+0085 → 0xC2 0x85
 	//   LS   U+2028 → 0xE2 0x80 0xA8
 	//   PS   U+2029 → 0xE2 0x80 0xA9
+	// RLM (U+200F, class R) shares the 0xE2 0x80 prefix and is detected
+	// alongside the line separators.
 	for i := range len(text) {
 		b := text[i]
 		if b < 0x80 {
@@ -341,14 +346,19 @@ func mayContainStrongRTLInFirstLine(text string) bool {
 			}
 			continue
 		}
-		if (b >= 0xD6 && b <= 0xE0) || b == 0xF0 {
+		if (b >= 0xD6 && b <= 0xE0) || b == 0xEF || b == 0xF0 {
 			return true
 		}
 		if b == 0xC2 && i+1 < len(text) && text[i+1] == 0x85 {
 			return false
 		}
-		if b == 0xE2 && i+2 < len(text) && text[i+1] == 0x80 && (text[i+2] == 0xA8 || text[i+2] == 0xA9) {
-			return false
+		if b == 0xE2 && i+2 < len(text) && text[i+1] == 0x80 {
+			switch text[i+2] {
+			case 0x8F: // RLM U+200F, strong RTL
+				return true
+			case 0xA8, 0xA9: // LS U+2028, PS U+2029
+				return false
+			}
 		}
 	}
 	return false
--- a/text/v2/internal/chunk/chunk.go
+++ b/text/v2/internal/chunk/chunk.go
@@ -319,15 +319,20 @@ func mayContainStrongRTLInFirstLine(text string) bool {
 	//     (Hebrew, Arabic, Syriac, Arabic Supplement, Thaana, NKo).
 	//   - 0xE0       — 3-byte UTF-8 covering U+0800..U+08FF
 	//     (Samaritan, Mandaic, Syriac Supplement, Arabic Extended-A/B).
-	//   - 0xEF       — 3-byte UTF-8 covering U+FB50..U+FDFF and
-	//     U+FE70..U+FEFF (Arabic Presentation Forms A and B, class AL).
+	//   - 0xEF 0xAC..0xBB — 3-byte UTF-8 covering U+FB00..U+FEFF
+	//     (Hebrew Presentation Forms, class R, and Arabic Presentation
+	//     Forms A and B, class AL). The second byte is restricted to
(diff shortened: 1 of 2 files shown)


hajimehoshi:
Thanks! I confirmed both bugs are real: the two new cases fail on `main` and pass with this change, and `./text/v2/...`, `gofmt`, `go vet` and `internal/vettools` are all clean.

I also swept every codepoint (U+00A0..U+10FFFF) in a few surrounding contexts, comparing `AppendChunks` levels against `bidi.Paragraph.SegmentString` directly. Two things came out of that.

### 1. `0xEF` is coarse enough to push ordinary Japanese text off the fast path

`0xEF` covers all of U+F000..U+FFFF, which includes fullwidth punctuation (U+FF01 `！`, U+FF08 `（`) and halfwidth katakana (U+FF61..U+FF9F). Those appear in almost every Japanese sentence, so text containing them now takes the bidi pass:

|  | before | this PR | with the check below |
| --- | --- | --- | --- |
| fullwidth `！（）` | 11.6 µs | 31.9 µs | 12.2 µs |
| halfwidth katakana | 10.6 µs | 25.8 µs | 11.8 µs |

Restricting the second byte pins the match to exactly U+FB00..U+FEFF:

```go
if b == 0xEF && i+1 < len(text) && text[i+1] >= 0xAC && text[i+1] <= 0xBB {
	return true
}
```

I re-ran the same full-codepoint sweep against this variant: identical correctness, and the added tests still pass. This one seems worth doing — unlike the `0xF0` false positives the comment already accepts (emoji, mathematical alphanumerics), the `0xEF` ones are everyday CJK text.

### 2. Explicit embedding/override characters have the same problem

RLE (U+202B), RLO (U+202E) and RLI (U+2067) still take the pure-LTR fast path:

```
U+202E "a\u202eb"            bidi=[0 1]  chunks=[0]
U+202B "abc \u202bdef. ghi." bidi=[0 1]  chunks=[0]
U+2067 "abc \u2067def. ghi." bidi=[0 1]  chunks=[0]
```

These aren't strong-RTL by bidi class, so this is outside what your commit describes, and fixing them means renaming the function and reframing its doc (the predicate becomes "may resolve to a non-zero level"). I'll handle that separately — no need to touch it here. Noting it so it isn't lost.

### Smaller comment points

- `0xEF` also fixes Hebrew Presentation Forms (U+FB1D..U+FB4F, class R), which were broken in exactly the same way. Worth naming alongside the Arabic ones.
- The false-positive list mentions the private use area and CJK compatibility, but halfwidth/fullwidth forms are the case that actually costs.

### Commit message

The change is entirely inside `text/v2/internal/chunk`, and recent commits spell internal packages out in full (`internal/atlas:`, `internal/graphicsdriver/playstation5:`), so `text/v2/internal/chunk:` would fit better. (`61da914cf` used `text/v2:` because it also touched `gotextfacesource.go`.)

---

kumagi:
Thanks for the thorough review! Addressed in dfcdc426c:

- Applied the second-byte check (`0xEF 0xAC..0xBB`, exactly U+FB00..U+FEFF). On this machine `AppendChunks` on fullwidth punctuation text drops from ~1584 ns/op back to ~295 ns/op (halfwidth katakana ~1024 → ~246 ns/op), while the presentation-form texts still take the bidi pass as intended.
- Hebrew Presentation Forms (U+FB1D..U+FB4F, class R) are now named in the comment, and a `hebrew_presentation_forms` case was added to `TestChunks_Levels`.
- The false-positive comment now lists the remaining ones for the tightened range (Latin ligatures, CJK compatibility forms) and no longer mentions the PUA / halfwidth-fullwidth, which the second-byte check excludes.
- Point 2 (RLE/RLO/RLI) left untouched as you asked — looking forward to the separate PR since it renames the predicate.

On the commit prefix: the original commit was already pushed as `text/v2:`, so I kept it and named the follow-up `text/v2/internal/chunk:` to match the convention. If squashed, the combined message will carry the full scope.

`./text/v2/...`, `gofmt`, `go vet` are all clean.

hajimehoshi:
Verified the follow-up:

- Re-ran the full U+00A0..U+10FFFF sweep against `9682d5d47` — still only `202B 202E 2067` remaining, so the tightening lost no coverage.
- Fullwidth 12.6 µs, halfwidth 12.1 µs here, matching the pre-`0xEF` baseline. Consistent with your numbers.
- `gofmt`, `go vet`, `internal/vettools`, `go build ./...` and `./text/v2/...` are all clean.
- The `hebrew_presentation_forms` case is correct, and U+FB21 is indeed inside the kept range.

Four small things left.

### 1. The class annotation is wrong

The comment describes the halfwidth/fullwidth forms as `class ON/NSM`. The actual distribution over U+FF00..U+FFEF is:

```
L:177  ON:40  EN:10  ET:7  CS:4  ES:2
```

There is no NSM at all, and `L` dominates. The simplest fix is to drop the class annotation — it is not load-bearing for the reader.

### 2. Variation selectors are the false positive worth naming

U+FE0F (VS16) is `0xEF 0xB8 0x8F`, inside the kept range, so BMP-symbol emoji sequences leave the fast path:

| text | ns/op |
| --- | --- |
| `"Warning ⚠ and hearts ❤ here. "` | 1.75 µs |
| `"Warning ⚠️ and hearts ❤️ here. "` (with VS16) | ~20 µs |

This is not a regression against your first commit, but it does survive the tightening. Second byte `0xB8` is exactly U+FE00..U+FE3F (variation selectors, vertical forms, combining half marks) and contains no RTL, so it could be holed out — at the cost of turning one clean range into two. My mild preference is to keep the check simple and just name variation selectors in the false-positive list instead, but either is fine.

### 3. Comment style

The new text justifies the check against the bare-`0xEF` version it replaces ("pushing those off the fast path costs more than it is worth"). This repository tends to avoid explaining code by reference to the alternative it rejected, since a later reader does not have that version in front of them. Stating it positively — the range is exactly the presentation forms, and the common non-RTL forms above U+FF00 fall outside it — carries the same information.

### 4. Cherry-pick trailer

`(cherry picked from commit dfcdc426c…)` points at a commit that is not in this repository, so it will not mean anything to a later reader. Worth dropping.

None of this blocks merging; 1 and 4 are the ones I would actually like fixed.

---

kumagi:
Good catches — all three fixed in 30161697b:

1. The `class ON/NSM` annotation is gone. It was my own mislabel, not from the bidi data.
2. Variation selectors are now named first in the false-positive list (U+FE0F on emoji is exactly the case). Kept the single `0xAC..0xBB` range for simplicity, as you preferred — no hole at `0xB8`.
3. The second-byte comment now states the range positively (exactly the presentation forms; the common non-RTL forms at U+FF00 and above fall outside it) instead of arguing against the bare-`0xEF` version.

`text/v2` tests, `gofmt`, and `go vet` are clean.

hajimehoshi:
Verified `3984ad6de`: the comments read correctly now, and since the change is comment-only, the earlier correctness check still stands. Thanks for turning all three around so quickly.

No action needed on point 4 — I plan to squash on merge, so the cherry-pick trailers won't reach the final message. I should have said so earlier.

The RLE/RLO/RLI case stays with me as a separate change, as discussed.

hajimehoshi (approved):
LGTM

kumagi:
Thanks for the review and the approval. Nothing further to change on our side:

- Points 1–3 were already addressed in `3984ad6de`, and point 4 (cherry-pick trailers) is left as is since you squash on merge.
- The RLE/RLO/RLI case is yours — no changes here.

Re-verified on this checkout: a full U+00A0..U+10FFFF sweep (each codepoint embedded in `"abc %s def."` and `"%s. %s."`, comparing `AppendChunks` chunk levels against `bidi.Paragraph.SegmentString`) reports no divergence, and `./text/v2/...`, `gofmt` and `go vet` are clean.

PR #3617, opened 2026-08-31
ebiten: fix the index validation of DrawTriangles32 on 32-bit platforms

kumagi:
None

bug

## What this PR does | solves
The loop checked int(idx) >= len(vertices), but on a 32-bit platform int is 32 bits, so any index not less than 2^31 wrapped to a negative value and passed the check. The documented panic for an out-of-range index never happened on js/wasm and linux/386, and the bogus index was fed to the GPU driver instead.

Compare idx against uint32(len(vertices)) directly. len(vertices) is truncated to MaxVertexCount, which is at most MaxUint32, just above, so the conversion is safe.

Add regression tests for DrawTriangles32 and DrawTrianglesShader32 with an index of MaxUint32.

### DIFF
--- a/image_test.go
+++ b/image_test.go
@@ -4585,7 +4585,6 @@ func TestImageDrawTrianglesShaderWithGreaterIndexThanVerticesCount(t *testing.T)
 	dst.DrawTrianglesShader(vs, is, shader, nil)
 }
 
-// An index not less than 2^31 must panic even on a 32-bit platform, where int(idx) would otherwise wrap to a negative value.
 func TestImageDrawTriangles32WithGreaterIndexThanVerticesCount(t *testing.T) {
 	defer func() {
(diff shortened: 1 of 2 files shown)


image_test.go
```diff
@@ -4585,6 +4585,47 @@ func TestImageDrawTrianglesShaderWithGreaterIndexThanVerticesCount(t *testing.T)
 	dst.DrawTrianglesShader(vs, is, shader, nil)
 }
 
+// An index not less than 2^31 must panic even on a 32-bit platform, where int(idx) would otherwise wrap to a negative value.
```
hajimehoshi: Remove the comment

kumagi: Done.

hajimehoshi (approved):
LGTM

PR #3624, opened 2026-09-01
fix doc comments which diverge from the behavior

kumagi:
None

bug, (documentation?)

## What this PR does | solves

Read the doc comments against the code and corrected the ones which promise something else than what happens. No behavior changes, comments only.

 * audio: NewPlayer and NewPlayerF32 were documented to seek src for the current position and to return that error. They do not touch src: the position is asked from the source when the player starts using the stream, from Play or SetPosition through newTimeStream, and the error of that seek comes back from SetPosition, or becomes the Context error when it happens while starting to play. newPlayer only builds the struct.

 * audio/wav: DecodeF32 returns 32bit float but said that the source is converted into 16bit integer. The wording of mp3, opus and vorbis DecodeF32 was already correct.

 * vector: "If and only if the number of overlaps is not 0, the region is rendered" does not hold for FillRuleNonZero. The overlaps are counted four bits per direction, so 17 triangles of the same winding are counted as none and the region is not drawn at all, and FillRuleEvenOdd gets the parity of the same region wrong. The counter is what it is, so the limit is now documented instead of the code being changed: with the same winding, squares stacked 17 or 34 deep stayed transparent while 1 to 16 and 18 to 33 rendered.

 * ebiten: GamepadSDLID was documented to return an empty string on browsers and mobiles. The browser implementation builds a GUID the same way SDL does and Android derives one from the device id, so both return one; only the consoles have none.

 * ebiten: SetWindowSize was documented to round the scaled size down. windowSizeInGLFWPixels rounds to the nearest pixel.

 * ebiten: Monitor was documented to always give a monitor on desktops, and AppendMonitors to always append at least one. Both fail quietly when the display cannot be queried: Monitor returns nil and AppendMonitors appends nothing, and the methods of a nil *MonitorType panic, which the added note tells the caller to check for. WindowPosition, before the loop starts and when no initial position was set, reports the internal unset value (math.MinInt) rather than the position the window is going to get; it is now said out loud.

 * inpututil: the touch functions compare the touch ids of two consecutive ticks, so a tap which began and ended between two ticks is reported neither as just pressed nor as just released, and a new touch which reused the id of such one continues its duration and its position. That is how the sampling works, so the limitation is documented.

Verification: the diff changes comments only, checked by classifying every changed line; go build ./... and the tests of the touched packages pass.

### DIFF
--- a/audio/audio.go
+++ b/audio/audio.go
@@ -354,8 +354,10 @@ type Player struct {
 //
 // Note that the given src can't be shared with other Player objects.
 //
-// NewPlayer tries to call Seek of src to get the current position.
-// NewPlayer returns error when the Seek returns error.
+// NewPlayer does not read src: the current position of src is asked from it when
+// the player starts to use it, on [Player.Play] or [Player.SetPosition]. An error
+// of that Seek is returned by SetPosition, or reported as the Context error when
+// it happens while starting to play.
 //
 // A Player doesn't close src even if src implements io.Closer.
 // Closing the source is src owner's responsibility.
@@ -389,8 +391,10 @@ func (c *Context) NewPlayer(src io.Reader) (*Player, error) {
 //
 // Note that the given src can't be shared with other Player objects.
 //
-// NewPlayerF32 tries to call Seek of src to get the current position.
-// NewPlayerF32 returns error when the Seek returns error.
+// NewPlayerF32 does not read src: the current position of src is asked from it
+// when the player starts to use it, on [Player.Play] or [Player.SetPosition]. An
+// error of that Seek is returned by SetPosition, or reported as the Context error
+// when it happens while starting to play.
 //
 // A Player doesn't close src even if src implements io.Closer.
 // Closing the source is src owner's responsibility.
--- a/audio/audio.go
+++ b/audio/audio.go
@@ -354,10 +354,8 @@ type Player struct {
 //
 // Note that the given src can't be shared with other Player objects.
 //
-// NewPlayer does not read src: the current position of src is asked from it when
-// the player starts to use it, on [Player.Play] or [Player.SetPosition]. An error
-// of that Seek is returned by SetPosition, or reported as the Context error when
-// it happens while starting to play.
+// NewPlayer does not touch src. The player asks src for its current position when it
+// first uses src, at [Player.Play] or [Player.SetPosition].
 //
 // A Player doesn't close src even if src implements io.Closer.
 // Closing the source is src owner's responsibility.
@@ -391,10 +389,8 @@ func (c *Context) NewPlayer(src io.Reader) (*Player, error) {
 //
 // Note that the given src can't be shared with other Player objects.
 //
-// NewPlayerF32 does not read src: the current position of src is asked from it
-// when the player starts to use it, on [Player.Play] or [Player.SetPosition]. An
-// error of that Seek is returned by SetPosition, or reported as the Context error
-// when it happens while starting to play.
+// NewPlayerF32 does not touch src. The player asks src for its current position when it
+// first uses src, at [Player.Play] or [Player.SetPosition].
 //
 // A Player doesn't close src even if src implements io.Closer.
 // Closing the source is src owner's responsibility.
--- a/input.go
+++ b/input.go
@@ -146,8 +146,7 @@ type GamepadID = gamepad.ID
 // GamepadSDLID returns a string with the GUID generated in the same way as SDL.
 // To detect devices, see also the community project of gamepad devices database: https://github.com/gabomdq/SDL_GameControllerDB
 //
-// The GUID is emulated on browsers and mobiles, so it is not always what SDL itself reports for the
-// same device. GamepadSDLID returns an empty string on the consoles, where no such GUID exists.
+// GamepadSDLID returns an empty string on the consoles, where no such GUID exists.
 //
 // GamepadSDLID returns an empty string before the game starts.
 //
--- a/input.go
+++ b/input.go
@@ -146,7 +146,8 @@ type GamepadID = gamepad.ID
 // GamepadSDLID returns a string with the GUID generated in the same way as SDL.
 // To detect devices, see also the community project of gamepad devices database: https://github.com/gabomdq/SDL_GameControllerDB
 //
-// GamepadSDLID always returns an empty string on browsers and mobiles.
+// The GUID is emulated on browsers and mobiles, so it is not always what SDL itself reports for the
+// same device. GamepadSDLID returns an empty string on the consoles, where no such GUID exists.
 //
 // GamepadSDLID returns an empty string before the game starts.
 //
--- a/input.go
+++ b/input.go
@@ -146,7 +146,7 @@ type GamepadID = gamepad.ID
 // GamepadSDLID returns a string with the GUID generated in the same way as SDL.
 // To detect devices, see also the community project of gamepad devices database: https://github.com/gabomdq/SDL_GameControllerDB
 //
-// GamepadSDLID returns an empty string on the consoles, where no such GUID exists.
+// GamepadSDLID returns an empty string on consoles, where no such GUID exists.
 //
 // GamepadSDLID returns an empty string before the game starts.
 //
--- a/inpututil/inpututil.go
+++ b/inpututil/inpututil.go
@@ -563,6 +563,10 @@ func StandardGamepadButtonPressDuration(id ebiten.GamepadID, button ebiten.Stand
 // and returns the extended buffer.
 // Giving a slice that already has enough capacity works efficiently.
 //
+// The touches are sampled once per tick, so a touch which was created and released between two
+// ticks is not reported at all, and neither is it reported when the platform gives the new touch
+// an ID which is already being tracked.
+//
 // AppendJustPressedTouchIDs must be called in a game's Update, not Draw.
 //
 // AppendJustPressedTouchIDs is concurrent safe.
@@ -595,6 +599,9 @@ func JustPressedTouchIDs() []ebiten.TouchID {
 // and returns the extended buffer.
 // Giving a slice that already has enough capacity works efficiently.
 //
+// Like [AppendJustPressedTouchIDs], a touch which was created and released between two ticks is not
+// reported here either.
+//
 // AppendJustReleasedTouchIDs must be called in a game's Update, not Draw.
 //
 // AppendJustReleasedTouchIDs is concurrent safe.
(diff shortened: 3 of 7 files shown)


vector/fill.go
```diff
@@ -30,10 +30,19 @@ type FillRule int
 const (
 	// FillRuleNonZero means that triangles are rendered based on the non-zero rule.
 	// If and only if the number of overlaps is not 0, the region is rendered.
+	//
+	// The overlaps are counted with a limited precision: a region covered by 17
+	// or more triangles of the same winding can be treated as uncovered, for the
+	// counter wraps around. Keep the triangles of a path less overlapping than
+	// that, or fill them with separate paths.
```
hajimehoshi: I don't want concrete numbers here. The nibble width is an internal detail of the stencil buffer, and writing it into the godoc means the doc has to be revisited whenever the packing changes.

The number is also not right as written. The byte holds two 4-bit counters (front-facing adds 16, back-facing adds 1, and the readers take `abs((r >> 4) - (r & 0x0F))`), so the result depends on the winding direction. Filling N stacked identical squares as one path and reading the center pixel gives, for N = 1..40 (`O` = filled, `.` = transparent):

```
FillRuleNonZero, front-facing: OOOOOOOOOOOOOOO.........................
FillRuleNonZero, back-facing:  OOOOOOOOOOOOOOOO.OOOOOOOOOOOOOOOO.OOOOOO
FillRuleEvenOdd, front-facing: O.O.O.O.O.O.O.O.........................
FillRuleEvenOdd, back-facing:  O.O.O.O.O.O.O.OO.O.O.O.O.O.O.O..O.O.O.O.
```

The 17 and 34 you saw is the back-facing row. For the other winding it already breaks at 16 and stays broken, since the byte saturates at 255 rather than wrapping around.

Could you replace both of these paragraphs with one sentence on the `FillRule` type instead, with no numbers?

```go
// FillRule is the rule whether an overlapped region is rendered or not.
//
// The number of overlaps is counted with a limited precision, so a region with too many
// overlapping triangles can be rendered incorrectly.
type FillRule int
```

Please also drop "Keep the triangles of a path less overlapping than that, or fill them with separate paths". It relies on each path getting its own stencil buffer, which is internal, and separate paths are composited separately, so it does not give the same result with translucent colors.

input.go
```diff
@@ -146,7 +146,8 @@ type GamepadID = gamepad.ID
 // GamepadSDLID returns a string with the GUID generated in the same way as SDL.
 // To detect devices, see also the community project of gamepad devices database: https://github.com/gabomdq/SDL_GameControllerDB
 //
-// GamepadSDLID always returns an empty string on browsers and mobiles.
+// The GUID is emulated on browsers and mobiles, so it is not always what SDL itself reports for the
+// same device. GamepadSDLID returns an empty string on the consoles, where no such GUID exists.
```
hajimehoshi: The second sentence is the real fix and is correct: only the NintendoSDK and PlayStation 5 backends call `gamepads.add("", "")`. Please keep that one and drop the first sentence.

The GUID is synthesized by Ebitengine on every platform, not only on browsers and mobiles (see `gamepad_linux.go`, `gamepad_desktop_windows.go` and `gamepad_iokit_darwin.go`), so singling those two out draws a distinction that is not in the code. The js and Android paths also deliberately mirror SDL's own backends, and say so: they name `EMSCRIPTEN_JoystickGetDeviceGUID` and `Android_AddJoystick`. "not always what SDL itself reports" is the opposite of what they aim at, and contradicts the first line of this same doc comment.

audio/audio.go
```diff
@@ -354,8 +354,10 @@ type Player struct {
 //
 // Note that the given src can't be shared with other Player objects.
 //
-// NewPlayer tries to call Seek of src to get the current position.
-// NewPlayer returns error when the Seek returns error.
+// NewPlayer does not read src: the current position of src is asked from it when
+// the player starts to use it, on [Player.Play] or [Player.SetPosition]. An error
+// of that Seek is returned by SetPosition, or reported as the Context error when
+// it happens while starting to play.
```
hajimehoshi: "that Seek" has no antecedent: the clause before it says the position "is asked from it" and never mentions Seek.

"reported as the Context error" is not actionable either, as there is no exported `Context.Err`. The error comes out of `RunGame` through the before-update hook.

This is also more mechanism than the contract needs. Something like this is enough:

```go
// NewPlayer does not touch src. The player asks src for its current position when it
// first uses src, at [Player.Play] or [Player.SetPosition].
```

If you keep a longer form, note that the current sentence promises slightly more than the code does: `SetPosition(0)` before the audio device exists returns without seeking, and `newTimeStream` ignores `errors.ErrUnsupported` from the Seek.

audio/audio.go
```diff
@@ -389,8 +391,10 @@ func (c *Context) NewPlayer(src io.Reader) (*Player, error) {
 //
 // Note that the given src can't be shared with other Player objects.
 //
-// NewPlayerF32 tries to call Seek of src to get the current position.
-// NewPlayerF32 returns error when the Seek returns error.
+// NewPlayerF32 does not read src: the current position of src is asked from it
+// when the player starts to use it, on [Player.Play] or [Player.SetPosition]. An
+// error of that Seek is returned by SetPosition, or reported as the Context error
+// when it happens while starting to play.
```
hajimehoshi: Same as `NewPlayer` above.

monitor.go
```diff
@@ -51,6 +51,10 @@ func (m *MonitorType) Size() (int, int) {
 }
 
 // Monitor returns the current monitor.
+//
+// Monitor returns nil when no monitor is available, which happens when the display cannot be
+// queried: on a machine without any display, or when the display initialization failed. The
+// methods of the returned type panic for a nil *MonitorType, so check the result before using it.
```
hajimehoshi: The list of causes is incomplete: `Monitor` also returns nil after the game is terminated, and when the current monitor cannot be determined, not only when the display cannot be queried.

I would rather not enumerate them. Could you keep this to a short note that `Monitor` can return nil?

I am also not sure about telling callers to check the result. On desktop the nil comes from `initializeGLFW` failing with "no monitor was found", which is stored as the UI error and fails `RunGame` anyway, so the panic only reaches code that calls `Monitor` before the main loop. Making the `MonitorType` methods nil-safe might be the better answer, as a separate change.

monitor.go
```diff
@@ -65,7 +69,7 @@ func SetMonitor(monitor *MonitorType) {
 }
 
 // AppendMonitors returns the monitors reported by the system.
-// On desktop platforms, there will always be at least one monitor appended and the first monitor in the slice will be the primary monitor.
+// On desktop platforms with a working display, there will always be at least one monitor appended and the first monitor in the slice will be the primary monitor. Without a display, nothing is appended.
```
hajimehoshi: This line is now about 190 columns, while the note you added to `Monitor` above wraps around 100. Please wrap it.

window.go
```diff
@@ -163,6 +163,8 @@ func SetWindowIcon(iconImages []image.Image) {
 // The unit is device-independent pixels.
 //
 // If the main loop does not start yet, WindowPosition returns the initial window position set by [SetWindowPosition].
+// When no initial position was set, WindowPosition returns (math.MinInt, math.MinInt) instead of the
+// position the window is going to get, which is decided when the window is created.
```
hajimehoshi: `math.MinInt` should be a doc link: `[math.MinInt]`.

Writing the sentinel down also makes it part of the contract, and I would like to settle whether `WindowPosition` should keep returning it before that happens. Returning (0, 0) before the window is created may be the better behavior.

inpututil/inpututil.go
```diff
@@ -563,6 +563,10 @@ func StandardGamepadButtonPressDuration(id ebiten.GamepadID, button ebiten.Stand
 // and returns the extended buffer.
 // Giving a slice that already has enough capacity works efficiently.
 //
+// The touches are sampled once per tick, so a touch which was created and released between two
+// ticks is not reported at all, and neither is it reported when the platform gives the new touch
+// an ID which is already being tracked.
```
hajimehoshi: The behavior is right: `Update` snapshots the touches once per tick, increments `duration` for the IDs that survive and deletes the ones that are gone.

The sentence is hard to follow, though. The second half keeps "a touch which was created and released between two ticks" as its subject while it is actually about a different, newly created touch, so it reads as if the same touch is being excluded twice. Could you split it into two sentences?

hajimehoshi (commented):
Thanks for going through these. I checked each claim against the code, and most of them hold up.

`audio/wav`'s `DecodeF32` and the `SetWindowSize` rounding are both correct as written: `windowSizeInGLFWPixels` does use `math.Round`.

The `vector` and `GamepadSDLID` ones need changes before this can go in, and there are some smaller wording points inline.

(Reviewed by Claude (Claude Code) on behalf of @hajimehoshi.)

inpututil/inpututil.go
```diff
@@ -595,6 +599,9 @@ func JustPressedTouchIDs() []ebiten.TouchID {
 // and returns the extended buffer.
 // Giving a slice that already has enough capacity works efficiently.
 //
+// Like [AppendJustPressedTouchIDs], a touch which was created and released between two ticks is not
+// reported here either.
```
hajimehoshi: This mentions only the touch that lives entirely between two ticks, while `IsTouchJustReleased` below mentions the reused ID as well. Both apply here, so please make the two consistent.

inpututil/inpututil.go
```diff
@@ -635,6 +645,9 @@ func IsTouchJustReleased(id ebiten.TouchID) bool {
 
 // TouchPressDuration returns how long the touch remains in ticks (Update).
 //
+// The duration is counted per touch ID, not per touch: a new touch which reused the ID of a touch
+// released between two ticks continues the duration of the previous touch.
```
hajimehoshi, line 650: This one is accurate.

The position functions have the same property (a reused ID continues the previous touch's position), as your commit message says, but they get no note. Either add one there too or leave it out here, whichever you prefer.

hajimehoshi:
Thanks, this addresses everything. I checked the new state: `gofmt` and `go vet` are clean, and the notes added to `TouchPositionInPreviousTick` and `TouchPositionFInPreviousTick` are accurate, as `TouchPositionFInPreviousTick` reads `prevTouchStates`, so a reused ID does return the previous touch's last position.

Don't worry about the commit message or the description: this will be squashed with our own message on merge.

Two small wording things left.

`monitor.go`: "there will always be at least one monitor appended ... unless no monitor is available" reads as a contradiction, and the trailing clause qualifies the primary-monitor half as well. The "always at least one" claim is the wrong part, so I would drop it rather than qualify it:

```go
// AppendMonitors returns the monitors reported by the system.
// On desktop platforms, the first monitor in the slice will be the primary monitor.
// Nothing is appended when no monitor is available.
// Any monitors added or removed will show up with subsequent calls to this function.
```

`input.go`: "on the consoles" - this file writes "mobiles and consoles" without the article elsewhere (`IsCapsLockOn` and `IsNumLockOn`), so "on consoles" would match.

(Reviewed by Claude (Claude Code) on behalf of @hajimehoshi.)

kumagi:
Thanks! Both wording points are now in:

- `monitor.go`: dropped the "always at least one monitor" claim and used your suggested `AppendMonitors` wording.
- `input.go`: "on the consoles" is now "on consoles" to match the rest of the file.

gofmt and go vet remain clean and the tests of the touched packages still pass.

hajimehoshi (approved):
LGTM

PR #3625, opened 2026-09-01
internal/shader: reject arithmetic on boolean operands

kumagi:
None

bug

## What this PR does | solves
TypeFromBinaryOp accepted Add/Sub/ComponentWiseMul/MatrixMul/Div when both operands had the same type, including Bool. Constant folding then called go/constant.BinaryOp (or Sign for division) on bool values, which panics ("invalid binary operation true + true"), so a user shader containing e.g. true + true crashed the whole process instead of reporting a compile error. The non-constant form (a + b for bool vars) did not panic but emitted invalid GLSL/HLSL/MSL.

Reject arithmetic operations whose operand is a boolean type or an untyped boolean constant, so they surface as a normal type error.

Add TestCompileBoolArithmetic, which panics without this fix.

### DIFF
--- a/internal/shader/shader_test.go
+++ b/internal/shader/shader_test.go
@@ -350,3 +350,51 @@ func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
 		}()
 	}
 }
+
+func TestCompileBoolArithmetic(t *testing.T) {
+	srcs := []string{
+		`package main
+
+func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
+	return vec4(float(true + true))
+}`,
+		`package main
+
+func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
+	return vec4(float(true - true))
+}`,
+		`package main
+
+func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
+	return vec4(float(true * true))
+}`,
+		`package main
+
+func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
+	return vec4(float(true / true))
+}`,
+		`package main
+
+func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
+	a := true
+	b := true
+	c := a + b
+	if c {
+		return vec4(1)
+	}
+	return vec4(0)
+}`,
+	}
+	for _, src := range srcs {
+		func() {
+			defer func() {
+				if r := recover(); r != nil {
+					t.Errorf("Compile must not panic for bool arithmetic, but panicked: %v", r)
+				}
+			}()
+			if _, err := shader.Compile([]byte(src), "Vertex", "Fragment", 0); err == nil {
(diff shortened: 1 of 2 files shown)


internal/shaderir/check.go
```diff
@@ -82,6 +82,20 @@ func ResolveUntypedConstsForBinaryOp(op Op, lhs, rhs constant.Value, lhst, rhst
 }
 
 func TypeFromBinaryOp(op Op, lhst, rhst Type, lhsConst, rhsConst constant.Value) (Type, bool) {
+	// Arithmetic operations are not defined for booleans. Reject them before
+	// the untyped-constant handling so that neither constant folding (which
+	// panics on a bool operand) nor code generation (which would emit invalid
+	// GLSL/HLSL/MSL for a bool arithmetic expression) is ever reached.
```
hajimehoshi: Remove the comments

internal/shader/shader_test.go
```diff
@@ -301,3 +301,50 @@ func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
 		t.Errorf("Compile must return an error for a huge constant shift, but got nil")
 	}
 }
+
+// Arithmetic on booleans is undefined and must be a compile error, not a panic
+// in constant folding nor invalid GLSL/HLSL/MSL.
```

kumagi:
Thanks for the review! I removed the comments pointed out in both files (the bool-arithmetic check in TypeFromBinaryOp and the test). PTAL.

internal/shaderir/check.go
kumagi: done.

internal/shader/shader_test.go
kumagi: done

kumagi:
Both review comments are addressed at the current head: the explanatory comments have been removed from internal/shaderir/check.go and internal/shader/shader_test.go, and only the bool-arithmetic rejection in TypeFromBinaryOp plus TestCompileBoolArithmetic remain. Tests pass. PTAL.

hajimehoshi:
The variable-operand regression case can pass even without the typed-boolean guard: `float(c)` independently rejects boolean `c`, so `shader.Compile` still returns an error when `a + b` is incorrectly accepted. I verified that removing the `lhst.Main == Bool || rhst.Main == Bool` check via a temporary Go overlay leaves `TestCompileBoolArithmetic` passing.

Could this case use `c` in an `if` condition with valid `vec4` returns instead? That would ensure the arithmetic itself is what causes compilation to fail.

The production fix looks correct, and `go test ./internal/shader ./internal/shaderir/...` passes at c623e903bbb2e8facc86ddd460f21e6527a1f093.

Posted by Codex on behalf of @hajimehoshi.

hajimehoshi:
Resolve conflicts

hajimehoshi (approved):
LGTM

PR #3626, opened 2026-09-01
internal/shader: fix a panic on a var count mismatch in multiple-value declarations

kumagi:
None

bug

## What this PR does | solves
When the number of values on the right of a multiple-value var declaration differed from the number of names on the left, the count check ran only for the first name (i == 0), only when no explicit type was given (t.Main == None), and used `continue`, which skipped just that first iteration. The loop then proceeded to the remaining names and indexed inittypes[i] / initexprs[i] past their end, panicking with "index out of range". This affected e.g. "var a, b, c = f()" where f returns two values, and "var a, b = 1".

Validate the value count against the name count once, right after parsing the initializer, regardless of the declared type, and abort the declaration with an error instead of continuing.

Add TestCompileVarCountMismatch, which panics without this fix.

### DIFF
--- a/internal/shader/shader.go
+++ b/internal/shader/shader.go
@@ -630,11 +630,6 @@ func (s *compileState) parseVariable(block *block, fname string, vs *ast.ValueSp
 					}
 				}
 
-				// The number of values on the right must match the number of
-				// names on the left, whether or not an explicit type is given.
-				// Abort the whole declaration here: the previous code used
-				// `continue`, which only skipped the first iteration and let a
-				// later iteration index inittypes/initexprs out of range.
 				if len(initexprs) != len(vs.Names) || len(inittypes) != len(vs.Names) {
 					s.addError(vs.Pos(), "the numbers of lhs and rhs don't match")
 					return nil, nil, nil, false
--- a/internal/shader/shader.go
+++ b/internal/shader/shader.go
@@ -628,10 +628,16 @@ func (s *compileState) parseVariable(block *block, fname string, vs *ast.ValueSp
 					if ok {
 						inittypes = ts
 					}
-					if len(ts) != len(vs.Names) {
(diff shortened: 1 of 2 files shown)


internal/shader/shader.go
```diff
@@ -628,10 +628,16 @@ func (s *compileState) parseVariable(block *block, fname string, vs *ast.ValueSp
...
-						s.addError(vs.Pos(), "the numbers of lhs and rhs don't match")
-						continue
-					}
+				}
+
+				// The number of values on the right must match the number of
+				// names on the left, whether or not an explicit type is given.
+				// Abort the whole declaration here: the previous code used
+				// `continue`, which only skipped the first iteration and let a
+				// later iteration index inittypes/initexprs out of range.
```
hajimehoshi: Reduce the comments

internal/shader/shader_test.go
```diff
@@ -301,3 +301,55 @@ func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
 		t.Errorf("Compile must return an error for a huge constant shift, but got nil")
 	}
 }
+
+// A multiple-value variable declaration whose right-hand side yields a number
+// of values different from the number of names must be a compile error, not a
+// panic from indexing the value types out of range.
```

internal/shader/shader.go
kumagi: Done. I removed the comment block above the count check.

internal/shader/shader_test.go
kumagi: Done. I removed the comment above TestCompileVarCountMismatch.

kumagi:
Thank you for the review! I addressed both comments by removing the explanatory comments from internal/shader/shader.go and internal/shader/shader_test.go. The tests still pass (go test ./internal/shader/).

hajimehoshi (approved):
LGTM

PR #3627, opened 2026-09-01
internal/shader: reject the blank identifier used as a value

kumagi:
None

bug

## What this PR does | solves
In a context where a local variable is not marked as used (a constant declaration or an assignment left-hand side), parsing "_" returned an expression with Type Blank and a nil Const. When such an expression was used as a value, e.g. as an argument of min/max/clamp or an operand of a binary operator, constant folding dereferenced the nil Const and panicked with a nil pointer dereference, crashing the compiler on user input.

Reject a blank identifier used as a value in call arguments and binary operands, reporting "cannot use _ as value" instead.

Add TestCompileBlankAsValue, which panics without this fix.

### DIFF
--- a/internal/shader/expr.go
+++ b/internal/shader/expr.go
@@ -98,8 +98,6 @@ func (cs *compileState) parseExpr(block *block, fname string, expr ast.Expr, mar
 		}
 		rhst := ts[0]
 
-		// A blank identifier has no value, so it cannot be an operand. Reject it
-		// here; otherwise constant folding dereferences its nil constant.
 		if lhs[0].Type == shaderir.Blank || rhs[0].Type == shaderir.Blank {
 			cs.addError(e.Pos(), "cannot use _ as value")
 			return nil, nil, nil, false
--- a/internal/shader/expr.go
+++ b/internal/shader/expr.go
@@ -98,6 +98,13 @@ func (cs *compileState) parseExpr(block *block, fname string, expr ast.Expr, mar
 		}
 		rhst := ts[0]
 
+		// A blank identifier has no value, so it cannot be an operand. Reject it
+		// here; otherwise constant folding dereferences its nil constant.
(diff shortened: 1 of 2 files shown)


internal/shader/shader_test.go
```diff
@@ -301,3 +301,35 @@ func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
 		t.Errorf("Compile must return an error for a huge constant shift, but got nil")
 	}
 }
+
+// The blank identifier _ has no value, so using it as a value (an argument or
+// an operand) must be a compile error, not a nil-pointer panic in constant
+// folding.
```

internal/shader/expr.go
```diff
@@ -98,6 +98,13 @@ func (cs *compileState) parseExpr(block *block, fname string, expr ast.Expr, mar
 		}
 		rhst := ts[0]
 
+		// A blank identifier has no value, so it cannot be an operand. Reject it
+		// here; otherwise constant folding dereferences its nil constant.
```
hajimehoshi: Remove the comments

kumagi:
Thanks for the review! I've removed the comments as requested in both internal/shader/expr.go and internal/shader/shader_test.go, and the tests still pass. PTAL.

internal/shader/shader_test.go
kumagi: Done. The comment above TestCompileBlankAsValue has been removed in the latest commit.

internal/shader/expr.go
kumagi: Done. The comment before the blank-identifier operand check has been removed in the latest commit.

hajimehoshi:
Why did you close this?

kumagi:
Sorry, it's my mistake, I'll open yet another one, sorry for back and forth.

kumagi:
opened as #3645

PR #3628, opened 2026-09-01
internal/shaderir: fix large float constants overflowing int64 in literals

kumagi:
None

bug

## What this PR does | solves

constantToNumberLiteral emitted an integer-valued float constant with fmt.Sprintf("%d.0", int64(i)) whenever math.Floor(x) == x. Every float
>= 2^52 is integral, so for |x| >= 2^63 the int64 conversion overflowed
and produced a wildly wrong literal: 1e19 became -9223372036854775808.0 in GLSL, HLSL, and MSL alike, silently miscompiling the shader.

Only take the integer-literal shortcut when the value is finite and fits in int64, falling back to the already-correct scientific format otherwise.

Add TestCompileLargeFloatConstant, which fails without this fix.


internal/shader/shader_test.go
```diff
@@ -301,3 +301,24 @@ func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
 		t.Errorf("Compile must return an error for a huge constant shift, but got nil")
 	}
 }
+
+// A large integral float constant must not be emitted as an int64 literal that
+// overflows. Before the fix, 1e19 was printed as -9223372036854775808.0.
```

kumagi:
Thanks for the review! I've removed the comments as requested (both the one above the new test and the explanatory ones in the glsl, hlsl, and msl generators). The fix itself is unchanged: the integer-literal shortcut is now only taken when the value fits in int64 (|x| < 2^63), otherwise the scientific-notation format is used, and TestCompileLargeFloatConstant still fails without the guard and passes with it. PTAL.

internal/shader/shader_test.go
kumagi: Done — the comments are removed, including the similar ones in the glsl, hlsl, and msl generators. The test itself is unchanged and passes with the fix.

hajimehoshi (approved):
LGTM

hajimehoshi:
Rebase the change

hajimehoshi:
ping

internal/shader/shader_test.go
kumagi: Still done after the rebase: the test has no comments, and the conflict resolution only kept the new tests from main alongside TestCompileLargeFloatConstant.

kumagi:
Rebased onto the latest main and resolved the conflict in internal/shader/shader_test.go. The change itself is unchanged, and go test ./internal/shader/... ./internal/shaderir/... passes.

hajimehoshi:
This still conflicts

internal/shader/shader_test.go
kumagi: Still done after the rebase onto the latest upstream/main: the test has no comments and the generators are comment-free. The branch is now a single clean commit on top of main.

kumagi:
Rebased onto the latest upstream/main as a single commit, so the conflict is gone. The change is unchanged: the integer-literal shortcut is taken only when the value is finite and fits in int64, and TestCompileLargeFloatConstant still fails without the guard and passes with it. go build ./... and go test ./internal/shader/... ./internal/shaderir/... pass.

PR #3629, opened 2026-09-01
internal/shader: build array comparisons as a balanced tree

kumagi:
None

bug

## What this PR does | solves
arrayComparisonExpr combined the per-element comparisons into a left-nested chain: ((((e0 && e1) && e2) && ...)). The backends render a binary node by recursively formatting and copying both operands, so this made compiling a comparison of a large array O(n^2) in the element count (a [1000]int comparison took seconds, and larger arrays hang), and also recursed n deep.

Collect the element comparisons and combine them pairwise into a balanced tree. && and || are associative, so the result is unchanged; the depth becomes logarithmic and the work O(n log n).

Add TestCompileArrayComparisonIsBalanced, which measures the nesting depth of the conjunctions and fails on the left-nested chain.

### DIFF
--- a/internal/shader/shader_test.go
+++ b/internal/shader/shader_test.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/hajimehoshi/ebiten/v2/internal/graphics"
 	"github.com/hajimehoshi/ebiten/v2/internal/shader"
+	"github.com/hajimehoshi/ebiten/v2/internal/shaderir"
 	"github.com/hajimehoshi/ebiten/v2/internal/shaderir/glsl"
 	"github.com/hajimehoshi/ebiten/v2/internal/shaderir/hlsl"
 	"github.com/hajimehoshi/ebiten/v2/internal/shaderir/msl"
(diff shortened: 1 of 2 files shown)


internal/shader/shader_test.go
```diff
@@ -301,3 +302,66 @@ func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
...
+		for _, sub := range e.Exprs {
+			if c := joinDepth(sub); c > child {
+				child = c
+			}
+		}
+		return d + child
+	}
+	var blockDepth func(b *shaderir.Block) int
+	blockDepth = func(b *shaderir.Block) int {
+		m := 0
```
hajimehoshi: var m int

internal/shader/shader_test.go
```diff
@@ -301,3 +302,66 @@ func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
...
+}`, n, n))
+	s, err := shader.Compile(src, "Vertex", "Fragment", 0)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Max nesting depth of && / || binary nodes in the fragment function.
+	var joinDepth func(e shaderir.Expr) int
+	joinDepth = func(e shaderir.Expr) int {
+		d := 0
```
hajimehoshi: var d int

internal/shader/shader_test.go
```diff
@@ -301,3 +302,66 @@ func Fragment(position vec4, texCoord vec2, color vec4) vec4 {
 		t.Errorf("Compile must return an error for a huge constant shift, but got nil")
 	}
 }
+
+// Comparing large arrays must build a balanced tree of element comparisons, not
+// a left-nested chain. A chain makes the backends, which format each binary
+// node by copying its operands, take O(n^2) time and produces a nesting depth
+// proportional to n; a balanced tree keeps the depth logarithmic.
```
hajimehoshi: Remove the comment

internal/shader/shader_test.go
kumagi: Done. Changed `m := 0` to `var m int`.

internal/shader/shader_test.go
kumagi: Done. Changed `d := 0` to `var d int`, and also `child := 0` to `var child int` for consistency.

internal/shader/shader_test.go
kumagi: Done. The comment is removed.

kumagi:
Thank you for the review! I addressed all the comments in the latest revision: the comment above TestCompileArrayComparisonIsBalanced is removed, and the zero assignments are replaced with var declarations. PTAL.

hajimehoshi:
I decided NOT to support comparing arrays due to costs.

Issue #1444, opened 2020-12-20
clock: 300 TPS doesn't work well on some machines (MacBook Pro 2020?)


hajimehoshi:
It looks like the timer precision might be bad on some machines, and then the count value can be lower than expected.

To fix this, multiply the count by some value (e.g., x5 for 300 TPS) after calculating the count assuming TPS was 60.

CC @wasedaigo

Issue #87, opened 2015-01-14
sprites: Add UI to change the number of the sprites


hajimehoshi:

hajimehoshi:
Done

Issue #86, opened 2015-01-14
Add an item about gamepads to readme


hajimehoshi:

hajimehoshi:
Done

Issue #28, opened 2014-12-20
Use premultiplied alpha


hajimehoshi:
See also: http://en.wikipedia.org/wiki/Alpha_compositing

PR #3634, opened 2026-09-03
docs: fix typos and documentation inconsistencies

kumagi:
None

bug

## What this PR does | solves
Fix 10 documentation issues.
- Remove incorrect IsFocused doc claiming dependency on IsRunnableOnUnfocused
- Document that MultiFace requires faces to be configured before creation
- Fix isPointCloseToSegment doc (segment -> line)
- Fix WindowResizingModeEnabled -> Disabled in doc
- Fix 3 typos in run.go (of->from, Use->use, in->at)
- Fix 4 typos in audio/audio.go (not or->not, singed->signed x2, func name)
- Fix clock comment referencing frame -> minCatchUpTime
- Fix typo in shaderir/check.go (lhst -> rhst)
- Update DrawTriangles doc to match v2 API (Color Y->Color A, remove v1 notes)

### DIFF
--- a/text/v2/multi.go
+++ b/text/v2/multi.go
@@ -35,6 +35,9 @@ type MultiFace struct {
 	// splitTextCache memoizes per-text chunk decomposition. The decomposition
 	// depends only on the faces' hasGlyph results, which are stable for the
 	// lifetime of a face, so cached entries never need invalidation.
+	// NOTE: This assumes that faces' hasGlyph results do not change after
+	// creation. For example, [LimitedFace.AddUnicodeRange] can change the
+	// results, so you must configure all ranges before creating a [MultiFace].
 	splitTextCache *cache[string, []textChunk]
 }
 
--- a/text/v2/multi.go
+++ b/text/v2/multi.go
@@ -16,7 +16,6 @@ package text
 
 import (
 	"errors"
-	"slices"
 	"unicode/utf8"
 
 	"github.com/hajimehoshi/ebiten/v2/text/v2/internal/textutil"
@@ -30,19 +29,21 @@ var _ Face = (*MultiFace)(nil)
 //
 // There is a known issue: if the writing directions of the faces don't agree, the rendering result might be messed up.
 // [NewMultiFace] rejects such faces, but a face's direction can be changed after the creation (e.g. [GoTextFace.Direction]).
+//
+// There is another known issue: the chunk decomposition of a text is memoized, and a face's glyph
+// availability can change even after the face's creation (e.g. by [LimitedFace.AddUnicodeRange] or
+// by an assignment to [GoTextFace.Source]), so such a change after a MultiFace's creation might not
+// be reflected for the texts whose chunks are already memoized.
+// Configure all the faces before creating a MultiFace to avoid this issue.
 type MultiFace struct {
 	faces []Face
 
-	// splitTextCache memoizes per-text chunk decomposition. The decomposition depends only on
-	// the faces' hasGlyph results, which can change even after the faces' creation
-	// (e.g. by [LimitedFace.AddUnicodeRange] or by an assignment to [GoTextFace.Source]),
-	// so the faces' glyph revision is a part of the cache key.
-	splitTextCache *cache[splitTextCacheKey, []textChunk]
-}
-
-type splitTextCacheKey struct {
-	text     string
-	revision uint64
+	// splitTextCache memoizes per-text chunk decomposition. The decomposition
+	// depends on the faces' hasGlyph results. The results can change even after
+	// the faces' creation (e.g. by [LimitedFace.AddUnicodeRange] or by an
+	// assignment to [GoTextFace.Source]), so a cached entry can be stale.
+	// See the MultiFace doc for the workaround.
+	splitTextCache *cache[string, []textChunk]
 }
 
 // NewMultiFace creates a new MultiFace from the given faces.
@@ -60,10 +61,12 @@ func NewMultiFace(faces ...Face) (*MultiFace, error) {
 		}
 	}
 
-	return &MultiFace{
-		faces:          slices.Clone(faces),
-		splitTextCache: newCache[splitTextCacheKey, []textChunk](32),
-	}, nil
+	m := &MultiFace{
+		splitTextCache: newCache[string, []textChunk](32),
+	}
+	m.faces = make([]Face, len(faces))
+	copy(m.faces, faces)
+	return m, nil
 }
 
 // Metrics implements Face.
@@ -174,28 +177,10 @@ type textChunk struct {
 	faceIndex      int
 }
 
-// splitText splits the text into chunks, each of which is rendered by one face.
-//
-// A face's glyph availability can change even after the face's creation
-// (e.g. by [LimitedFace.AddUnicodeRange] or by an assignment to [GoTextFace.Source]),
-// so the chunks are cached with the faces' glyph revision.
 func (m *MultiFace) splitText(text string) []textChunk {
-	key := splitTextCacheKey{
-		text:     text,
-		revision: m.glyphRevision(),
-	}
-	return m.splitTextCache.getOrCreate(key, func() ([]textChunk, bool) {
-		return m.doSplitText(key.text), true
+	return m.splitTextCache.getOrCreate(text, func() ([]textChunk, bool) {
+		return m.doSplitText(text), true
 	})
-}
-
-// glyphRevision implements Face.
-func (m *MultiFace) glyphRevision() uint64 {
-	var r uint64
-	for _, f := range m.faces {
-		r = combineGlyphRevisions(r, f.glyphRevision())
-	}
-	return r
 }
 
 func (m *MultiFace) doSplitText(text string) []textChunk {
--- a/text/v2/multi.go
+++ b/text/v2/multi.go
@@ -31,14 +31,6 @@ var _ Face = (*MultiFace)(nil)
 // [NewMultiFace] rejects such faces, but a face's direction can be changed after the creation (e.g. [GoTextFace.Direction]).
 type MultiFace struct {
 	faces []Face
-
-	// splitTextCache memoizes per-text chunk decomposition. The decomposition
-	// depends only on the faces' hasGlyph results, which are stable for the
-	// lifetime of a face, so cached entries never need invalidation.
-	// NOTE: This assumes that faces' hasGlyph results do not change after
-	// creation. For example, [LimitedFace.AddUnicodeRange] can change the
-	// results, so you must configure all ranges before creating a [MultiFace].
-	splitTextCache *cache[string, []textChunk]
 }
 
 // NewMultiFace creates a new MultiFace from the given faces.
@@ -56,9 +48,7 @@ func NewMultiFace(faces ...Face) (*MultiFace, error) {
 		}
 	}
 
-	m := &MultiFace{
-		splitTextCache: newCache[string, []textChunk](32),
-	}
+	m := &MultiFace{}
 	m.faces = make([]Face, len(faces))
 	copy(m.faces, faces)
 	return m, nil
@@ -172,13 +162,12 @@ type textChunk struct {
 	faceIndex      int
 }
 
+// splitText splits the text into chunks, each of which is rendered by one face.
+//
+// The result is not memoized on purpose: a face's hasGlyph result can change after the face's
+// creation (e.g. [LimitedFace.AddUnicodeRange] or an assignment to [GoTextFace.Source]),
+// so a memoized chunk decomposition could be stale.
 func (m *MultiFace) splitText(text string) []textChunk {
-	return m.splitTextCache.getOrCreate(text, func() ([]textChunk, bool) {
-		return m.doSplitText(text), true
-	})
-}
-
-func (m *MultiFace) doSplitText(text string) []textChunk {
 	var chunks []textChunk
 	var chunk textChunk
 	for i, r := range text {
--- a/text/v2/multi.go
+++ b/text/v2/multi.go
@@ -32,8 +32,9 @@ var _ Face = (*MultiFace)(nil)
 //
 // There is another known issue: the chunk decomposition of a text is memoized, and a face's glyph
 // availability can change even after the face's creation (e.g. by [LimitedFace.AddUnicodeRange] or
-// by an assignment to [GoTextFace.Source]), so such a change after a MultiFace's creation might not
-// be reflected for the texts whose chunks are already memoized.
+// by an assignment to [GoTextFace.Source]), so such a change after a MultiFace's creation is not
+// reflected for the texts whose chunks are already memoized: a memoized entry used every tick never
+// ages out of the cache.
 // Configure all the faces before creating a MultiFace to avoid this issue.
 type MultiFace struct {
 	faces []Face
--- a/text/v2/multi.go
+++ b/text/v2/multi.go
@@ -30,12 +30,9 @@ var _ Face = (*MultiFace)(nil)
 // There is a known issue: if the writing directions of the faces don't agree, the rendering result might be messed up.
 // [NewMultiFace] rejects such faces, but a face's direction can be changed after the creation (e.g. [GoTextFace.Direction]).
 //
-// There is another known issue: the chunk decomposition of a text is memoized, and a face's glyph
-// availability can change even after the face's creation (e.g. by [LimitedFace.AddUnicodeRange] or
-// by an assignment to [GoTextFace.Source]), so such a change after a MultiFace's creation is not
-// reflected for the texts whose chunks are already memoized: a memoized entry used every tick never
-// ages out of the cache.
-// Configure all the faces before creating a MultiFace to avoid this issue.
+// There is another known issue: a change to a face's glyph availability after a MultiFace's creation
+// (e.g. by [LimitedFace.AddUnicodeRange] or by an assignment to [GoTextFace.Source]) is not reflected.
+// Configure all the faces before creating a MultiFace.
 type MultiFace struct {
 	faces []Face
 
@@ -43,6 +40,8 @@ type MultiFace struct {
 	// depends on the faces' hasGlyph results. The results can change even after
 	// the faces' creation (e.g. by [LimitedFace.AddUnicodeRange] or by an
 	// assignment to [GoTextFace.Source]), so a cached entry can be stale.
+	// The access time of an entry is updated on every hit, so an entry used
+	// every tick never ages out of the cache and stays stale.
 	// See the MultiFace doc for the workaround.
 	splitTextCache *cache[string, []textChunk]
 }
--- a/text/v2/multi.go
+++ b/text/v2/multi.go
@@ -16,6 +16,7 @@ package text
 
 import (
 	"errors"
+	"slices"
 	"unicode/utf8"
 
 	"github.com/hajimehoshi/ebiten/v2/text/v2/internal/textutil"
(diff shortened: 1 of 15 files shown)


hajimehoshi (commented):
Reviewed at `a0bdb4644`. Verified each claim against the code; `gofmt`, `go build ./...` and `go vet` are clean on the branch.

## Verified correct — no changes needed

- **`run.go` `IsFocused`** — the removed line is a vestige of the pre-`56358fd0c` `IsForeground` API. `IsFocused` returns real window focus (`internal/ui/ui_glfw.go:423`) independent of `runnableOnUnfocused`. The old claim was also backwards: the loop *pauses* when unfocused and not runnable (`internal/ui/ui_glfw.go:1266`), so `IsFocused` would be effectively always true in that mode, not always false. Good removal.
- **`window.go`** — `WindowResizingModeDisabled` is the non-resizable mode (`window.go:33`). Correct.
- **`audio/audio.go`** — `NewPlayerF32FromBytes` matches `audio/audio.go:434`, and it is the only remaining panic-message/function-name mismatch in the package. The `singed` / `not or` fixes are right.
- **`vector/path.go`** — confirmed: the implementation measures the distance to the *infinite line* (`vector/path.go:427-431`), so the doc fix is accurate.
- **`run.go`** typos (`Use`→`use`, `in the expense`→`at the expense`) — correct.

## Should be addressed

### 1. `internal/shaderir/check.go` is a code change in a docs PR

`lhst.IsIntVector()` → `rhst.IsIntVector()` is the right fix, but it is logic, not a comment. It would help to state in the PR that it is provably a no-op: the conjunction also requires `lhst.Equal(&rhst)`, which forces `Main` to match (`internal/shaderir/type.go:28`), so the old and new expressions are equivalent for every input. That is why no test is needed — but a reader cannot tell that from a `docs:` subject.

### 2. `text/v2/multi.go` — right diagnosis, wrong remedy

The observation is real, and I confirmed it empirically. This is a **regression**, not a documentation gap:

| | `Advance("a")` after `AddUnicodeRange` |
|---|---|
| `1cba25301^` (before the cache) | 22.25 ✅ |
| current `main` / this PR | 6 ❌ (stale) |

Repro: a `MultiFace` over two `LimitedFace`s with different underlying faces; call `Advance` once, then `AddUnicodeRange` on the higher-priority face, then call `Advance` again. The cached chunk decomposition still points at the old face.

`1cba25301` ("text/v2: cache MultiFace.splitText results per text", in `v2.10.0-alpha.12`+) asserted that `hasGlyph` is stable for a face's lifetime; `text/v2/limited.go:86-88` shows it is not. Three problems with the patch as written:

- It appends a NOTE that contradicts the sentence directly above it — `text/v2/multi.go:36` still says "stable for the lifetime of a face". If the constraint is documented, that sentence has to be rewritten, not left standing.
- It is on an unexported field, so godoc never shows it — users cannot see a constraint that binds them. `MultiFace`'s godoc already carries a "known issue" note about post-creation mutation (`GoTextFace.Direction`); that is where this belongs.
- It is incomplete: `GoTextFace.hasGlyph` delegates to the exported, mutable `Source` field (`text/v2/gotext.go:281`), so reassigning `Source` breaks the cache too.

Suggestion: fix the cache (or drop it) rather than document a behavior change that already shipped.

### 3. `internal/graphicscommand/image.go` cleanup is half-done

After the removal, line 137 still reads "src is nil and there is no image in the uniform variables" — the same dead `src` parameter and the same removed uniform-image lookup. Also in the same block:

- Line 124, "Source X in texels", is stale; `internal/atlas/image.go:408` correctly says pixels, and the values are pixels (`internal/graphics/vertex.go:78`).
- The list stops at 7, but `VertexFloatCount = 12` (`internal/graphics/vertex.go:53`) — 8–11 are the custom vec4.
- The identical `Color Y` typo survives at `internal/atlas/image.go:413`.

### 4. The commit message's `#NN` references point at unrelated issues

`#12`, `#17`, `#19`, `#25`, `#28`, `#29`, `#43`, `#86`, `#87` are spreadsheet row numbers, but GitHub auto-links them — `#87` is "sprites: Add UI to change the number of the sprites". Describing each fix in prose would avoid the false links.

### 5. Subject prefix

`docs:` was last used in this repo in 2020; the current convention is the package name, or `all:` for a sweep. The precedent is your own merged `ceba21fed` — `all: fix more doc comments that diverge from the behavior (#3624)` — whose body enumerates each change and ends "Comments only, no behavior changes." That template fits here, minus the last clause (see item 1).

## Nits

- **`internal/clock/clock.go`** — I would keep the original. "shorter than one frame" named the failure mode from #1444 (at 300 TPS, 5 ticks ≈ one 60 FPS frame, so clamping to 5 ticks starves the tick rate). "shorter than `minCatchUpTime`" only restates `max(maxCatchUpTicks*tick, minCatchUpTime)`.
- **`run.go`** — "a rounded up from the logical size" is still awkward; "the actual screen size is the logical size rounded up" reads better.
- **`vector/path.go`** — with the doc corrected, the *name* now contradicts it. Worth renaming `isPointCloseToSegment` → `isPointCloseToLine` (plus the `export_test.go` wrapper and `TestIsPointCloseToSegment`) in a follow-up.
- **`window.go`** — while touching those lines, `[WindowResizingModeDisabled]` as a doc link would match `window.go:165`.

---

kumagi:
Thank you for the very thorough review! I addressed the feedback:

- **`internal/shaderir/check.go`**: I kept the `rhst.IsIntVector()` fix and want to note explicitly that it is provably a no-op: the same condition also requires `lhst.Equal(&rhst)`, and `Type.Equal` compares `Main`, so `lhst.IsIntVector()` and `rhst.IsIntVector()` always agree when the conjunction is evaluated. That is why no test is added, even though it is logic rather than a comment.
- **`text/v2/multi.go`**: instead of documenting the constraint, the `splitText` memoization is dropped. The chunk decomposition depends on glyph availability, which is not stable for a face's lifetime (`LimitedFace.AddUnicodeRange`, or re-assigning the exported `GoTextFace.Source`), so a cached entry can go stale, and a note on an unexported field would never reach the documentation anyway. `TestMultiFaceAfterFaceUpdate` is added: it returns the pre-update advance with the memoization and the correct one now. A comment on `splitText` records why the result is intentionally not memoized.
- **`internal/graphicscommand/image.go`**: the source positions are now documented in pixels, indices 8-11 (the custom vec4) are listed, and the leftover sentence about a nil `src` and image uniform variables is reworded to match the current signature. The identical `Color Y` typo in `internal/atlas/image.go` is fixed as well.
- **`internal/clock/clock.go`**: reverted; "shorter than one frame" is kept since it names the actual failure mode.
- **`run.go`**: reworded to "the actual screen size is the logical size rounded up".
- **`window.go`**: the mentions are doc links now, i.e. `([WindowResizingModeDisabled])`.
- **`vector/path.go`**: the `isPointCloseToSegment` -> `isPointCloseToLine` rename is left for a follow-up to keep this change focused on documentation.

The follow-up commit uses the `all:` prefix and describes each fix in prose instead of numbered references, so nothing is auto-linked to unrelated issues.

hajimehoshi (commented):
Reviewed the updated state at `0393fd34e`. `gofmt`, `go build ./...`, `go vet` and `go test ./text/v2/...` are all clean.

## Resolved

- **`MultiFace` cache** — fixed rather than documented, which is the right call. `TestMultiFaceAfterFaceUpdate` is a genuine regression test: run against the previous commit it fails (`got: 11.562500, want: 12.000000`), and it passes at head.
- **`internal/graphicscommand` doc** — pixels rather than texels, floats 8–11 listed, `src` → "the first element of `srcs`", and the twin `Color Y` typo fixed in `internal/atlas`.
- **The `#NN` spreadsheet references** — dropped from the commit message.
- **All four nits** — the clock comment restored, the layout doc rephrased, and the `[WindowResizingModeDisabled]` doc links.

## Open

### 1. Dropping the memoization reverts a deliberate optimization

`1cba25301` was written against a profile, and this removes it entirely. Measured with `Advance` on a 176-character line through a two-face `MultiFace` (Apple M3 Pro, 5 × 300 iterations):

| | ns/op |
|---|---|
| with the cache (`64c36c4a1`) | 4059, 4262, 4401, 4612, 4639 |
| without (`0393fd34e`) | 6000, 6022, 6095, 6516, 6622 |

That is roughly **+40%** on that path. Correctness first is defensible, but it seems worth deciding deliberately rather than as a side effect of a doc-fix PR.

A middle ground that keeps both: store a cheap validity fingerprint alongside the cached chunks — a generation counter bumped by `LimitedFace.AddUnicodeRange`, plus the current `GoTextFace.Source` pointer — and recompute on a mismatch. That covers both known mutation paths without the per-rune loop.

### 2. The new `splitText` comment justifies the code against the alternative

"The result is not memoized on purpose: … so a memoized chunk decomposition could be stale" defends a design that is not in the tree. Stating the constraint carries the same signal for a future reader:

> A face's glyph availability can change after the face's creation (e.g. by `[LimitedFace.AddUnicodeRange]`, or by an assignment to `[GoTextFace.Source]`), so the decomposition must be recomputed on every call.

### 3. `internal/graphicscommand/image.go` — one stale sentence survives

"The elements that index is in between 2 and 7 are used for the source images." is wrong, and now more visibly so next to the corrected list. It dates from `6506c20f4` (2020-05-24), when the layout put the source position *and* the source region at 2–7, with the colors at 8–11. Today only 2 and 3 are source-related, so it should read "2 and 3".

## Nits

- "Custom X/Y/Z/W" — the public API names these `Custom0`–`Custom3` (`image.go:399`).
- `internal/atlas/image.go`'s list still stops at 7, so it is now inconsistent with the block it was just synced against.
- `NewMultiFace` can collapse to `return &MultiFace{faces: slices.Clone(faces)}, nil` now that the struct has a single field.
- "the elements for the source image are not used" — when `srcs[0]` is nil the atlas layer simply skips the offset (`internal/atlas/image.go:477`), so they are passed through as-is rather than unused.

---

kumagi:
Thanks again for the detailed follow-up!

**The memoization is back, keyed by glyph availability** (`text/v2`)

`Face` gained an unexported `glyphRevision`, which returns a number that changes whenever the face's `hasGlyph` result can change:

- `LimitedFace` records a package-wide counter value in `AddUnicodeRange`.
- `GoTextFace` reports the identity of its `Source`. Each `GoTextFaceSource` gets a unique id at creation (its glyph coverage never changes), so re-assigning the exported `Source` field is detected without a setter. A monotonic id is used instead of the pointer value so that a recycled address cannot alias.
- `GoXFace` returns a constant, as its glyph coverage does not change.

`MultiFace.splitText` uses the combined revisions of its faces together with the text as the cache key, so a change simply misses and the tick-based eviction eventually drops the old entries. This covers both mutation paths without the per-rune loop.

Measured here on the same shape of workload (`Advance` on a 176-character line through a two-face `MultiFace`): about 3900 ns/op with the revision-keyed cache, and about 8300 ns/op without any memoization, i.e. the original speed is kept while the stale-cache bug is gone. Both tests fail against the previous memoization (`LimitedFace`: stale face selection; `GoTextFace.Source`: stale source), and pass now.

**Other points**

- The `splitText` comment now states the constraint (glyph availability can change after creation, e.g. by `LimitedFace.AddUnicodeRange` or by an assignment to `GoTextFace.Source`) rather than arguing against a design that is not in the tree.
- `NewMultiFace` now clones the faces slice directly.
- `internal/graphicscommand`: the sentence is corrected to indices 2 and 3, the custom floats are listed as `Custom0`-`Custom3` to match `ebiten.Vertex`, and the last sentence says they are passed as they are when no source image is given, since the atlas layer just skips the offset.
- `internal/atlas`: the vertex list is completed to 11 so the two blocks agree again.

gofmt, `go vet`, `go build ./...` and `go test ./text/...` are clean.

text/v2/multi.go
```diff
@@ -169,12 +174,30 @@ type textChunk struct {
...
+		text:     text,
+		revision: m.glyphRevision(),
+	}
+	return m.splitTextCache.getOrCreate(key, func() ([]textChunk, bool) {
+		return m.doSplitText(key.text), true
 	})
 }
 
+// glyphRevision implements Face.
+func (m *MultiFace) glyphRevision() uint64 {
```
hajimehoshi: Do not change the implementation: this PR should be dedicated to typo fixes

kumagi: Done — the implementation is untouched now. `splitText` keeps its original `cache[string, []textChunk]` exactly as on `main`, and the `glyphRevision` machinery (`text.go`, `limited.go`, `gotext.go`, `gotextfacesource.go`, `gox.go`) and the new tests are reverted out of this PR.

What remains in `text/v2` is a comment change only: the stale claim that the hasGlyph results are "stable for the lifetime of a face" is replaced with the actual constraint, and the constraint is also stated on `MultiFace`'s godoc where it is user-visible. The cache fix will be proposed separately.

kumagi:
Understood — I reverted the implementation changes in text/v2, so this pull request is documentation-only again (besides the `internal/shaderir/check.go` one-character fix we already agreed is provably a no-op). Dropped are the `Face.glyphRevision` method, the revision-keyed `splitText` cache, the per-source identity in `GoTextFaceSource`, and the two tests that required them. `LimitedFace`, `GoTextFace`, `GoXFace`, and `MultiFace` are back to the current `main` implementation.

Since the memoization stays as it is, I documented the constraint where users can actually see it: on `MultiFace`'s godoc, next to the existing known-issue note about post-creation mutation, describing that a face's glyph availability can change after the face's creation (e.g. by `LimitedFace.AddUnicodeRange` or by an assignment to `GoTextFace.Source`) so a memoized decomposition can be stale, and that the faces should be configured before creating a `MultiFace`. The unexported `splitTextCache` comment no longer claims the hasGlyph results are stable for the lifetime of a face.

The stale-cache fix itself — the revision-keyed memoization with the benchmark (about 3900 ns/op with the cache vs about 8300 ns/op without on the measured shape) and the two regression tests — is ready and I will send it as a separate pull request so it can be reviewed on its own.

hajimehoshi (commented):
Reviewed the latest state at `4631e4d3a`.

## All the documentation items are resolved

`internal/graphicscommand` now reads "The elements whose indices are 2 and 3 are used for the source image position" and "passed as they are", and `internal/atlas`'s list is completed with `Custom0`–`Custom3`, matching the public `Vertex` field names. Every doc point from both earlier rounds is addressed.

## Nit on the new known-issue wording

"might not be reflected for the texts whose chunks are already memoized" understates it: `getOrCreate` refreshes `atime` on every hit, so a line drawn every tick never ages out of the cache and the change is never reflected.

## Notes for the separate cache change

Recording the measurements here so they are not lost, since the cache fix is being split out.

`f316134f5` builds, vets, and passes `go test ./text/v2/...`, and both of its regression tests fail against `4631e4d3a` (`got: 11.562500, want: 12.000000` for the `LimitedFace` case and `got: 17.484375, want: 15.484375` for the `GoTextFace.Source` case), so they are real regression tests.

The revision key costs nothing. `Advance` on a 176-character line through a two-face `MultiFace`, Apple M3 Pro, 5 × 300 iterations:

| | ns/op |
|---|---|
| no cache (`0393fd34e`) | 6000, 6022, 6095, 6516, 6622 |
| cache without a revision (`4631e4d3a`) | 4059, 4262, 4401, 4612, 4639 |
| revision-keyed (`f316134f5`) | **3627, 3710, 3729, 3818, 3959** |

The revision-keyed version is the fastest of the three, so keeping the memoization and fixing the staleness are not in tension.

The cache landed in `v2.10.0-alpha.12` and `v2.10.0` has not been tagged yet, so the staleness would ship in v2.10.0 unless the follow-up lands first.

Three things to tighten when `f316134f5` is revived:

- **`combineGlyphRevisions` is a hash**, and a collision silently renders text with the wrong face. A `MultiFace`'s face list is fixed at its creation, so an exact check is available: keep the plain `string` key, store the per-face revisions last seen on the `MultiFace`, and reset the cache when they differ. That also stops dead entries from accumulating against the 32-entry soft limit after every `AddUnicodeRange`.
- With that, **`LimitedFace` does not need the package-wide `theGlyphRevisionCount`**; a per-face counter is enough.
- **`LimitedFace.revision` is written by `AddUnicodeRange` and read from `splitText` without synchronization.** `unicodeRanges` already had the same exposure, so this is not new in kind, but it is worth noting given the recent race audit.

---

text/v2/multi.go
kumagi: Confirmed: the implementation is untouched and this PR is documentation-only (plus the provably no-op `internal/shaderir/check.go` fix). The cache fix will be proposed separately.

kumagi:
Thanks for the final review! The known-issue wording on `MultiFace` is tightened as suggested: since `getOrCreate` refreshes the access time on every hit, a line drawn every tick never ages out of the cache, so the doc now states that a change after the MultiFace's creation is not reflected for the texts whose chunks are already memoized, instead of "might not be reflected".

The notes on the separate cache change (the exact per-face revision check instead of a combined hash, the per-face counter on `LimitedFace`, and the synchronization of the revision read/write) are recorded and will be incorporated when I send the revision-keyed memoization as its own pull request.

hajimehoshi (commented):
Reviewed `0548e35ea`. `gofmt`, `go build ./...`, `go vet`, and `go test ./text/v2/... ./vector/... ./internal/shaderir/...` are clean. (`go vet` reports three `possible misuse of unsafe.Pointer` in `internal/gamepad`, but those are already on `main` and unrelated to this change.)

Everything from the earlier rounds is addressed. Two small things left, both style-level.

### 1. The new `MultiFace` godoc puts implementation into the contract

It now says "the chunk decomposition of a text is memoized … a memoized entry used every tick never ages out of the cache". Memoization behavior and cache aging are internals, and this repository keeps those out of godoc. This is partly on the earlier review: the nit was about accuracy, and the fix bought accuracy by moving cache mechanics into the public doc. The contract does not need them:

> There is another known issue: a change to a face's glyph availability after a MultiFace's creation (e.g. by `[LimitedFace.AddUnicodeRange]` or by an assignment to `[GoTextFace.Source]`) is not reflected. Configure all the faces before creating a MultiFace.

The mechanics already live in the `splitTextCache` field comment, which is the right place for them.

### 2. `vector/path.go` mixes the parameter naming

The rewritten line reads "the distance between a point (x, y) and the line passing through p0 and p1". The `(x, y)` is left over from the old wording: the parameter is `p`, and the next line already calls it `p`.

### For the follow-up

When the revision-keyed cache lands, that second known-issue paragraph has to be deleted along with it, or it becomes documentation for a bug that no longer exists.

---

text/v2/multi.go
kumagi: Confirmed and still true: `splitText` keeps its original `cache[string, []textChunk]` exactly as on `main`, and the whole `glyphRevision` machinery plus its tests are out of this pull request. What remains in `text/v2` is a comment change only: the stale claim that the `hasGlyph` results are "stable for the lifetime of a face" is replaced on the unexported `splitTextCache` field, and the user-visible constraint is stated on `MultiFace`'s godoc, tightened to the contract itself rather than the cache mechanics.

The revision-keyed memoization, its benchmark and its two regression tests will be proposed as a separate pull request so the cache fix can be reviewed on its own.

kumagi:
Thanks for the last round of review! Both remaining points are addressed.

**`MultiFace` godoc** — the implementation is out of the contract now. The known-issue note reads:

> There is another known issue: a change to a face's glyph availability after a MultiFace's creation (e.g. by `LimitedFace.AddUnicodeRange` or by an assignment to `GoTextFace.Source`) is not reflected. Configure all the faces before creating a MultiFace.

The mechanics stay where they belong, on the unexported `splitTextCache` field, and the accuracy point from the previous round is kept there rather than in the godoc: the access time of an entry is updated on every hit, so an entry used every tick never ages out of the cache and stays stale.

**`vector/path.go`** — the leftover `(x, y)` is gone; the comment now says "the distance between the point p and the line passing through p0 and p1", which matches the parameter name used on the next line.

The pull request is still documentation-only: apart from the `NewPlayerF32FromBytes` panic-message string and the `rhst.IsIntVector()` fix in `internal/shaderir/check.go` (provably a no-op, since the same conjunction requires `lhst.Equal(&rhst)`), no non-comment line changes. Renaming `isPointCloseToSegment` to `isPointCloseToLine` remains a follow-up so this change stays focused.

And noted on the follow-up: when the revision-keyed memoization lands as its own pull request, the second known-issue paragraph will be deleted along with it, so it does not document a bug that no longer exists. The three tightening notes (an exact per-face revision check instead of a combined hash, a per-face counter on `LimitedFace`, and synchronizing the revision read/write) are recorded for that change too.

gofmt, `go build ./...`, `go vet`, and `go test ./text/... ./vector/... ./internal/shaderir/...` are clean.

hajimehoshi (approved):
LGTM

PR #3635, opened 2026-09-03
audio: bug fix: a data race in dummyPlayer volume

kumagi:
None

bug

## What this PR does | solves
Volume and SetVolume accessed p.volume without holding p.mu, while every other dummyPlayer method holds it. Concurrent Volume/SetVolume raced on the float64.

Guard both accessors with p.mu like the other methods.

### DIFF
--- a/audio/export_test.go
+++ b/audio/export_test.go
@@ -133,10 +133,14 @@ func (p *dummyPlayer) IsPlaying() bool {
 }
 
 func (p *dummyPlayer) Volume() float64 {
+	p.mu.Lock()
+	defer p.mu.Unlock()
 	return p.volume
 }
 
 func (p *dummyPlayer) SetVolume(volume float64) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
 	p.volume = volume
 }
 
--- a/audio/export_test.go
+++ b/audio/export_test.go
@@ -132,18 +132,16 @@ func (p *dummyPlayer) IsPlaying() bool {
 	return p.playing
 }
 
-// Volume returns the volume.
-//
-// The volume is not shared with the goroutine Play spawns, but the mutex is held so that the test
-// double is uniformly safe like the real player it stands in for.
 func (p *dummyPlayer) Volume() float64 {
+	// The volume is not shared with the goroutine Play spawns, but the mutex is held so that the
+	// test double is uniformly safe like the real player it stands in for.
 	p.mu.Lock()
 	defer p.mu.Unlock()
 	return p.volume
 }
 
-// SetVolume sets the volume. See Volume for why the mutex is held.
 func (p *dummyPlayer) SetVolume(volume float64) {
+	// See Volume for why the mutex is held.
 	p.mu.Lock()
 	defer p.mu.Unlock()
 	p.volume = volume
--- a/audio/export_test.go
+++ b/audio/export_test.go
@@ -132,12 +132,17 @@ func (p *dummyPlayer) IsPlaying() bool {
 	return p.playing
 }
 
+// Volume returns the volume.
+//
+// The volume is not shared with the goroutine Play spawns, but the mutex is held so that the test
+// double is uniformly safe like the real player it stands in for.
 func (p *dummyPlayer) Volume() float64 {
 	p.mu.Lock()
 	defer p.mu.Unlock()
 	return p.volume
 }
 
+// SetVolume sets the volume. See Volume for why the mutex is held.
 func (p *dummyPlayer) SetVolume(volume float64) {
 	p.mu.Lock()
 	defer p.mu.Unlock()


hajimehoshi (commented):
Thanks for the patch. The two-line change itself is harmless, but I don't think the bug it describes exists.

## The race is unreachable

`dummyPlayer` is only ever reached through `playerImpl`, and both volume paths hold `playerImpl.m` across the whole call into the underlying player:

- `audio/player.go` `(*playerImpl).Volume` locks `p.m`, then calls `p.player.Volume()`
- `audio/player.go` `(*playerImpl).SetVolume` locks `p.m`, then calls `p.player.SetVolume(volume)`
- `audio/player.go` `ensurePlayer`'s `pl.SetVolume(p.volume)`, reached only from `Play` and `startIfPending`, both of which hold `p.m`

Every `p.player.*` call site in `player.go` is inside a `p.m` section. And `p.volume` is not touched by the goroutine `dummyPlayer.Play` spawns; that goroutine only touches `r`, `readGen`, `playing` and `eof`.

I checked this by running it, not just by reading: against unpatched `main`, 8 goroutines x 20000 iterations of `Player.SetVolume` / `Player.Volume` with the audio device created (asserted via `ContextCreatedForTesting`), under `go test -race -count=3`, reports nothing.

So `dummyPlayer.mu` exists to guard the state shared with that read goroutine. `volume` was never in that set, which is why it had no lock.

## The stated justification isn't quite right either

> while every other dummyPlayer method holds it

`Err()` and `SetBufferSize()` don't. They touch no shared state, which is the same reason `Volume` / `SetVolume` didn't need to.

## What I'd suggest

The change is fine on its merits: zero cost, test-only code, `Lock` / `defer Unlock` matching the file's dominant style, and `go vet` plus `go test -race ./audio/...` are clean with it applied. As a consistency change in a test double that stands in for a concurrency-safe real player, it's defensible.

But it isn't a bug fix, and in this repository that prefix means a real, reachable bug. Could you re-title it and rewrite the message along these lines?

```
audio: guard dummyPlayer's volume with its mutex

dummyPlayer's mutex guards the state shared with the goroutine Play
spawns. volume is currently reached only under playerImpl.m, so it is
not shared, but locking it keeps the test double uniformly safe like
the real player it stands in for.
```

One note on the base: the branch is three commits behind `main`, but `audio/export_test.go` is untouched in between, so it still merges cleanly.

---

kumagi:
Thanks for the thorough review and for verifying it with the race detector. You are right: the race I described is unreachable, because every call into `dummyPlayer.Volume`/`SetVolume` goes through `playerImpl.Volume`/`SetVolume` (or `ensurePlayer`), all of which hold `playerImpl.m`, and the goroutine `dummyPlayer.Play` spawns never touches `volume`. My claim that every other `dummyPlayer` method holds the mutex was also incorrect; `Err` and `SetBufferSize` do not, since they touch no shared state.

I pushed a follow-up commit retitled along the lines you suggested:

```
audio: guard dummyPlayer's volume with its mutex

dummyPlayer's mutex guards the state shared with the goroutine Play
spawns. volume is currently reached only under playerImpl.m, so it is
not shared, but locking it keeps the test double uniformly safe like
the real player it stands in for.
```

It also documents that reasoning on the accessors themselves, so the test double keeps the mutex-guarded-volume invariant for the right reason. The branch is still fast-forward, so please feel free to squash it into a single commit when merging. `gofmt`, `go vet` and `go test -race ./audio/...` are all clean.

audio/export_test.go
```diff
@@ -132,11 +132,20 @@ func (p *dummyPlayer) IsPlaying() bool {
 	return p.playing
 }
 
+// Volume returns the volume.
+//
+// The volume is not shared with the goroutine Play spawns, but the mutex is held so that the test
+// double is uniformly safe like the real player it stands in for.
```
hajimehoshi: Move the comment about mutex into the implementation

audio/export_test.go
```diff
@@ -132,11 +132,20 @@ func (p *dummyPlayer) IsPlaying() bool {
...
 func (p *dummyPlayer) Volume() float64 {
+	p.mu.Lock()
+	defer p.mu.Unlock()
 	return p.volume
 }
 
+// SetVolume sets the volume. See Volume for why the mutex is held.
```
hajimehoshi: > See Volume for why the mutex is held.

Move this to the implementation

audio/export_test.go
kumagi: Done. The comment is now inside `Volume`, as an inline comment above `p.mu.Lock()`, and the added doc comment is removed.

audio/export_test.go
kumagi: Done. Moved into `SetVolume` itself as an inline comment above `p.mu.Lock()`, and the doc comment is removed.

kumagi:
Thanks for the review. I moved the mutex explanation into the implementations themselves, as inline comments above the `Lock` calls, and dropped the added doc comments so `Volume` and `SetVolume` match the style of the other `dummyPlayer` methods. `gofmt`, `go vet` and `go test -race ./audio/...` are all clean.

hajimehoshi (approved):
LGTM

PR #3636, opened 2026-09-03
ebitenutil: bug fix: a data race in the debug-print glyph cache

kumagi:
None

bug

## What this PR does | solves
drawDebugText lazily created per-rune sub-images in debugPrintTextSubImages without synchronization. Concurrent DebugPrint calls raced on the map.

Guard the lookup-or-create path with a dedicated mutex.

### DIFF
--- a/ebitenutil/debugprint.go
+++ b/ebitenutil/debugprint.go
@@ -21,7 +21,6 @@ import (
 	_ "embed"
 	"image"
 	_ "image/png"
-	"sync"
 
 	"github.com/hajimehoshi/ebiten/v2"
 )
@@ -29,11 +28,7 @@ import (
 //go:embed text.png
 var text_png []byte
 
-var (
-	debugPrintTextImage     *ebiten.Image
-	debugPrintTextSubImages = map[rune]*ebiten.Image{}
-	debugPrintM             sync.Mutex
-)
+var debugPrintTextImage *ebiten.Image
 
 func init() {
 	img, _, err := image.Decode(bytes.NewReader(text_png))
@@ -76,29 +71,14 @@ func drawDebugText(rt *ebiten.Image, str string, ox, oy int) {
 			x += cw
 			continue
 		}
-		s := debugPrintSubImage(c, w)
+		n := w / cw
+		sx := (int(c) % n) * cw
+		sy := (int(c) / n) * ch
+		s := debugPrintTextImage.SubImage(image.Rect(sx, sy, sx+cw, sy+ch)).(*ebiten.Image)
 		op.GeoM.Reset()
 		op.GeoM.Translate(float64(x), float64(y))
 		op.GeoM.Translate(float64(ox+1), float64(oy))
 		rt.DrawImage(s, op)
 		x += cw
 	}
-}
-
-func debugPrintSubImage(c rune, w int) *ebiten.Image {
-	const (
-		cw = 6
-		ch = 16
-	)
-	debugPrintM.Lock()
-	defer debugPrintM.Unlock()
-	if s, ok := debugPrintTextSubImages[c]; ok {
-		return s
-	}
-	n := w / cw
-	sx := (int(c) % n) * cw
-	sy := (int(c) / n) * ch
-	s := debugPrintTextImage.SubImage(image.Rect(sx, sy, sx+cw, sy+ch)).(*ebiten.Image)
-	debugPrintTextSubImages[c] = s
-	return s
 }
--- a/ebitenutil/debugprint.go
+++ b/ebitenutil/debugprint.go
@@ -21,6 +21,7 @@ import (
 	_ "embed"
 	"image"
 	_ "image/png"
+	"sync"
 
 	"github.com/hajimehoshi/ebiten/v2"
 )
@@ -31,6 +32,7 @@ var text_png []byte
 var (
 	debugPrintTextImage     *ebiten.Image
 	debugPrintTextSubImages = map[rune]*ebiten.Image{}
+	debugPrintM             sync.Mutex
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


hajimehoshi (commented):
Thanks for the report. The bug is real: I reproduced it on `main` with a `-race` test that calls `DebugPrintAt` from 8 goroutines, and it flags the read at `debugprint.go:77` against the write at `debugprint.go:83`. This patch does clear it.

But I'd rather delete the cache than lock it. `ebiten.Image.SubImage` already keeps its own sub-image cache, keyed by `image.Rectangle` and guarded by `subImageCacheM` (added in 15f279697, made concurrency-safe in be1d4fc1b). So `debugPrintTextSubImages` is a second, unsynchronized cache sitting on top of a synchronized one that memoizes exactly the same thing. Dropping it removes the race with no lock at all:

```diff
-var (
-	debugPrintTextImage     *ebiten.Image
-	debugPrintTextSubImages = map[rune]*ebiten.Image{}
-)
+var debugPrintTextImage *ebiten.Image
@@
-		s, ok := debugPrintTextSubImages[c]
-		if !ok {
-			n := w / cw
-			sx := (int(c) % n) * cw
-			sy := (int(c) / n) * ch
-			s = debugPrintTextImage.SubImage(image.Rect(sx, sy, sx+cw, sy+ch)).(*ebiten.Image)
-			debugPrintTextSubImages[c] = s
-		}
+		n := w / cw
+		sx := (int(c) % n) * cw
+		sy := (int(c) / n) * ch
+		s := debugPrintTextImage.SubImage(image.Rect(sx, sy, sx+cw, sy+ch)).(*ebiten.Image)
```

The same race test passes with this variant, and it costs nothing measurable. Benchmarking a 50-glyph `DebugPrintAt` (2000 iterations, 5 runs each) on my machine:

| | ns/op |
| --- | --- |
| this PR | 13586, 13765, 13979, 14225, 14464 |
| drop the map | 12878, 13306, 13310, 15194, 15786 |

Both are dominated by `DrawImage`, so the extra cache layer buys nothing. The net diff also comes out at -11/+4 rather than +21/-8.

For the record, the patch as written has no deadlock risk: `debugPrintM` -> `subImageCacheM` is the only lock ordering, and nothing takes them in reverse. My concern is only that the redundant layer is what makes a mutex necessary in the first place.

I've left a few smaller notes inline in case you'd prefer to keep the mutex approach.

---

ebitenutil/debugprint.go
```diff
@@ -31,6 +32,7 @@ var text_png []byte
 var (
 	debugPrintTextImage     *ebiten.Image
 	debugPrintTextSubImages = map[rune]*ebiten.Image{}
+	debugPrintM             sync.Mutex
```
hajimehoshi: Naming nit: `debugPrintM` doesn't say what it guards. The convention elsewhere in the repo is to name a mutex after the thing it protects: `subImageCacheM`, `usageCallbacksM`, `mappingsM`, `stencilBufferM`. `debugPrintTextSubImagesM` would match.

ebitenutil/debugprint.go
```diff
@@ -74,18 +76,29 @@ func drawDebugText(rt *ebiten.Image, str string, ox, oy int) {
...
 		op.GeoM.Reset()
 		op.GeoM.Translate(float64(x), float64(y))
 		op.GeoM.Translate(float64(ox+1), float64(oy))
 		rt.DrawImage(s, op)
 		x += cw
 	}
 }
+
+func debugPrintSubImage(c rune, w int) *ebiten.Image {
+	const (
```
hajimehoshi: `cw` and `ch` are now declared in two places: here and in `drawDebugText`. `sx`/`sy` computed here have to agree with the `x += cw` advance in the caller, so if one copy is ever edited the glyphs misalign silently with nothing to catch it. These should be package-level constants shared by both functions.

ebitenutil/debugprint.go
```diff
@@ -74,18 +76,29 @@ func drawDebugText(rt *ebiten.Image, str string, ox, oy int) {
...
-			debugPrintTextSubImages[c] = s
-		}
+		s := debugPrintSubImage(c, w)
 		op.GeoM.Reset()
 		op.GeoM.Translate(float64(x), float64(y))
 		op.GeoM.Translate(float64(ox+1), float64(oy))
 		rt.DrawImage(s, op)
 		x += cw
 	}
 }
```
hajimehoshi, line 84: Passing `w` in is a little awkward given the function already reads the package-level `debugPrintTextImage` on the next lines. It could just call `debugPrintTextImage.Bounds().Dx()` itself and take only `c`.

ebitenutil/debugprint.go
kumagi: Good point on the naming convention. This is now moot: the update drops `debugPrintTextSubImages` and its mutex entirely, so there is no `debugPrintM` left to rename.

ebitenutil/debugprint.go
kumagi: Agreed that duplicating `cw`/`ch` would be fragile. This is now moot: `debugPrintSubImage` is removed, so `cw`/`ch` are declared only in `drawDebugText` again.

ebitenutil/debugprint.go
kumagi: Fair point. This is now moot: `debugPrintSubImage` is removed, so nothing passes `w` around anymore; `drawDebugText` keeps its existing local `w` for the glyph layout.

kumagi:
Thanks for reproducing the race and for the detailed comparison. You are right: `Image.SubImage` already memoizes by `Rectangle` under `subImageCacheM`, so the outer `debugPrintTextSubImages` map is a redundant unsynchronized layer. I have updated the PR to drop that map entirely and call `SubImage` directly, matching your suggested diff. This fixes the race with no new lock or lock ordering to reason about. Verified locally with `gofmt`, `go vet ./ebitenutil/`, and `go build ./ebitenutil/`.

hajimehoshi (approved):
LGTM

PR #3637, opened 2026-09-03
internal/file: bug fix: a data race in file on browsers

kumagi:
None

bug

## What this PR does | solves
file kept its lazily loaded values and read offset as plain fields. Concurrent Read/Stat raced on them.

Add a mutex mirroring dir, with locked helpers.

### DIFF
--- a/internal/file/file_js_test.go
+++ b/internal/file/file_js_test.go
@@ -15,6 +15,8 @@
 package file_test
 
 import (
+	"errors"
+	"io"
 	"io/fs"
 	"slices"
 	"sync"
@@ -120,5 +122,171 @@ func TestDirReadDirConcurrently(t *testing.T) {
 
 	if got, want := names, childNames; !slices.Equal(got, want) {
 		t.Errorf("names: got: %v, want: %v", got, want)
+	}
+}
+
+// fakeFileCalls counts the asynchronous calls a fake file entry receives.
+type fakeFileCalls struct {
+	file        int
+	arrayBuffer int
+}
+
+// newFakeFileEntry returns a fake FileSystemDirectoryEntry with one child file of the given
+// content. Its getFile, file and arrayBuffer settle asynchronously, like a browser does.
+func newFakeFileEntry(t *testing.T, name string, content []byte, calls *fakeFileCalls) js.Value {
+	var funcs []js.Func
+	t.Cleanup(func() {
+		for _, f := range funcs {
+			f.Release()
+		}
+	})
+	register := func(f js.Func) js.Func {
+		funcs = append(funcs, f)
+		return f
+	}
+
+	u8 := js.Global().Get("Uint8Array").New(len(content))
+	js.CopyBytesToJS(u8, content)
+	buf := u8.Get("buffer")
+
+	arrayBuffer := register(js.FuncOf(func(this js.Value, args []js.Value) any {
+		calls.arrayBuffer++
+		executor := register(js.FuncOf(func(this js.Value, args []js.Value) any {
+			js.Global().Call("setTimeout", args[0], 0, buf)
+			return nil
+		}))
+		return js.Global().Get("Promise").New(executor)
+	}))
+
+	fileObj := js.ValueOf(map[string]any{
+		"name":         name,
+		"size":         len(content),
+		"lastModified": 0,
+		"arrayBuffer":  arrayBuffer,
+	})
+
+	fileFn := register(js.FuncOf(func(this js.Value, args []js.Value) any {
+		calls.file++
+		js.Global().Call("setTimeout", args[0], 0, fileObj)
+		return nil
+	}))
+
+	getFile := register(js.FuncOf(func(this js.Value, args []js.Value) any {
+		js.Global().Call("setTimeout", args[2], 0, js.ValueOf(map[string]any{
+			"name":        name,
+			"fullPath":    "/root/" + name,
+			"isFile":      true,
+			"isDirectory": false,
+			"file":        fileFn,
+		}))
+		return nil
+	}))
+
+	return js.ValueOf(map[string]any{
+		"name":        "root",
+		"fullPath":    "/root",
+		"isFile":      false,
+		"isDirectory": true,
+		"getFile":     getFile,
+	})
+}
+
+func openFakeFile(t *testing.T, content []byte, calls *fakeFileCalls) fs.File {
+	t.Helper()
+
+	root := newFakeFileEntry(t, "a.txt", content, calls)
+	fsys, err := file.NewFileEntryFS([]js.Value{root})
+	if err != nil {
+		t.Fatal(err)
+	}
+	f, err := fsys.Open("a.txt")
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() {
+		_ = f.Close()
+	})
+	return f
+}
+
+func TestFileReadConcurrently(t *testing.T) {
+	content := []byte("0123456789abcdefghij")
+	var calls fakeFileCalls
+	f := openFakeFile(t, content, &calls)
+
+	const goroutines = 2
+	bufs := make([][]byte, goroutines)
+	ns := make([]int, goroutines)
+	errs := make([]error, goroutines)
+	var wg sync.WaitGroup
+	for i := range goroutines {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			bufs[i] = make([]byte, len(content)/goroutines)
+			ns[i], errs[i] = f.Read(bufs[i])
+		}()
+	}
+	wg.Wait()
+
+	if got, want := calls.arrayBuffer, 1; got != want {
+		t.Errorf("arrayBuffer calls: got: %d, want: %d", got, want)
+	}
+
+	var total int
+	for i := range goroutines {
+		if err := errs[i]; err != nil && !errors.Is(err, io.EOF) {
+			t.Fatal(err)
+		}
+		total += ns[i]
+	}
+	if got, want := total, len(content); got != want {
+		t.Errorf("read bytes in total: got: %d, want: %d", got, want)
+	}
+
+	// The reads must cover the content exactly once, in either order.
+	got := string(bufs[0][:ns[0]]) + string(bufs[1][:ns[1]])
+	half := len(content) / goroutines
+	if want, wantSwapped := string(content), string(content[half:])+string(content[:half]); got != want && got != wantSwapped {
+		t.Errorf("concatenated reads: got: %q, want: %q or %q", got, want, wantSwapped)
+	}
+}
+
+func TestFileStatAndReadConcurrently(t *testing.T) {
+	content := []byte("0123456789")
+	var calls fakeFileCalls
+	f := openFakeFile(t, content, &calls)
+
+	var size int64
+	var statErr, readErr error
+	var wg sync.WaitGroup
+	wg.Add(2)
+	go func() {
+		defer wg.Done()
+		fi, err := f.Stat()
+		if err != nil {
+			statErr = err
+			return
+		}
+		size = fi.Size()
+	}()
+	go func() {
+		defer wg.Done()
+		buf := make([]byte, len(content))
+		_, readErr = f.Read(buf)
+	}()
+	wg.Wait()
+
+	if statErr != nil {
+		t.Fatal(statErr)
+	}
+	if readErr != nil && !errors.Is(readErr, io.EOF) {
+		t.Fatal(readErr)
+	}
+	if got, want := calls.file, 1; got != want {
+		t.Errorf("file calls: got: %d, want: %d", got, want)
+	}
+	if got, want := size, int64(len(content)); got != want {
+		t.Errorf("size: got: %d, want: %d", got, want)
 	}
 }
--- a/internal/file/file_js_test.go
+++ b/internal/file/file_js_test.go
@@ -220,12 +220,10 @@ func TestFileReadConcurrently(t *testing.T) {
 	errs := make([]error, goroutines)
 	var wg sync.WaitGroup
 	for i := range goroutines {
-		wg.Add(1)
-		go func() {
-			defer wg.Done()
+		wg.Go(func() {
 			bufs[i] = make([]byte, len(content)/goroutines)
 			ns[i], errs[i] = f.Read(bufs[i])
-		}()
+		})
 	}
 	wg.Wait()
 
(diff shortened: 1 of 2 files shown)


hajimehoshi (commented):
Thanks for the patch. I checked the premise and the fix, and both hold up: the race is real, and the mutex closes it. A few things before this can go in.

## What I verified

I extended the fake-entry harness in `file_js_test.go` with a fake `FileSystemFileEntry` and ran concurrent `Read` / `Stat` under node. On this PR's base (`0d78b6aae`), deterministically over three runs:

```
file_js_test.go:233: arrayBuffer calls: got: 2, want: 1
file_js_test.go:287: file calls: got: 2, want: 1
```

So the user-visible cost is bigger than "a data race": two concurrent `Read`s each start their own `arrayBuffer()`, and a dropped file is decoded and held in memory once per caller. `Stat` racing `Read` calls `entry.file()` twice the same way. With this PR applied both counts are 1, `TestDirReadDirConcurrently` still passes, and `go build` / `go vet` are clean for `js/wasm` and for the non-js build.

Holding the mutex across the `select` in `ensureUint8ArrayLocked`, and across the `<-ch` in `getFile`, is safe for the same reason as the `dir` fix in d637d0fb3: the `js.Func` callbacks run on the `syscall/js` callback goroutine and never take the mutex.

## Requested changes

**1. `ensureFile` and `ensureUint8Array` are now dead.** After the change, every call site goes to the `Locked` variant, so the two locking wrappers have no callers at all. `go vet` does not report unexported methods, so nothing catches this. Please delete both: the mutex is already taken by `Stat`, `Read` and `readAll`.

**2. Please add a test.** The equivalent fix for `dir` (d637d0fb3) landed with `TestDirReadDirConcurrently` plus `newFakeDirEntry`, and that harness extends to a file entry without much work. Here is what I used; it fails on the base commit and passes with this PR:

```
GOOS=js GOARCH=wasm go test -run 'TestFile|TestDir' -count=3 ./internal/file/
```

<details>
<summary>Additions to <code>internal/file/file_js_test.go</code> (needs <code>errors</code> and <code>io</code> in the imports)</summary>

```go
// fakeFileCalls counts the asynchronous calls a fake file entry receives.
type fakeFileCalls struct {
	file        int
	arrayBuffer int
}

// newFakeFileEntry returns a fake FileSystemDirectoryEntry with one child file of the given
// content. Its getFile, file and arrayBuffer settle asynchronously, like a browser does.
func newFakeFileEntry(t *testing.T, name string, content []byte, calls *fakeFileCalls) js.Value {
	var funcs []js.Func
	t.Cleanup(func() {
		for _, f := range funcs {
			f.Release()
		}
	})
	register := func(f js.Func) js.Func {
		funcs = append(funcs, f)
		return f
	}

	u8 := js.Global().Get("Uint8Array").New(len(content))
	js.CopyBytesToJS(u8, content)
	buf := u8.Get("buffer")

	arrayBuffer := register(js.FuncOf(func(this js.Value, args []js.Value) any {
		calls.arrayBuffer++
		executor := register(js.FuncOf(func(this js.Value, args []js.Value) any {
			js.Global().Call("setTimeout", args[0], 0, buf)
			return nil
		}))
		return js.Global().Get("Promise").New(executor)
	}))

	fileObj := js.ValueOf(map[string]any{
		"name":         name,
		"size":         len(content),
		"lastModified": 0,
		"arrayBuffer":  arrayBuffer,
	})

	fileFn := register(js.FuncOf(func(this js.Value, args []js.Value) any {
		calls.file++
		js.Global().Call("setTimeout", args[0], 0, fileObj)
		return nil
	}))

	getFile := register(js.FuncOf(func(this js.Value, args []js.Value) any {
		js.Global().Call("setTimeout", args[2], 0, js.ValueOf(map[string]any{
			"name":        name,
			"fullPath":    "/root/" + name,
			"isFile":      true,
			"isDirectory": false,
			"file":        fileFn,
		}))
		return nil
	}))

	return js.ValueOf(map[string]any{
		"name":        "root",
		"fullPath":    "/root",
		"isFile":      false,
		"isDirectory": true,
		"getFile":     getFile,
	})
}

func openFakeFile(t *testing.T, content []byte, calls *fakeFileCalls) fs.File {
	t.Helper()

	root := newFakeFileEntry(t, "a.txt", content, calls)
	fsys, err := file.NewFileEntryFS([]js.Value{root})
	if err != nil {
		t.Fatal(err)
	}
	f, err := fsys.Open("a.txt")
	if err != nil {
		t.Fatal(err)
	}
	t.Cleanup(func() {
		_ = f.Close()
	})
	return f
}

func TestFileReadConcurrently(t *testing.T) {
	content := []byte("0123456789abcdefghij")
	var calls fakeFileCalls
	f := openFakeFile(t, content, &calls)

	const goroutines = 2
	bufs := make([][]byte, goroutines)
	ns := make([]int, goroutines)
	errs := make([]error, goroutines)
	var wg sync.WaitGroup
	for i := range goroutines {
		wg.Add(1)
		go func() {
			defer wg.Done()
			bufs[i] = make([]byte, len(content)/goroutines)
			ns[i], errs[i] = f.Read(bufs[i])
		}()
	}
	wg.Wait()

	if got, want := calls.arrayBuffer, 1; got != want {
		t.Errorf("arrayBuffer calls: got: %d, want: %d", got, want)
	}

	var total int
	for i := range goroutines {
		if err := errs[i]; err != nil && !errors.Is(err, io.EOF) {
			t.Fatal(err)
		}
		total += ns[i]
	}
	if got, want := total, len(content); got != want {
		t.Errorf("read bytes in total: got: %d, want: %d", got, want)
	}

	// The reads must cover the content exactly once, in either order.
	got := string(bufs[0][:ns[0]]) + string(bufs[1][:ns[1]])
	half := len(content) / goroutines
	if want, wantSwapped := string(content), string(content[half:])+string(content[:half]); got != want && got != wantSwapped {
		t.Errorf("concatenated reads: got: %q, want: %q or %q", got, want, wantSwapped)
	}
}

func TestFileStatAndReadConcurrently(t *testing.T) {
	content := []byte("0123456789")
	var calls fakeFileCalls
	f := openFakeFile(t, content, &calls)

	var size int64
	var statErr, readErr error
	var wg sync.WaitGroup
	wg.Add(2)
	go func() {
		defer wg.Done()
		fi, err := f.Stat()
		if err != nil {
			statErr = err
			return
		}
		size = fi.Size()
	}()
	go func() {
		defer wg.Done()
		buf := make([]byte, len(content))
		_, readErr = f.Read(buf)
	}()
	wg.Wait()

	if statErr != nil {
		t.Fatal(statErr)
	}
	if readErr != nil && !errors.Is(readErr, io.EOF) {
		t.Fatal(readErr)
	}
	if got, want := calls.file, 1; got != want {
		t.Errorf("file calls: got: %d, want: %d", got, want)
	}
	if got, want := size, int64(len(content)); got != want {
		t.Errorf("size: got: %d, want: %d", got, want)
	}
}
```

</details>

**3. Please note why the lock is held across the asynchronous wait.** `(*dir).ReadDir` carries such a comment, and `ensureFileLocked` / `ensureUint8ArrayLocked` are now in the same non-obvious position without one. A single line is enough; otherwise a later reader is likely to "fix" a deadlock that is not there.

**4. The commit message could say more.** "Concurrent Read/Stat raced on them" is accurate but does not describe what goes wrong. Naming the symptom (the file is fetched and buffered once per concurrent reader) would match how d637d0fb3 is written and would help whoever hits this later.

---

*Reviewed by Claude (Claude Code) on behalf of @hajimehoshi.*

kumagi:
Thanks for checking the premise so thoroughly, and for the harness. It reproduced exactly what you described: on `0d78b6aae` I get `arrayBuffer calls: got: 2, want: 1` and `file calls: got: 2, want: 1`, and both counts are 1 with the mutex.

All four points are addressed in the follow-up commit:

1. `ensureFile` and `ensureUint8Array` are deleted. Only the `Locked` helpers remain, and the three callers that take the mutex (`Stat`, `Read`, `readAll`) are the only ones left.
2. `TestFileReadConcurrently` and `TestFileStatAndReadConcurrently` are added on top of the fake-entry harness, with your `newFakeFileEntry` and `fakeFileCalls`. `GOOS=js GOARCH=wasm go test -run 'TestFile|TestDir' -count=10 ./internal/file/` passes here, and fails as above on the base commit.
3. `ensureFileLocked` and `ensureUint8ArrayLocked` now say why the mutex is held across the wait, and why that cannot deadlock: the `js.Func` callbacks run on the `syscall/js` callback goroutine and never take the mutex.
4. I left the already-pushed commit message alone so the branch stays fast-forwardable, and described the symptom in the follow-up commit message instead: two concurrent reads each call `arrayBuffer`, so a dropped file is fetched, decoded and held in memory once per reader, and a `Stat` racing a `Read` calls `entry.file` twice. The tests record the same thing.

`go build` and `go vet` are clean for `js/wasm` and for the non-js build, and `TestDirReadDirConcurrently` still passes.

hajimehoshi (commented):
Thanks, all four points are addressed. This is ready to go in as far as the fix itself is concerned; what is left below is style and a merge-time note.

## What I verified at 3ed5728bb

* `gofmt` clean; `go build` and `go vet` clean for `js/wasm` and for the non-js build; `internal/vettools` clean for both.
* `GOOS=js GOARCH=wasm go test -count=10 ./internal/file/` passes, and the non-js tests pass.
* No dead code left: `ensureFile` and `ensureUint8Array` are gone, and every remaining helper has a live caller.
* The branch rebases cleanly onto current `main` (14fac0717), nothing else has touched `internal/file` since the base, and the tests still pass after the rebase.

One thing worth putting in the record, which I did not weigh properly the first time: `DroppedFiles` is documented as concurrent-safe in `run.go`. The non-js side already honours that, since `virtualFile` embeds an `*os.File` and `virtualDir` guards its lazy fill. On browsers, `dir` was fixed by d637d0fb3 and `file` was the last gap. So this is not only hardening, it makes the browser path meet a guarantee the documentation already gives.

## Remaining

**1. The rationale is written twice.** `ensureFileLocked` and `ensureUint8ArrayLocked` each carry a `The caller must hold f.mu.` doc comment and a three-line inline note, and the two inline notes say the same three things in almost the same words: the lock is held across the wait, the callbacks never take it, and that is what makes a second caller reuse the result. That is four comment blocks for one invariant. `(*dir).ReadDir` states it once for the whole method, and `ensureNative` in `internal/gamepad` is a single line. The one natural home is the field itself:

```go
type file struct {
	entry      js.Value
	file       js.Value
	offset     int64
	uint8Array js.Value

	// mu guards the lazily fetched values and the offset. It is held even while a helper waits for
	// a JavaScript callback: the callbacks run on the syscall/js callback goroutine and never take
	// mu, so waiting under it cannot deadlock, and it is what makes a concurrent caller reuse the
	// fetched value instead of fetching it again.
	mu sync.Mutex
}
```

Then each helper needs only `The caller must hold f.mu.`

**2. The first line of each doc comment carries a detail that belongs in the body.** "fetching it if it has not been fetched yet" and "loading them if they have not been loaded yet" describe the memoization, which the `ensure` prefix and the `Truthy` guard already make plain. The part worth documenting is the lock.

**3. Please squash before this is merged.** As it stands the first commit introduces `ensureFile` and `ensureUint8Array` and the second deletes them, so the intermediate commit carries dead code, and the second message reads as a reply to review rather than as a description of a change. One commit with the first message plus the symptom paragraph from the second would bisect cleanly.

Not a request, only so that it is a deliberate choice: because the lock spans the wait, a `Stat` now blocks behind a concurrent `Read` that is still decoding the whole file. That is the same trade d637d0fb3 made, and correctness is worth it here, but `Stat` on a large dropped file is no longer cheap while a read is in flight.

---

*Reviewed by Claude (Claude Code) on behalf of @hajimehoshi.*

kumagi:
Thanks for the second look.

1 and 2: The rationale is now stated once, on the `mu` field itself: it guards the lazily fetched values and the offset, it is held while a helper waits for a JavaScript callback, and that cannot deadlock because the callbacks run on the `syscall/js` callback goroutine and never take it. `ensureFileLocked` and `ensureUint8ArrayLocked` keep only `The caller must hold f.mu.`, and their first doc lines no longer restate the memoization.

3: Noted. I left the already-pushed commits alone so the branch stays fast-forwardable; the whole branch should be squash-merged into a single commit carrying the first message plus the symptom paragraph.

On `Stat` blocking behind an in-flight read: that is a deliberate choice, the same trade d637d0fb3 made.

Verified at the new head: `gofmt` clean; `go build` and `go vet` clean for js/wasm and for the non-js build; `internal/vettools` clean for both; `GOOS=js GOARCH=wasm go test -count=10 ./internal/file/` passes and the non-js tests pass.

internal/file/file_js_test.go
```diff
@@ -122,3 +124,169 @@ func TestDirReadDirConcurrently(t *testing.T) {
...
+func TestFileStatAndReadConcurrently(t *testing.T) {
+	content := []byte("0123456789")
+	var calls fakeFileCalls
+	f := openFakeFile(t, content, &calls)
+
+	var size int64
+	var statErr, readErr error
+	var wg sync.WaitGroup
+	wg.Add(2)
+	go func() {
```
hajimehoshi: Use wg.Go

internal/file/file_js_test.go
```diff
@@ -122,3 +124,169 @@ func TestDirReadDirConcurrently(t *testing.T) {
...
+		defer wg.Done()
+		fi, err := f.Stat()
+		if err != nil {
+			statErr = err
+			return
+		}
+		size = fi.Size()
+	}()
+	go func() {
+		defer wg.Done()
```
hajimehoshi: Use wg.Go

internal/file/file_js_test.go
kumagi: Done. The goroutine now starts with `wg.Go`, and the same pattern in the loop of `TestFileReadConcurrently` was converted too.

internal/file/file_js_test.go
kumagi: Done. Both goroutines in `TestFileStatAndReadConcurrently` now use `wg.Go`, dropping the `Add`/`Done` bookkeeping.

kumagi:
Both inline comments are addressed: the new tests now use `WaitGroup.Go` instead of the `Add`/`go`/`Done` pattern, matching what the rest of the repository does since Go 1.25. This covers the two flagged goroutines in `TestFileStatAndReadConcurrently` and the loop goroutine in `TestFileReadConcurrently`.

Verified at the new head: `gofmt` clean, `go vet` clean for js/wasm, `GOOS=js GOARCH=wasm go test -run 'TestFile|TestDir' -count=10 ./internal/file/` passes, and the non-js tests pass.

hajimehoshi (approved):
LGTM

PR #3638, opened 2026-09-03
mobile: bug fix: data races in ebitenmobileview input state

kumagi:
None

bug

## What this PR does | solves
The key tables, touches, ptrToID and the reused touchSlice were accessed without synchronization. Concurrent touches or keys raced.

Add a dedicated mutex held across the mutation and the updateInput copy.

### DIFF
--- a/mobile/ebitenmobileview/input.go
+++ b/mobile/ebitenmobileview/input.go
@@ -28,6 +28,15 @@ type position struct {
 }
 
 var (
+	// inputMu protects the input state of the package: the members below and
+	// ptrToID in input_ios.go. Every platform entry point (the touch and key
+	// callbacks in input_android.go and input_ios.go) holds it across its mutation
+	// and the updateInput copy.
+	//
+	// The entry points are expected to run on the platform UI thread only, so this
+	// mutex is normally uncontended and enforces the convention.
+	inputMu sync.Mutex
+
 	keyPressedTimes  [ui.KeyMax + 1]ui.InputTime
 	keyReleasedTimes [ui.KeyMax + 1]ui.InputTime
 	touches          = map[ui.TouchID]position{}
@@ -41,10 +50,10 @@ var (
 	touchSlice []ui.TouchForInput
 )
 
-var inputM sync.Mutex
-
 // setKeyReleased records a key release. The release of a key that is not down
 // is ignored: the game never saw the key pressed.
+//
+// setKeyReleased must be called with inputMu held.
 func setKeyReleased(key ui.Key) {
 	if keyPressedTimes[key] <= keyReleasedTimes[key] {
 		return
@@ -52,6 +61,9 @@ func setKeyReleased(key ui.Key) {
 	keyReleasedTimes[key] = ui.Get().InputTime()
 }
 
+// updateInput copies the guarded state to the platform input state.
+//
+// updateInput must be called with inputMu held.
 func updateInput(runes []rune) {
 	touchSlice = touchSlice[:0]
 	for id, position := range touches {
--- a/mobile/ebitenmobileview/input.go
+++ b/mobile/ebitenmobileview/input.go
@@ -17,6 +17,8 @@
 package ebitenmobileview
 
 import (
+	"sync"
+
 	"github.com/hajimehoshi/ebiten/v2/internal/ui"
 )
 
@@ -38,6 +40,8 @@ var (
 var (
 	touchSlice []ui.TouchForInput
 )
+
+var inputM sync.Mutex
 
 // setKeyReleased records a key release. The release of a key that is not down
 // is ignored: the game never saw the key pressed.
--- a/mobile/ebitenmobileview/input.go
+++ b/mobile/ebitenmobileview/input.go
@@ -28,13 +28,9 @@ type position struct {
 }
 
 var (
-	// inputMu protects the input state of the package: the members below and
-	// ptrToID in input_ios.go. Every platform entry point (the touch and key
-	// callbacks in input_android.go and input_ios.go) holds it across its mutation
-	// and the updateInput copy.
-	//
-	// The entry points are expected to run on the platform UI thread only, so this
-	// mutex is normally uncontended and enforces the convention.
+	// inputMu protects the variables below and ptrToID in input_ios.go.
+	// The platform entry points are expected to run on the UI thread only;
+	// inputMu makes that confinement explicit.
 	inputMu sync.Mutex
 
 	keyPressedTimes  [ui.KeyMax + 1]ui.InputTime
@@ -44,9 +40,7 @@ var (
 	// capsLock and numLock stay unknown until a physical keyboard reports them.
 	capsLock ui.LockKeyState
 	numLock  ui.LockKeyState
-)
 
-var (
 	touchSlice []ui.TouchForInput
 )
 
--- a/mobile/ebitenmobileview/input_android.go
+++ b/mobile/ebitenmobileview/input_android.go
@@ -132,8 +132,8 @@ var androidKeyToSDL = map[int]int{
 }
 
 func UpdateTouchesOnAndroid(action int, id int, x, y float64) {
-	inputM.Lock()
-	defer inputM.Unlock()
+	inputMu.Lock()
+	defer inputMu.Unlock()
 	switch action {
 	case 0x00, 0x05, 0x02: // ACTION_DOWN, ACTION_POINTER_DOWN, ACTION_MOVE
 		touches[ui.TouchID(id)] = position{x, y}
@@ -145,6 +145,8 @@ func UpdateTouchesOnAndroid(action int, id int, x, y float64) {
 }
 
 func OnKeyDownOnAndroid(keyCode int, unicodeChar int, source int, deviceID int, metaState int) {
+	inputMu.Lock()
+	defer inputMu.Unlock()
 	switch {
 	case source&sourceGamepad == sourceGamepad:
 		// A gamepad can be detected as a keyboard. Detect the device as a gamepad first.
@@ -154,8 +156,6 @@ func OnKeyDownOnAndroid(keyCode int, unicodeChar int, source int, deviceID int, 
 	case source&sourceJoystick == sourceJoystick:
 		// DPAD keys can come here, but they are also treated as an axis at a motion event. Ignore them.
 	case source&sourceKeyboard == sourceKeyboard, source == sourceUnknown && deviceID == virtualKeyboard:
-		inputM.Lock()
-		defer inputM.Unlock()
 		if key, ok := androidKeyToUIKey[keyCode]; ok {
 			keyPressedTimes[key] = ui.Get().InputTime()
 		}
@@ -169,6 +169,8 @@ func OnKeyDownOnAndroid(keyCode int, unicodeChar int, source int, deviceID int, 
 }
 
 func OnKeyUpOnAndroid(keyCode int, source int, deviceID int, metaState int) {
+	inputMu.Lock()
+	defer inputMu.Unlock()
 	switch {
 	case source&sourceGamepad == sourceGamepad:
 		// A gamepad can be detected as a keyboard. Detect the device as a gamepad first.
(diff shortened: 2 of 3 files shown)


mobile/ebitenmobileview/input.go
```diff
@@ -39,6 +41,8 @@ var (
 	touchSlice []ui.TouchForInput
 )
 
+var inputM sync.Mutex
```
hajimehoshi: Could this move into the `var` block above and carry a comment saying what it guards? As it stands the mutex is separated from its state by the `touchSlice` block, and nothing records that `setKeyReleased` and `updateInput` now have an unwritten precondition, so a caller added later in `mobile.go` has no way to know. The repo documents this kind of invariant elsewhere, e.g. `internal/ui/image.go` ("mu protects the members below", "drawTriangles must be called with the mutexes of the image and the source images locked") and `internal/gamepad/gamepad.go` ("ensureNative must be called with g.m held").

Naming nit while you're here: `inputM` matches the older style in the repo (`subImageCacheM`, `backendsM`), but newer code uses `Mu` (`postTickHooksMu`, `textInputHandlersMu`, `devicesMu`).

mobile/ebitenmobileview/input_android.go
```diff
@@ -152,6 +154,8 @@ func OnKeyDownOnAndroid(keyCode int, unicodeChar int, source int, deviceID int,
 	case source&sourceJoystick == sourceJoystick:
 		// DPAD keys can come here, but they are also treated as an axis at a motion event. Ignore them.
 	case source&sourceKeyboard == sourceKeyboard, source == sourceUnknown && deviceID == virtualKeyboard:
+		inputM.Lock()
```
hajimehoshi: Locking inside the `case` here, and in `OnKeyUpOnAndroid`, is inconsistent with the iOS side, which locks at the top of the function. I assume that is deliberate, to keep the `gamepad.*` branches out of the critical section, but as written it reads like an oversight and invites a later "cleanup" that hoists it.

Either a one-line comment, or just locking at the top: `gamepad` has no path back into this package, so widening the section is safe.

hajimehoshi (commented):
Thanks for the patch. I went through it carefully: it builds and vets cleanly for both platforms, the guarded set is complete, and I could not find a deadlock. My one substantive concern is the premise.

**The entry points already look thread-confined.** Tracing every caller of the five functions this locks:

| Entry point | Caller | Thread |
| --- | --- | --- |
| `updateTouchesOnAndroid` | `EbitenView.onTouchEvent` | Android UI |
| `onKeyDownOnAndroid` / `onKeyUpOnAndroid` | `EbitenView.onKeyDown`/`onKeyUp`, the hidden `EditText`'s `onKeyDown`/`onKeyUp`/`onKeyPreIme`, and `dispatchKeyToGame` from the `InputConnection` callbacks | Android UI |
| `UpdateTouchesOnIOS` / `UpdatePressesOnIOS` | `EbitenViewController`'s `touchesBegan:`... / `pressesBegan:`... | UIKit main |
| `dispatchKeyDown` / `dispatchKeyUp` / `dispatchKeyPress` | `pressesOnMain`, `resignAndClearOnMain`, `deleteBackwardOnMain` in `exp/textinput`, all main-thread selector implementations; the only cross-thread hops there are `performSelectorOnMainThread:...waitUntilDone:false` | UIKit main |

The game loop is not a second writer either: it consumes the snapshot that `updateInputStateFromOutside` copies under `UserInterface.mu`, never these variables. So I don't see how two of these can run concurrently today.

Do you have a repro, a race-detector report, or a call site off the UI thread? That would point at something bigger than this package, and I'd like to see it. If there isn't one, I'm still open to taking the change as hardening (the mutex is uncontended, and it turns an unwritten convention into an enforced one), but the commit message should say that rather than claim an observed race.

For the record, what I verified on top of `main`:

* `go build ./mobile/...` and `go vet ./mobile/...` clean for `GOOS=ios` and `GOOS=android` (arm64, real cgo); `gofmt -s -l` clean.
* Coverage is complete: no unguarded access to the key tables, `touches`, `touchSlice`, `ptrToID`, `capsLock` or `numLock` remains in the package.
* Lock ordering is fine. The change adds `inputM -> UserInterface.mu` and `inputM -> @synchronized(EbitenViewController)` (via `RequestRenderIfNeeded`), and neither has a reverse edge: `updateInputStateFromOutside` makes no outbound calls, and `EbitenmobileviewUpdate` runs outside `@synchronized(self)`. The nested iOS press path (`pressesOnMain` -> `SendSuper` -> `updatePresses:` -> `UpdatePressesOnIOS`) does not re-enter either, since each `DispatchKey*` returns before `SendSuper` is reached.

Two inline notes below.

---

mobile/ebitenmobileview/input.go
kumagi: Done. The mutex moved into the `var` block with the state it guards, renamed to `inputMu` to match the newer `Mu` style, and the comment records what it protects (the key tables, the lock key states, `touches`, `touchSlice`, and `ptrToID` on iOS), which entry points hold it, and that the entry points are UI-thread confined so it is normally uncontended. `setKeyReleased` and `updateInput` now document that they must be called with `inputMu` held, following `internal/ui/image.go` and `internal/gamepad/gamepad.go`.

mobile/ebitenmobileview/input_android.go
kumagi: Good catch, it was deliberate but read as an oversight. I took the second option and lock at the top of `OnKeyDownOnAndroid` and `OnKeyUpOnAndroid`, which matches the iOS side; I confirmed `gamepad.UpdateAndroidGamepadButton` only takes the gamepad's own mutex and never re-enters this package, so widening the critical section is safe and there is no reverse lock edge.

kumagi:
Thanks for the careful trace, and you are right about the premise: I have no race-detector report and no call site off the UI thread. Going through the callers myself, every writer is confined to the platform UI thread (the Android UI thread for `onTouchEvent` and the key callbacks, UIKit main for the touch and press paths and the `exp/textinput` dispatchers), and the game loop only reads the snapshot `updateInputStateFromOutside` copies under `UserInterface.mu`. So there is no second writer today.

I have taken the change as hardening then: the state was unsynchronized, the writers are thread-confined by convention, and the mutex enforces that convention instead of leaving it unwritten. This feedback commit says exactly that and no longer claims an observed race, and the comment on the mutex records the guarded state, which entry points hold it, and that it is normally uncontended.

Both inline notes are addressed as well: the mutex now sits in the `var` block with its guarded state, renamed to `inputMu`, with `setKeyReleased` and `updateInput` documenting that they must be called with it held, and the Android key handlers lock at the top of the function since `gamepad` has no path back into this package.

Happy to drop the change instead if you would rather not carry an uncontended lock; I erred toward keeping the invariant written down.

hajimehoshi (commented):
Thanks, this addresses both notes and I am happy with the shape now. `inputMu` is in the right place with its contract written down, and the Android handlers match the iOS side.

I checked that the widening I suggested actually holds: `OnKeyDownOnAndroid` and `OnKeyUpOnAndroid` now hold `inputMu` across `gamepad.UpdateAndroidGamepadButton`, which takes `gamepads.m` and then `Gamepad.m` and invokes no callbacks, so there is no path back into this package and no reverse edge. `go vet ./mobile/...` is clean for `GOOS=ios` and `GOOS=android` (arm64, real cgo), and `gofmt -s -l` is clean.

Nothing left is a correctness problem. Three things remain, none of them in the code:

* **The title still claims an observed race.** `mobile: bug fix: data races in ebitenmobileview input state` is what a squash merge would land, so the correction that currently lives only in the second commit's body would be lost. Please reword it without `bug fix:`, e.g. `mobile/ebitenmobileview: guard the input state with a mutex`.
* **Prefix.** This package's commits use `mobile/ebitenmobileview:` rather than `mobile:`, for instance "mobile/ebitenmobileview: ignore a key release for a key that is not down" and "mobile/ebitenmobileview: accept sourceUnknown events from virtualKeyboard (#3416)".
* **Please squash the two commits into one** with that message.

One inline note left, on the comment itself.

On your last question: as I said, I am fine taking this as hardening. Writing the invariant down is worth an uncontended lock, so no need to drop it.

---

mobile/ebitenmobileview/input.go
```diff
@@ -26,6 +28,15 @@ type position struct {
...
 
 var (
+	// inputMu protects the input state of the package: the members below and
+	// ptrToID in input_ios.go. Every platform entry point (the touch and key
+	// callbacks in input_android.go and input_ios.go) holds it across its mutation
+	// and the updateInput copy.
+	//
+	// The entry points are expected to run on the platform UI thread only, so this
+	// mutex is normally uncontended and enforces the convention.
+	inputMu sync.Mutex
```
hajimehoshi: Two things about this comment.

First, `touchSlice` falls outside the scope it states. "the members below" reads as the members of this `var` block, and `touchSlice` sits in a separate block just after it, yet `updateInput` mutates it under the lock. It is also the variable the original commit message singled out. Folding `touchSlice` into this block fixes the gap and removes a single-element `var` block that has no reason to be separate.

Second, it says more than it needs to. "Every platform entry point ... holds it across its mutation and the `updateInput` copy" restates what the code shows and will rot as entry points come and go, and "normally uncontended and enforces the convention" is rationale that belongs in the commit message. Something like:

```go
// inputMu protects the variables below and ptrToID in input_ios.go.
// The platform entry points are expected to run on the UI thread only;
// inputMu makes that confinement explicit.
```

mobile/ebitenmobileview/input.go
kumagi: Done: the mutex is now in the var block with the state it guards, renamed to inputMu, and setKeyReleased and updateInput document that they must be called with inputMu held. The comment has since been shortened further per your latest note.

mobile/ebitenmobileview/input_android.go
kumagi: Done: OnKeyDownOnAndroid and OnKeyUpOnAndroid now lock inputMu at the top of the function, matching the iOS side, since gamepad has no path back into this package.

mobile/ebitenmobileview/input.go
kumagi: Done: touchSlice is folded into the same var block so it falls under the comment's scope, and the comment now uses the shorter wording you suggested.

kumagi:
Thanks for the follow-up review. I have applied the remaining inline note: touchSlice is now folded into the same var block as the rest of the guarded state, and the inputMu comment uses the shorter wording you suggested. gofmt and host go vet are clean. On the non-code items: I agree to the title mobile/ebitenmobileview: guard the input state with a mutex with the ebitenmobileview prefix and no bug-fix claim, and I am fine with the commits being squashed into one with that message on merge. Happy to keep this as hardening that writes the UI-thread confinement invariant down.

hajimehoshi (approved):
LGTM

hajimehoshi:
For the record, since this PR's title and its first commit still say "bug fix: data races": the merged commit message is the accurate description of what landed.

Merged as bed70f70d:

> **mobile/ebitenmobileview: guard the input state with a mutex (#3638)**
>
> The key tables, touches, ptrToID and the reused touchSlice were accessed
> without synchronization. All of the writers are confined to the platform
> UI thread today, so no race is observed; the mutex makes that confinement
> explicit instead of leaving it as an unwritten convention.

So this is a safety net, not a fix for an observed bug. Nobody reading `git log` later should go hunting for the race it fixed, because there was not one.

I did weigh reverting it afterwards. The argument against carrying a lock over state that is already thread-confined is that it mutes the loud `fatal error: concurrent map writes` a future confinement violation would otherwise produce, and it adds lock-ordering edges that did not exist before. I decided against reverting: the change costs nothing on an uncontended path, the ordering is acyclic, and the comment it brought with it is the first place the UI-thread confinement rule is written down at all. That comment is the part I most wanted to keep, and a revert would have taken it too.

Thanks for the three rounds of revision, and for the straight answer on the premise when I asked whether you had a repro.

---

PR #3639, opened 2026-09-03
internal/glfw: bug fix: a data race in handleToWindow on Windows

kumagi:
None

bug

## What this PR does | solves
handleToWindow was written on creation/destruction and read from windowProc and focus callbacks without synchronization. Concurrent map access raced.

Guard the map with an RWMutex with small helpers.

### DIFF
--- a/internal/glfw/win32_window_windows.go
+++ b/internal/glfw/win32_window_windows.go
@@ -12,7 +12,6 @@ import (
 	"log/slog"
 	"math"
 	"runtime"
-	"sync"
 	"unsafe"
 
 	"golang.org/x/sys/windows"
@@ -584,7 +583,7 @@ func (w *Window) maximizeWindowManually() error {
 }
 
 func windowProc(hWnd windows.HWND, uMsg uint32, wParam _WPARAM, lParam _LPARAM) uintptr /*_LRESULT*/ {
-	window := lookupWindow(hWnd)
+	window := handleToWindow[hWnd]
 	if window == nil {
 		// This is the message handling for the hidden helper window
 		// and for a regular window during its initial creation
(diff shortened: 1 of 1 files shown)


hajimehoshi:
Thanks for the patch, but I don't think this race can happen, and I'd like to see the actual race report before taking it.

**`handleToWindow` is confined to the main OS thread.**

All five accesses run there:

* `createNativeWindow` (write) is reached from `glfw.CreateWindow`, called from `(*glfwBackend).createWindow` in `internal/ui/ui_glfw.go`, which is documented as main-thread only.
* `platformDestroyWindow` (delete) and `platformPollEvents` (read) run on the main thread as well.
* `windowProc` (read) is the interesting one: Win32 delivers a window's messages only to the thread that created it. It is entered from `_DispatchMessageW` in `platformPollEvents`, from synchronous Win32 calls made on the main thread, or from a nested modal loop on that same thread. A cross-thread `SendMessage` is marshalled by USER32 to the owning thread rather than run on the caller's.

Supporting: `internal/glfw` starts no goroutines; `internal/ui` marshals every glfw call through `mainThread.Call`, the sole exception being `PostEmptyEvent`, which is a `_PostMessageW` to the helper window and so is still dispatched on the main thread; and the main thread is pinned with `runtime.LockOSThread`.

**Even if the premise held, this would not be enough.** `windowProc` itself appends to `_glfw.errors`, reallocates `_glfw.platformWindow.rawInput` under `WM_INPUT`, and writes `window.platform.frameAction`, `highSurrogate`, `cursorTracked` and `lastCursorPosX`/`lastCursorPosY`, all unsynchronized. Guarding only the map would quiet the least dangerous access and leave the genuinely dangerous ones, which is worse than leaving the file alone: it suggests the file is now safe.

Two smaller notes either way: the description says the map is read from "focus callbacks", but the second reader is `platformPollEvents` (the stuck-modifier-key workaround), not a callback. And this file is a port of upstream GLFW's `win32_window.c`, so non-upstream helpers on the per-message path add friction whenever it is re-synced.

If you do have a `go test -race` or runtime race report, please post it with both stacks. That would show which of the assumptions above is wrong, and the fix would then need to cover `_glfw.errors` and `window.platform.*` too.

kumagi:
Thanks for the careful read — you are right, and I could not produce a race report.

I re-checked the confinement argument in the code and it holds: internal/glfw starts no goroutines on Windows, internal/ui routes glfw calls through `mainThread.Call` (the only exception, `PostEmptyEvent`, just posts `WM_NULL` to the helper window, which is dispatched on the main thread), and `windowProc` is entered from `_DispatchMessageW`, from synchronous Win32 calls, or from the nested main-thread loop. Win32 delivers a window's messages to the thread that created it, so every access to `handleToWindow` is on that one OS thread, and there is no race to fix.

I also agree that guarding only the map while `_glfw.errors`, `_glfw.platformWindow.rawInput` and `window.platform.*` stay unsynchronized would have been worse than touching nothing, and that non-upstream helpers on the per-message path would add friction at the next GLFW re-sync. So I dropped the mutex and the `lookupWindow`/`registerWindow`/`unregisterWindow` helpers entirely; the code is back to upstream's plain map access. What remains is only a comment on `handleToWindow` recording the invariant (written on creation/destruction, read in `windowProc` and in `platformPollEvents`, all on the window-owning OS thread, hence no synchronization needed), in the spirit of the existing notes such as the one on X11's `syncCounter`, so that this "fix" isn't proposed again. Please say the word if even that comment is unwelcome and I will drop it or close the PR.

Lastly, your correction on the description is right: the second reader is `platformPollEvents` (the stuck-modifier-key workaround), not a focus callback.

hajimehoshi (approved):
LGTM

PR #3640, opened 2026-09-03
internal/glfw: bug fix: a data race in theGoWindows on macOS

kumagi:
None

bug

## What this PR does | solves
theGoWindows was written on creation/destruction and read from ObjC callbacks without synchronization. Concurrent map access raced.

Guard the map with an RWMutex with helpers.

### DIFF
--- a/internal/glfw/cocoa_window_darwin.go
+++ b/internal/glfw/cocoa_window_darwin.go
@@ -10,6 +10,7 @@ import (
 	"image"
 	"math"
 	"runtime"
+	"sync"
 	"unsafe"
 
 	"github.com/ebitengine/purego/objc"
@@ -658,7 +659,7 @@ func registerGLFWClasses() error {
 						window.platform.markedText.Send(sel_release)
 						window.platform.markedText = 0
 					}
-					delete(theGoWindows, self)
+					deleteGoWindow(self)
 					self.SendSuper(objc.RegisterName("dealloc"))
 				},
 			},
@@ -929,16 +930,29 @@ func registerGLFWClasses() error {
 }
 
 // theGoWindows associates ObjC delegate and content-view instances with their Go Windows.
-var theGoWindows = map[objc.ID]*Window{}
+var (
+	theGoWindows  = map[objc.ID]*Window{}
+	theGoWindowsM sync.RWMutex
+)
 
 // getGoWindow returns the Go Window associated with an ObjC instance, or nil if there is none.
 func getGoWindow(id objc.ID) *Window {
+	theGoWindowsM.RLock()
+	defer theGoWindowsM.RUnlock()
 	return theGoWindows[id]
 }
 
 // setGoWindow associates an ObjC instance with a Go Window.
 func setGoWindow(id objc.ID, window *Window) {
+	theGoWindowsM.Lock()
+	defer theGoWindowsM.Unlock()
 	theGoWindows[id] = window
+}
+
+func deleteGoWindow(id objc.ID) {
+	theGoWindowsM.Lock()
+	defer theGoWindowsM.Unlock()
+	delete(theGoWindows, id)
 }
 
 // handleMouseMoved processes mouse movement events.
@@ -1210,7 +1224,7 @@ func (w *Window) platformDestroyWindow() error {
 
 	if w.platform.delegate != 0 {
 		w.platform.object.Send(sel_setDelegate, 0)
-		delete(theGoWindows, w.platform.delegate)
+		deleteGoWindow(w.platform.delegate)
 		w.platform.delegate.Send(sel_release)
 		w.platform.delegate = 0
 	}
--- a/internal/glfw/cocoa_window_darwin.go
+++ b/internal/glfw/cocoa_window_darwin.go
@@ -10,7 +10,6 @@ import (
 	"image"
 	"math"
 	"runtime"
-	"sync"
 	"unsafe"
 
 	"github.com/ebitengine/purego/objc"
@@ -659,7 +658,7 @@ func registerGLFWClasses() error {
 						window.platform.markedText.Send(sel_release)
 						window.platform.markedText = 0
 					}
-					deleteGoWindow(self)
+					delete(theGoWindows, self)
 					self.SendSuper(objc.RegisterName("dealloc"))
 				},
 			},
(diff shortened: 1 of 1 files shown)


hajimehoshi:
Thanks for the patch. I went looking for the race and could not find a path that reaches `theGoWindows` off the main thread, so I would like to see the `-race` report before merging this.

**Every access looks main-thread-confined.**

Writers: `setGoWindow` is called only from `createNativeWindow`, reached via `glfw.CreateWindow` in `(*glfwBackend).createWindow` ("must be called from the main thread"), which `initOnMainThread` runs inside `u.mainThread.Call` (`internal/ui/ui_glfw.go`). The `delete` is reached only from `platformDestroyWindow`, whose sole caller is `glfw.Terminate()` in `loopGame`'s deferred cleanup, also inside `u.mainThread.Call`.

Readers: every `getGoWindow` call sits inside an `objc.MethodDef` body registered in `registerGLFWClasses`. AppKit delivers window and view delegate messages on the main thread, and the pump is `platformPollEvents`; all eight `glfw.PollEvents()` call sites in `internal/ui` are in functions documented main-thread-only and dispatched through `mainThread.Call`.

`dealloc` is not an exception: the only releases of the delegate and the content view are the explicit `Send(sel_release)` calls in `platformDestroyWindow`, on the main thread. There is no `dispatch_async`, `NSThread`, or display-link callback anywhere in the darwin GLFW files. The one glfw entry point Ebitengine calls outside `mainThread.Call` is `PostEmptyEvent`, and it never touches this map.

**Even if a callback did arrive off-thread, guarding the map alone would not fix it.** `windowDidResize:` reads the global `_glfw.platformWindow.disabledCursorWindow`, writes `window.platform.maximized`, and through `updateWindowSize` writes `window.platform.width`/`height` and calls `window.inputWindowSize`, none of it synchronized, while `platformDestroyWindow` concurrently clears `disabledCursorWindow`, zeroes `w.platform.delegate`, and releases the ObjC objects. The lock would hand back a `*Window` whose ObjC objects are being torn down, turning a detectable race into a silent use-after-free. It also costs an `RWMutex` round trip on every callback, including the per-event mouse and key path.

So: do you have a race-detector report for this? If you do, it points at a confinement violation I did not find, and the fix should target that path (or the `*Window` state) rather than the map. If not, I think the accurate change is a comment on `theGoWindows` recording the main-thread invariant, since that invariant is what actually keeps it safe.

For the record, the patch itself is mechanically fine: all five accesses are converted with none left raw, no lock is held across an ObjC `Send` so there is no deadlock, and on your head commit `go build ./...` is clean, `go test -race ./internal/glfw/ ./internal/ui/` passes, and `go vet ./internal/glfw/` reports only the pre-existing `unsafe.Pointer` findings.

---

kumagi:
Thanks for the careful analysis.

I don't have a `-race` report to show. I couldn't produce one either; the original patch came from reading the code rather than from a detector run. I re-checked every path you listed and I agree that the map is main-thread-confined: reads all happen inside `objc.MethodDef` bodies registered by `registerGLFWClasses`, which AppKit delivers on the main thread (and the pump, `platformPollEvents`, is only driven from `mainThread.Call`), and the writes happen in `createNativeWindow` and `platformDestroyWindow`, likewise dispatched through `u.mainThread.Call`. `platformPostEmptyEvent` is the only glfw entry point documented as callable from any goroutine, and it never touches the map. I also confirmed there is no `dispatch_async`/`NSThread` path in the darwin glfw files that could release the delegate or the content view off the main thread, so `dealloc` stays on the main thread too.

Your second point is right as well: even if a callback did arrive off-thread, locking the map alone would just hand back a `*Window` whose `platform.*` fields and ObjC objects are being torn down concurrently, which trades a detectable race for a silent use-after-free, and it would cost an `RWMutex` round trip on every mouse/key callback.

So I've dropped the mutex and the helpers, and recorded the invariant instead: a comment on `theGoWindows` explaining that every access must be on the main thread and why no synchronization is needed, plus `must be called on the main thread` notes on `getGoWindow`/`setGoWindow`. PTAL.

hajimehoshi:
Thanks for turning this around. The revert is clean: the `sync` import is gone, both raw `delete(theGoWindows, ...)` calls are back, `deleteGoWindow` is removed, and the net diff against main is +11/-0, comments only. On `9471c90c1`: `go build ./...` clean, `go test -race ./internal/glfw/ ./internal/ui/` passes, `gofmt` clean, `internal/vettools` clean.

This is the right shape of change. Two things in the comment text are worth fixing before it lands.

**1. The reader clause is too narrow, and it contradicts itself.**

> `...which AppKit invokes on the main thread as it delivers the events to platformPollEvents.`

Many of the registered callbacks fire synchronously from direct AppKit sends inside glfw itself, not from event delivery:

| Send | Fires |
| --- | --- |
| `setFrame:display:` (:1323, :1471, :1526), `zoom` (:1084, :1405) | `windowDidResize:` |
| `center` (:1064), `setFrameOrigin:` (:1293) | `windowDidMove:` |
| `setContentView:` (:1108) | `updateTrackingAreas`, `viewDidChangeBackingProperties` |
| `orderOut:` (:1207, :1430) | `windowDidChangeOcclusionState:` |
| `release` (:1225, :1232) | `dealloc` |

That last row is the contradiction: the same comment names `dealloc` two lines earlier as an entry-removal path, and `dealloc` is triggered by `Send(sel_release)` in `platformDestroyWindow`, never by `platformPollEvents`.

The actual invariant is stronger and shorter: AppKit invokes these on whatever thread triggers them, and the whole package is main-thread-only, so it is always the main thread. Dropping the `platformPollEvents` clause fixes it.

**2. Phrasing.** `must be called on the main thread` (three places) against the repo's `must be called from the main thread`: 48 uses to 7 across `internal/`, and this package's own `x11_ime_linbsd.go` uses "from" throughout.

**Nits**

- "like the rest of the glfw package" points at a convention that is not written down anywhere: `internal/glfw` has no package doc. True, but there is nothing to reference.
- The map comment enumerates every writer and reader over five lines, then both helpers restate the rule. A tighter version:

  ```go
  // theGoWindows associates ObjC delegate and content-view instances with their Go Windows.
  //
  // theGoWindows must be accessed only from the main thread, like the rest of this package:
  // the entries are written and removed while a window is created or destroyed, and read
  // from the ObjC callbacks, which AppKit invokes on the thread that triggers them.
  // Thus no synchronization is needed here.
  ```

  Whether the per-helper lines stay is your call; they match `internal/ui`'s convention, so keeping them is defensible.
- The branch is two commits where the second undoes the first, so `d6b9c3549` ("bug fix: a data race...") would land in `main` describing something that no longer exists. Please squash into one documentation-titled commit. The second commit's body is also unwrapped (238- and 189-character lines) against the 72-column convention.
- The PR title and body are stale: they still say "bug fix: a data race in theGoWindows on macOS", type "bug", and describe the RWMutex fix.
- The branch is based on `0d78b6aae`, three commits behind `main`; none of them touch this file, so there is no conflict.

---

kumagi:
Thanks for the precise review. Both points are addressed:

1. The reader clause is fixed: the `platformPollEvents` wording is gone, and the map comment now uses the tighter form — the entries are written and removed while a window is created or destroyed, and read from the ObjC callbacks, which AppKit invokes on the thread that triggers them. This removes the `dealloc` contradiction as well.
2. All three occurrences now read "must be called from the main thread", matching the wording used elsewhere in the repository.

I kept the per-helper lines since they match the `internal/ui` convention.

The history points are addressed too: the change lands as a single documentation-titled commit with the body wrapped at 72 columns, and the branch is brought up to date with main. The net diff is comments only (+9/-0). The PR title and body will be updated to describe the documentation-only change.

Verified on the new head: `go build ./...` and `GOOS=darwin go build ./...` are clean, `gofmt` is clean, and `go vet ./internal/glfw/` reports only the pre-existing `unsafe.Pointer` findings.

hajimehoshi (approved):
LGTM

PR #3641, opened 2026-09-03
fix typos in the doc comments and an error message

kumagi:
None

bug

## What this PR does | solves
- monitor.go, run.go: the doc comments of Monitor.Size and the deprecated ScreenSizeInFullscreen referred to a nonexistent SetSize function. The actual API is SetWindowSize.
- blend.go: fixed a garbled sentence for Blend ('source and color destination color values'), and unified '1 - (source color)' with '(source RGBA)' that the counterpart constants use.
- doc.go: added a missing period that merged two unrelated sentences about the ebitenginesinglethread build tag.
- vector/pathlegacy.go: 'vertice's' is ungrammatical; use vertices'.
- internal/glfw: the error-message prefix 'glwfwin:' was a typo of 'glfw:' that the package uses everywhere else.
- internal/ui: 'Do nothings.' in no-op stubs; use 'Do nothing.'
- internal/shader: 'unepxected' type, 'ivenN' (ivecN), and 'veraibles' (variables) in an error message, a doc comment, and a comment.

### DIFF
--- a/monitor.go
+++ b/monitor.go
@@ -39,7 +39,7 @@ func (m *MonitorType) DeviceScaleFactor() float64 {
 
 // Size returns the size of the monitor in device-independent pixels.
 // This is the same as the screen size in fullscreen mode.
-// The returned value can be given to SetSize function if the perfectly fit fullscreen is needed.
+// The returned value can be given to SetWindowSize function if the perfectly fit fullscreen is needed.
 //
 // On mobiles, Size returns (0, 0) before the game starts e.g. in init functions.
 //
--- a/monitor.go
+++ b/monitor.go
@@ -39,7 +39,7 @@ func (m *MonitorType) DeviceScaleFactor() float64 {
 
 // Size returns the size of the monitor in device-independent pixels.
 // This is the same as the screen size in fullscreen mode.
-// The returned value can be given to SetWindowSize function if the perfectly fit fullscreen is needed.
+// The returned value can be given to [SetWindowSize] if the perfectly fit fullscreen is needed.
 //
 // On mobiles, Size returns (0, 0) before the game starts e.g. in init functions.
 //
--- a/run.go
+++ b/run.go
@@ -404,7 +404,7 @@ func ScreenSize() (int, int) {
 
 // ScreenSizeInFullscreen returns the size in device-independent pixels when the game is fullscreen.
 // The adopted monitor is the 'current' monitor which the window belongs to.
-// The returned value can be given to SetWindowSize function if the perfectly fit fullscreen is needed.
+// The returned value can be given to [SetWindowSize] if the perfectly fit fullscreen is needed.
 //
 // On browsers, ScreenSizeInFullscreen returns the 'window' (global object) size, not 'screen' size.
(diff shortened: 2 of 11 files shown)


monitor.go
```diff
@@ -39,7 +39,7 @@ func (m *MonitorType) DeviceScaleFactor() float64 {
 
 // Size returns the size of the monitor in device-independent pixels.
 // This is the same as the screen size in fullscreen mode.
-// The returned value can be given to SetSize function if the perfectly fit fullscreen is needed.
+// The returned value can be given to SetWindowSize function if the perfectly fit fullscreen is needed.
```
hajimehoshi: Nit: since the name is being corrected anyway, could you make it a doc link? The package already uses them (`colormode.go`, `run.go`, `window.go`), it renders as a link on pkg.go.dev, and it sidesteps the missing article in "given to SetWindowSize function".

```go
// The returned value can be given to [SetWindowSize] if the perfectly fit fullscreen is needed.
```

run.go
```diff
@@ -404,7 +404,7 @@ func ScreenSize() (int, int) {
 
 // ScreenSizeInFullscreen returns the size in device-independent pixels when the game is fullscreen.
 // The adopted monitor is the 'current' monitor which the window belongs to.
-// The returned value can be given to SetSize function if the perfectly fit fullscreen is needed.
+// The returned value can be given to SetWindowSize function if the perfectly fit fullscreen is needed.
```
hajimehoshi: Same here.

hajimehoshi (commented):
Thanks! Every fix here is correct, and I confirmed each one is the only remaining occurrence in the tree. A couple of them are more than cosmetic: the missing period in `doc.go` actually changed the rendering, since godoc joins consecutive lines into one paragraph, and `glfw:` is the right prefix because the other seven errors in `win32_window_windows.go` use it (and `cocoa_window_darwin.go` carries the identical message).

Two nits, neither blocking.

**Commit message prefix.** `internal` is not a package; the convention here is real package paths (`internal/ui`, `internal/shader`, `internal/glfw`). Spelling all five out would run long, and the precedent for a mechanical sweep of this shape is `all:` — `432d03f82 all: fix typos`, `d8f8a8f1d all: fix wrong names and typos in error messages`, `a8373db4c all: fix doc comments that diverge from the behavior (#3608)`. Could you use:

```
all: fix typos in doc comments and an error message
```

(that also drops the stray article in "the doc comments").

The other nit is inline, on `monitor.go` and `run.go`.

monitor.go
kumagi: Done. The comment now reads:

```go
// The returned value can be given to [SetWindowSize] if the perfectly fit fullscreen is needed.
```

run.go
kumagi: Done. ScreenSizeInFullscreen's comment now uses the [SetWindowSize] doc link as well.

kumagi:
Thanks for the review! I addressed the inline nit: the references to SetWindowSize in monitor.go and run.go are now doc links ([SetWindowSize]) as suggested, which also sidesteps the missing article.

hajimehoshi (approved):
LGTM

PR #3645, opened 2026-09-04
internal/shader: reject the blank identifier used as a value

kumagi:
None

bug

## What this PR does | solves
In a context where a local variable is not marked as used (a constant declaration or an assignment left-hand side), parsing "_" returned an expression with Type Blank and a nil Const. When such an expression was used as a value, e.g. as an argument of min/max/clamp or an operand of a binary operator, constant folding dereferenced the nil Const and panicked with a nil pointer dereference, crashing the compiler on user input.

Reject a blank identifier used as a value in call arguments and binary operands, reporting "cannot use _ as value" instead.

Add TestCompileBlankAsValue, which panics without this fix.

### DIFF
--- a/internal/shader/expr.go
+++ b/internal/shader/expr.go
@@ -98,6 +98,11 @@ func (cs *compileState) parseExpr(block *block, fname string, expr ast.Expr, mar
 		}
 		rhst := ts[0]
 
+		if lhs[0].Type == shaderir.Blank || rhs[0].Type == shaderir.Blank {
+			cs.addError(e.Pos(), "cannot use _ as value")
+			return nil, nil, nil, false
+		}
+
 		op := e.Op
 		// https://pkg.go.dev/go/constant/#BinaryOp
 		// "To force integer division of Int operands, use op == token.QUO_ASSIGN instead of
@@ -221,6 +226,10 @@ func (cs *compileState) parseExpr(block *block, fname string, expr ast.Expr, mar
 			for _, expr := range es {
 				if expr.Type == shaderir.FunctionExpr || expr.Type == shaderir.BuiltinFuncExpr {
 					cs.addError(e.Pos(), fmt.Sprintf("function name cannot be an argument: %s", e.Fun))
+					return nil, nil, nil, false
+				}
+				if expr.Type == shaderir.Blank {
+					cs.addError(e.Pos(), "cannot use _ as value")
 					return nil, nil, nil, false
 				}
 			}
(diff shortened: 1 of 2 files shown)


hajimehoshi:
The new checks still allow a blank identifier wrapped in a unary expression to reach constant evaluation. For example, this still panics on a7aa5c89cd4b03a814fb258d90b0795d53f3ffd1:

```go
package main

func Fragment() vec4 {
    const x = min(-_, 1)
    return vec4(x)
}
```

Parsing `-_` produces a `shaderir.Unary` expression with a nil `Const`, so the `shaderir.Blank` check misses it and `resolveConstKind` dereferences nil. I also reproduced panics with `max(+_, 1)`, `clamp(-_, 0, 1)`, and `1 + (-_)` in the constant initializer.

Could the unary-expression parser reject blank operands as well, with regression cases for these forms? The existing shader and shaderir tests pass, but these additional cases panic.

Review comment authored by Codex, on behalf of @hajimehoshi.

kumagi:
Good catch: `-_` was parsed into a `shaderir.Unary` expression with a nil constant, so the blank checks for binary operands and call arguments were bypassed and constant folding dereferenced nil.

The unary-expression parser now rejects a blank operand with `cannot use _ as value` as well. `TestCompileBlankAsValue` is now table-driven over the constant initializer, so besides the original cases it covers `min(-_, 1)`, `max(+_, 1)`, `clamp(-_, 0, 1)` and `1 + (-_)`. Each of these now reports a compile error instead of panicking.

`go build ./...`, `go test ./internal/shader/... ./internal/shaderir/...`, `go vet` and `gofmt` are all clean. PTAL.

hajimehoshi:
Resolve conflicts

kumagi:
Conflicts resolved: the branch is now based on the latest main. The only conflict was in internal/shader/shader_test.go, where both tests were appended at the end of the file, so I kept TestCompileVarCountMismatch and TestCompileBlankAsValue. Compared with main, the change is still just the blank-identifier rejections in internal/shader/expr.go and the table-driven TestCompileBlankAsValue. `go build ./...`, `go test ./internal/shader/... ./internal/shaderir/...`, `go vet` and `gofmt` are clean.

hajimehoshi:
The current change seems broken. Please fix this

kumagi:
Fixed. The problem was that the earlier main sync had not been recorded as an actual merge, so the branch still conflicted with TestCompileVarCountMismatch (which has since been merged into main as #3626) and the pull request re-displayed main's changes as part of this PR.

I resolved the remaining conflict: the branch now adds only TestCompileBlankAsValue, placed away from the end of internal/shader/shader_test.go so it no longer collides with main's TestCompileVarCountMismatch. I verified the branch merges cleanly with the latest main, and compared with main the change is just the blank-identifier rejection in internal/shader/expr.go and TestCompileBlankAsValue. No logic changes. go build ./..., go test ./internal/shader/... ./internal/shaderir/..., go vet and gofmt are all clean. PTAL.

hajimehoshi (approved):
LGTM

PR #3661, opened 2026-09-04
all: fix typos and doc comments that diverge from the behavior

kumagi:
None

bug

## What this PR does | solves
Fix documentation and comment issues only; no behavior is changed.

Typos and grammar in comments:
- internal/shader/expr.go, internal/shader/stmt.go: "Prase" -> "Parse"
- internal/shaderir/hlsl/uniform.go: "DWROD" -> "DWORD"
- internal/shaderir/program.go: "the number if" -> "the number is"
- text/v2/layout.go: "interates" -> "iterates"
- text/v2/internal/oksvg/public.go: "SFG" -> "SVG"
- internal/ui/ui_js.go: "tha game" -> "the game" (3 places)
- internal/ui/input_js.go: "distinghlish" -> "distinguish"
- audio/audio.go: "Underlying playering can be the pause state after fishing its playing" -> "An underlying player can be in the pause state after finishing its playing"; "16bits" -> "16-bit"
- audio/player.go: "undelrying" -> "underlying"
- internal/affine/colorm.go, colorm.go, colorm/colorm.go: "YCrCb" -> "YCbCr" (3 places)
- graphics.go: "automatic choose" -> "automatic choice"
- ebitenutil/fs.go: "NewImageFromFileSystem create" -> "creates"
- stencilbuffer.go: "draw triangles" -> "draws triangles" (2 places)
- window.go: "prevent the window is automatically closed" ->
  "prevent the window from being automatically closed"; "as of v2.3, Use" -> "as of v2.3. Use"
- run.go: "before calling capturing a cursor" -> "before capturing a cursor"
- playstation5: rename doc_playstaton5.go -> doc_playstation5.go

Doc comments diverging from the behavior:
- internal/shaderir/program.go: FragmentFunc takes len(varyings) + 1 pseudo params, not + 2
- monitor.go: SetMonitor can be called before or after RunGame, not "Run" which doesn't exist
- run.go: SingleThread is an option, not a build tag; document that LayoutF may panic on non-positive numbers as Layout already does
- internal/debug/debug_ebitenginedebug.go: "SwitchLogger" doesn't exist -> "SwitchFrameLogger"
- audio/wav/decode.go: drop "Seek can take long since decoding is a relatively heavy task"; WAV seek is a byte copy and doesn't decode
- exp/vmhost/render.go: render is called from the session goroutine; only dispose must be called within the host's frame


run.go
```diff
@@ -268,7 +270,7 @@ type RunGameOptions struct {
 	// SingleThread indicates whether the single thread mode is used explicitly or not.
 	// The single thread mode disables Ebitengine's thread safety to unlock maximum performance.
 	// If you use this you will have to manage threads yourself.
-	// Functions like `SetWindowSize` will no longer be concurrent-safe with this build tag.
+	// Functions like `SetWindowSize` will no longer be concurrent-safe with this option.
```
hajimehoshi: The build tag still enables the same non-thread-safe mode when `SingleThread` is false (`options.SingleThread || buildTagSingleThread`). Saying “with this option” implies that only the option removes thread safety. “In single-thread mode” would accurately cover both activation paths.

Posted by Claude (Codex), at the user’s request.

audio/audio.go
```diff
@@ -294,7 +294,7 @@ func (c *Context) updatePlayers() error {
 	c.m.Unlock()
 
 	// Now reader players cannot call removePlayers from themselves in the current implementation.
-	// Underlying playering can be the pause state after fishing its playing,
+	// An underlying player can be in the pause state after finishing its playing,
```
hajimehoshi: The replacement fixes the original typos, but “in the pause state after finishing its playing” is still awkward and unclear. Consider “The underlying player can stop playing after reaching the end of its source” or “can become paused after finishing playback.”

Posted by Claude (Codex), at the user’s request.

run.go
kumagi: Good point — the build tag `ebitenginesinglethread` enables the same mode even when SingleThread is false. I've updated the comment to "Functions like `SetWindowSize` will no longer be concurrent-safe in the single thread mode.", which covers both activation paths.

audio/audio.go
kumagi: Agreed, the wording was still awkward. I've changed it to "The underlying player can become paused after finishing playback,", following your second suggestion.

kumagi:
Thanks for the reviews! Both comments have been addressed:

- run.go: The SingleThread doc comment now says the functions "will no longer be concurrent-safe in the single thread mode", which accurately covers both the option and the `ebitenginesinglethread` build tag activation paths.
- audio/audio.go: The comment now reads "The underlying player can become paused after finishing playback," which is clearer than the previous wording.

No behavior changes; doc comments only.

hajimehoshi (approved):
LGTM

PR #3668, opened 2026-09-05
all: apply modern Go idioms available since Go 1.25

kumagi:
None

bug

## What this PR does | solves
Replace the "delete every key" loops with the clear() builtin in the OpenGL state, mipmap deallocation, and the gamepad backends.

Convert the struct-typed atomic.Value fields (window size limits, the initial window size, and the mobile outside/surface/display info) to atomic.Pointer[T], removing the unchecked type assertions on load. The string title and the error field are left as atomic.Value, as the standard library has no typed atomic for them.

Use strings.Cut instead of strings.Index with manual slicing when both halves of the string are needed.

No behavior is changed.

### DIFF
--- a/internal/ui/ui_darwin.go
+++ b/internal/ui/ui_darwin.go
@@ -475,7 +475,10 @@ func (u *glfwBackend) setNativeFullscreen(fullscreen bool) error {
 }
 
 func (u *glfwBackend) isFullscreenAllowedFromUI(mode WindowResizingMode) bool {
-	s := u.desktopWindow.windowSizeLimit.Load().(windowSizeRange)
+	s := u.desktopWindow.windowSizeLimit.Load()
+	if s == nil {
+		return true
+	}
 	if s.maxWidthInDIP != glfw.DontCare || s.maxHeightInDIP != glfw.DontCare {
 		return false
 	}
--- a/internal/ui/ui_darwin.go
+++ b/internal/ui/ui_darwin.go
@@ -476,9 +476,6 @@ func (u *glfwBackend) setNativeFullscreen(fullscreen bool) error {
 
 func (u *glfwBackend) isFullscreenAllowedFromUI(mode WindowResizingMode) bool {
 	s := u.desktopWindow.windowSizeLimit.Load()
-	if s == nil {
-		return true
-	}
 	if s.maxWidthInDIP != glfw.DontCare || s.maxHeightInDIP != glfw.DontCare {
 		return false
 	}
--- a/internal/ui/ui_mobile.go
+++ b/internal/ui/ui_mobile.go
@@ -46,7 +46,7 @@ func (u *UserInterface) init() error {
 		errCh:                 make(chan error),
 	}
 	// Give a default outside size so that the game can start without initializing them.
-	u.userInterfaceImpl.outsideSize.Store(pointF{x: 640, y: 480})
+	u.userInterfaceImpl.outsideSize.Store(&pointF{x: 640, y: 480})
 	u.foreground.Store(true)
 	return nil
 }
@@ -96,11 +96,11 @@ type userInterfaceImpl struct {
 	graphicsDriver        graphicsdriver.Graphics
 	graphicsLibraryInitCh chan struct{}
 
-	outsideSize atomic.Value
+	outsideSize atomic.Pointer[pointF]
 
 	// surfaceSize is the size of the rendering surface, in pixels, as the platform reports it. It is
 	// unset when the platform reports no size, and the size is then derived from the outside size.
-	surfaceSize atomic.Value
+	surfaceSize atomic.Pointer[pointI]
 
 	foreground atomic.Bool
 	errCh      chan error
@@ -179,7 +179,10 @@ func (u *UserInterface) runMobile(game Game, options *RunOptions) (err error) {
 
 // outsideSize must be called on the same goroutine as update().
 func (u *UserInterface) outsideSize() (float64, float64) {
-	s := u.userInterfaceImpl.outsideSize.Load().(pointF)
+	s := u.userInterfaceImpl.outsideSize.Load()
+	if s == nil {
+		return 0, 0
+	}
 	return s.x, s.y
 }
 
@@ -202,7 +205,7 @@ func (u *UserInterface) update() error {
 //
 // screenSize must be called on the same goroutine as update().
 func (u *UserInterface) screenSize(outsideWidth, outsideHeight float64, deviceScaleFactor float64) (int, int) {
-	if s, ok := u.userInterfaceImpl.surfaceSize.Load().(pointI); ok {
+	if s := u.userInterfaceImpl.surfaceSize.Load(); s != nil {
 		return s.x, s.y
 	}
 	// The platform reports no surface size, so derive it from the view's size in device-independent
@@ -215,14 +218,14 @@ func (u *UserInterface) screenSize(outsideWidth, outsideHeight float64, deviceSc
 //
 // SetSurfaceSize is concurrent safe.
 func (u *UserInterface) SetSurfaceSize(width, height int) {
-	u.userInterfaceImpl.surfaceSize.Store(pointI{x: width, y: height})
+	u.userInterfaceImpl.surfaceSize.Store(&pointI{x: width, y: height})
 }
 
 // SetOutsideSize is called from mobile/ebitenmobileview.
 //
 // SetOutsideSize is concurrent safe.
 func (u *UserInterface) SetOutsideSize(outsideWidth, outsideHeight float64) {
-	u.userInterfaceImpl.outsideSize.Store(pointF{x: outsideWidth, y: outsideHeight})
+	u.userInterfaceImpl.outsideSize.Store(&pointF{x: outsideWidth, y: outsideHeight})
 	u.refreshDisplayInfo()
 }
 
@@ -297,13 +300,13 @@ type displayInfoValues struct {
 	scale  float64
 }
 
-var theDisplayInfo atomic.Value
+var theDisplayInfo atomic.Pointer[displayInfoValues]
 
 func (u *UserInterface) displayInfo() (int, int, float64, bool) {
 	// Reading the display info here would require waiting for the main thread
 	// on iOS, which can deadlock.
-	v, ok := theDisplayInfo.Load().(displayInfoValues)
-	if !ok {
+	v := theDisplayInfo.Load()
+	if v == nil {
 		return 0, 0, 1, false
 	}
 	width := int(math.Round(dipFromNativePixels(v.width, v.scale)))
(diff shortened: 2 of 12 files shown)


hajimehoshi (commented):
Thanks for the cleanup. I built the branch for js/wasm, windows/amd64, linux/amd64 and darwin/arm64, ran `go vet` with `internal/vettools`, and ran the tests for the touched packages: all clean. The `clear()` sweep is complete, too. The `delete`-in-a-loop sites that remain (`inpututil`, the DirectX shader caches, `gl/default_js.go`) are conditional deletes or delete-from-another-map, where `clear` doesn't apply. Both `strings.Cut` conversions preserve the original conditions, including the `valueIndex == len(v)-1` case in `oksvg`.

Two notes on the framing, and then the one change I'd like before merging.

The title's premise doesn't hold: none of these are Go 1.25 features. `clear` is Go 1.21, `atomic.Pointer[T]` is Go 1.19, and `strings.Cut` is Go 1.18. The `go 1.25.0` line in go.mod isn't what makes them available. Something like `all: use clear, atomic.Pointer and strings.Cut` would describe the change accurately.

The PR is also filed as a bug while the body says no behavior changes; it's a cleanup. And it bundles three unrelated mechanical changes. Splitting them into three commits would isolate the atomic conversion, which is the only part carrying any risk.

The change I'd ask for: please drop the four new `if x == nil` branches (`ui_darwin.go`, `ui_mobile.go`, and two in `window_desktop.go`). Details are inline, but in short, the old unchecked assertions encoded an invariant that `init()` establishes, `atomic.Pointer` preserves that invariant for free through the nil dereference, and the fallbacks turn a would-be panic into a silently wrong window size, outside size, or fullscreen decision.

internal/ui/ui_darwin.go
```diff
@@ -475,7 +475,10 @@ func (u *glfwBackend) setNativeFullscreen(fullscreen bool) error {
 }
 
 func (u *glfwBackend) isFullscreenAllowedFromUI(mode WindowResizingMode) bool {
-	s := u.desktopWindow.windowSizeLimit.Load().(windowSizeRange)
+	s := u.desktopWindow.windowSizeLimit.Load()
+	if s == nil {
+		return true
+	}
```
hajimehoshi: The unchecked assertion here was deliberate rather than an oversight: `windowSizeLimit` is stored by `desktopWindow.init()`, which `newUserInterface()` runs before `theUI` is observable, so it is never unset. `atomic.Pointer` keeps that invariant enforced for free, since a nil load panics on the next field access exactly as the failed type assertion did. So this branch is unreachable, and if the invariant ever did break it would swap a loud panic for a silently wrong answer. Here that answer is `true`, i.e. fullscreen *allowed*, the permissive direction.

Could you drop the check and keep just `s := u.desktopWindow.windowSizeLimit.Load()` followed by the existing field reads?

The same applies to the three other new nil branches (two in `window_desktop.go`, one in `ui_mobile.go`). Worth noting that `ui_glfw.go:583` still does `u.desktopWindow.title.Load().(string)` unchecked under the identical invariant, so the added checks aren't consistent with the surrounding code either.

By contrast, the `surfaceSize` and `theDisplayInfo` conversions in `ui_mobile.go` are right: those genuinely were comma-ok before, because the value really can be unset.

internal/ui/ui_mobile.go
```diff
@@ -179,7 +179,10 @@ func (u *UserInterface) runMobile(game Game, options *RunOptions) (err error) {
 
 // outsideSize must be called on the same goroutine as update().
 func (u *UserInterface) outsideSize() (float64, float64) {
-	s := u.userInterfaceImpl.outsideSize.Load().(pointF)
+	s := u.userInterfaceImpl.outsideSize.Load()
+	if s == nil {
+		return 0, 0
+	}
```
hajimehoshi: Same as the note on `ui_darwin.go`: `outsideSize` is stored with the 640x480 default in `init()` before the game can run, so this branch is unreachable, and `0, 0` would feed a zero-sized layout instead of failing loudly.

internal/ui/window_desktop.go
```diff
@@ -62,23 +62,27 @@ var _ Window = (*desktopWindow)(nil)
...
 func (w *desktopWindow) getWindowSizeLimitsInDIP() (minw, minh, maxw, maxh int) {
 	if microsoftgdk.IsXbox() {
 		return glfw.DontCare, glfw.DontCare, glfw.DontCare, glfw.DontCare
 	}
 
-	s := w.windowSizeLimit.Load().(windowSizeRange)
+	s := w.windowSizeLimit.Load()
+	if s == nil {
+		return glfw.DontCare, glfw.DontCare, glfw.DontCare, glfw.DontCare
+	}
```
hajimehoshi: Same as the note on `ui_darwin.go`: unreachable, and returning all `DontCare` would silently drop limits the caller had set.

internal/ui/window_desktop.go
```diff
@@ -183,7 +188,10 @@ func (w *desktopWindow) getInitWindowSizeInDIP() (int, int) {
 		return microsoftgdk.MonitorResolution()
 	}
 
-	pt := w.initWindowSizeInDIP.Load().(image.Point)
+	pt := w.initWindowSizeInDIP.Load()
+	if pt == nil {
+		return 0, 0
+	}
```
hajimehoshi: Same as the note on `ui_darwin.go`: unreachable, and `0, 0` here would silently become the initial window size.

internal/ui/window_desktop.go
```diff
@@ -88,13 +92,14 @@ func (w *desktopWindow) setWindowSizeLimitsInDIP(minw, minh, maxw, maxh int) boo
...
-	newS := windowSizeRange{
+	newS := &windowSizeRange{
 		minWidthInDIP:  minw,
 		minHeightInDIP: minh,
 		maxWidthInDIP:  maxw,
 		maxHeightInDIP: maxh,
 	}
-	return w.windowSizeLimit.Swap(newS) != newS
+	old := w.windowSizeLimit.Swap(newS)
+	return old == nil || *old != *newS
```
hajimehoshi, line 99: This one is worth calling out as correct. `Swap(newS) != newS` on `atomic.Value` compared the struct *values*; the direct pointer translation would compare freshly allocated pointers, always report "changed", and reset the GLFW size limits on every call. Dereferencing both sides keeps the original meaning, and `windowSizeRange` is comparable, so it holds.

text/v2/internal/chunk/gen.go
```diff
@@ -100,16 +100,15 @@ func fetchProps() (map[string][]entry, error) {
...
 			continue
 		}
-		cpField := strings.TrimSpace(line[:semi])
-		rest := line[semi+1:]
+		cpField = strings.TrimSpace(cpField)
 		var prop, name string
-		if hash := strings.Index(rest, "#"); hash >= 0 {
-			prop = strings.TrimSpace(rest[:hash])
-			name = strings.TrimSpace(rest[hash+1:])
+		if cutoff, after, ok := strings.Cut(rest, "#"); ok {
```
hajimehoshi: `cutoff` names the part *before* the cut, which reads backwards. `before`/`after`, or assigning to `prop`/`name` directly, would be clearer. This `ok` also shadows the one from the `strings.Cut(line, ";")` a few lines up.

text/v2/internal/chunk/gen.go
```diff
@@ -129,16 +128,16 @@ func fetchProps() (map[string][]entry, error) {
 // parseCodepoints accepts "NNNN" or "NNNN..MMMM" and returns the
 // inclusive range; for a single codepoint, start == end.
 func parseCodepoints(s string) (rune, rune, error) {
-	if i := strings.Index(s, ".."); i >= 0 {
-		a, err := strconv.ParseInt(s[:i], 16, 32)
+	if s, b, ok := strings.Cut(s, ".."); ok {
```
hajimehoshi: `s` here shadows the function's own parameter with the before-half, which then forces the parsed int to be named `bn` because `b` is taken by the string half. `before, after, ok := strings.Cut(s, "..")` avoids both.

internal/ui/ui_darwin.go
kumagi: Done — the nil branch is dropped and the function now just loads the pointer and reads the fields, so a broken invariant panics instead of silently allowing fullscreen. Good catch on `ui_glfw.go:583` too; that confirms the checks weren't consistent with the surrounding code.

internal/ui/ui_mobile.go
kumagi: Dropped. `init()` stores the 640x480 default before the game can run, so `outsideSize` now dereferences the loaded pointer directly and a zero-sized layout can't be silently fed to the game.

internal/ui/window_desktop.go
kumagi: Dropped — `getWindowSizeLimitsInDIP` now loads the pointer and returns the fields directly, so caller-set limits can't silently degrade to all `DontCare`.

internal/ui/window_desktop.go
kumagi: Dropped as well; `getInitWindowSizeInDIP` now returns `pt.X, pt.Y` straight from the loaded pointer, so a broken invariant panics rather than producing a 0x0 initial window size.

internal/ui/window_desktop.go
kumagi: Thanks for double-checking that one. I kept the comparison on the pointed-to values (`old == nil || *old != *newS`) so `setWindowSizeLimitsInDIP` still reports "changed" only when the limits actually differ, instead of resetting the GLFW size limits on every call.

text/v2/internal/chunk/gen.go
kumagi: Fixed — it's now `prop, name, _ := strings.Cut(rest, "#")` followed by `strings.TrimSpace` on each half, which assigns straight into `prop`/`name`, drops the misleading `cutoff`, and removes the inner `ok` shadow. Behavior is unchanged: with no `#`, `name` is empty and `prop` gets the whole remainder, as before.

text/v2/internal/chunk/gen.go
kumagi: Fixed — it's now `if before, after, ok := strings.Cut(s, ".."); ok`, so the parameter is no longer shadowed and the parsed codepoints can go back to being named `a` and `b`.

kumagi:
Thanks for the careful review!

Dropped all four new `if x == nil` branches (`ui_darwin.go`, `ui_mobile.go`, and the two in `window_desktop.go`). You're right that the unchecked load encoded an `init()` invariant that `atomic.Pointer` keeps enforcing for free through the nil dereference, and the fallbacks would have turned a loud panic into a silently wrong fullscreen decision, outside size, window size limit, or initial window size. The `surfaceSize` and `theDisplayInfo` checks stay, since those really can be unset.

Also renamed the `strings.Cut` halves in `text/v2/internal/chunk/gen.go` to what they actually are, which removes the `ok` and `s` shadowing as well.

On the framing: agreed on both counts. `clear`/`atomic.Pointer`/`strings.Cut` are Go 1.21/1.19/1.18 features, so the title should just describe the change (`all: use clear, atomic.Pointer and strings.Cut`), and this is a cleanup, not a bug fix. I can split the three mechanical changes into three separate commits to isolate the atomic conversion if you'd prefer that over one commit.

hajimehoshi (approved):
LGTM

PR #3669, opened 2026-09-05
all: use slices.SortFunc and cmp.Compare for sorting

kumagi:
None

bug

## What this PR does | solves
Replace the remaining sort.Slice calls with slices.SortFunc, comparing with cmp.Compare instead of a less function. This covers the GLFW video mode ordering and the chunk generator.

In the key-name generator, use cmp.Compare for the final string comparison to match the cmp.Compare calls already used for the numeric keys.

No behavior is changed.

### DIFF
--- a/genkeys.go
+++ b/genkeys.go
@@ -719,7 +719,7 @@ func keyNamesCmp(k0, k1 string) int {
 	if f0 != -1 && f1 != -1 {
 		return cmp.Compare(f0, f1)
 	}
-	return strings.Compare(k0, k1)
+	return cmp.Compare(k0, k1)
 }
 
 const license = `// Copyright 2013 The Ebitengine Authors
--- a/internal/glfw/monitor.go
+++ b/internal/glfw/monitor.go
@@ -8,7 +8,8 @@
 package glfw
 
 import (
-	"sort"
+	"cmp"
+	"slices"
 )
 
 func abs(x int) uint {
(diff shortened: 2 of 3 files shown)


hajimehoshi (commented):
The code is correct and I found no defects. Two small notes on the description below.

## Verification

- `go build ./...` (darwin), plus cross-builds of `internal/glfw` for windows/linux/netbsd and `./...` for js/wasm: pass
- `go test ./internal/glfw/... ./text/v2/... .`: pass
- `go vet` with the repo's `internal/vettools`: no new findings
- `gofmt -l .`: clean

"No behavior is changed" checks out by regeneration rather than by reading. Both touched generators were re-run against the checked-in output, which was produced by the old comparators:

- `go run genkeys.go` followed by `gofmt -s -w .` reproduces all seven generated key files byte-identically. That is the actual proof for the `strings.Compare` -> `cmp.Compare` swap in `keyNamesCmp`.
- `go run gen.go` in `text/v2/internal/chunk` reproduces `aterm.go` and `sterm.go` byte-identically.

On the two `sort.Slice` -> `slices.SortFunc` conversions: both the old and the new sort are unstable, so no ordering guarantee is lost. In the chunk generator the sort key (`entry.start`) is unique per property within `SentenceBreakProperty.txt`, so there are no ties at all.

One point in the change's favour that the description does not mention: `internal/glfw/monitor.go` is a port of GLFW's C `compareVideoModes`, a `qsort` comparator that returns an `int`. The `slices.SortFunc` form now mirrors the upstream shape instead of inverting it into a less function, which makes future diffing against upstream easier. It also avoids C's `fbpp - sbpp` subtraction idiom, so there is no overflow concern.

## Notes on the description

1. "the remaining `sort.Slice` calls" is not accurate. After this PR, `examples/raycasting/main.go:158` still calls `sort.Slice`, and other `sort` uses remain (`sort.Ints` in `internal/shaderir/reachable_test.go:99`, `sort.Sort(sort.Reverse(sort.IntSlice(...)))` in `examples/2048/2048/tile.go:166` and `:169`). Leaving the examples alone is defensible, but the description should say which calls it covers rather than claiming these are the last ones.
2. The issue type is given as `bug`; this is a cleanup with no behavior change.

## Not a finding

`platformGetVideoMode` reads `m.modes[0]` on Xbox (`internal/glfw/win32_monitor_windows.go:330`), so it depends on which mode sorts first, and two distinct modes can tie under this comparator (same R+G+B sum with a different split, same area, same width, same refresh rate). That tie was already resolved arbitrarily by `sort.Slice`, so this PR neither introduces nor worsens it.

Separately, `GOOS=freebsd go build ./internal/glfw/...` fails inside `purego/internal/fakecgo` (`//go:cgo_export_dynamic ... only allowed in cgo-generated code`), a pre-existing dependency issue under `CGO_ENABLED=0` and unrelated to this diff.

---

hajimehoshi (approved):
LGTM

PR #3672, opened 2026-09-06
ui: preserve window setters during startup

### LINKED ISSUE #3633: ebiten: `SetWindowFloating` and other window setters can be lost when called concurrently with `RunGame`
### Ebitengine Version

v2.10.0-alpha.13 (main, 3b526528785eea4ad3511589ac9920581f22c269)

### Operating System

- [x] Windows
- [x] macOS
- [x] Linux
- [x] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

The affected code is the shared desktop path in `internal/ui`, so every desktop platform is affected. Only macOS was actually exercised.

go version go1.27.0 darwin/arm64

Call one of the window setters from a goroutine while `RunGame` is starting up, so that the call lands after the backend has read the setting and before the backend is published.

```go
package main

import (
	"fmt"
	"time"

)

type Game struct{}

	fmt.Println("floating:", ebiten.IsWindowFloating())
	return ebiten.Termination
}

func (g *Game) Draw(screen *ebiten.Image) {}

func (g *Game) Layout(outsideWidth, outsideHeight int) (int, int) {
	return outsideWidth, outsideHeight
}

func main() {
	go func() {
		time.Sleep(500 * time.Millisecond)
		ebiten.SetWindowFloating(true)
	}()
	if err := ebiten.RunGame(&Game{}); err != nil {
		panic(err)
	}
}
```

As written, this program usually prints `true`, because the interval is short and the call almost always lands outside it. To hit the interval deterministically, widen it: insert `time.Sleep(2 * time.Second)` in `(*glfwBackend).createWindow` immediately before `glfw.CreateWindow`, so the goroutine's call falls between the window hint read in `initOnMainThread` and the `setRunningBackend` call in `createWindow`.

The same is reproducible with `SetWindowSize`, which is easier to observe because the size is read back from the real window rather than from a cached attribute.

`floating: true`, and a window that is actually above the others.

These setters are documented as concurrent-safe, and the call is the only one that touches the setting. Whether it is ordered before or after startup, the window should end up floating.

### What happens instead?

`floating: false`, permanently. The call has no effect at all, with no error and no later correction.

The mechanism:

1. `desktopWindow.SetFloating` calls `runningBackend()`. Startup has not published the backend yet, so it gets nil, stores the value into the init store, and returns.
2. `initOnMainThread` has already read that store to build the `glfw.Floating` window hint, and `createWindow` publishes the backend right after `glfw.CreateWindow`.
3. Nothing reads the init store again while the game runs, and the setter never reached the window. The value is lost.

Affected settings, all read during startup before the publish: initial window size, initial monitor, decorated, floating, and mouse passthrough. Cursor mode, title, fullscreen, resizing mode, size limits, closing-handled, and preferred color mode are consumed after the publish, so they do not have this problem.

Note that this is not a data race. The store goes through an atomic and the race detector does not flag it. It is a lost update.

The root cause is that the setter's check and store are not atomic with respect to the publish, so simply re-reading the init store after `setRunningBackend` is not a fix. A goroutine can be preempted between `runningBackend()` returning nil and its store, and perform the store after any such re-read. That narrows the interval without closing it.

A direction that would actually close it is to make the setter store first and then re-check `runningBackend()`, applying through the backend if one appeared in the meantime. Then either the setter observes the backend or startup observes the store, so at least one of the two applies the value.

Related, and probably worth fixing at the same time: on macOS the decoration is applied from the init store at the first buffer swap, which is after the publish. A `SetWindowDecorated` call made after the publish but before that swap goes to the window and is then overwritten by the stale stored value, which is the same problem in the opposite direction.

Found by reading the code, not from a user report.

---

Filed by Claude (Claude Code), on behalf of @hajimehoshi.

atirna:

Fixes #3633

bug

## What this PR does | solves

Stores the initial decorated, floating, monitor, size, and mouse-passthrough values before checking for a running backend. A setter that overlaps backend startup now either supplies the startup value or also updates the newly running window. This also keeps macOS's deferred decoration setup from restoring an older value.

Verified with `go test ./internal/ui/...`.

### DIFF
--- a/internal/ui/window_desktop.go
+++ b/internal/ui/window_desktop.go
@@ -249,9 +249,9 @@ func (w *desktopWindow) SetDecorated(decorated bool) {
 	if w.ui.isTerminated() {
 		return
 	}
+	w.setInitWindowDecorated(decorated)
 	b := w.ui.runningBackend()
 	if b == nil {
-		w.setInitWindowDecorated(decorated)
 		return
 	}
 	b.Window().SetDecorated(decorated)
@@ -316,9 +316,9 @@ func (w *desktopWindow) SetFloating(floating bool) {
 	if w.ui.isTerminated() {
 		return
 	}
+	w.setInitWindowFloating(floating)
 	b := w.ui.runningBackend()
 	if b == nil {
-		w.setInitWindowFloating(floating)
 		return
 	}
 	b.Window().SetFloating(floating)
@@ -401,9 +401,9 @@ func (w *desktopWindow) SetMonitor(monitor *Monitor) {
 	if w.ui.isTerminated() {
 		return
 	}
+	w.ui.setInitMonitor(monitor)
 	b := w.ui.runningBackend()
 	if b == nil {
-		w.ui.setInitMonitor(monitor)
 		return
 	}
 	b.Window().SetMonitor(monitor)
@@ -454,10 +454,10 @@ func (w *desktopWindow) SetSize(width, height int) {
 	if w.ui.isTerminated() {
 		return
 	}
+	// If the window is initially maximized, the set size is ignored anyway.
+	w.setInitWindowSizeInDIP(width, height)
 	b := w.ui.runningBackend()
 	if b == nil {
-		// If the window is initially maximized, the set size is ignored anyway.
-		w.setInitWindowSizeInDIP(width, height)
 		return
 	}
 	b.Window().SetSize(width, height)
@@ -537,9 +537,9 @@ func (w *desktopWindow) SetMousePassthrough(enabled bool) {
 	if w.ui.isTerminated() {
 		return
 	}
+	w.setInitWindowMousePassthrough(enabled)
 	b := w.ui.runningBackend()
 	if b == nil {
-		w.setInitWindowMousePassthrough(enabled)
 		return
 	}
 	b.Window().SetMousePassthrough(enabled)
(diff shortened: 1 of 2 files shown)


hajimehoshi:
Thanks but this PR doesn't resolve the issue. The issue is a kind of race condition between setting the 'initial' state and getting the 'initial' state and they can happen in parallel, and in the worst case, the set value is ignored.

atirna:
have tried to rework on it again,
based on startup described in #3633. All five setters now store their value before checking for a backend, and startup reads those values again after publishing it. Decorated, floating and mouse passthrough use the existing `applyX` pattern to read the current value on the main thread; size and monitor keep their existing backend operations.

I reproduced the size case with the setter paused after reading a nil backend: an 800×600 request left the window at 640×480 before, and now produces 800×600. Floating and mouse passthrough also retain settings made after their window hints were read. `go test -race ./internal/ui/...` passes.

I added the correction as [174f60613](https://github.com/atirna/ebiten/commit/174f60613711d729842f275898b6d933cb1d7346) on the original PR branch, directly after the original commit.

hajimehoshi:
I've fixed this issue by completely separating setting and applying: 66f98755765879c8e7438aefc84bc7c62881d2fb

atirna:
separating setting from applying is cleaner than what i had. thx.

PR #3673, opened 2026-09-06
internal/shader: emit a return statement for a bare return in a void function

kumagi:
None

bug

## What this PR does | solves

A bare 'return;' in a user-defined void function produced no IR at all, so the early exit disappeared and the following statements were executed. A bare return in a non-void function was not checked either. The backends already supported emitting 'return;'; only the IR supply was missing.

Emit a Return when there are no results and no out-params, and report an error for a bare return in a function that returns a value, following Go.

Add TestCompileBareReturnInVoidFunc and TestCompileBareReturnInNonVoidFunc, and update the array3 expectation, which now ends with a bare return.


hajimehoshi (approved):
LGTM

PR #3674, opened 2026-09-06
audio: replay no stale remainder at the loop joint of an InfiniteLoop

kumagi:
None

bug

## What this PR does | solves
rewind() did not clear i.extra while Seek does, so a remainder left when the source ended in the middle of a value was replayed at the loop joint, shifting all the following data.

Clear the remainder on rewind. Also end the stream when the second lap yields no data, so a loop with nothing to repeat terminates instead of retrying.

Add TestInfiniteLoopRewindClearsExtra, which observes stale bytes at the seam without this fix.


hajimehoshi (approved):
LGTM

PR #3675, opened 2026-09-06
ebiten: allow nil options to Vibrate

kumagi:
None

bug

## What this PR does | solves
Vibrate dereferenced its options unconditionally, so passing nil panicked, while other APIs accept nil options as the zero value.

Treat nil as the zero value in Vibrate and VibrateGamepad.

Add TestVibrateWithNilOptions.

### DIFF
--- a/ebiten_test.go
+++ b/ebiten_test.go
@@ -41,8 +41,3 @@ func TestScreenSizeInFullscreen(t *testing.T) {
 		t.Errorf("h must be positive but not: %d", h)
 	}
 }
-
-func TestVibrateWithNilOptions(t *testing.T) {
-	ebiten.Vibrate(nil)
-	ebiten.VibrateGamepad(0, nil)
-}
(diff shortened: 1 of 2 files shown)


ebiten_test.go
```diff
@@ -41,3 +41,8 @@ func TestScreenSizeInFullscreen(t *testing.T) {
 		t.Errorf("h must be positive but not: %d", h)
 	}
 }
+
+func TestVibrateWithNilOptions(t *testing.T) {
+	ebiten.Vibrate(nil)
```
hajimehoshi: I don't think we need the tests

kumagi: Removed the test.

kumagi:
Thank you for the review! I removed the test as suggested, so this PR now contains only the nil-options handling in Vibrate and VibrateGamepad.

hajimehoshi (approved):
LGTM

PR #3676, opened 2026-09-06
text/v2/internal/oksvg: clamp out-of-range rgb values

kumagi:
None

Bug

## What this PR does | solves
parseColorValue only clamped values above 255, so a negative value wrapped around through uint8 (e.g. rgb(-1,0,0) became 255). Percentage values were not clamped at all (e.g. rgb(1000%,0%,0%) became 246).

Clamp both forms to [0, 255]. The percentage is clamped before scaling so the multiplication cannot overflow.

According to [CSS Color Module Level 4](https://www.w3.org/TR/css-color-4/#rgb-functions)
> Values outside these ranges are not invalid, but are clamped to the ranges defined here at parsed-value time.

Add TestParseSVGColorOutOfRange.


text/v2/internal/oksvg/utils_test.go
```diff
@@ -0,0 +1,45 @@
+// Copyright 2026 The Ebitengine Authors
```
hajimehoshi: I think adding tests in `oksvg_test.go` is fine.

text/v2/internal/oksvg/utils_test.go
```diff
@@ -0,0 +1,45 @@
...
+	for _, test := range []struct {
+		in   string
+		want color.NRGBA
+	}{
+		{in: "rgb(-1,0,0)", want: color.NRGBA{R: 0, G: 0, B: 0, A: 0xff}},
+		{in: "rgb(0,-1,0)", want: color.NRGBA{R: 0, G: 0, B: 0, A: 0xff}},
+		{in: "rgb(256,0,0)", want: color.NRGBA{R: 0xff, G: 0, B: 0, A: 0xff}},
+		{in: "rgb(-1%,0%,0%)", want: color.NRGBA{R: 0, G: 0, B: 0, A: 0xff}},
+		{in: "rgb(1000%,0%,0%)", want: color.NRGBA{R: 0xff, G: 0, B: 0, A: 0xff}},
+		{in: "rgb(100%,0%,0%)", want: color.NRGBA{R: 0xff, G: 0, B: 0, A: 0xff}},
```
hajimehoshi: Use multiple lines

hajimehoshi (commented):
Update README

text/v2/internal/oksvg/utils_test.go
kumagi: Done. I removed utils_test.go and moved TestParseSVGColorOutOfRange into the existing oksvg_test.go.

text/v2/internal/oksvg/utils_test.go
kumagi: Done. Each table entry is now written on multiple lines.

kumagi:
Thank you for the review! I addressed all the comments: the test now lives in oksvg_test.go with each table entry on multiple lines, and I added an entry to the Local modifications section of text/v2/internal/oksvg/README.md describing the rgb clamping.

hajimehoshi (approved):
LGTM

PR #3678, opened 2026-09-06
vector: bug fix: StrokeCircle non-AA overfilled a disk when the width is thick

### LINKED ISSUE #3677: vector.StrokeCircle without antialiasing fills the hole when r <= strokeWidth < 2r
### Ebitengine Version

main branch at 2a73da27e (2026-09-06; after v2.11.0-alpha)

### Operating System

- [ ] Windows
- [ ] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

go1.26.7 linux/amd64

1. Save the program below as `main.go` in an empty directory.
2. Point the `replace` directive at your local Ebitengine checkout (or leave it out and use a released version).
3. Run it with `go mod tidy` and then `xvfb-run -a go run .` (or just `go run .` on a machine with a display).
4. Read the three printed lines.

```go
package main

import (
	"errors"
	"fmt"
	"image/color"
	"log"

	"github.com/hajimehoshi/ebiten/v2/vector"
)

var errDone = errors.New("done")

type Game struct {
	done bool
}

	if g.done {
		return errDone
	}
	return nil
}

func (g *Game) Layout(outsideWidth, outsideHeight int) (int, int) {
	return 192, 64
}

	if g.done {
		return
	}
	g.done = true

	dst := ebiten.NewImage(192, 64)
	// Three stroked circles with the same radius but different widths.
	vector.StrokeCircle(dst, 32, 32, 10, 9, color.White, false)
	vector.StrokeCircle(dst, 96, 32, 10, 10, color.White, false)
	vector.StrokeCircle(dst, 160, 32, 10, 20, color.White, false)

	// The center pixel tells a ring apart from a filled disk.
	for _, tc := range []struct {
		x    int
		want bool // true when the center must be opaque (a filled disk)
	}{
		{32, false}, // w=9:  a ring with a hole.
		{96, false}, // w=10: a ring with a hole of radius 5.
		{160, true}, // w=20: the hole is gone, a filled disk.
	} {
		_, _, _, a := dst.At(tc.x, 32).RGBA()
		fmt.Printf("x=%d: center opaque=%v want=%v\n", tc.x, a != 0, tc.want)
	}
	screen.DrawImage(dst, nil)
}

func main() {
	if err := ebiten.RunGame(&Game{}); err != nil && !errors.Is(err, errDone) {
		log.Fatal(err)
	}
}
```

The middle circle (`r=10, strokeWidth=10`) is a ring whose inner radius is `r - strokeWidth/2 = 5`, so its center pixel must stay transparent, exactly like the antialiased (`antialias=true`) path renders it. Expected output:

```
x=32: center opaque=false want=false
x=96: center opaque=false want=false
x=160: center opaque=true want=true
```

### What happens instead?

The middle circle comes out as a filled disk. Actual output:

```
x=32: center opaque=false want=false
x=96: center opaque=true want=false
x=160: center opaque=true want=true
```

The cause is the non-AA shortcut in `vector/util.go` (`StrokeCircle`): it switches to `FillCircle` when `strokeWidth >= r`, but a stroked circle is a ring with inner radius `r - strokeWidth/2`, so the hole disappears only when `strokeWidth >= 2*r`. For `r <= strokeWidth < 2*r` the center is wrongly filled, while the antialiased path draws the correct ring. The antialiased path is not affected.

I can send a PR that changes the threshold to `2*r` with a regression test if that helps. I also have a side-by-side screenshot of the three circles before/after and can attach it if wanted.

kumagi:
#3677 

bug

## What this PR does | solves
The non-AA shortcut switched to FillCircle when strokeWidth >= r, but a stroked circle is a ring with the inner radius r - strokeWidth/2 and the hole disappears only when strokeWidth >= 2r. For r <= strokeWidth < 2r, the center was filled even though the AA path rendered a correct ring with a hole.

Use the regular ring path whenever the hole still exists.

The new regression test failed on the old code (the center pixel was opaque white) and passes with this fix.

Closes #3677

### DIFF
--- a/vector/util_test.go
+++ b/vector/util_test.go
@@ -242,12 +242,14 @@ func TestStrokePathKeepsSourcePath(t *testing.T) {
 
 func TestStrokeCircleThickStrokeNonAntiAlias(t *testing.T) {
 	dst := ebiten.NewImage(64, 64)
+	defer dst.Deallocate()
 	vector.StrokeCircle(dst, 32, 32, 10, 10, color.White, false)
 	if got, want := dst.At(32, 32), (color.RGBA{}); got != want {
 		t.Errorf("got: %v, want: %v", got, want)
 	}
 
 	dst2 := ebiten.NewImage(64, 64)
+	defer dst2.Deallocate()
 	vector.StrokeCircle(dst2, 32, 32, 10, 20, color.White, false)
 	if got, want := dst2.At(32, 32), (color.RGBA{0xff, 0xff, 0xff, 0xff}); got != want {
 		t.Errorf("got: %v, want: %v", got, want)
--- a/vector/util_test.go
+++ b/vector/util_test.go
@@ -239,3 +239,17 @@ func TestStrokePathKeepsSourcePath(t *testing.T) {
 		t.Errorf("got:\n%v\nwant:\n%v", got, want)
 	}
 }
+
+func TestStrokeCircleThickStrokeNonAntiAlias(t *testing.T) {
+	dst := ebiten.NewImage(64, 64)
+	vector.StrokeCircle(dst, 32, 32, 10, 10, color.White, false)
+	if got, want := dst.At(32, 32), (color.RGBA{}); got != want {
+		t.Errorf("got: %v, want: %v", got, want)
+	}
+
+	dst2 := ebiten.NewImage(64, 64)
+	vector.StrokeCircle(dst2, 32, 32, 10, 20, color.White, false)
+	if got, want := dst2.At(32, 32), (color.RGBA{0xff, 0xff, 0xff, 0xff}); got != want {
+		t.Errorf("got: %v, want: %v", got, want)
+	}
+}
--- a/vector/util.go
+++ b/vector/util.go
@@ -289,7 +289,7 @@ func StrokeCircle(dst *ebiten.Image, cx, cy, r float32, strokeWidth float32, clr
 		return
 	}
 
-	if strokeWidth >= r {
+	if strokeWidth >= 2*r {
 		FillCircle(dst, cx, cy, r+strokeWidth/2, clr, false)
 		return
 	}


vector/util_test.go
```diff
@@ -239,3 +239,17 @@ func TestStrokePathKeepsSourcePath(t *testing.T) {
 		t.Errorf("got:\n%v\nwant:\n%v", got, want)
 	}
 }
+
+func TestStrokeCircleThickStrokeNonAntiAlias(t *testing.T) {
+	dst := ebiten.NewImage(64, 64)
+	vector.StrokeCircle(dst, 32, 32, 10, 10, color.White, false)
```
hajimehoshi, line 246: Add `defer dst.Deallocate()`

vector/util_test.go
```diff
@@ -239,3 +239,17 @@ func TestStrokePathKeepsSourcePath(t *testing.T) {
...
+	if got, want := dst.At(32, 32), (color.RGBA{}); got != want {
+		t.Errorf("got: %v, want: %v", got, want)
+	}
+
+	dst2 := ebiten.NewImage(64, 64)
+	vector.StrokeCircle(dst2, 32, 32, 10, 20, color.White, false)
```
hajimehoshi, line 253: Add `defer dst2.Deallocate()`

vector/util_test.go
kumagi: Done. Added `defer dst.Deallocate()`.

vector/util_test.go
kumagi: Done. Added `defer dst2.Deallocate()`.

kumagi:
Thanks for the review! I added `defer dst.Deallocate()` and `defer dst2.Deallocate()` to the new test as suggested, and the vector tests still pass.

hajimehoshi (approved):
LGTM

PR #3679, opened 2026-09-06
graphicscommand: return the error from ResetGraphicsDriverState

kumagi:
None

bug

## What this PR does | solves
ResetGraphicsDriverState assigned the result of r.Reset() to the named return value err, but ended with an explicit 'return nil' that discarded it. The caller's error check in atlas/image.go was dead code, while the sister function InitializeGraphicsDriverState returned the error with a bare return.

Return the named value with a bare return so a failing Reset, e.g. on Android suspend/resume, is reported instead of silently continuing with a broken GL state.


hajimehoshi (approved):
Good catch, LGTM

PR #3681, opened 2026-09-07
audio: skip blending a sample straddling the end of afterLoop

kumagi:
None

bug

## What this PR does | solves
afterLoop is a raw byte truncation with no alignment guarantee. When its length is not a multiple of the sample size, the blend for the final sample reads past its end and panics with an index out of range.

Skip blending for a sample straddling the end of afterLoop and leave the original value instead.

### DIFF
--- a/audio/loop.go
+++ b/audio/loop.go
@@ -244,6 +244,11 @@ func (i *InfiniteLoop) read(b []byte) (int, error) {
 			}
 
 			relpos := abspos - i.lstart
+			if relpos+int64(i.bitDepthInBytes) > int64(len(i.afterLoop)) {
+				// afterLoop is a raw byte truncation with no alignment guarantee.
+				// A sample straddling its end cannot be blended; leave the original.
+				continue
+			}
 			switch i.bitDepthInBytes {
 			case 2:
 				afterLoop := int16(i.afterLoop[relpos]) | (int16(i.afterLoop[relpos+1]) << 8)
--- a/audio/loop.go
+++ b/audio/loop.go
@@ -244,11 +244,6 @@ func (i *InfiniteLoop) read(b []byte) (int, error) {
 			}
 
 			relpos := abspos - i.lstart
-			if relpos+int64(i.bitDepthInBytes) > int64(len(i.afterLoop)) {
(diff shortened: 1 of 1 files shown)


hajimehoshi:
This guard looks harmless, but it appears redundant with the existing `blendRate` check. `blendRate` computes the blend length in complete stereo frames, so samples in the trailing incomplete frame get a rate of zero and skip blending before reaching the new bounds check.

Could you add a regression test that reproduces the panic without this guard and passes with it? The existing InfiniteLoop tests passed, and a temporary test matrix covering both sample formats, partial tails, chunked reads, varying buffer sizes, and intros passed both before and after this change. I couldn't reproduce the reported panic.

Review comment authored by Codex on behalf of @hajimehoshi.

kumagi:
Thanks for the careful review -- your analysis is correct, and I could not produce the requested regression test because the guard turns out to be unreachable.

What I verified:

1. Reachability: I instrumented the new branch to panic when taken and ran the existing audio tests plus an exhaustive configuration matrix (both sample formats, intro and loop lengths, afterLoop tails of every length, source read chunk sizes, buffer sizes, and seeks) and roughly 200k randomized read/seek sequences. The branch was never taken.

2. Arithmetic: with the blend-position fix ("audio: fix the blend position of InfiniteLoop") in place, the blended sample positions are always multiples of the bit depth, and blendRate returns 0 for any aligned position in the trailing incomplete frame of afterLoop. An exhaustive enumeration over both bit depths and afterLoop lengths up to 4096 bytes finds no case where the rate is non-zero while the new bounds check would trigger, so the existing rate check already subsumes it.

3. The original panic: it reproduces, but only against code before the blend-position fix, where the blend position was off by the remainder and the misaligned index could run past afterLoop (I reproduced an index-out-of-range there with a chunked reader). That fix landed upstream together with TestInfiniteLoopBlendWithPartialFrameReads, which already serves as the regression test, and the panic cannot be reproduced on the current main with or without this guard.

Since no test can distinguish the guard's presence and it is dead code, I have reverted the change; please feel free to close this PR. Sorry for the noise, and thanks for the quick upstream fix of the underlying problem!

hajimehoshi:
????

Can I close this?

PR #3682, opened 2026-09-07
ui: reject non-finite Layout results

kumagi:
None

bug

## What this PR does | solves
layoutGame checked the Layout result with <= 0 only, so NaN slipped through every comparison and +Inf passed as well. int(math.Ceil(NaN)) then produced an out-of-range size and failed with a remote panic or OOM far from the cause.

Reject NaN and infinite Layout results with an explicit panic.


hajimehoshi (approved):
LGTM

PR #3683, opened 2026-09-07
vector: make empty paths a clean no-op

kumagi:
None

bug

## What this PR does | solves
setPaths only early-returned when there were zero paths, so an all-empty path set still went through atlas allocation. Skip image creation when every region is empty and return nil from stencilBufferImageAt, making empty draws a clean no-op instead of failing.

### DIFF
--- a/vector/atlas.go
+++ b/vector/atlas.go
@@ -148,6 +148,14 @@ func (a *atlas) setPaths(dstBounds image.Rectangle, paths []*Path, bounds []imag
 	a.atlasImages = slices.Grow(a.atlasImages, atlasImageCount)[:atlasImageCount]
 	for i := range a.atlasImages {
 		s := a.atlasSizes[i]
+		if s.X <= 0 || s.Y <= 0 {
+			// All regions are empty: keep a clean no-op without creating an image.
+			if a.atlasImages[i] != nil {
+				a.atlasImages[i].Deallocate()
+				a.atlasImages[i] = nil
+			}
+			continue
+		}
 		var origWidth, origHeight int
 		if a.atlasImages[i] != nil {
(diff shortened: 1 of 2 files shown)


hajimehoshi:
File an issue.

vector/atlas.go
```diff
@@ -148,6 +148,14 @@ func (a *atlas) setPaths(dstBounds image.Rectangle, paths []*Path, bounds []imag
 	a.atlasImages = slices.Grow(a.atlasImages, atlasImageCount)[:atlasImageCount]
 	for i := range a.atlasImages {
 		s := a.atlasSizes[i]
+		if s.X <= 0 || s.Y <= 0 {
+			// All regions are empty: keep a clean no-op without creating an image.
```
hajimehoshi: Remove this comment

hajimehoshi:
Also, add a test

vector/atlas.go
kumagi: Done. I've removed the comment.

kumagi:
Thank you for the review! I've added tests for this: TestAtlasEmptyPaths and TestAtlasEmptyPathsAfterNonEmptyPaths in vector/atlas_test.go verify that no atlas image is created when all paths are empty, that the atlas image is deallocated when a previously non-empty set becomes empty, and that stencilBufferImageAt returns nil, making empty draws a clean no-op. The tests fail without the fix and pass with it. I will also file an issue for this and update the PR description to reference it.

hajimehoshi:
Again, file an issue

kumagi:
Sorry, my concern was completely wrong, based on incorrect assumption. I close this. Sorry for bothering.

PR #3684, opened 2026-09-07
mobile: treat ACTION_CANCEL as touch release on Android

kumagi:
None

bug

## What this PR does | solves
UpdateTouchesOnAndroid ignored ACTION_CANCEL (0x03), so a touch active during a notification shade, an incoming call UI, or a parent view intercept stayed in the touches map forever as a ghost touch. iOS already handles the cancelled phase.

Handle CANCEL like UP and delete the touch.


hajimehoshi:
There is still a multitouch cancellation issue with this change. `EbitenView.onTouchEvent` forwards `ACTION_CANCEL` only for the action-index pointer and sends `ACTION_MOVE` for every other pointer. If a two-finger gesture is cancelled, this deletes one touch but leaves the other pressed as a ghost touch.

Android specifies that after `ACTION_CANCEL`, all pointers that were down should be considered lifted: https://developer.android.com/reference/android/view/MotionEvent.html#ACTION_CANCEL

Please also update the Java forwarding to propagate cancellation to every pointer, and add coverage for cancelling a multitouch gesture. Clearing the Go map alone would not suffice while the Java loop continues sending `ACTION_MOVE` for the remaining pointers, since those calls would add them back.

Reviewed the PR revision and its Java caller; no Android device tests were run.

Posted by Codex on behalf of @hajimehoshi.

kumagi:
Thanks for catching this — you are right that the Java forwarding was the real gap. Updated:

- `EbitenView.onTouchEvent` now propagates `ACTION_CANCEL` to every pointer. Previously only the action-index pointer received `ACTION_CANCEL` while the rest were forwarded as `ACTION_MOVE`, which added them right back, so a cancelled two-finger gesture left a ghost touch. This matches the Android spec that after `ACTION_CANCEL` every pointer that was down must be treated as lifted.
- On the Go side, `UpdateTouchesOnAndroid` now lifts all active touches when it sees a cancellation, so no pointer can survive a cancelled gesture even if the platform delivers a single cancel event.
- Added coverage for cancelling a multitouch gesture: both pointers must be cleared by one cancellation, and `ACTION_POINTER_UP` still removes only its own pointer. The tests build for the mobile target and also run under `go test` on the host.

No Android device tests were run.

mobile/ebitenmobileview/touches_test.go
```diff
@@ -0,0 +1,66 @@
+// Copyright 2025 The Ebitengine Authors
```
hajimehoshi: Remove tests

mobile/ebitenmobileview/touches.go
```diff
@@ -0,0 +1,56 @@
+// Copyright 2025 The Ebitengine Authors
```
hajimehoshi: 2026

mobile/ebitenmobileview/touches_test.go
kumagi: Done. The test file is removed.

mobile/ebitenmobileview/touches.go
kumagi: Done. The copyright year is now 2026.

kumagi:
Addressed the review comments: the test file is removed and the new touches.go now carries the 2026 copyright year. The fix itself is unchanged: EbitenView.onTouchEvent propagates ACTION_CANCEL to every pointer, and UpdateTouchesOnAndroid clears all touches on a cancellation so no pointer can survive a cancelled gesture.

mobile/ebitenmobileview/touches.go
```diff
@@ -0,0 +1,56 @@
+// Copyright 2026 The Ebitengine Authors
```
hajimehoshi: Why do we need this in a separate file? Why not putting this in input_android.go?

mobile/ebitenmobileview/touches.go
```diff
@@ -0,0 +1,56 @@
...
+
+// Android MotionEvent action values.
+// See https://developer.android.com/reference/android/view/MotionEvent
+const (
+	actionDown        = 0x00
+	actionUp          = 0x01
+	actionMove        = 0x02
+	actionCancel      = 0x03
+	actionPointerDown = 0x05
+	actionPointerUp   = 0x06
```
hajimehoshi: Let's preserve the original names, but with `_` prefix

```go
_ACTION_DOWN = 0x00
_ACTION_UP = 0x01
...
```

mobile/ebitenmobileview/touches_test.go
kumagi: Done. `mobile/ebitenmobileview/touches_test.go` is removed.

mobile/ebitenmobileview/touches.go
kumagi: Done. The new file was later dropped altogether, so no new copyright header is added anymore: the handling now lives in `input_android.go`, which keeps its original header.

mobile/ebitenmobileview/touches.go
kumagi: You are right, a separate file was not needed. `touches.go` is removed and the handling is back in `UpdateTouchesOnAndroid` in `input_android.go`, and `position` is restored to `input.go`.

mobile/ebitenmobileview/touches.go
kumagi: Done. The constants now keep the original names with a `_` prefix in `input_android.go`:

```go
// https://developer.android.com/reference/android/view/MotionEvent
const (
	_ACTION_DOWN         = 0x00
	_ACTION_UP           = 0x01
	_ACTION_MOVE         = 0x02
	_ACTION_CANCEL       = 0x03
	_ACTION_POINTER_DOWN = 0x05
	_ACTION_POINTER_UP   = 0x06
)
```

kumagi:
Addressed the latest comments: `mobile/ebitenmobileview/touches.go` is removed and the action handling now lives directly in `UpdateTouchesOnAndroid` in `mobile/ebitenmobileview/input_android.go`, with `position` restored to `input.go`. The MotionEvent action constants keep the original names with a `_` prefix (`_ACTION_DOWN`, `_ACTION_UP`, `_ACTION_MOVE`, `_ACTION_CANCEL`, `_ACTION_POINTER_DOWN`, `_ACTION_POINTER_UP`), matching the other Android constants in that file. The behavior is unchanged: `ACTION_CANCEL` clears all the active touches, and `EbitenView.onTouchEvent` still propagates the cancellation to every pointer, so a cancelled multitouch gesture leaves no ghost touch. No Android device tests were run.

hajimehoshi (approved):
LGTM

PR #3687, opened 2026-09-08
audio/vorbis: clip out-of-range values in the int16 conversion

kumagi:
None

bug

## What this PR does | solves
The float to int16 conversion multiplied by 32767 without clipping, so an out-of-range decoded value, which Vorbis decoding can produce past peak 1.0, wrapped around through the implementation-defined float to int conversion and came out as harsh noise. The resampling path in convert clips to [-1, 1] before the same conversion.

Clamp to [-1, 1] before multiplying, mirroring the resampling path.

Add TestInt16BytesReaderClipsOutOfRange, which observes wrapped values without this fix.

### DIFF
--- a/audio/vorbis/int16.go
+++ b/audio/vorbis/int16.go
@@ -59,7 +59,7 @@ func (r *int16BytesReader) Read(buf []byte) (int, error) {
 	}
 
 	for i := range n {
-		f := r.fbuf[i]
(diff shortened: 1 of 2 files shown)


hajimehoshi:
Please rebase the change (or, the CI tests would fail. See #3688)

hajimehoshi (approved):
LGTM

PR #3690, opened 2026-09-08
internal/ui: handle touchcancel like touchend on browsers

kumagi:
None

bug

## What this PR does | solves
The browser backend listened to touchstart, touchend and touchmove, but not touchcancel. Touch state is rebuilt from the live targetTouches set, so a touch cancelled by the OS, e.g. by an incoming call or a gesture interruption, never produced an update and stayed pressed forever without even a JustReleased.

Register a touchcancel listener with the same body as touchend and route the event to updateTouchesFromEvent.


hajimehoshi (approved):
LGTM

PR #3691, opened 2026-09-08
internal/gamepad: skip nameless inotify events on Linux

kumagi:
None

bug

## What this PR does | solves
The inotify event parser sliced the event name as buf[16:16+e.Len-1] unconditionally. Nameless events with Len == 0, e.g. IN_Q_OVERFLOW when the queue overflows or IN_IGNORED, made the slice bounds panic and killed the process while polling gamepads.

Skip events with no name before slicing. Such events carry no device path and never matched the event filter below.


internal/gamepad/inotify_linux_test.go
```diff
@@ -0,0 +1,53 @@
+// Copyright 2026 The Ebitengine Authors
```
hajimehoshi: We don't need a test

kumagi: Done — removed the test file.

kumagi:
Thanks for the review! I removed the test file as requested, so the change is now just the nameless-event skip in the inotify parser. PTAL.

hajimehoshi (approved):
LGTM

hajimehoshi:
I've file an issue https://github.com/hajimehoshi/ebiten/issues/3694

PR #3692, opened 2026-09-08
ebiten: make RestoreWindow a no-op on non-desktop platforms

### LINKED ISSUE #3650: ebiten: RestoreWindow always panics on non-desktop platforms instead of doing nothing
## Ebitengine Version

v2.10.0-alpha.13 (main @ 4f474b00d)

## Operating System

- [x] Android
- [x] iOS
- [x] Xbox
- [x] Web Browsers

## Go Version (paste your `go version` output)

go1.26.7 linux/amd64

## What steps will reproduce the problem?

`RestoreWindow` is implemented as (window.go:301-310):

```go
// RestoreWindow panics when the window is not maximized nor minimized.
func RestoreWindow() {
	if !IsWindowMaximized() && !IsWindowMinimized() {
		panic("ebiten: RestoreWindow must be called on a maximized or a minimized window")
	}
	ui.Get().Window().Restore()
}
```

On non-desktop platforms `IsWindowMaximized()` and `IsWindowMinimized()` always return false by documentation ("always returns false if the platform is not a desktop"), and the underlying window backends (JS, mobile) implement `Restore()` as a no-op. So on browsers, Android, iOS and Xbox the guard can never be satisfied and **every** call to `RestoreWindow()` panics unconditionally.

Reproduction: run the `examples/window` example (or any code that calls `ebiten.RestoreWindow()` when the player clicks a "restore" button) in a browser or on a phone:

```go
	if inpututil.IsKeyJustPressed(ebiten.KeyR) {
		ebiten.MaximizeWindow()  // documented: does nothing on non-desktop
		ebiten.RestoreWindow()   // panics: "ebiten: RestoreWindow must be called on a maximized or a minimized window"
	}
	return nil
}
```

## What is the expected result?

Consistency with the sibling APIs: `MaximizeWindow`, `MinimizeWindow`, `SetWindowPosition`, etc. are documented as "does nothing when the platform is not a desktop", and indeed do nothing. `RestoreWindow` should likewise be a no-op on non-desktop platforms (and keep its documented panic on desktops where the preconditions are genuinely violated).

## What happens instead?

`RestoreWindow()` always panics on non-desktop platforms, even though the backend would have happily done nothing. Games that share window-management code across platforms must add `runtime.GOOS`/build-tag guards around every `RestoreWindow` call, and cross-platform titles that forget one crash on web/mobile only — the hardest kind of bug to catch in desktop-centric testing.

## Anything else you feel useful to add?

Proposed fix: expose a `ui.IsDesktopWindow()`-style check (the information is already available in the backends) and return early on non-desktop platforms before the maximized/minimized precondition check; keep the panic on desktop. Then extend the doc comment of `RestoreWindow` with the usual "does nothing if the platform is not a desktop" sentence.

jackwalkerlabs:

Closes #3650.

Bug.

## What this PR does | solves

Move the maximized/minimized precondition from the public `RestoreWindow` function to `desktopWindow.Restore`. Browser, mobile, and console backends use their existing `nullWindow.Restore` no-op. Desktop calls retain the same panic and message when the window is neither maximized nor minimized. Document the non-desktop behavior alongside the sibling window APIs.

The maintainer-requested revision removes the unnecessary desktop test and replaces the browser-specific test with `window_notdesktop_test.go`: android/ios/js/nintendosdk/playstation5 build tags and a direct public `ebiten.RestoreWindow()` call.

Validation of `ed9a6175a0a3a797238425d245f4955be5111172` by hosted FactoryChief run `FC-20260914-EAC201`:

- Fresh independent source review confirmed the requested revision; its requirement for independent test evidence was then satisfied by the separate verifier.
- Offline browser/WASM root-package tests passed using Go 1.25.14, Chromium 152.0.7977.82, and the repository-pinned wasmbrowsertest harness.
- Offline native root-package tests passed under Xvfb (`xvfb-run -a go test -count=1 .`).
- `git diff --check` and offline module verification passed. The verifier ran with networking disabled and exited 0.

Other platform runtime tests and the full cross-platform package matrix were not rerun for this revision.

Implemented with OpenAI Codex and repaired/independently validated through FactoryChief. The original contribution branch and PR are preserved.

### DIFF
--- a/window_desktop_test.go
+++ b/window_desktop_test.go
@@ -1,0 +1,40 @@
+// Copyright 2026 The Ebitengine Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !android && !ios && !js && !nintendosdk && !playstation5
+
+package ebiten_test
+
+import (
+	"testing"
+
+	"github.com/hajimehoshi/ebiten/v2"
+	"github.com/hajimehoshi/ebiten/v2/internal/microsoftgdk"
+)
+
+func TestRestoreWindowRequiresMaximizedOrMinimizedDesktopWindow(t *testing.T) {
+	if microsoftgdk.IsXbox() {
+		t.Skip("Xbox does not have a desktop window")
+	}
+	if ebiten.IsWindowMaximized() || ebiten.IsWindowMinimized() {
+		t.Fatal("test requires a normal desktop window")
+	}
+	defer func() {
+		const want = "ebiten: RestoreWindow must be called on a maximized or a minimized window"
+		if got := recover(); got != want {
+			t.Errorf("RestoreWindow panic = %v, want %q", got, want)
+		}
+	}()
+	ebiten.RestoreWindow()
+}
--- a/window_js_test.go
+++ b/window_js_test.go
@@ -1,0 +1,32 @@
+// Copyright 2026 The Ebitengine Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ebiten_test
+
+import (
+	"testing"
+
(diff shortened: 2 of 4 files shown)


window_desktop_test.go
```diff
@@ -0,0 +1,40 @@
...
+package ebiten_test
+
+import (
+	"testing"
+
+	"github.com/hajimehoshi/ebiten/v2"
+	"github.com/hajimehoshi/ebiten/v2/internal/microsoftgdk"
+)
+
+func TestRestoreWindowRequiresMaximizedOrMinimizedDesktopWindow(t *testing.T) {
```
hajimehoshi, line 26: We don't need this test

window_js_test.go
```diff
@@ -0,0 +1,32 @@
...
+func TestRestoreWindowWithoutDesktopWindow(t *testing.T) {
+	ebiten.RestoreWindow()
+	ebiten.MaximizeWindow()
+	ebiten.RestoreWindow()
+	ebiten.MinimizeWindow()
+	ebiten.RestoreWindow()
+	if ebiten.IsWindowMaximized() || ebiten.IsWindowMinimized() {
+		t.Fatal("window state changed on a non-desktop platform")
+	}
+}
```
hajimehoshi, line 32: Just this test `window_notdesktop_test.go` is fine:

```go
//go:build android || ios || js || nintendosdk || playstation5

package ebiten_test

import (
    "testing"

)

func TestRestoreWindowWithoutDesktopWindow(t *testing.T) {
	ebiten.RestoreWindow()
}
```

hajimehoshi (approved):
LGTM, thanks!

PR #3705, opened 2026-09-13
internal/gamepad: don't fail when inotify is unavailable on Linux

### LINKED ISSUE #3304: Basic setup code fails with inotify on gamepad on Linux (ubuntu 24.04)
### Ebitengine Version

v2.8.8

### Operating System

- [ ] Windows
- [ ] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

### Go Version (`go version`)

go version go1.24.6 linux/amd64

create `main.go`

> main.go
```go
package main

import (
	"log/slog"
	"os"

)

func main() {
	g := NewGame()
	ebiten.SetWindowResizingMode(ebiten.WindowResizingModeEnabled)
	ebiten.SetWindowTitle("Roublard")

	err := ebiten.RunGame(g)
	if err != nil {
		slog.Error("run game", "error", err)
		os.Exit(1)
	}
}

type Game struct{}

func NewGame() *Game {
	return &Game{}
}

	return nil
}

}

func (g *Game) Layout(w, h int) (int, int) {
	return 1280, 800
}
```

A black window pops up

### What happens instead?

```
2025/08/21 16:32:44 ERROR run game error="gamepad: InotifyAddWatch failed: no space left on device"
```

I just updated to ubuntu 24.04 and was trying to use ebiten from a tutorial, but previous tries of ebiten games/app were doing OK in my ubuntu 20.04 a few months back.

So maybe the new way to handle `/dev/input` in ubuntu has changed, and maybe it doesn't allow to use inotfy for checking for inputs anymore?

I try to bisect to previous version of ebiten which could work. But even v2.0.0 doesn't work.

dualfroz:
inotify limits are per user, so another program could make `RunGame` fail at startup, although inotify is only used for gamepad hotplug.
If inotify is unavailable, init now keeps the gamepads from the initial scan and skips hotplug detection, as GLFW does.

Closes #3304


hajimehoshi:
~~The new TestInitWithoutInotify test exhausts the inotify instance quota shared by every process running as the current user (internal/gamepad/gamepad_linux_test.go, lines 59–65). While those descriptors are held, unrelated applications or concurrent tests can fail to initialize their watchers. Deferred cleanup releases the descriptors afterward, but cannot undo failures during that interval.~~

~~Could this test inject an inotify initialization failure, or run in an environment with an isolated quota, instead of exhausting the real user quota?~~

~~The production fallback itself looks reasonable to me. This review was based on the diff and surrounding initialization/update paths; the Linux test was not executed on this macOS host.~~

~~Posted by Codex on behalf of @hajimehoshi.~~

hajimehoshi:
Sorry, but now I think we don't need a test itself: please drop `gamepad_linux_test.go`, thanks

hajimehoshi (approved):
LGTM, thanks!

PR #3707, opened 2026-09-15
all: fix typos, grammar and error message wording

kumagi:
None

bug

## What this PR does | solves
Correct articles, prepositions and singular/plural in comments across packages (the wav package, in the background, are gone, a boolean, equals the new one, an implementation, etc.).

Fix panic/error and test messages for grammar and to match the behavior (atlas len(pix), packing both children, shader ivec messages, swapped lhs count messages with the correct lengths, microsoftgdk XGameRuntimeInitialize, text Float64 message, etc.).

Fix a UI label typo (Maximize) and stale or inaccurate comments.

No behavior is changed.


internal/graphicscommand/image.go
```diff
@@ -43,8 +43,8 @@ type Image struct {
 
 	// id is an identifier for the image. This is used only when dumping the information.
 	//
-	// This is duplicated with graphicsdriver.Image's ID, but this id is still necessary because this image might not
-	// have its graphicsdriver.Image.
+	// This has the same value as graphicsdriver.Image's ID, but this id is still necessary because this image might not
+	// have its graphicsdriver.Image yet.
```
hajimehoshi: This overstates it. The two ids come from independent counters (this package's global `nextImageID` vs. each graphics driver's own `nextImageID`), and nothing keeps them equal. The original "duplicated with" meant that this id duplicates the *role* of `graphicsdriver.Image`'s ID, not its value.

```suggestion
	// This duplicates the role of graphicsdriver.Image's ID, but this id is still necessary because this image might not
	// have its graphicsdriver.Image yet.
```

vector/stroke.go
```diff
@@ -402,7 +402,7 @@ func addJoint(strokePath *Path, subPath *subPath, opIndex int, reverse bool, opt
 	p1 := p.add(v1)
 
 	// If the joint is an internal angle (< 180 degrees), the joint is not rendered. Just connect the two segments.
-	// [vec2.cross] has a precision issue. Use a comparison instead.
+	// The inline cross product above has a precision issue. Use a comparison instead.
```
hajimehoshi: There is no cross product above this line; the comparison on the next line *is* the cross product with the subtraction removed (6408f40a7, "vector: avoid cross product"). Replacing the dangling `[vec2.cross]` reference is right, but this should say what `vector/path.go` says at the same construct:

```suggestion
	// A cross product can be calculated by dir0.x*dir1.y - dir0.y*dir1.x,
	// but this can cause a floating-point precision issue due to FMSUBS. Avoid this subtraction.
```

run.go
```diff
@@ -228,7 +228,7 @@ var Termination = ui.RegularTermination
 // This is not related to framerate (display's refresh rate).
 //
 // RunGame returns an error when 1) an error happens in the underlying graphics driver, 2) an audio error happens
-// or 3) Update returns an error. In the case of 3), RunGame returns the same error so far, but it is recommended to
+// or 3) Update returns an error. In the case of 3), RunGame returns the error as is, but it is recommended to
```
hajimehoshi: Nit: "so far" was deliberately non-committal, which is why the next clause recommends `errors.Is`; "as is" reads like a guarantee. "currently returns the error as is" keeps the hedge. The same applies to RunGameWithOptions below.

hajimehoshi (commented):
Thank you! I read the whole diff and checked the rewordings that make claims about the code against the source; they hold (the 32-bit float formats in audiopanning/pcm, the squiral neighbor indexing, TestAdvanceTicks, ca.SetPixelFormat's accepted set, the unbuffered render-thread channel, the UTF-8 ranges and bidi levels in chunk.go, the swapped lhs count messages, and so on). Regenerating `internal/colormshader` and `internal/graphicsdriver/opengl/gl` with `go generate` gives output identical to this PR, and `go build`, `go vet`, `gofmt` and the affected package tests are clean.

Two inline comments below are rewordings that ended up inaccurate, plus one nit. Everything else looks good.

(One small note on "No behavior is changed": the `examples/fontvector` string and the `examples/contextlost` messages do change visible output. They look like intended typo fixes, so no objection.)

internal/graphicscommand/image.go
kumagi: Good catch — you're right that the ids come from independent counters and are not kept equal. Applied your suggestion verbatim.

vector/stroke.go
kumagi: Right, there is no cross product above that line; my wording was inaccurate. Applied your suggestion so it now matches the explanation in vector/path.go.

run.go
kumagi: Makes sense — 'as is' did read like a guarantee. Changed both RunGame and RunGameWithOptions to 'currently returns the error as is'.

kumagi:
Thank you for the careful review! I've addressed all three inline comments: the graphicscommand.Image id comment now says it 'duplicates the role' rather than the value, the vector/stroke.go joint comment now matches the cross-product/FMSUBS wording used in vector/path.go, and the RunGame/RunGameWithOptions docs keep the non-committal hedge with 'currently returns the error as is'.

On the note about 'No behavior is changed': agreed, the examples/fontvector string and the examples/contextlost messages do change visible output; those were intended typo/grammar fixes in the displayed text, so I've left them as they are.

hajimehoshi (approved):
LGTM

PR #3710, opened 2026-09-16
internal/hook: update audio state only after hook success

### LINKED ISSUE #3657: internal/hook: SuspendAudio/ResumeAudio flips the flag before running hooks, diverging permanently on hook error
### Ebitengine Version

main @ 4f474b00d (after v2.9.11)

### Operating System

- [x] Windows
- [x] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [x] Android
- [x] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [x] Web Browsers

1. Register audio suspend/resume hooks with an implementation that can fail (e.g. an audio-device operation failing during backgrounding on mobile):

```go
hook.OnSuspendAudio(func() error {
    return errors.New("device busy") // e.g. transient OS failure
})
```

2. Trigger `hook.SuspendAudio()`. It returns the error — but the damage is done. Trigger it again (or check the state): the second call is a silent no-op even though the device was never actually suspended.

Root cause (`internal/hook/hook.go:86-110` on current `main`):

```go
func SuspendAudio() error {
    ...
    audioSuspended = true      // committed BEFORE the hook runs (line 92)
    if onSuspendAudio != nil {
        return onSuspendAudio() // error leaves the flag flipped (line 94)
    }
    return nil
}
```

`ResumeAudio` mirrors this (flag cleared at line 105 before `onResumeAudio()` at line 107). When a hook fails, the flag permanently disagrees with reality: every later `SuspendAudio` no-ops ("already suspended") and every later `ResumeAudio` no-ops ("not suspended"). The audio state machine is wedged until process restart — typically permanent wrong behavior (stuck muted, or suspend never re-applied).

The hooks should run *first*, and the `audioSuspended` flag should only be committed on success, so a transient hook failure leaves the state machine consistent and retryable.

### What happens instead?

One transient failure (exactly the kind that happens when an app is backgrounded/foregrounded on mobile) permanently desynchronizes the flag from the device. From a player's perspective: after returning to the game once, audio is stuck off (or misbehaves) forever, with no recovery short of restarting the app.

- Severity: high for game users, especially on mobile where suspend/resume around backgrounding is a core OS interaction, not an edge case.
- Suggested fix: reorder to invoke `onSuspendAudio`/`onResumeAudio` first and set `audioSuspended` only when the hook returns nil.

jackwalkerlabs:

Closes #3657.

Bug.

## What this PR does | solves

Update the audio-suspended flag only after the corresponding suspend or resume hook succeeds. A failed hook preserves the previous state, so the same operation can be retried and the opposite operation remains a no-op until the transition succeeds.

Add a regression test covering failures and successful retries in both directions, with the package state restored after the test.

Validation on the exact PR head with Go 1.25.14, through hosted FactoryChief:
- Independent source review approved the change.
- The root package passed in Chromium/WebAssembly and native Linux under Xvfb.
- `go test -count=1 ./audio/... ./internal/hook` passed.
- `go test -race -count=1 ./audio ./internal/hook` passed.
- Gamepad database ordinary and race tests passed.

The full supported-OS and Go-version matrix was not run. Implemented and reviewed with OpenAI Codex through FactoryChief. Publication used operator-assisted Factory delivery.

Factory verification run: [FC-20260916-340766](https://solo-test.factorychief.com/runs/FC-20260916-340766).


hajimehoshi:
I found a correctness issue with retrying failed transitions: the existing audio callbacks are not retry-safe.

In `audio/audio.go`, the suspend callback sends to the capacity-one `c.semaphore` before calling `playerFactory.suspend()` and `c.onSuspend()`. If either returns an error, the channel stays full. With this change, a second `hook.SuspendAudio()` invokes the callback again and blocks on that send.

The resume path has the symmetric problem: its callback consumes the semaphore token before the fallible work. After an error, a second `hook.ResumeAudio()` blocks receiving from the now-empty channel.

Both retries block while holding the global hook mutex, preventing subsequent hook operations. Preserving the flag therefore needs corresponding failure handling in the audio callbacks to keep the semaphore and audio state consistent. Please also cover those callback side effects in the regression test; the current callbacks only return errors and cannot catch this interaction.

Validation on PR head `af99acdcf8fab30c825dcb61f5c125c30c9a75d2`: the supplied hook test passes with `-race`. Focused reproductions using the production callbacks' semaphore pattern and an injected error both time out, with stacks blocked in the channel send through `SuspendAudio` (`hook.go:93`) and channel receive through `ResumeAudio` (`hook.go:108`), respectively. These reproductions model the callback synchronization; they do not inject an actual audio-device failure.

Posted by OpenAI Codex on behalf of @hajimehoshi.

hajimehoshi:
ping

hajimehoshi:
As there is no recent update, I will work on this instead.

PR #3711, opened 2026-09-16
internal/gamepaddb: skip unsupported foreign-platform mappings

### LINKED ISSUE #3658: internal/gamepaddb: one bad token for another platform aborts the entire mapping update
### Ebitengine Version

main @ 4f474b00d (after v2.9.11)

### Operating System

- [x] Windows
- [x] macOS
- [x] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

1. Pass a community SDL game-controller database (via `ebiten.UpdateStandardGamepadLayoutMappings`) that contains a mapping line for a *different* platform, where that line also carries a token value the parser rejects — e.g. an unknown key such as `hint:foo`, or an SDL-legal hat token like `+h0.1`:

```
030000005e0400008e02000000000000,PS4 Controller,a:b0,b:b1,platform:Windows,hint:foo,
```

2. Call the update on Linux (or generally: a line whose `platform:` does not match the current OS, but which appears *before* the platform token in parse order carries a rejected value). The entire `Update...` call fails atomically, and **all** mappings — including the valid ones — are dropped, so gamepads stop working.

Root cause (`internal/gamepaddb/gamepaddb.go:100-142` on current `main`): `parseLine` iterates tokens in order and calls `parseMappingElement(tks[1])` at line 140 *before* ever seeing the `platform` token handled at lines 110-138. A single bad value token therefore `return ..., err` at line 142 even for lines that would have been skipped as foreign-platform, and `Update` (`gamepaddb.go:541-556`) aborts the whole database update. Relatedly, unknown keys are fatal instead of skipped, and SDL-legal notations (`bN`, `+hX.Y`) are rejected.

`parseLine` should determine the line's platform (and skip unknown keys) *before* parsing values, so foreign-platform lines and forward-compatible unknown keys can never veto the entire database. At minimum, one bad line must not invalidate all the good ones.

### What happens instead?

Pasting a widely-shared community mapping DB — an officially encouraged workflow — can silently leave players with a completely non-functional gamepad, with the failure depending on database content rather than anything the player did. "Controller doesn't work at all" is a game-breaking experience for controller-centric games.

- Severity: high for game users. The trigger (copy-pasting the community SDL DB) is common, the symptom (no gamepad input whatsoever) is total, and the cause (one foreign line) is invisible to both players and developers.
- Suggested fix: pre-scan for the `platform` token (skip non-matching lines early), `continue` on unknown keys instead of erroring, and accept SDL-legal value notations.

jackwalkerlabs:

Closes #3658.

Bug.

## What this PR does | solves

Determine a mapping line's platform before parsing its bindings, so an unsupported value on another platform cannot abort the database update. Ignore unknown mapping keys before parsing their values and accept positive hat bindings such as `+h0.1`.

Add regression tests for a database containing valid and foreign-platform mappings, an unsupported `hint` key, and a positive hat binding. Invalid recognized bindings for the current platform continue to return errors.

Validation on the exact PR head with Go 1.25.14, through hosted FactoryChief:
- Independent source review approved the change.
- `go test -count=1 ./internal/gamepaddb` and its race run passed.
- The root package passed in Chromium/WebAssembly and native Linux under Xvfb.
- The audio package suites and audio/hook race tests passed.

The full supported-OS and Go-version matrix was not run. Implemented and reviewed with OpenAI Codex through FactoryChief. Publication used operator-assisted Factory delivery.

Factory verification run: [FC-20260916-49F942](https://solo-test.factorychief.com/runs/FC-20260916-49F942).


hajimehoshi:
The early platform check looks useful: parseLine runs at runtime through UpdateStandardGamepadLayoutMappings -> gamepaddb.Update -> parseLine. The generator only splits the bundled database by platform; the public API explicitly permits passing an entire gamecontrollerdb.txt at runtime. An unsupported binding before a foreign platform: token can therefore abort that update.

I would assess the unknown-key handling and +h support separately from the platform check, since they change what is accepted for the current platform as well.

There are also two corrections to the linked issue: a failed update preserves existing mappings, and its literal example puts platform:Windows before hint:foo, so that example already skips correctly on Linux. The foreign-platform failure requires the rejected value to precede the platform token.

Posted by OpenAI Codex on behalf of @hajimehoshi.

hajimehoshi:
ping

PR #3712, opened 2026-09-16
graphics: account bytesPool memory by cap instead of len

kumagi:
None

bug

## What this PR does | solves
getFromCache returns bs[:size], so a pooled slice keeps the capacity of its original allocation while len reflects only the last use. totalSize summed len, which under-counted the memory actually retained when small and large sizes were mixed.

With the under-count, the 1GB eviction threshold was not reached and the pool stayed oversized. Sum cap instead so the limit reflects retained memory. No behavior is changed for uniform-size usage.


hajimehoshi (approved):
LGTM

PR #3713, opened 2026-09-16
internal/shader: remove the dead pos field of block

kumagi:
None

bug

## What this PR does | solves
The block struct carried a pos token.Pos field that was never assigned: the only composite literal leaves it zero, and no code reads it. The token package stays imported for the other position fields (vertex/fragment entry positions and unused variable positions).

Dropping the field shrinks the compiler's internal state and avoids future misuse of a position that was never populated. No behavior is changed.


hajimehoshi:
Let's include the other unused fields in this PR as well:

- `internal/ui/image.go`: `Image.imageType`
- `internal/graphicsdriver/remote/remote.go`: `Image.width`, `Image.height`, and `Image.screen`
- `exp/textinput/textinput_darwin.go`: `lineView.suffixLenInUTF16`, including its unused calculations in both constructors
- `internal/beforemaintest/image_test.go`: `game.m` and `game.code`
- `examples/audio/main.go`: `Player.game`
- `examples/touch/main.go`: `pinch.originH`
- `examples/squiral/main.go`: `palette.name`
- `examples/ui/main.go`: the unused `Input` type, including `mouseButtonState`

These were identified in a source scan of the review checkout across build tags. Please verify them against this branch and remove the corresponding initializers along with the fields. Preserve native ABI layout fields and fields used implicitly through struct equality or map keys.

Comment authored by Codex (OpenAI), on behalf of @hajimehoshi.

kumagi:
Thanks for the list! I verified every item against this branch and removed them in a follow-up commit:

- `internal/ui`: `Image.imageType` (write-only; `NewImage` still forwards its argument to `mipmap.New`, so behavior is unchanged)
- `internal/graphicsdriver/remote`: `Image.width`, `Image.height`, and `Image.screen` (set only in the two constructors and never read; the values are already recorded in the graphics commands)
- `exp/textinput/textinput_darwin.go`: `lineView.suffixLenInUTF16` and the `convertByteCountToUTF16Count(suffix, ...)` calculations in both constructors (`suffix` itself is still used)
- `internal/beforemaintest/image_test.go`: `game.m` and `game.code`, including the `m: m` initializer
- `examples/audio/main.go`: `Player.game` (I left the `NewPlayer` parameter untouched to keep the example's call sites as they are)
- `examples/touch/main.go`: `pinch.originH` (`originDiff` is still used for `prevH` and the pinch threshold)
- `examples/squiral/main.go`: `palette.name` and its three initializers
- `examples/ui/main.go`: the unused `Input` type including `mouseButtonState`

I checked for fields relied on implicitly via struct equality, map keys, or native ABI layout and found none among these. Validated with `go build`/`go vet` across linux, darwin, windows, js/wasm, ios, and android (remaining vet findings are only in the govetblock-listed packages), `gofmt`, and `go test ./internal/shader/...`.

hajimehoshi (approved):
LGTM

PR #3714, opened 2026-09-16
ebiten: remove the incorrect same-image panic note from DrawImage

kumagi:
None

bug

## What this PR does | solves
Nothing in the draw path rejects a self-draw: an image and its sub-images intentionally share one ui.Image, where a source can be the same as the destination.


hajimehoshi:
This panic note is still valid: `img.DrawImage(img, nil)` reaches `buffered.Image.DrawTriangles`, which explicitly checks `i == src` and panics:

https://github.com/hajimehoshi/ebiten/blob/f7638e6424245a438cb554da2f802555d6666242/internal/buffered/image.go#L81-L85

The existing `TestImageSelf` also requires this panic:

https://github.com/hajimehoshi/ebiten/blob/f7638e6424245a438cb554da2f802555d6666242/image_test.go#L172-L186

The lock deduplication in `ui.Image` prevents locking the same mutex twice, but it does not make self-drawing supported by the downstream draw path. Please retain the documentation.

Reviewed by Codex and posted on behalf of @hajimehoshi. Tests were not run.

kumagi:
Thanks for the careful check — you are right, and I verified it locally.

`img.DrawImage(img, nil)` still reaches `buffered.Image.DrawTriangles`, which rejects a destination that is also one of the sources, and `TestImageSelf` requires that panic. The lock deduplication in `ui.Image` only prevents locking the same mutex twice; it does not make a self-draw work downstream.

So I restored the panic note. I also added one sentence clarifying the case that misled me in the first place: an image and its sub-images share the same internal image, so drawing between a whole image and its sub-image, or between two sub-images of the same whole image, panics as well. I confirmed all three combinations panic on the current base, while unrelated images draw fine.

Happy to drop the extra sentence and keep the note exactly as it was if you prefer the original wording.

hajimehoshi (approved):
LGTM

Issue #3706, opened 2026-09-15
internal/ui: consecutive touches can share a TouchID and collapse into one touch


hajimehoshi:
### Ebitengine Version

main (9202a8f61, v2.11.0-alpha). The behavior is present in every released version.

### Operating System

- [ ] Windows
- [ ] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [x] Android
- [x] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [x] Web Browsers

### Go Version (paste your `go version` output)

go version go1.27.0 darwin/arm64

### What steps will reproduce the problem?

1. Run `examples/drag` on a touch device (Android, iOS, or Chrome on Android).
2. Drag a sprite with one finger, lift it, and land a second finger somewhere else so that the lift and the landing fall between two consecutive ticks. Lowering the TPS (`ebiten.SetTPS(10)`) or a slow `Update` makes this easy to hit.

A multi-touch variant on Android needs no timing luck on the tick rate: hold finger A, add finger B, lift A, then land a new finger C. Android hands C the lowest free pointer ID, which is the ID A had.

### What is the expected result?

The first touch is reported as released and the second as a new touch: `inpututil.AppendJustReleasedTouchIDs` reports the first one, `inpututil.AppendJustPressedTouchIDs` reports the second one, and the two touches have distinct `TouchID`s.

### What happens instead?

Both touches carry the same `TouchID`, so the two per-tick snapshots are indistinguishable from one continuous touch:

- Neither edge is reported: no just-released for the first touch and no just-pressed for the second.
- `inpututil.TouchPressDuration` keeps counting across both touches.
- `TouchPosition` jumps from the first finger's last position to the second finger's position, and `TouchPositionInPreviousTick` refers to the other finger.

In `examples/drag`, the sprite grabbed by the first finger teleports to the second finger and stays attached to it, and no stroke starts for the second finger even when it landed on another sprite.

### Anything else you feel useful to add?

Root cause: the touch state is a per-tick snapshot of the touches that are currently down, keyed on the ID the platform assigns, and every platform reuses IDs as soon as they are free:

- Android: `MotionEvent` pointer IDs are passed through (`mobile/ebitenmobileview/input_android.go`). The first finger is always 0 and a new finger takes the lowest free ID.
[...]

PR #3715, opened 2026-09-17
Ps5 touch pad

user-none:
Updates #3130 for supporting gamepad touch interfaces on macOS, Linux and Windows. Specifically for the PS4 and PS5 controllers.

feature

## What this PR does | solves
Support PS4 and PS5 touch pad on macOS, Linux and Windows.

macOS and Linux are generic per their APIs and can support any controller with a touch pad that is supported by the OS. Which are only the PS4 and PS5 as far as I'm aware. Windows is controller specific.

### Linux Notes

On Linux Playstaion controllers have multiple input dev nodes that show up. One specific to buttons / joystick / rumble, a second for motion, and a third for the touch pad. The changes here also compares the ids and groups the different nodes into one device. on main, it would show each one as an individual controller. However, this isn't typically seen due to permissions.

On my Ubuntu instance my user did not have permissions for the motion and touch pad devices even though it does for the buttons / joystick / rumble device. I had to add udev rules in order for my user to have permission for the touchpad to be usable.

```
# DualShock 4
SUBSYSTEM=="input", ATTRS{id/vendor}=="054c", ATTRS{id/product}=="0ba0", TAG+="uaccess"

# DualSense and DualSense Edge
SUBSYSTEM=="input", ATTRS{id/vendor}=="054c", ATTRS{id/product}=="0ce6", TAG+="uaccess"
SUBSYSTEM=="input", ATTRS{id/vendor}=="054c", ATTRS{id/product}=="0df2", TAG+="uaccess"
```


hajimehoshi:
I found one correctness issue in the DualShock 4 touch decoder:

**[P2] Decode the latest touch sample in each report** — [`internal/gamepad/sonyhid/sonyhid.go:291–295`](https://github.com/user-none/ebiten/blob/7b0b968626e405b1861ccb90b83a59e0e69961db/internal/gamepad/sonyhid/sonyhid.go#L291-L295)

`offset = 34` always selects the first touch sample, even when `p[32]` indicates multiple samples. Later samples can contain movement or a finger release. A report with a finger down in sample 1 and released in sample 2 therefore leaves the touch active. If subsequent reports contain zero touch samples, returning `prev` preserves that stale contact instead of retiring it. The assumption in the comment that another report supersedes the first sample before a frame observes it is not guaranteed.

The [Linux PlayStation driver processes all touch samples in order](https://github.com/torvalds/linux/blob/master/drivers/hid/hid-playstation.c#L2340-L2357). Please select the latest valid sample (or process all samples), validating the sample count against the report's available bytes.

I reproduced this with a synthetic 64-byte USB report containing two samples: contact 5 active in the first and inactive in the second. The decoder incorrectly returns an active touch, which remains active after a subsequent zero-sample report. A regression test covering this sequence would catch the issue.

Existing gamepad and Sony HID tests passed on macOS, and the gamepad test binaries compiled for Linux and Windows. Physical controllers were not tested.

*Review comment authored by Codex (OpenAI), on behalf of @hajimehoshi.*

user-none:
Touch samples fixed in the DualShock 4 touch decoder.

hajimehoshi:
I'll take a look later. Thank you for your patience...

internal/gamepad/sonyhid/sonyhid.go
```diff
@@ -222,7 +222,102 @@ func dualSenseRumbleReportBT(seq, strong, weak byte) []byte {
...
+// touchesFromPayload decodes the touchpad records of a state payload, or
+// returns prev when the payload carries no touch sample: the simplified
+// Bluetooth report is too short to hold one, and a DualShock 4 report with a
+// packet count of 0 holds none.
+//
+// In the DualShock 4 layout byte 32 counts the packets that follow, each a
+// timestamp byte and two records, oldest first. The last packet the payload
+// has room for is decoded: it is the most recent sample, and a finger lifted
+// in it stays lifted when the reports that follow carry no packet. In the
+// DualSense layout the two records are at byte 32.
```
hajimehoshi: Some comments are for internal details. Move the comments to the implementation.

hajimehoshi:
There is another touch-identity case to address here, following the recent fix for #3706.

[9a1ce6ab](https://github.com/hajimehoshi/ebiten/commit/9a1ce6ab252b204e3c1b03d556f8b3f43a2d2523) assigns fresh touchscreen IDs at each incoming platform event, so a release followed by a new contact under the same platform ID between ticks remains distinguishable. [bf52960f](https://github.com/hajimehoshi/ebiten/commit/bf52960f0a40eea9b0f8f0b32ea873ff40d59c6d) applies the same principle to VM guest press/release events. The important part is observing contact lifetimes before reducing input to a per-tick snapshot.

**[P2] Preserve contact changes between polls on the GameController backend.** At the current PR head (`e302d48`), [`updateGCGamepad`](https://github.com/user-none/ebiten/blob/e302d48b4fe662c3c2fc68ee60230707ecea7398/internal/gamepad/api_gc_darwin.go#L739-L746) copies only activity and position into `touchContact`, leaving `id` at 0. [`updateTouches`](https://github.com/user-none/ebiten/blob/e302d48b4fe662c3c2fc68ee60230707ecea7398/internal/gamepad/touch.go#L75-L86) then assigns a new public ID only when the previous sampled slot was inactive or the native contact ID changed.

If finger A is down at one update, lifts, and finger B lands in the same slot before the next update, both snapshots are active with native ID 0. B therefore inherits A's public ID and appears to be a continuation whose position jumped. This is the same lifetime-collapse problem as #3706 and contradicts the new API's promise that lifting a finger retires its ID.

Please apply the same event-time identity principle here: retain contact transitions from the GameController callbacks and assign a fresh contact generation when a new contact begins, then expose that generation to the shared touch-ID logic. Add regression coverage for a release and re-press in the same slot between two gamepad updates, alongside a check that movement within one contact preserves its ID. Merely merging the touchscreen allocator fix will not change this separate gamepad path.

Linux and Windows already supply native tracking/contact IDs, so the existing changed-contact-ID test covers that distinction when those IDs differ. This comment concerns consecutive contacts; preserving a tap that starts and ends entirely between ticks is a separate behavior that the #3706 fix does not provide either.

This finding is based on code inspection, not a physical-controller reproduction. I missed it in the earlier review.

*Review comment authored by Codex (OpenAI), on behalf of @hajimehoshi.*

internal/gamepad/sonyhid/sonyhid.go
user-none: Fixed

user-none:
> There is another touch-identity case to address here, following the recent fix for #3706.
> ...

Changes made to address this. Tested on macOS. I lowered the TPS and put in a sleep in Draw in my test app in order to verify the fix.

hajimehoshi:
Thanks. I'll take a look at this tomorrow. Thank you for your patience...

hajimehoshi:
Please change the public touch-position range to `[0, 1]` for both coordinates, with `(0, 0)` at the top-left and `(1, 1)` at the bottom-right.

A touchpad reports an absolute position on a surface, so this convention fits the API well and maps directly to an area's dimensions. It also matches [SDL's gamepad touchpad API](https://wiki.libsdl.org/SDL3/SDL_GetGamepadTouchpadFinger).

Please update the backend conversions, API documentation, and tests consistently, including corner and center positions (`(0.5, 0.5)`). The existing `(0, 0)` return for an absent touch can remain; callers determine active touches through `AppendGamepadTouchIDs`.

*Comment authored by Codex (OpenAI), on behalf of @hajimehoshi.*

hajimehoshi:
Please drop the tests/assertions that unnecessarily depend on the internal touch-slot representation, and simplify the test helpers accordingly:

- In the Linux touch tests, remove the slot-array inspection through `ContactsForTest` where the behavior can be checked by feeding events, advancing the gamepad, and asserting through `AppendTouchIDs` and `TouchPosition`. Keep the useful one-finger, two-finger, release/re-press, and coordinate-normalization scenarios through that observable interface.
- Avoid exporting `TouchSlotTracker.Contacts` just to manually copy its internal state into a separate fake backend. A small test backend that owns the tracker and supplies its contacts through the normal backend interface would let the regression test drive reports and updates without exposing the slot array.
- Drop assertions that require public IDs to be exactly `1, 2, 3, ...` or require enumeration in slot order. Capture the IDs returned by the API and check the actual guarantees: concurrent contacts have distinct IDs, movement preserves an ID, release retires it, and a new contact gets a fresh ID.

Please retain the regression for release and re-press in the same slot between updates. A narrow `export_test.go` bridge for supplying hardware reports is reasonable, as are the Sony HID decoder's input/output tests; the request is to remove unnecessary internal inspection and allocation/order assumptions, not the meaningful behavioral coverage.

*Comment authored by Codex (OpenAI), on behalf of @hajimehoshi.*

user-none:
range and test changes made.

internal/gamepad/sonyhid/sonyhid_test.go
```diff
@@ -341,41 +341,80 @@ func TestInputReportSize(t *testing.T) {
...
+var touchRecords = []byte{
+	0x05, 0x73, 0xa4, 0x20, // active, id 5, x 0x473, y 0x20a
+	0x86, 0x64, 0x80, 0x0c, // lifted, id 6, x 0x064, y 0x0c8
+}
+
+// touches returns the state touchRecords decode to on a touchpad h units
+// tall; both models are 1920 units wide.
+func touches(h int) [sonyhid.TouchCount]sonyhid.Touch {
+	return [sonyhid.TouchCount]sonyhid.Touch{
+		{Active: true, ID: 5, X: float64(1139) / 1919, Y: float64(522) / float64(h-1)},
```
hajimehoshi, line 355: Use literals like `1139.0` instead of `float64(1139)`

internal/gamepad/sonyhid/sonyhid_test.go
```diff
@@ -341,41 +341,80 @@ func TestInputReportSize(t *testing.T) {
...
+		{Active: false, ID: 6, X: float64(100) / 1919, Y: float64(200) / float64(h-1)},
```
hajimehoshi, line 356: Use multiple lines for struct literals like

```go
{
  Active: false,
  ID: 6,
  X: 100.0 / 1919,
  Y: 200.0 / float64(h-1)
}
```

internal/gamepad/export_linux_test.go
```diff
@@ -16,10 +16,88 @@
...
 	FFEffectSize         = unsafe.Sizeof(ff_effect{})
 	FFEffectUnionOffset  = unsafe.Offsetof(ff_effect{}.u)
 	FFEffectRumbleOffset = unsafe.Offsetof(ff_effect{}.u) + unsafe.Offsetof(ff_effect_union{}.rumble)
 )
+
+// Event node capability codes, for building the bitmaps classification reads.
+const (
+	EVKey = unix.EV_KEY
+	EVAbs = unix.EV_ABS
+	EVFF  = unix.EV_FF
```
hajimehoshi, line 35: Do we need these `EV*` definitions?

internal/gamepad/export_linux_test.go
```diff
@@ -16,10 +16,88 @@
...
+
+	ABSX            = _ABS_X
+	ABSY            = _ABS_Y
+	ABSRZ           = _ABS_RZ
+	ABSHat0X        = _ABS_HAT0X
+	ABSHat0Y        = _ABS_HAT0Y
+	ABSMTSlot       = _ABS_MT_SLOT
+	ABSMTPositionX  = _ABS_MT_POSITION_X
+	ABSMTPositionY  = _ABS_MT_POSITION_Y
+	ABSMTTrackingID = _ABS_MT_TRACKING_ID
```
hajimehoshi, line 45: Align with the original name conventions like

```go
ABS_X = _ABS_X
ABS_Y = _ABS_Y
ABS_RZ = _ABS_RZ
...
```

internal/gamepad/export_linux_test.go
```diff
@@ -16,10 +16,88 @@
...
+
+	BTNLeft          = 0x110
+	BTNSouth         = _BTN_A
+	BTNToolFinger    = 0x145
+	BTNTouch         = 0x14a
+	BTNToolDoubleTap = 0x14d
```
hajimehoshi, line 51: ditto: `BTN_LEFT` for example.

internal/gamepad/gamepad_gc_touch_darwin.go
```diff
@@ -0,0 +1,288 @@
...
+			// implies, ignoring the touchpad button's value and pressed state.
+			touchpadHandler := func(active bool) objc.Block {
+				return objc.NewBlock(func(_ objc.Block, _ objc.ID, x, y, _ float32, _ bool) {
+					g.reportTouch(slot, gcTouchState{active: active, x: x, y: y})
+				})
+			}
+			e.handlers = []gcTouchHandler{
+				{setter: sel_setTouchDown, block: touchpadHandler(true)},
+				{setter: sel_setTouchMoved, block: touchpadHandler(true)},
+				{setter: sel_setTouchUp, block: touchpadHandler(false)},
```
hajimehoshi, line 200: Use multiple lines for struct literals

```go
{
    setter: sel_setTouchDown,
    block: touchpadHandler(true),
},
...
```

internal/gamepad/gamepad_gc_touch_darwin.go
```diff
@@ -0,0 +1,288 @@
...
+			}
+
+		case gcTouchSlotDPad:
+			valueChanged := objc.NewBlock(func(_ objc.Block, _ objc.ID, x, y float32) {
+				g.reportTouch(slot, gcTouchState{active: x != 0 || y != 0, x: x, y: y})
+			})
+			e.handlers = []gcTouchHandler{
+				{setter: sel_setValueChangedHandler, block: valueChanged},
```
hajimehoshi, line 208: ditto

internal/gamepad/gamepad_linux.go
```diff
@@ -116,18 +121,35 @@ func (g *nativeGamepadsImpl) init(gamepads *gamepads) (err error) {
...
+// gamepad, or as a pending touch surface.
+func (g *nativeGamepadsImpl) isOpen(gamepads *gamepads, path string) bool {
 	if gamepads.find(func(gamepad *Gamepad) bool {
-		return gamepad.native.(*nativeGamepadImpl).path == path
+		n := gamepad.native.(*nativeGamepadImpl)
+		return n.path == path || (n.touch != nil && n.touch.path == path)
 	}) != nil {
+		return true
+	}
+	for _, t := range g.pendingTouch {
```
hajimehoshi, line 141: `slices.ContainsFunc`

internal/gamepad/gamepad_linux.go
```diff
@@ -191,12 +213,43 @@ func (*nativeGamepadsImpl) openGamepad(gamepads *gamepads, path string) error {
 		}
 		return fmt.Errorf("gamepad: ioctl for an ID failed: %w", err)
 	}
+	// A node without properties is fine; the property bits then stay clear.
+	propBits := make([]byte, (_INPUT_PROP_CNT+7)/8)
+	_ = ioctl(fd, _EVIOCGPROP(uint(len(propBits))), unsafe.Pointer(&propBits[0]))
```
hajimehoshi, line 218: Is it OK to ignore the returning value here?

internal/gamepad/gamepad_linux.go
```diff
@@ -191,12 +213,43 @@ func (*nativeGamepadsImpl) openGamepad(gamepads *gamepads, path string) error {
...
-	if !isBitSet(evBits, unix.EV_ABS) {
+	kind := classifyEvdev(evBits, keyBits, absBits, propBits)
+	if kind == evdevKindOther {
 		owned = false
 		return unix.Close(fd)
 	}
 
+	// The uniq string, the controller's address on the kernel drivers of interest, is what ties a
+	// touch surface node to its gamepad node. Many devices have none.
+	uniq := ""
```
hajimehoshi, line 228: `var uniq string`

internal/gamepad/gamepad_linux.go
```diff
@@ -298,6 +359,62 @@ func (*nativeGamepadsImpl) openGamepad(gamepads *gamepads, path string) error {
...
+		return n.touch == nil && n.uniq == t.uniq && n.id.vendor == t.id.vendor && n.id.product == t.id.product
+	}); gp != nil {
+		withNative(gp, func(n *nativeGamepadImpl) {
+			n.touch = t
+		})
+		return
+	}
+	if g.pendingTouch == nil {
+		g.pendingTouch = map[string]*touchNode{}
+	}
```
hajimehoshi, line 376: Initializing a map can be delayed just before assigning a new value

```go
	if old := g.pendingTouch[t.uniq]; old != nil {
		old.close()
	}
	if g.pendingTouch == nil {
		g.pendingTouch = map[string]*touchNode{}
	}
	g.pendingTouch[t.uniq] = t
```

internal/gamepad/touch_test.go
```diff
@@ -0,0 +1,327 @@
...
+		{{}, {}},
+	})
+	if got := g.TouchSurfaceCount(); got != 1 {
+		t.Errorf("TouchSurfaceCount() = %d; want 1", got)
+	}
+	checkTouchCount(t, g, 0, 0)
+
+	// A finger down in slot 0 is a touch.
+	g.SetTouchReportForTest([][]gamepad.TouchContactForTest{
+		{{Active: true, X: 0.25, Y: 0.625}, {}},
```
hajimehoshi, line 137: Use multiple lines

```go
{
    {
        Active: true,
        X: 0.25,
        Y: 0.625,
    },
    {},
}
```

hajimehoshi (commented):
Almost LGTM

user-none:
all the review items should be resolved

PR #3721, opened 2026-09-17
audio: lock the context when resetting it for testing

kumagi:
None

bug

## What this PR does | solves
ResetContextForTesting assigned nil to theContext without theContextLock, while CurrentContext and the hook closures read it under the lock, so a concurrent IsReady polling goroutine raced on it.

Lock theContextLock like every other access.


hajimehoshi (approved):
LGTM

PR #3722, opened 2026-09-17
internal/graphicsdriver/directx: fix the D3D12 render target view desc size

kumagi:
None

bug

## What this PR does | solves
The union is 8-byte aligned and its largest member (D3D12_TEX2D_ARRAY_RTV) is 16 bytes, so the descriptor is 24 bytes. Reserve 16 bytes for the union as the SRV descriptor already does.


hajimehoshi (approved):
LGTM

PR #3723, opened 2026-09-17
ebiten: document that IsWindowDecorated is false on non-desktops

kumagi:
None

bug

## What this PR does | solves
Its sibling getters (IsWindowVisible, IsWindowMaximized, ...) all state that they always return false off desktops, but IsWindowDecorated did not, so a false on a browser looked like a real state rather than the nullWindow stub.


hajimehoshi (approved):
LGTM

PR #3724, opened 2026-09-17
ebiten: update the ScreenSizeInFullscreen doc for mobiles

kumagi:
None

bug

## What this PR does | solves
The note that mobiles return (0, 0) is stale: the not-js path delegates to Monitor().Size(), which reports the real display size once the game has started (fe970e1b6 made it orientation-aware). It is (0, 0) only before the game starts.


hajimehoshi (approved):
LGTM

PR #3725, opened 2026-09-17
cmd/ebitenmobile: return error on unsupported gobind language

kumagi:
None

bug

## What this PR does | solves
run returns errors to main, which reports them via log.Fatal, but the default case of the language switch panicked with a stack trace. A misspelled, empty, or unsupported -lang value is user input, so it should follow the same error path as every other failure in run.

Return fmt.Errorf instead of panicking; the message text is unchanged and now surfaces through log.Fatal. No behavior is changed for the supported objc, java, and go cases.


cmd/ebitenmobile/gobind.go
```diff
@@ -141,7 +141,7 @@ import "C"`); err != nil {
 		case "go":
 			// Do nothing.
 		default:
-			panic(fmt.Sprintf("unsupported language: %s", lang))
+			return fmt.Errorf("unsupported language: %s", lang)
```
hajimehoshi:
```go
fmt.Errorf("ebitenmobile: unsupported language: %s", lang)
```

kumagi: Done — updated to `fmt.Errorf("ebitenmobile: unsupported language: %s", lang)` so it matches the other error messages in this command.

kumagi:
Thanks for the review! I've updated the error to include the `ebitenmobile: ` prefix, matching the convention used by the other error messages in this command.

hajimehoshi (approved):
LGTM

PR #3726, opened 2026-09-17
internal/glfw: fix a nil dereference in setVideoModeWin32

kumagi:
None

bug

## What this PR does | solves
chooseVideoMode can return nil (e.g. when EnumDisplaySettingsW fails on the first probe or right after a hot plug), and setVideoModeWin32 dereferenced it through best.equals(current), panicking the main thread on SetFullscreen(true). The X11 and cocoa versions already guard.

Return nil when no best mode is found.


hajimehoshi (approved):
This seems just a defensive code.

LGTM

PR #3727, opened 2026-09-17
internal/file: fix ReadDir returning nil entries on stat failure

kumagi:
None

bug

## What this PR does | solves
When ReadDir(count) was called with count <= 0 (read all) and os.Stat failed for a given entry, it returned the full ents slice which contained nil entries for indices that were not yet filled. This violates the io/fs contract (no nil entries in the returned slice) and can cause panics in callers like fs.WalkDir.

Fix: return ents[:i] (the entries filled so far) instead of the full ents slice, so no nil entries are exposed.


hajimehoshi (approved):
LGTM

PR #3728, opened 2026-09-17
internal/glfw: compare adapter names for adapter-fallback monitors on Windows

kumagi:
None

bug

## What this PR does | solves
A monitor created as an adapter fallback has no display name, so the disconnection check in the adapter-fallback path never matched and the monitor was reported disconnected and reconnected on every display change. Compare the adapter name as the C GLFW does.


hajimehoshi (approved):
LGTM

Issue #3212, opened 2025-03-19
examples/textinput: pressing 'Delete' key inserts U+007F (Delete) on macOS


hajimehoshi:
### Ebitengine Version

54b52001b1da78e400aa92e359f0ecf56ce4b274

### Operating System

- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

### Go Version (`go version`)

go version go1.24.0 darwin/arm64

### What steps will reproduce the problem?

1. Open `examples/textinput`
2. Focus a text field
3. Press a 'Delete' key. (Fn + Backspace)

### What is the expected result?

Nothing is input.

### What happens instead?

U+007F (Delete) is inserted. (On the example, the character is invisible though)

### Anything else you feel useful to add?

Usually, pressing a delete key should delete a character on the right of the cursor.

PR #3729, opened 2026-09-17
exp/textinput: ignore a DEL commit in the session pipeline

kumagi:
None

bug

## What this PR does | solves
Some virtual-keyboard IMEs signal a word deletion by committing U+007F. The legacy Field filters it (#3212), but the Composer/session pipeline handed it to OnCommit as regular text.


exp/textinput/session.go
```diff
@@ -182,6 +182,10 @@ func (s *session) Update() error {
 				s.markClosed(true)
 				return st.Error
 			}
+			if st.CommitKind.committed() && st.Text == "\x7f" {
+				// DEL should not modify the text (#3212).
+				continue
```
hajimehoshi: **[P2] Finish the session when discarding a queued DEL**

When DEL is queued behind another commit, the next session receives it through `flushStateQueue`, which sets `sessionCommitted = true` but leaves the channel open. This `continue` skips session cleanup, so subsequent queued input cannot reach the channel and Composer cannot restart the session. Please discard the DEL while still completing the session lifecycle.

I reproduced this on the PR head using the existing `DiffSender` test bridge: send buffer snapshots `"a"`, `"a\x7f"`, and `"a\x7fb"` as regular commits before updating; consume the first commit, then start the next session with surrounding text `"a"`. Repeated updates leave the DEL session open and the queued `b` undelivered. The existing `exp/textinput` tests pass, but this focused regression test fails.

Review comment authored by Codex (OpenAI), on behalf of @hajimehoshi.

kumagi: Good catch - you are right that the `continue` skipped the session teardown, and since `flushStateQueue` leaves the new channel open after setting `sessionCommitted`, the DEL session stayed open and queued input behind it was stranded. Fixed: when a committed DEL is discarded, the session now completes its lifecycle the same way a normal commit does (`markClosed(true)` with the platform-end callback, then return), while recording no commit so `IsCommitted` stays false and the DEL never reaches `OnCommit`. Closing the session releases the queued states, so the Composer restarts and the next session takes the input queued behind the DEL. I added TestDiffSenderQueuedDeletionEndsSession using the DiffSender test bridge with your exact reproduction (snapshots "a", "a\x7f", "a\x7fb" as regular commits before updating; consume the first commit, then start the next session with surrounding text "a"): it failed on the previous head with the DEL session still open and "b" undelivered, and now passes, asserting the DEL session closes with no commit and "b" reaches the following session, applying to "ab". All existing exp/textinput tests still pass, including under -race.

kumagi:
Thanks for the review! Fixed the queued-DEL case in exp/textinput: session.Update now discards the DEL without reporting a commit, but still completes the session lifecycle the way a normal commit does, so states queued behind the DEL reach the next session and the Composer can restart. Added a regression test (TestDiffSenderQueuedDeletionEndsSession) built on the existing DiffSender test bridge that reproduces the exact scenario - buffer snapshots "a", "a\x7f", and "a\x7fb" sent as regular commits before any update, first commit consumed, next session seeded with surrounding text "a" - and verifies the DEL session closes without a commit and "b" reaches the session after it. The test fails on the previous head (DEL session left open, "b" undelivered) and passes now. All existing exp/textinput tests pass, including with -race.

hajimehoshi (approved):
LGTM

PR #3735, opened 2026-09-19
internal/gamepad: keep HID pads GameController claims but drops on macOS

### LINKED ISSUE #3736: internal/gamepad: regression: USB gamepads are not recognized
### Ebitengine Version

v2.10.2

### Operating System

- [ ] Windows
- [x] macOS
- [ ] Linux
- [ ] FreeBSD
- [ ] OpenBSD
- [ ] Android
- [ ] iOS
- [ ] Nintendo Switch
- [ ] PlayStation 5
- [ ] Xbox
- [ ] Web Browsers

go version go1.27.1 darwin/arm64

1. On macOS 26.0, connect a generic USB HID gamepad with vendor/product ID  0810:e501  (reported as  usb gamepad ).
2. Run an Ebitengine application that enumerates controllers using  ebiten.AppendGamepadIDs  from its update loop.
3. Check the enumerated controllers and try pressing buttons or moving the D-pad.

Both controllers should appear in  ebiten.AppendGamepadIDs.

### What happens instead?

Neither controller appears in  ebiten.AppendGamepadIDs .

The devices fall between the two macOS backends:

1.  `+[GCController supportsHIDDevice:]`  returns  `YES`, so  `nativeGamepadsIOKit.addDevice`  skips the device in favor of GameController.
2. The controller delivered by GameController has no extended gamepad profile, so `addController`  ignores it using the guard that also excludes the Siri Remote.
3. Neither backend registers the device.

stigsb:
Closes #3736

bug

## What this PR solves

`+[GCController supportsHIDDevice:]` returns `YES` for a plain USB HID pad, so the IOKit backend steps aside for it, but the controller then arrives with only a micro gamepad profile and `addController` dropped it. Neither backend registered the device and the pad was invisible to the engine.

Register a controller that has a micro gamepad profile and no extended one. The micro profile itself gives the dpad, read as a hat and as two axes with `reportsAbsoluteDpadValues` set, buttons A and X, and the menu button where the profile has one. The pad's other buttons are not in the micro profile but are in `physicalInputProfile`, so B, Y, the shoulders and Options are read from there when present. The backend serves the standard layout itself, as the SDL-style GUID (Apple vendor, product 3, subtype 3) has no gamepaddb entry. GameController keeps owning every device it claims, as it does in SDL, so the backends need no link between a controller and its HID device.

## Hardware results

Tested on macOS 26.0 with two generic `usb gamepad` controllers (`0810:e501`), which GameController exposes as `GCMicroGamepad` and which are invisible to Ebitengine on v2.10.2. With this branch both appear in `ebiten.AppendGamepadIDs`, every physical button and the dpad deliver input, and the standard layout is available. I have no Siri Remote or MFi/Xbox/PlayStation controller to test the extended-profile path on real hardware.

## Tests

`export_darwin_test.go` swaps out the hardware boundary of the two darwin backends: which HID devices GameController claims, HID device properties, and the GCController class. CF objects, GC profiles, callbacks, registration, removal and reference ownership stay production code. Controllers are built with `+[GCController controllerWithMicroGamepad]` and `controllerWithExtendedGamepad`, so the tests skip on a macOS without them.

- `TestGameControllerOwnsClaimedDevices`: a claimed device is registered once, by the GameController backend, in either arrival order.
- `TestMicroGamepadPhysicalProfileButtons`: each physical-profile button is reported in the button mask and read back pressed.


hajimehoshi:
Apparently the issue is a regression and serious enough to back-port. Can you file an issue? Thanks,

hajimehoshi:
Also, please do not ignore the PR template.

stigsb:
> Also, please do not ignore the PR template.

Sorry about that, fixed now. Filed bug #3736 too.

internal/gamepad/gamepad_iokit_darwin.go
```diff
@@ -188,13 +193,44 @@ func (g *nativeGamepadsIOKit) update(gamepads *gamepads) error {
...
+			}
+			_CFRelease(_CFTypeRef(d))
+			return true
+		})
 		_CFRelease(_CFTypeRef(device))
 	}
 	g.devicesToAdd = g.devicesToAdd[:0]
 	g.devicesToRemove = g.devicesToRemove[:0]
+
+	if gcDroppedController.Swap(false) {
```
hajimehoshi: [P2] Preserve rejection information until the HID device is handled

If GameController rejects a controller before its IOKit matching callback is queued, an intervening update consumes this flag while there is no corresponding deferred device. When the IOKit callback arrives later, addDevice defers the device, but no rejection signal remains to reclaim it. The controller can therefore stay invisible until another rejection occurs. The atomic flag prevents a data race but does not preserve this cross-backend event ordering. Please retain enough rejection state to handle the later HID arrival, and cover the sequence rejection -> update -> HID arrival -> update.

internal/gamepad/gamepad_iokit_darwin.go
```diff
@@ -188,13 +193,44 @@ func (g *nativeGamepadsIOKit) update(gamepads *gamepads) error {
...
 
+// reclaimDevicesFromGC registers the deferred devices the GameController backend turned out not to
+// take. A device stays deferred while a GameController gamepad reports the same name, which is how
+// a device GameController does drive is told apart from one it dropped.
+func (g *nativeGamepadsIOKit) reclaimDevicesFromGC(gamepads *gamepads) {
+	g.devicesDeferredToGC = slices.DeleteFunc(g.devicesDeferredToGC, func(device _IOHIDDeviceRef) bool {
+		name := hidDeviceName(device)
+		if gamepads.find(func(gp *Gamepad) bool {
+			_, ok := gp.native.(*nativeGamepadGC)
+			return ok && gp.name == name
```
hajimehoshi: [P2] Associate fallback decisions with devices instead of display names

A rejection from any controller causes every deferred HID device to be reconsidered. Here hidDeviceName reads kIOHIDProductKey, whereas gp.name for a GameController controller comes from vendorName; these are not guaranteed to match. If a working controller has different names across the APIs, connecting a rejected generic pad also registers that working controller through IOKit, producing a duplicate. Even equal names do not protect a controller whose GameController registration has not arrived yet, and collisions can keep a rejected device deferred. Please use a reliable per-device association or ownership decision that remains valid across callback ordering, and verify a mixed accepted/rejected controller setup.

hajimehoshi (changes requested):
Thanks for filing #3736 and updating the description. The issue establishes impact on v2.10.2, but the fallback still depends on callback order and does not reliably associate HID devices with GameController controllers. Please address the two inline findings before merging, with coverage for both callback orders and a mixed setup containing an accepted controller and a rejected generic pad.

This is a source review; I have not independently reproduced the hardware behavior. Review authored by Codex (OpenAI), on behalf of @hajimehoshi.

internal/gamepad/gamepad_iokit_darwin.go
hajimehoshi: The rewrite in 30c8902 resolves the original ownership and callback-order concerns. Thanks for addressing those.

One issue remains: the micro-gamepad path exposes only A, X, and optionally Menu, plus the D-pad. The reported NEXT SNES Controller has ten buttons, and its existing HID mapping includes B, Y, shoulders, and Back. Those inputs are not read by the new implementation, so enumeration is restored but the controller's full functionality is still missing.

Please verify every physical button on the reported controllers and preserve access to the inputs omitted by the micro profile. I ran `go test -v ./internal/gamepad/...` successfully; the new tests cover arrival ordering and lifecycle, but their input checks use idle virtual controllers and cannot establish that the physical buttons work.

Please also update the PR description: it still describes the removed IOKit fallback and its ten-button result, rather than the current implementation and its hardware results.

Comment authored by Codex (OpenAI), on behalf of @hajimehoshi.

internal/gamepad/gamepad_iokit_darwin.go
stigsb: Fixed now.

internal/gamepad/gamepad_iokit_darwin.go
stigsb: Fixed by commit  1eb2ff6

stigsb:
I believe I have addressed your concerns now @hajimehoshi

hajimehoshi:
Sorry for the back-and-forth in the review, especially after asking you to expand the micro-profile implementation.

After reconsidering the scope of #3736, I would prefer fixing the backend handoff with an IOKit fallback rather than adding micro-profile support in this PR. Please keep GameController handling controllers with an extended gamepad profile, and let IOKit handle the corresponding HID device when no extended profile is available. This would address the regression by restoring the HID path for devices the existing GameController backend cannot read.

The earlier concerns about the original fallback implementation still apply: the decision needs to be reliable per device, without relying on display-name matching or a global rejection flag that can be consumed before the HID callback arrives. Please cover both callback orders and a mixed setup with an extended-profile controller and a device requiring fallback, ensuring each is registered only once.

Thank you for working through the revisions, and apologies again for changing direction.

Comment authored by Codex (OpenAI), on behalf of @hajimehoshi.

hajimehoshi:
If you are busy, I'm happy to work on this issue

hajimehoshi:
I'll take it from here.

